PR #1160, opened 2013-09-13
Android packages may have underscores

bootstraponline:
var str = "mFocusedApp=AppWindowToken{41ac2030 token=Token{41797308
ActivityRecord{419b9290 u0 com.example.webviewapp_18_api/.MainActivity}}}";

// undefined
/mFocusedApp.+ ([a-zA-Z0-9.]+)\/(.?[^}]+)}/.exec(str)

Fix #1159

### DIFF
--- a/android/adb.js
+++ b/android/adb.js
@@ -1254,7 +1254,7 @@ ADB.prototype.getFocusedPackageAndActivity = function(cb) {
   logger.info("Getting focused package and activity");
   this.requireDeviceId();
   var cmd = this.adbCmd + " shell dumpsys window windows"
-    , searchRe = new RegExp(/mFocusedApp.+ ([a-zA-Z0-9\.]+)\/(\.?[^\}]+)\}/);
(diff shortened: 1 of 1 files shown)


CMahesh:
I ran into the same issue 
<code>
org.openqa.selenium.SessionNotCreatedException: A new session could not be created. (Original error: Could not parse activity from dumpsys) (WARNING: The server did not provide any stacktrace information)
    Command duration or timeout: 56.75 seconds
    Build info: version: '2.44.0', revision: '76d78cf323ce037c5f92db6c1bba601c2ac43ad8', time: '2014-10-23 13:11:40'
    System info: host: 'mobserver5', ip: '10.147.116.98', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.10.3', java.version: '1.7.0_75'
    Driver info: io.appium.java_client.android.AndroidDriver
<code>

while running the appium from MAC using a simple app. It is not reproducable everything. But, only sometimes.

PR #1331, opened 2013-10-20
iOS logs for simulator and real device

ecomrobotics:
#### Setting up new features

``` sh
./reset.sh --ios --verbose
```
###### Supported log types : logcat, ios

``` python
self.driver.log_types
```
###### iOS logs for simulator and devices

``` python
self.driver.get_log("syslog")
```

TODO: "crashlog"
<br/>
#### Fixed server crash on ADB.prototype.getDevicesWithRetry

``` log
error: uncaughtException: Cannot call method 'restartAdb' of undefined
```
#### Fixed an issue in grunt configuration

``` log
maxBuffer increased to 2000*1024 due to problems with "maxBuffer exceeded" for selendroid building 
```


camillake:
Does get_log() support 'crashlog' now? I use python client to get_log('crashlog'), but return empty list on iOS device.

jlipps:
@camillake you can use the get log types method to figure out which log types are available

PR #1890, opened 2014-02-10
Add androidCoverage cap

bootstraponline:
This is the first half of #1889. The app is started with `-e coverage true` when the device cap `android-coverage` is set.

I'll update this PR with the second half (extracting coverage data) later today.

### DIFF
--- a/docs/caps.md
+++ b/docs/caps.md
@@ -21,6 +21,7 @@ Appium server capabilities
 |`app-wait-activity`| Activity name for the Android activity you want to wait for|`SplashActivity`|
 |`device-ready-timeout`| Timeout in seconds while waiting for device to become ready|`5`|
 |`compressXml`| [setCompressedLayoutHeirarchy(true)](http://developer.android.com/tools/help/uiautomator/UiDevice.html#setCompressedLayoutHeirarchy(boolean\))| `true`|
+|`androidCoverage`| Fully qualified instrumentation class. Passed to -w in adb shell am instrument -e coverage true -w | `com.my.Pkg/com.my.Pkg.instrumentation.MyInstrumentation`|
 
 --
 
--- a/docs/server-args.md
+++ b/docs/server-args.md
@@ -33,6 +33,7 @@ All flags are optional, but some are required in conjunction with certain others
 |`--app-activity`|null|(Android-only) Activity name for the Android activity you want to launch from your package (e.g., MainActivity)|`--app-activity MainActivity`|
 |`--app-wait-package`|false|(Android-only) Package name for the Android activity you want to wait for (e.g., com.example.android.myApp)|`--app-wait-package com.example.android.myApp`|
 |`--app-wait-activity`|false|(Android-only) Activity name for the Android activity you want to wait for (e.g., SplashActivity)|`--app-wait-activity SplashActivity`|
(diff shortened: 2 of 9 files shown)


su-agrawal:
@bootstraponline @jlipps 
I have tried the ant+emma setup as mentioned here https://github.com/appium/appium/blob/0.18.x/docs/en/android_coverage.md but no coverage files are generated.. Would the setup be different for Android+Gradle project. Pls help. Thanks!

bootstraponline:
The reason the ant/emma instructions have been removed is that Google has moved on to completely different technology. Google now recommends gradle/jacoco. You can see how jacoco is setup via gradle in [this droiddriver example](https://github.com/appium/droiddriver_examples/blob/eedcb261ddcabab5111cdc7c0f420428a0918f82/espresso_sample/espresso/BasicSample/app/build.gradle#L38).

The `mobile :endCoverage` and `EndEmmaBroadcast.java` should work the same since jacoco supports the emma dump command. I don't think anyone has tried this yet since appium 1.5 is the current priority.

su-agrawal:
@bootstraponline I have tried with Jacoco setup and I see report for unit tests. I have done the same setup by enabling jacoco and writing endemmabraoadcast. But running appium uia tests passes the tests but do not generate a coverage report for me in the sdcard. Would you be able to guide in that direction?

bootstraponline:
@su-agrawal Unfortunately it sounds like this feature is probably broken in appium. I use jacoco with unit tests and not via appium. There are no automated tests to verify appium's coverage support works correctly. I'd expect this to be resolved after the appium 1.5 release.

shubhansha21:
@bootstraponline any updates on this? Please let me know.

bootstraponline:
Google completely broke coverage support in Android N so I don't think this is going to happen

> Because Jack does not generate intermediate class files when compiling an app, tools that depend on these files do not currently work with Jack. Some examples of these tools are:
>
> Lint detectors that operate on class files
> Tools and libraries that require the app’s class files (such as instrumentation tests with JaCoCo)
>
> https://developer.android.com/preview/j8-jack.html

shashankjhs:
Tried passing androidCoverage capabilities while launching app using appium.but didn't able to generate code coverage.
help me out with this.

bootstraponline:
Jack is going away so code coverage should be feasible again. I don't work on appium anymore so it'd be up to the community to add this feature.

> https://android-developers.googleblog.com/2017/03/future-of-java-8-language-feature.html

shashankjhs:
Thanks for quick reply @bootstraponline,
I am trying with 
  :Appium 1.6.3
  :Api level 22(Android 5.2)
  :Java version 8
   
With this combination it should work or not ? or need to wait Appium community to update appium code.
Thanks a lot in advance.

bootstraponline:
I think this feature is currently broken in appium so you'd have to wait until that's fixed. Due to the nature of open source projects, there's no ETA.

shashankjhs:
Thanks for the info ...

jonathong:
Still no new info here?

bootstraponline:
I don't work on appium anymore so it'd be up to the community to add this feature.

I know that code coverage with espresso is working (non-appium). If one of the appium devs wants to work on this, I'm happy to provide advice.

bootstraponline:
I don't think code coverage is working with appium anywhere. It works if you use the underlying tools directly (uiautomator2/espresso).

PR #1910, opened 2014-02-14
add ability to scroll ios views

jlipps:
this works on ios7!

cc @dancuellar @dandoveralba

### DIFF
--- a/docs/gestures.md
+++ b/docs/gestures.md
@@ -79,6 +79,7 @@ In the case of these new mobile methods, `script` must be one of:
   * `mobile: flick`
   * `mobile: swipe`
   * `mobile: scrollTo`
+  * `mobile: scroll`
   * `mobile: shake`
 (The `mobile:` prefix allows us to route these requests to the appropriate endpoint).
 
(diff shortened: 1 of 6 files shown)


xudafeng:
:+1:

pavelzag:
Bumping up this discussion because https://github.com/appium/appium/blob/master/docs/gestures.md#scroll is broken
Are there any new for C# mobile:sweep implementation? Is it expected to be added in the upcoming versions?
Cheers, Pavel

docs/gestures.md
```diff
@@ -222,6 +227,26 @@ In these examples, note that the element parameter is always optional.
...
+    // continue testing
+  });
+  ```
+
+* **Java:**
+
+  ```java
+  JavascriptExecutor js = (JavascriptExecutor) driver;
+  HashMap<String, Double> scrollObject = new HashMap<String, Double>();
+  scrollObject.put("direction", "down");
```
gustavjr: Your HashMap is declared with < String, Double> and here your are passing < String, String> .

jlipps: thanks, will fix that.

jlipps: actually this is already fixed in master. this PR is ancient.

PR #3964, opened 2014-11-04
ios-webkit-debug-proxy launcher

sebv:
Goes round `Invalid message _rpc_applicationUpdated` error by killing/restarting proxy process.

@jlipps is it ok to add that, (Still needs documenting)


MJanach:
I'm getting a similar error to the one this fix was designed for (I think).  But I'm on Appium 1.3.6.  I am facing the same error mentioned here: https://discuss.appium.io/t/safari-on-ios-ios-webkit-debug-crashing-when-safarilauncher-launches-safari/1208

PR #4005, opened 2014-11-07
touch actions *do* work in webview

lukeis:


bhaskar-appium:
press(), longPress() are working fine with Native App on Android...But moveTo(),tap() aren't working.
I'm using element,not with co-ordinates...
Appium 1.3.4
Windows 7
Android 4.4.2, Java Bindings 2.1.0
Please suggest me in a better way.Thank you in advance :)

0x1mason:
@bhaskar-appium If you have an implementation question, please try asking at https://discuss.appium.io/.  You may want to include the code in question since it's difficult to evaluate the problem without it.

0x1mason:
I think the short-term solution to this issue in Android is to implement touch actions with [chrome-remote-interface](https://github.com/cyrus-and/chrome-remote-interface), perhaps as a stopgap.  Over the long-term, we should probably ditch the ChromeDriver server as a separate process.  We can use CD as a library and run a CD http listener from bootstrap.  That would solve our touch action problems, greatly simplify CD instance management, improve debugging, and generally increase our control over implementation.

lukeis:
gonna close this PR, not seemingly getting around to implementing this.

0x1mason:
@lukeis We'll try to get this in for the next release, v1.3.6.

lukeis:
@0x1mason why reopen? unless you're using this to track the issue?

I meant to say "I'm not seemingly getting around to implementing this correctly"

0x1mason:
@lukeis Perhaps I misread.  I thought it was just missing tests.  I'll go ahead an reclose.

doronzRumble:
can i understand that since this issue got closed, it wont be implemented as a feature for one of the upcoming releases?
is there a workaround for this to enable swiping in webapp?

lukeis:
@doronzRumble you can switch to Native and perform the action (given the coordinates of the webview items that you can fetch before switching to native)

chOOnz:
@lukeis Yes we can do that but it more a "hacky" way in my opinion. I don't like working on native context when I do everything else on the webview.

@0x1mason I think you forgot to implement it, we are on version 1.4.16 now...

jlipps:
@chOOnz this was closed as 'won't merge' for the time being. we might implement the feature in a future version of appium though

chOOnz:
@jlipps Sounds good for me. Is there an estimation when this will come in?

I really have trouble clicking on elements in my webview and I guess I'm not the only one.

jlipps:
@chOOnz as a workaround you can always get into the native context and do the tapping based on coordinates.

chOOnz:
@jlipps No I actually can't switch to the native context because we want to use the same tests inside a browser, there is no native context.

jlipps:
@chOOnz so when you call the getContexts command there is only one thing in it?

chOOnz:
@jlipps No, I can see the native context and the webview context. Working inside the native context is no option for us though so switching back and forth will not work for us. Otherwise I would switch to native context to perform the tap.

I already got it working on webview context by using the JSExecutor and trigger the touchend event. It's an ugly way to do it though but it's working for now.

jlipps:
ok glad you found a workaround; however i'm curious why the native context is no option for you?

chOOnz:
@jlipps We decided to use the native context instead of JSExecutor. I now use the touch actions of appium to tap on x, y coodinates. The problem I ran into is converting the webview coordinates of an element to the native coordinates. It seems to be a little bit tricky due to the fact that the small android menu on the top has different heights (depends on the devices resolution). Is there any solution to this problem?

jlipps:
@chOOnz does that menu not have height attributes?

chOOnz:
@jlipps By menu I mean the system menu provided by android itself. I didn't find a way yet to access it and then read attributes from it. I hope you know about which menu I am talking about. :D

PR #4044, opened 2014-11-12
iOS: Fix for a bug when driver.get() never returns for page with alert.

paymand:
Running:

``` python
driver.switch_to.context("WEBVIEW")
driver.get("http://twitch.tv")
alert = driver.switch_to.alert()
alert.dismiss()
```

Appium gets stuck after `driver.get` forever (or until it hits the command timeout and the session is closed).
The website in the test, _twitch.tv_, opens up an alert while we are checking for readyState and it blocks the javascript thread it seems. As a result the call to `checkPageIsReady` never returns while the alert is present.
Regardless, the `pageLoad` function is not supposed to take more than `waitMs` and this is what my fix tries to achieve by adding a new timeout.

### DIFF
--- a/lib/devices/ios/remote-debugger.js
+++ b/lib/devices/ios/remote-debugger.js
@@ -363,27 +363,37 @@ RemoteDebugger.prototype.navToUrl = function (url, cb) {
 
 RemoteDebugger.prototype.pageLoad = function () {
   clearTimeout(this.loadingTimeout);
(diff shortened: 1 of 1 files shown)


0x1mason:
@paymand If you like to keep the issue up for discussion, please create an Opera-specific issue. Thanks!

paymand:
@sebv: That's right. I totally forgot. I'll see if I can get a special dev build...

PR #4252, opened 2014-12-10
retry dump hierarchy if dump.xml wasn't created

aluedeke:
dump hierarchy sometimes succeeds but doesn't create a dump.xml file (usually during ui transitions).


aluedeke:
@jlipps @bootstraponline can you share more details about your test setup? what android version are you using? is it an arm or intel emulator? anything else special? would love to fix my setup to be able to contribute more stuff. Test execution takes more than 1 hour so i would love to get it running.

jlipps:
I’m running the tests now, fyi

jlipps:
Had some trouble with the emulators and got a lot of failures (I upgraded to yosemite). Trying again now

jlipps:
I'm blocked on testing this by #4368, but apart from this bug, all the tests were passing with your change! So I'm optimistic everything is good with it.

jlipps:
Since I'm having huge problems with my test env, @0x1mason can you pull this branch and see whether this passes the Android tests?

0x1mason:
@jlipps Yes.

lib/devices/android/bootstrap/src/io/appium/android/bootstrap/utils/ReflectionUtils.java
```diff
@@ -0,0 +1,109 @@
+package io.appium.android.bootstrap.utils;
```
0x1mason: This is sooooo much better.  The fact these were all instance methods was killing me.

lib/devices/android/bootstrap/src/io/appium/uiautomator/core/InteractionController.java
```diff
@@ -0,0 +1,49 @@
...
+  public static final String METHOD_PERFORM_MULTI_POINTER_GESTURE = "performMultiPointerGesture";
+
+  private final Object interactionController;
+
+  public InteractionController(Object interactionController) {
+    this.interactionController = interactionController;
+  }
+
+  public boolean sendKey(int keyCode, int metaState){
+    return (Boolean) invoke(method(CLASS_INTERACTION_CONTROLLER, METHOD_SEND_KEY, int.class, int.class), interactionController, keyCode, metaState);
```
0x1mason: If you're going to go to the trouble of creating a wrapper, can't you just define all your reflected methods in the ctor rather than repeatedly doing so for each invocation? Or is there a reason not to? It's not a huge deal either way tho.

aluedeke: no special reason why it is like that. i am just not a big fan of initializing stuff upfront which i probably don't need later on. if you care about  performance here i would prefer to build a tiny cache (Map) in ReflectionUtils. what do you think?

lib/devices/android/bootstrap/src/io/appium/uiautomator/core/AccessibilityNodeInfoDumper.java
```diff
@@ -0,0 +1,222 @@
...
+              Point size = new android.graphics.Point();
+              display.getSize(size);
+              width = size.x;
+              height = size.y;
+
+              serializer.attribute("", "rotation", Integer.toString(display.getRotation()));
+            }
+
+          if (root != null) {
+            dumpNodeRec(root, serializer, 0, width, height);
```
0x1mason: If this isn't API_18, width and height will be -1.

lib/devices/android/bootstrap/src/io/appium/uiautomator/core/AccessibilityNodeInfoDumper.java
```diff
@@ -0,0 +1,222 @@
...
+        try {
+            FileWriter writer = new FileWriter(dumpFile);
+            XmlSerializer serializer = Xml.newSerializer();
+            StringWriter stringWriter = new StringWriter();
+            serializer.setOutput(stringWriter);
+            serializer.startDocument("UTF-8", true);
+            serializer.startTag("", "hierarchy");
+            int width = -1;
+            int height = -1;
+            if(API.API_18){
```
0x1mason: It'd be nice if we can document what's special about API 18 and why it requires special handling.

lib/devices/android/bootstrap/src/io/appium/uiautomator/core/AccessibilityNodeInfoHelper.java
```diff
@@ -0,0 +1,47 @@
...
+        if (node == null) {
+            return null;
+        }
+        // targeted node's bounds
+        Rect nodeRect = new Rect();
+        node.getBoundsInScreen(nodeRect);
+        Rect displayRect = new Rect();
+        displayRect.top = 0;
+        displayRect.left = 0;
+        displayRect.right = width;
```
0x1mason: What happens if width and height are -1?

lib/devices/android/bootstrap/src/io/appium/uiautomator/core/InteractionController.java
0x1mason: This is fine.  I was more just curious as to why you did it the other way.  I doubt it has performance implications for Appium either way.

0x1mason:
Can you add a comment to the code explaining why we need to re-do Google's XML dump implementation?  It doesn't have to be long, but it should be clear enough to someone unfamiliar with this ticket as why we've taken the original code and modified it.

0x1mason:
@aluedeke Just curious--in what state is the root null that we don't want to abort serialization?

jlipps:
@aluedeke any thoughts? we'd love to bring this into 1.3.5 which is about ready to release, and I think @0x1mason has some good questions

bootstraponline:
> Can you add a comment to the code explaining why we need to re-do Google's XML dump implementation? It doesn't have to be long, but it should be clear enough to someone unfamiliar with this ticket as why we've taken the original code and modified it.

It'd also be nice to link to the relevant android issues.

lib/devices/android/bootstrap/src/io/appium/uiautomator/core/AccessibilityNodeInfoDumper.java
aluedeke: getDefaultDisplay was added in API Level 18. will add a comment

lib/devices/android/bootstrap/src/io/appium/uiautomator/core/AccessibilityNodeInfoHelper.java
aluedeke: there was actually a bug, the bounds attribute should only be serialized if API Level >=18, will provide a fix asap

aluedeke:
@0x1mason not sure about which null check your are talking. From my experiences the root is null when switching between two applications (as long as the second app isn't ready the root is null). This led to exceptions in the old code. Now it just returns an empty xml which can be properly handled by the clients.

lib/devices/android/bootstrap/src/io/appium/uiautomator/core/AccessibilityNodeInfoDumper.java
```diff
@@ -0,0 +1,233 @@
...
+              Display display = UiAutomatorBridge.getInstance().getDefaultDisplay();
+              Point size = new android.graphics.Point();
+              display.getSize(size);
+              width = size.x;
+              height = size.y;
+
+              serializer.attribute("", "rotation", Integer.toString(display.getRotation()));
+            }
+
+          if (root != null) {
```
0x1mason: I see what's happening now.  I got thrown off by the fact that the if_API_18 logic is executed regardless of whether or not root is null.  Since you're making changes anyway, can you move the display dimension logic within the != null block?  If not possible, then no worries.

aluedeke: check

aluedeke:
anyhing else open?

0x1mason:
No, I think that's it.  Apologies for throwing this at you a month after your submitted your pull request.  I hadn't looked at this previously because of the holidays.

aluedeke:
@0x1mason @jlipps can one of you can run the tests again? i have still some issues with the test setup

0x1mason:
@0x1mason @jlipps On it.

PR #4365, opened 2015-01-08
Workaround for issue where UiAUtomator fails to find visible elements. Fixes #4200

### LINKED ISSUE #4200: Restart UIAutomator to find missing elements
Read the end of [this issue](https://github.com/appium/java-client/issues/74#issuecomment-61511540).

According to @ranbena restarting UIAutomator works %100 of the time to fix an issue where sometimes elements aren't returned by getPAgeSource(). It's definitely a bug in Android UIAutomator, but it seems like we can work around it by restarting the UIAutomator process.

Right now, restarting the process causes the Appium session to fail. What should we consider for fixing these missing elements?

Options I've thought of so far:
1) Make a "restart UIAutomator" function available to clients.
2) Restart UIAutomator whenever an element isn't found (or add it to some retry logic)
3) Create a "Setting" (from our settings api) which when enabled restarts UIAutomator when elements aren't found, and doesn't restart UIAutomator when disabled.

Any more ideas?
@jlipps @bootstraponline @sebv @imurchie @0x1mason @JessicaSachs

0x1mason:
Some control updates fail to trigger AccessibilityEvents, resulting in stale AccessibilityNodeInfo instances. In these cases, UIAutomator will fail to locate visible elements. As a work-around, this change force clears the AccessibilityInteractionClient's cache and instigates another search. This technique also makes some Appium's searches conclude more quickly since it obviates the need for polling.  Since there is a general benefit, I do not believe users should have to opt-in.

### DIFF
--- a/lib/devices/android/bootstrap/src/com/android/uiautomator/common/ReflectionUtils.java
+++ b/lib/devices/android/bootstrap/src/com/android/uiautomator/common/ReflectionUtils.java
@@ -7,6 +7,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 
 import static io.appium.android.bootstrap.utils.API.API_18;
+import java.lang.reflect.InvocationTargetException;
 
 public class ReflectionUtils {
   private static Field enableField(final Class<?> clazz, final String field)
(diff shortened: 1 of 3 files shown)


jlipps:
:+1:

bootstraponline:
:+1: awesome fix. Between this and https://github.com/appium/appium/pull/4252, the next release of appium could be the best yet in terms of uiautomator reliability.

0x1mason:
@bootstraponline Thanks!  

Gonna merge.

PR #4387, opened 2015-01-12
use appium-support package for system detection

jlipps:
Starting to move some code over. I can't publish the appium-support package to NPM yet, though, so this can't be merged till that's done and the package.json ref updated

Issue #3177, opened 2014-07-20
Failed to switch back and forth between webview & native view


raycheung:
I'm using Protractor to drive Appium , and use wd-bridge to switch the context via the following function:

```
switch_context_to = (context) ->
  wdBrowser.contexts().then (contexts) ->
    throw "Unknown context: [#{context}] ! Available contexts: [#{contexts}]" unless context in contexts
    wdBrowser.context context
```

and protractor.conf.coffee is set with:

```
autoWebview: true
autoWebviewTimeout: 10
```

I'm able to switch the context from the webview to the `NATIVE_APP`, click through the buttons in the emulated camera view (of the android emulator), then switched back to the webview, the immediate next action is always fail with the following error:

```
debug: Proxying command to 127.0.0.1:9515
debug: Making http request with opts: {"url":"http://127.0.0.1:9515/wd/hub/session/50080a17d90e3b63eb84f3950b0e468b/execute_async","method":"POST","json":{"script":"try { return (function (selector, callback) {\n  var el = document.querySelector(selector);\n  try {\n    angular.element(el).injector().get('$browser').\n        notifyWhenNoOutstandingRequests(callback);\n  } catch (e) {\n    callback(e);\n  }\n}).apply(this, arguments); }\ncatch(e) { throw (e instanceof Error) ? e : new Error(e); }","args":["body"]}}
debug: Responding to client with error: {"status":28,"value":{"message":"A script did not complete before its timeout expired.","origValue":"asynchronous script timeout: result was not received in 0 seconds\n  (Session info: webview=)\n  (Driver info: chromedriver=2.10.267517,platform=Mac OS X 10.9.4 x86_64)"},"sessionId":"24d3ad6797e84db1c395ad7820dd7100"}
info: <-- POST /wd/hub/session/24d3ad6797e84db1c395ad7820dd7100/execute_async 500 345.137 ms - 320 
F

Failures:

  1) An new customer signs up and applies the first loan application
   Message:
     ScriptTimeoutError: A script did not complete before its timeout expired.
   Stacktrace:
     ScriptTimeoutError: A script did not complete before its timeout expired.
    at IncomingMessage.EventEmitter.emit (events.js:117:20)
    at _stream_readable.js:919:16
    at process._tickCallback (node.js:419:13)
```

bhawani2008:
Hello,
[...]

PR #4388, opened 2015-01-13
use appium-chromedriver package

jlipps:
This isn't ready to go yet, because:

1) appium-chromedriver needs a publish and to be added to package.json
2) that can't happen until I get publish rights to appium-support

But it's ready for feedback. The only thing that's really wonky is that the appium-chromedriver package uses an EventEmitter model whereas here to interface with the higher-level Appium code, we use callbacks. So at some point we'll want to standardize all devices on EventEmitter to clean up this wonkiness.


jlipps:
It's worth mentioning the Chrome and Webview tests pass

jlipps:
ok, I edited this and added a bunch of comments. It's ready for review, but note that it has the changes from #4387, and once that's merged it will no longer be a part of this pull request.

@sebv @imurchie @0x1mason please review.

I also wonder if I've come across a bug in our handling of android webviews. Every time we create a new chromedriver, we run a command that kills all chromedrivers on the port we want to use. But we want to have multiple chromedrivers active---one for each webview. However we don't do any cleverness with making sure each Chromedriver instance runs on a different port. So my hypothesis is that with the current code (in appium master), we kill all the non-active context chromedrivers whenever we switch to a new chromedriver context. It seemed odd to me that we'd be doing this, so I didn't do anything to fix it in this pull request, but it probably warrants investigation.

Also, I've been pushing code to https://github.com/appium/appium-chromedriver; you may also want to familiarize yourself with the "final" API. I don't think there's any work there left to do but you guys tell me.

0x1mason:
I think I'm experiencing the same or similar bug on windows:

```bat
info: [debug] dataDir set to: /data
info: [debug] Creating Chrome session
info: [debug] Set chromedriver binary as: C:\Users\eric\git\appium\build\chromed
river\windows\chromedriver.exe
info: [debug] Ensuring Chromedriver exists
info: [debug] Killing any old chromedrivers, running: FOR /F "usebackq tokens=5"
 %a in (`netstat -nao ^| findstr /R /C:"9515 "`) do (FOR /F "usebackq" %b in (`T
ASKLIST /FI "PID eq %a" ^| findstr /I chromedriver.exe`) do (IF NOT %b=="" TASKK
ILL /F /PID %b))
info: [debug] No old chromedrivers seemed to exist
info: [debug] Spawning chromedriver with: C:\Users\eric\git\appium\build\chromed
river\windows\chromedriver.exe
info: [debug] [CHROMEDRIVER] Starting ChromeDriver 2.13.307647 (5a7d0541ebc58e69
994a6fb2ed930f45261f3c29) on port 9515
Only local connections are allowed.
info: [debug] Making http request with opts: {"url":"http://127.0.0.1:9515/wd/hu
b/session","method":"POST","json":{"sessionId":null,"desiredCapabilities":{"chro
meOptions":{"androidPackage":"com.android.chrome","args":["--disable-translate"]
,"androidDeviceSerial":"emulator-5554"},"loggingPrefs":{"performance":"ALL"}}}}
info: [debug] [CHROMEDRIVER STDERR] [0.005][SEVERE]: bind() returned an error: O
nly one usage of each socket address (protocol/network address/port) is normally
 permitted. (0x2740)
info: [debug] [CHROMEDRIVER STDERR] [0.005][WARNING]: Unknown error 10038 mapped
 to net::ERR_FAILED
info: [debug] [CHROMEDRIVER] Port not available. Exiting...
info: [debug] Chromedriver exited with code 1
```

lib/devices/android/android-hybrid.js
```diff
@@ -118,101 +119,129 @@ androidHybrid.getProcessNameFromWebview = function (webview, cb) {
...
+
+  // check the status by sending a simple window-based command to ChromeDriver
+  // if there is an error, we want to recreate the ChromeDriver session
+  this.chromedriver.hasWorkingWebview().then(function (works) {
+    if (works) return cb();
+    logger.debug("ChromeDriver is not associated with a window. " +
+                 "Re-initializing the session.");
+    // catch any errors the restart process bubbles up
+    this.chromedriver.once('error', cb);
+    this.chromedriver.on('stateChanged', function (msg) {
```
0x1mason: Can we avoid using string literals here? It exposes implementation details, tightly coupling this invocation to a specific string in ChromeDriver and making it unnecessarily fragile.  It's easily avoided by using a constant instead.  We've already created constants for some of the states and events, iirc.

Same would apply to msg.state below.

lib/devices/android/android-hybrid.js
```diff
@@ -118,101 +119,129 @@ androidHybrid.getProcessNameFromWebview = function (webview, cb) {
...
-        this.cleanupChromedriver(this.chromedriver, function () {
-          setupNewChromeDriver(cb);
-        });
-      } else {
+androidHybrid.setupExistingChromedriver = function (context, cb) {
+  logger.debug("Found existing Chromedriver for context '" + context + "'." +
+               " Using it.");
+  this.rememberProxyState();
+  this.chromedriver = this.sessionChromedrivers[context];
+  this.proxyReqRes = this.chromedriver.proxyReq.bind(this.chromedriver);
```
0x1mason: curious if this is Function.prototype.bind() or a different bind.

lib/devices/android/android-hybrid.js
```diff
@@ -118,101 +119,129 @@ androidHybrid.getProcessNameFromWebview = function (webview, cb) {
...
+    _.each(this.args.specialChromedriverSessionArgs, function (val, option) {
+      logger.debug("This method is being deprecated. Apply chromeOptions " +
+                   "normally to pass along options,see sites.google.com/a/" +
+                   "chromium.org/chromedriver/capabilities for more info");
+      caps.chromeOptions[option] = val;
+    });
+  }
+  caps = this.decorateChromeOptions(caps);
+  // see note in chrome.js::createSession for explanation of this pattern
+  this.chromedriver.once('error', ocb);
```
0x1mason: See comment below on string literals

lib/devices/android/android-hybrid.js
0x1mason: Nm.  I see now.

lib/devices/device.js
```diff
@@ -34,7 +35,7 @@ Device.prototype.configure = function (args, caps) {
     // use a custom tmp dir to avoid loosing data and app
     // when computer is restarted
     this.args.tmpDir = process.env.APPIUM_TMP_DIR ||
-      (helpers.isWindows() ? process.env.TEMP : "/tmp");
+      (isWindows() ? process.env.TEMP : "/tmp");
```
0x1mason: Is there any reason that this dir already exists and may contain data from previous sessions?  Or that some other application is using a dir of the same name?  Might be nice to use a UUID rather than "/tmp".  Not a huge deal, but if we're changing this already.

lib/devices/android/android-hybrid.js
```diff
@@ -255,4 +287,38 @@ androidHybrid.initAutoWebview = function (cb) {
   }
 };
 
+// get the correct chromedriver executable path based on our system
+// TODO: don't download/build chromedriver in reset.sh, instead let this be
+// something that the appium-chromedriver package manages
+androidHybrid.initChromedriverPath = function (cb) {
+  if (this.args.chromedriverExecutable) {
```
0x1mason: Should we check this is a valid path?

lib/devices/device.js
0x1mason: Should we be using our new support.tempDir function here?

lib/devices/android/android-hybrid.js
0x1mason: We could also wrap the implementation, eg
```js
this.onStateChanged = (cb) => { this.on('stateChanged', cb); }
```

lib/devices/android/chrome.js
```diff
@@ -121,11 +117,15 @@ ChromeAndroid.prototype.pushAndUnlock = function (cb) {
...
   var caps = {
     chromeOptions: {
       androidPackage: this.args.appPackage
     }
   };
+  if (this.args.enablePerformanceLogging) {
+    caps.loggingPrefs = {performance: 'ALL'};
+  }
 
   var knownPackages = ["org.chromium.chrome.shell",
```
0x1mason: Out of scope for this ticket, but we should consolidate these collections with those in appium.js and put it in appium-support or some such.

lib/devices/android/android-hybrid.js
sebv: This is where the object model is not clean, but out of scope for this PR.

lib/devices/android/android-hybrid.js
sebv: First param is msg not error. I think here we need to move the promise bit we use in test to appium-support, or use a cb equivalent.

lib/devices/device.js
sebv: Yeah this should be done all in one place.

lib/devices/android/android-hybrid.js
```diff
@@ -3,14 +3,20 @@
...
   , errors = require('../../server/errors.js')
+  , path = require('path')
+  , system = require('appium-support').system
+  , isWindows = system.isWindows()
+  , isLinux = system.isLinux()
+  , exec = require('child_process').exec
   , UnknownError = errors.UnknownError
   , async = require('async')
-  , Chromedriver = require('./chromedriver.js')
+  , Chromedriver = require('appium-chromedriver').default
```
sebv: :+1:  This is the correct pattern.

lib/devices/android/android-hybrid.js
```diff
@@ -3,14 +3,20 @@
 var logger = require('../../server/logger.js').get('appium')
   , _ = require('underscore')
   , errors = require('../../server/errors.js')
+  , path = require('path')
+  , system = require('appium-support').system
+  , isWindows = system.isWindows()
+  , isLinux = system.isLinux()
```
sebv: Don't really need this.

jlipps:
btw, I think the bug I uncovered while working on this is the same as in #3177

jlipps:
Making a note of review comments since i'm about to push the conflict resolution:

- [ ] don't use string literals, maybe just wrap eventemitter implementation
- [ ] use support.tempDir

jlipps:
ok, code has been updated and is ready for final review. chrome/webview tests pass on emus and real device.

jlipps:
looking for a +1 from @sebv, @0x1mason, @imurchie

jlipps:
oh, also ping @Jonahss

0x1mason:
lgtm

Jonahss:
:)

sebv:
lgtm

jlipps:
hi all, i cleaned this up so the model is easier to understand. all tests still pass. should be a straight up improvement, but please feel free to have another look. will merge tomorrow if no objections

lib/devices/android/android-hybrid.js
jlipps: just responding to this now: the appium-chromedriver package ensures it's a valid path

PR #4402, opened 2015-01-16
Handle iOS 8.2 remote debugger

imurchie:
In the last two beta releases of iOS 8.2 the remote debugger returns a valid app connection but no content. The context code would then disconnect the remote debugger, there being no windows, and the whole thing would cycle like that until it timed out. So check for that condition, and retry.


jlipps:
good changes, thanks!

jlipps:
gotta love the remote debugger

PR #4445, opened 2015-01-26
ensure we drop into the right continuation cb

jlipps:
...when selecting hybrid context in ios (fix #4408)

PR #4453, opened 2015-01-28
Fix for the Safari Startup Crashes

sebv:
Appium now waits specifically for the Apple page instead of the most recent page.

I've also fixed real-safari, no need to press the reload button.
- other minor stuffs.

ping @jlipps @0x1mason

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -1525,23 +1525,42 @@ IOS.prototype.initAutoWebview = function (cb) {
   }
 };
 
-IOS.prototype.navToInitialWebview = function (cb) {
-  if (this.args.udid) {
-    setTimeout(function () {
-      // the latest window is the one we want.
-      this.navToLatestAvailableWebview(cb);
-    }.bind(this), 6000);
-  } else {
-    this.navToLatestAvailableWebview(cb);
-  }
+IOS.prototype.getContextsAndViews = function (cb) {
+  this.listWebFrames(function (err, webviews) {
+    if (err) return cb(err);
+    var ctxs = [{id: this.NATIVE_WIN}];
+    this.contexts = [this.NATIVE_WIN];
+    _.each(webviews, function (view) {
+      ctxs.push({id: this.WEBVIEW_BASE + view.id, view: view});
+      this.contexts.push(view.id.toString());
+    }.bind(this));
+    cb(null, ctxs);
+  }.bind(this));
 };
 
-IOS.prototype.navToLatestAvailableWebview = function (cb) {
-  if (parseInt(this.iOSSDKVersion, 10) >= 7 && !this.args.udid && this.capabilities.safari) {
-    this.navToViewThroughFavorites(cb);
-  } else {
-    this.navToView(cb);
-  }
+IOS.prototype.getLatestWebviewContextForTitle = function (titleRegex, cb) {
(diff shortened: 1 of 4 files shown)


sebv:
Also tested it in conjunction with #4445

0x1mason:
This seems awfully fragile.  How confident can we be that the magic `apple` string is a constant in this case?

submodules/io.appium.gappium.sampleapp
```diff
@@ -1 +1 @@
-Subproject commit 3fb5e71b1bc149a57de341b1afe5872c0a2227b9
```
0x1mason: Did you intend to change this?  Doesn't look related.

lib/devices/ios/ios.js
```diff
@@ -1525,23 +1525,43 @@ IOS.prototype.initAutoWebview = function (cb) {
...
-      // the latest window is the one we want.
-      this.navToLatestAvailableWebview(cb);
-    }.bind(this), 6000);
-  } else {
-    this.navToLatestAvailableWebview(cb);
-  }
+IOS.prototype.getContextsAndViews = function (cb) {
+  this.listWebFrames(function (err, webviews) {
+    if (err) return cb(err);
+    var ctxs = [{id:this.NATIVE_WIN}];
```
jlipps: spacing: `id: this.NATIVE_WIN`

lib/devices/ios/ios.js
```diff
@@ -1525,23 +1525,43 @@ IOS.prototype.initAutoWebview = function (cb) {
...
-    this.navToViewThroughFavorites(cb);
-  } else {
-    this.navToView(cb);
-  }
+IOS.prototype.getLatestWebviewContextForTitle = function (titleRegex, cb) {
+  this.getContextsAndViews(function (err, contexts) {
+    if (err) return cb(err);
+    var matchingCtx;
+    _(contexts).each(function (ctx) {
+      if (ctx.view &&  (ctx.view.title || "").match(titleRegex)) {
```
jlipps: spacing here after `&&`

submodules/io.appium.gappium.sampleapp
jlipps: Yeah, this takes Gappium back a commit, I don't think that's right.

jlipps:
@0x1mason it is a constant in this case because it's built into the Safari app. It will only possibly change when we have a new release of iOS, which requires testing and code updates in Appium anyway; so we'll catch any breaking changes at that point.

I reviewed the code, I think it's a great change and will fix the issue, but there are a few small things that need fixing.

submodules/io.appium.gappium.sampleapp
sebv: Yeah that's wrong.

sebv:
Fixed and rebase. 

@0x1mason I've tried on iOS81+iOS61emulators, real devices, a iOS81 emulator with chinese locale. I guess we'll need to keep an eye on this to see if we are missing some use cases.

jlipps:
we only need to run the safari, context, and hybrid tests for this change

jlipps:
and let's not forget to merge #4445 before we merge this.

jlipps:
I think we should merge this and test ASAP.

0x1mason:
Sounds good.

PR #4467, opened 2015-02-01
Add default packages to uiautomator id search

### LINKED ISSUE #4416: Different Locating by id (ressource-id) with Appium and Selendroid
Hello all,

I want to use Appium to automate an App, which is developed for iOS and Android.
Regarding Android the App has several small webviews included and supports Android versions back to 2.3.7. 
Since most users aren't on Android 4.4 I need to use Selendroid to test most of the configurations. When time goes by our users will migrate their Android versions and I can switch to Appium completely.

I know there are differences in locating elements in Selendroid and Appium, for example different nodes when using xpath. What strikes me as odd is that Appium and Selendroid differ when trying to locate elements by ID.

Here is and example (using Java, PageFactory and Annotations)

Here is *Appium*:
```
@AndroidFindBy(id = "de.myApp.client.android:id/welcome_button_login")
private MobileElement btn_Login ;
```

And here is *Selendroid*:
```
@AndroidFindBy(id = "welcome_button_login")
private MobileElement btn_Login ;
```
I personally prefer the way Selendroid does it, because it locates by the specified ID only and not by the complete package and id name. This makes my life easier especially since our app is shipped to different countries and sadly with different package names (e.g. fr.myApp.client.android or es.myApp.client.android). 

Therefore my question is:

Is there a reason why Appium does not locate by the ID alone but needs the whole package name or will this be changed in a future release?

Thank you very much,

  mgehlen

bootstraponline:
Given an element with resource id: com.example.android.testing.espresso.BasicSample:id/textToBeChanged

Now we can find this element using only id('textToBeChanged')
In addition to checking the default package name, the android package is also searched:

android:id/textToBeChanged

Fix #4416 

I'm working on DroidDriver and the need to fully qualify the ids was getting annoying.

### DIFF
--- a/lib/devices/android/android.js
+++ b/lib/devices/android/android.js
@@ -99,9 +99,9 @@ Android.prototype.start = function (cb, onDie) {
   async.series([
     this.initJavaVersion.bind(this),
     this.initAdb.bind(this),
+    this.packageAndLaunchActivityFromManifest.bind(this),
     this.initUiautomator.bind(this),
     this.prepareDevice.bind(this),
-    this.packageAndLaunchActivityFromManifest.bind(this),
     this.checkApiLevel.bind(this),
     this.pushStrings.bind(this),
     this.processFromManifest.bind(this),
--- a/lib/devices/android/bootstrap/src/io/appium/android/bootstrap/Bootstrap.java
+++ b/lib/devices/android/bootstrap/src/io/appium/android/bootstrap/Bootstrap.java
@@ -2,6 +2,7 @@ package io.appium.android.bootstrap;
 
 import com.android.uiautomator.testrunner.UiAutomatorTestCase;
 import io.appium.android.bootstrap.exceptions.SocketServerException;
+import io.appium.android.bootstrap.handler.Find;
 
 /**
  * The Bootstrap class runs the socket server.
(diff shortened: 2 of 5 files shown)


bootstraponline:
```
$ mocha -t 60000 -R spec /Users/user/woven/appium/test/functional/android/apidemos/find/basics-specs.js


  apidemo - find - basics
    ✓ should find a single element by resource-id with implicit package (1371ms)
    ✓ should find multiple elements by resource-id with implicit package (314ms)
    ✓ should find multiple elements by resource-id  with implicit package even when theres just one (72ms)


  3 passing (21s)
```

jlipps:
:+1: though maybe let's hold off on merging since we're about to publish 1.3.5

bootstraponline:
> Can one of the admins verify this patch?

what does this even mean? who is an `admin` on github? we have `owners` and `collaborators`

jlipps:
I think this is a cloudbees thing @sebv is working on. Seems like when one of us replies to this, then additional builds are run?

jlipps:
added to 1.3.6 (cc @0x1mason)

sebv:
retest this please

0x1mason:
retest this please

0x1mason:
@sebv NullRef during CI.  Filed a ticket.

0x1mason:
Retest this please

0x1mason:
@sebv I don't want to merge this until we get the null ref sorted. Any ideas on what might be the issue?

jlipps:
not sure why that build had an NPE. i checked out your branch and ran the tests, it looks good.

PR #4519, opened 2015-02-10
iOS flick: Support for element flick in the webview

scf2k:
This PR implements flick functionality in the webview when element is specified. Flick starts from the center of the element and finishes at the point calculated using specified offsets. Under the hood the flick() method  is used from the uiauto/mechanic patch.

### DIFF
--- a/lib/devices/ios/ios-controller.js
+++ b/lib/devices/ios/ios-controller.js
@@ -755,9 +755,8 @@ iOSController.clickCoords = function (coords, cb) {
   }
 };
 
-iOSController.clickWebCoords = function (cb) {
-  var coords = this.curWebCoords
-    , wvCmd = "au.getElementsByType('webview')"
+iOSController.translateWebCoords = function (coords, cb) {
+  var wvCmd = "au.getElementsByType('webview')"
     , webviewIndex = this.webContextIndex();
 
   // add static offset for safari in landscape mode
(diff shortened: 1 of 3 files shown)


scf2k:
@jlipps please take a look

scf2k:
@0x1mason please take a look

jlipps:
Hi @scf2k code looks good! A couple things:

1) I'd really love a test of this somehow in our ios webview specs
2) You've got a commit from @sebv in here, maybe if you rebase and force-push it'll get rid of it.

trigger.txt
```diff
@@ -1,2 +1,2 @@
 This is used to test github webhooks.
-1
+0
```
0x1mason: What's this?

sebv:
test this please

0x1mason:
@scf2k Do you think you could write a test for this?

scf2k:
@0x1mason I'm preparing it

0x1mason:
:+1: great stuff, thanks!

scf2k:
Added a test.

While making it I noticed that translateWebCoords (ex clickWebCoords) works differently in safari and test web view app. There is a condition which checks for iOS8 and in this case `serviceBarHeight` is set to 0, 20 otherwise. This is correct for safari. But in webview app it should be always 0 to calculate proper result value.

scf2k:
@0x1mason please review

scf2k:
@0x1mason @jlipps sorry I have to ping you :D

jlipps:
All the code looks good to me, thanks @scf2k! @0x1mason is running the current release so I'll let him merge it at an appropriate moment.

scf2k:
great thanks! :)

Issue #4414, opened 2015-01-20
Appium 1.3.4 fails to start when using robotium backend on Android Real Device


aluedeke:
Hey,

i am trying to use Appium with the Robotium backend but the session creation always fails with the following log. I am using a Lollipop real device for testing on an linux host. Any ideas how to solve this:

```
info: Client User-Agent string: Apache-HttpClient/4.3.4 (java 1.5)
info: [debug] Using local app from desired caps: /home/aluedeke/Development/source/appium/gyg/src/test/resources/getyourguide-playstore.apk
info: [debug] Creating new appium session 4e9f54e9-9289-408f-9d1c-f7ab41d24a7c
info: [debug] Starting selendroid server
info: [debug] Getting Java version
info: Java version is: 1.8.0_25
info: [debug] Checking whether selendroid is built yet
info: [debug] Selendroid server exists!
info: [debug] Using fast reset? true
info: [debug] Preparing device for session
info: [debug] Checking whether app is actually present
info: [debug] Checking whether adb is present
info: [debug] Using adb from /home/aluedeke/Development/tools/android-sdk-linux/platform-tools/adb
info: Retrieving device
info: [debug] Trying to find a connected android device
info: [debug] Getting connected devices...
info: [debug] executing cmd: /home/aluedeke/Development/tools/android-sdk-linux/platform-tools/adb devices
info: [debug] 1 device(s) connected
info: Found device 07217e2423b790db
info: [debug] Setting device id to 07217e2423b790db
info: [debug] Waiting for device to be ready and to respond to shell commands (timeout = 5)
info: [debug] executing cmd: /home/aluedeke/Development/tools/android-sdk-linux/platform-tools/adb -s 07217e2423b790db wait-for-device
info: [debug] executing cmd: /home/aluedeke/Development/tools/android-sdk-linux/platform-tools/adb -s 07217e2423b790db shell "echo 'ready'"
info: [debug] Starting logcat capture
info: [debug] Checking whether aapt is present
info: [debug] Using aapt from /home/aluedeke/Development/tools/android-sdk-linux/build-tools/21.1.2/aapt
info: [debug] Checking if has internet permission from manifest.
info: [debug] executing cmd: /home/aluedeke/Development/tools/android-sdk-linux/build-tools/21.1.2/aapt dump badging /home/aluedeke/Development/source/appium/gyg/src/test/resources/getyourguide-playstore.apk
[...]

PR #4541, opened 2015-02-16
remove testZip step for selendroid

aluedeke:
fix for #4414


lib/helpers.js
```diff
@@ -41,70 +41,28 @@ exports.copyLocalZip = function (localZipPath, cb) {
...
-  logger.debug("Testing zip archive: " + zipPath);
-  if (exports.isWindows()) {
-    if (fs.existsSync(zipPath)) {
-      logger.debug("Zip archive tested clean");
-      cb(null, true);
+    if (exports.isWindows()) {
+      var zip = new AdmZip(zipPath);
+      zip.extractAllTo(path.dirname(zipPath), true);
+      logger.debug("Unzip successful");
+      cb(null, null);
```
jlipps: same thing here: `cb();`

lib/helpers.js
```diff
@@ -41,70 +41,28 @@ exports.copyLocalZip = function (localZipPath, cb) {
...
+          logger.debug("Unzip successful");
+          cb(null, stderr);
         } else {
-          logger.error("Zip file " + zipPath + " was not valid");
+          logger.error("Unzip threw error " + err);
           logger.error("Stderr: " + stderr);
           logger.error("Stdout: " + stdout);
-          cb("Zip archive did not test successfully, check appium server logs " +
-             "for output", false);
+          cb("Archive could not be unzipped, check appium logs.", null);
```
jlipps: wait, why is all this code duplicated between appium and appium-adb? it should definitely not be.

aluedeke: confused me as well but had no time to investigate deeper. when using selendroid the code in appium-adb is called (if i remember correctly)

jlipps:
Sorry @aluedeke but with a decent-sized change like this I think the right thing to do is refactor this out, probably so that it's in a separate package entirely (that can be used by appium-adb as well as appium). I really don't like having this duplicated code, and I'm not sure how it got that way.

paymand:
I don't think this is necessary anymore. The latest unzip version on Ubuntu (6.0-12ubuntu1.3) has the fix and testZip step doesn't fail anymore.

PR #4550, opened 2015-02-17
Fix issue 4532: Check if plist file exists before parsing

slk9:
Original issue: appium crash while parsing SafariLauncher's not existing plist file

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -1243,6 +1243,10 @@ IOS.prototype.parseLocalizableStrings = function (/* language, stringFile, cb */
   if (!fs.existsSync(strings)) {
     strings = path.resolve(this.args.app, this.args.localizableStringsDir, stringFile);
   }
+  if (!fs.existsSync(strings)) {
+     logger.warn("Plist file does not exist at " + strings);
+     return cb();
+  }
 
   parsePlistFile(strings, function (err, obj) {
     if (err) {


jlipps:
I believe this will cause a bug. In the case where the file doesn't exist, you call back with `null` as err and `null` as the object. Since err won't exist in the called back context, those contexts will proceed to assume the plist object exists and try to operate on it. This will cause a crash.

I think what we want to do here is call back with an actual error, and let the called-back contexts handle what to do with it then.

slk9:
I have updated commit by moving file exists check upper level to IOS.prototype.parsLocalizableStrings.
Otherwise appium crash on parsePlistFile (no logs :().

lib/devices/ios/ios.js
```diff
@@ -43,6 +43,10 @@ var parseXmlPlistFile = function (plistFilename, cb) {
 };
 
 var parsePlistFile = function (plist, cb) {
+  if (!fs.existsSync(plist)) {
```
jlipps: shouldn't we delete this block now?

lib/devices/ios/ios.js
```diff
@@ -1243,6 +1247,10 @@ IOS.prototype.parseLocalizableStrings = function (/* language, stringFile, cb */
   if (!fs.existsSync(strings)) {
     strings = path.resolve(this.args.app, this.args.localizableStringsDir, stringFile);
   }
+  if (!fs.existsSync(strings)) {
+    logger.warn("Plist file does not exists at " + strings);
```
jlipps: "does not exist"

slk9:
Done:
https://github.com/slk9/appium/commit/195daf7836ac87c6f72306ac7fdcaf40754fad15

jlipps:
:+1: will let @0x1mason merge when ready

PR #4554, opened 2015-02-17
Fixed fail to find xcode -- fixes #2884

0x1mason:
This change should provide a more robust way to find xcode.  I'm not sure if the problem is xcode-select or Node's stdout buffering, but this code side steps that altogether.

### DIFF
--- a/lib/devices/ios/xcode.js
+++ b/lib/devices/ios/xcode.js
@@ -4,19 +4,62 @@ var logger = require('../../server/logger.js').get('appium')
   , fs = require('fs')
   , path = require('path')
   , async = require('async')
-  , exec = require('child_process').exec;
+  , exec = require('child_process').exec
+  , helpers = require('../../helpers.js')
+  , env = process.env;
 
+var XCODE_SUBDIR = "/Contents/Developer";
 var xcode = {};
 
+var _hasExpectedSubDir = function (path) {
+  return path.substring(path.length - XCODE_SUBDIR.length) === XCODE_SUBDIR;
+};
+
+var _getPathFromSymlink = function (stdout, cb) {
+  // Node's invocation of xcode-select sometimes flakes and returns an empty string.
+  // Not clear why. As a workaround, Appium can reliably deduce the version in use by checking
+  // the locations xcode-select uses to store the selected version's path. This should be 100%
+  // reliable so long as the link locations remain the same. However, since we're relying on
+  // hardcoded paths, this approach will break the next time Apple changes the symlink location.
+  var symlnkPath = "/var/db/xcode_select_link";
+  var legacySymlnkPath = "/usr/share/xcode-select/xcode_dir_path"; //  Xcode < 5.x (?)
+
+  // xcode-select allows users to override its settings with the DEVELOPER_DIR env var,
+  // so check that first
+  if (helpers.notNullEmptyOrUndefined(env.DEVELOPER_DIR)) {
+    var dir = _hasExpectedSubDir(env.DEVELOPER_DIR) ?
+                env.DEVELOPER_DIR :
+                env.DEVELOPER_DIR + XCODE_SUBDIR;
+    cb(null, dir);
+  } else if (fs.existsSync(symlnkPath)) {
+    helpers.readlink(symlnkPath, cb);
+  } else if (fs.existsSync(legacySymlnkPath)) {
+    helpers.readlink(legacySymlnkPath, cb);
+  } else {
+    // We should only get here is we failed to capture xcode-select's stdout and our
+    // other checks failed. Either Apple has moved the symlink to a new location or the user
+    // is using the default install. 99.999% chance it's the latter, so issue a warning
+    // should we ever hit the edge case.
+    logger.warn("xcode-select returned an invalid path: " + stdout +
+                ". No links to alternate versions were detected. Using Xcode's " +
+                "default location.");
+    cb(null, "/Applications/Xcode.app" + XCODE_SUBDIR);
+  }
+};
+
 xcode.getPath = function (cb) {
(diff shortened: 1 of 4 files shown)


0x1mason:
@sebv Is the CI failure a build flake?

sebv:
Yeah looks like npm install failed.

lib/helpers.js
```diff
@@ -361,3 +366,11 @@ exports.notNullOrUndefined = function (val) {
 
   return isDefined;
 };
+
+// reads a symlink
+exports.readlink = function (path, cb) {
+  exec('readlink ' + path, function(err, stdout) {
```
jlipps: I think this could be written more compactly as:

```js
exec('readlink' + path, cb);
```

lib/helpers.js
```diff
@@ -349,6 +349,11 @@ exports.truncateDecimals = function (number, digits) {
 };
 
 // return true if the the value is not undefined, null, or NaN.
+exports.notNullEmptyOrUndefined = function (val) {
```
jlipps: this name is so long I will likely never use it. what about `exports.valuey` or `exports.hasValue`?

lib/devices/ios/xcode.js
```diff
@@ -29,18 +72,26 @@ xcode.getPath = function (cb) {
 
 xcode.getVersion = function (cb) {
   xcode.getPath(function (err, xcodePath) {
-    if (err  !== null) {
-      cb(err, null);
+    if (err === null && !_hasExpectedSubDir(xcodePath)) {
+      err = new Error("Expected a path ending with '" + XCODE_SUBDIR + "', but got " + xcodePath);
+    }
+    if (err !== null) {
+      return cb(err, null);
```
jlipps: since you're returning here, we don't need to put everything after this in an `else` (so we can save indentation space)

lib/helpers.js
0x1mason: yes, you're right

lib/devices/ios/xcode.js
```diff
@@ -29,18 +72,26 @@ xcode.getPath = function (cb) {
...
+      // we want to read the CFBundleShortVersionString from Xcode's plist.
+      // It should be in /[root]/XCode.app/Contents/
+      var plistPath = xcodePath.replace(XCODE_SUBDIR, "/Contents/Info.plist");
+      if (!fs.existsSync(plistPath)) {
+        return cb(new Error("Could not get Xcode version: " + plistPath + " does not exist on disk."), null);
       } else {
-        exec(JSON.stringify(xcodebuildPath) + ' -version', { maxBuffer: 524288, timeout: 3000 }, function (err, stdout) {
+        var cmd = "/usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' " + plistPath;
+        exec(cmd, { maxBuffer: 524288, timeout: 3000 }, function (err, stdout) {
+          if (err !== null) return cb(err, null);
```
jlipps: our idiom is usually `if (err) return cb(err);`

1) in many cases `err` is undefined, not null, when it doesn't exist
2) if we're returning an error, we don't need to return a null value for the success object

lib/helpers.js
0x1mason: hasValue might work.  Will come up with something.

lib/devices/ios/xcode.js
```diff
@@ -4,19 +4,62 @@ var logger = require('../../server/logger.js').get('appium')
...
+    helpers.readlink(legacySymlnkPath, cb);
+  } else {
+    // We should only get here is we failed to capture xcode-select's stdout and our
+    // other checks failed. Either Apple has moved the symlink to a new location or the user
+    // is using the default install. 99.999% chance it's the latter, so issue a warning
+    // should we ever hit the edge case.
+    logger.warn("xcode-select returned an invalid path: " + stdout +
+                ". No links to alternate versions were detected. Using Xcode's " +
+                "default location.");
+    cb(null, "/Applications/Xcode.app" + XCODE_SUBDIR);
```
jlipps: do we not at least want to check that this path exists?

lib/devices/ios/xcode.js
```diff
@@ -4,19 +4,62 @@ var logger = require('../../server/logger.js').get('appium')
...
+  }
+};
+
 xcode.getPath = function (cb) {
-  exec('xcode-select --print-path', { maxBuffer: 524288, timeout: 3000 }, function (err, stdout, stderr) {
+  exec('xcode-select -p', { maxBuffer: 524288, timeout: 3000 }, function (err, stdout, stderr) {
```
jlipps: does the `-p` shortcut work with older xcodes too? for some reason it feels newer to me

lib/devices/ios/xcode.js
```diff
@@ -29,18 +72,26 @@ xcode.getPath = function (cb) {
 
 xcode.getVersion = function (cb) {
   xcode.getPath(function (err, xcodePath) {
-    if (err  !== null) {
-      cb(err, null);
+    if (err === null && !_hasExpectedSubDir(xcodePath)) {
```
jlipps: i would just say `!err` instead of `err === null`

lib/devices/ios/xcode.js
```diff
@@ -29,18 +72,26 @@ xcode.getPath = function (cb) {
...
+      err = new Error("Expected a path ending with '" + XCODE_SUBDIR + "', but got " + xcodePath);
+    }
+    if (err !== null) {
```
jlipps: `if (err) {`

jlipps:
why did you close this one?

0x1mason:
Oh, because the CI failed and I wanted to confirm it was a build flake and
not my changes.

jlipps:
ok i will let you address all the comments here in the new one then

PR #4576, opened 2015-02-21
Updated to node-idevice to v0.1.4.

paymand:


paymand:
@0x1mason: Could we please get this merged?

PR #4594, opened 2015-02-24
New MobileSafari logic bypassing Apple page

sebv:
With slow connection there is a race condition between the initial Apple page and the first page requested, this fixes it by waiting for the Apple page to finish loading before the new session request returns.

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -111,6 +111,7 @@ IOS.prototype.init = function () {
   this.webElementIds = [];
   this.implicitWaitMs = 0;
   this.asyncWaitMs = 0;
+  this.pageLoadMs = 60000;
   this.asyncResponseCb = null;
   this.returnedFromExecuteAtom = {};
   this.executedAtomsCounter = 0;
@@ -1543,7 +1544,7 @@ IOS.prototype.getLatestWebviewContextForTitle = function (titleRegex, cb) {
     if (err) return cb(err);
     var matchingCtx;
     _(contexts).each(function (ctx) {
-      if (ctx.view && (ctx.view.title || "").match(titleRegex)) {
+      if (ctx.view && ctx.view.url !== "about:blank" && (ctx.view.title || "").match(titleRegex)) {
         matchingCtx = ctx;
       }
     });
@@ -1551,17 +1552,55 @@ IOS.prototype.getLatestWebviewContextForTitle = function (titleRegex, cb) {
   });
 };
 
+// Right now we don't necessarily wait for webview
+// and frame to load, which leads to race conditions and flakiness
+// , let see if we can transition to something better
+IOS.prototype.useNewSafari = function () {
+  return this.iOSSDKVersion === '8.1' && this.args.platformVersion === '8.1' && !this.args.udid && this.capabilities.safari;
+};
+
 IOS.prototype.navToInitialWebview = function (cb) {
   var timeout = 0;
   if (this.args.udid) timeout = parseInt(this.iOSSDKVersion, 10) >= 8 ? 4000 : 6000;
   if (timeout > 0) logger.debug('Waiting for ' + timeout + ' ms before navigating to view.');
   setTimeout(function () {
-    if (parseInt(this.iOSSDKVersion, 10) >= 7 && !this.args.udid && this.capabilities.safari) {
+    if (this.useNewSafari()) {
+      return this.typeAndNavToUrl(cb);
+    } else if (parseInt(this.iOSSDKVersion, 10) >= 7 && !this.args.udid && this.capabilities.safari) {
       this.navToViewThroughFavorites(cb);
     } else {
       this.navToViewWithTitle(/.*/, cb);
     }
   }.bind(this), timeout);
+};
+
+IOS.prototype.typeAndNavToUrl = function (cb) {
+  var initialUrl = this.args.safariInitialUrl || 'http://127.0.0.1:' + this.args.port + '/welcome';
+  var oldImpWait = this.implicitWaitMs;
+  this.implicitWaitMs = 7000;
+  function noArgsCb(cb) { return function (err) { cb(err); }; }
+  async.waterfall([
+    this.findElement.bind(this, 'name', 'URL'),
+    function (res, cb) {
+      this.implicitWaitMs = oldImpWait;
+      this.nativeTap(res.value.ELEMENT, noArgsCb(cb));
+    }.bind(this),
+    this.findElements.bind(this, 'name', 'Address'),
+    function (res, cb) {
+      var addressEl = res.value[res.value.length -1].ELEMENT;
+      this.setValueImmediate(addressEl, initialUrl, noArgsCb(cb));
+    }.bind(this),
+    this.findElement.bind(this, 'name', 'go'),
+    function (res, cb) {
+      this.nativeTap(res.value.ELEMENT, noArgsCb(cb));
+    }.bind(this)
+  ], function () {
+    this.navToViewWithTitle(/.*/i, function (err) {
+      if (err) return cb(err);
+      // Waits for page to finish loading.
+      this.remote.pageUnload(cb);
+    }.bind(this));
+  }.bind(this));
 };
 
 IOS.prototype.navToViewThroughFavorites = function (cb) {
--- a/lib/devices/ios/remote-debugger.js
+++ b/lib/devices/ios/remote-debugger.js
@@ -12,8 +12,8 @@ var net = require('net')
   , bufferpack = require('bufferpack')
   , uuid = require('node-uuid')
   , noop = function () {}
-  , assert = require('assert');
-
+  , assert = require('assert')
+  , Args = require("vargs").Constructor;
 // ====================================
 // CONFIG
 // ====================================
@@ -65,6 +65,15 @@ RemoteDebugger.prototype.init = function (debuggerType, onDisconnect) {
     }
   };
 
+};
+
+// temporarily using separate method to avoid conflict with webkit remote
+// debugger
+RemoteDebugger.prototype.configureExtraOpts = function (opts) {
+  opts = opts || {};
+  this.useNewSafari = opts.useNewSafari || false;
+  this.pageLoadMs = opts.pageLoadMs;
+  this.logger.debug('useNewSafari --> ' + this.useNewSafari);
 };
 
 // ====================================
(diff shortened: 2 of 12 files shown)


sebv:
@jlipps can you have look, cc @0x1mason

0x1mason:
LGTM

jlipps:
hmm, I thought we didn't _want_ to wait for the apple page to load? just for the webview to become active so we can start using it

0x1mason:
Do we stop the load before redirecting?  Could that achieve the same thing?

sebv:
We don't wait right know and it doesn't work, it makes little difference with fast connections.

The only thing I havn't tested in 8.0, 7.1, maybe it should be 8.1 specific.

@0x1mason I don't think the load to be stopped.

jlipps:
I guess my understanding of this situation may have been off; it was my understanding that all we had to do was tap the apple button to get a webview context that we could use---and we don't care about loading anything in that webview context even though safari is trying to.

My concern with this is it puts a dependency on loading an entire Apple website on every single Appium session anyone runs. Is your external internet flaking? Oh well, there goes your test!

jlipps:
If this change is really necessary, perhaps it should be hidden under a server flag or desired cap.

sebv:
Basically you have a webview with async stuff going on and we don't spin so there is a risk. With fast connection the page loading is quicker than a round-trip to the user so we are safe. (The concrete condition is for the 'apple.com' remote debugger messages to be received before the next `url` request starts processing.)

This happens for me when I enable SC and the whole internet  is tunneled back through my upstream  connection.

jlipps:
I wonder if there's a different thing that we can spin for than page loaded. Maybe we could check that `document.readyState` is 'loading' or 'interactive' instead of 'complete' for this case

sebv:
Won't make a difference, see the log: https://gist.github.com/sebv/c81e8d71ca7d6c202d46 the loaded interactive and interactive states are premature.

jlipps:
@sebv I don't see how that shows they are 'premature'. What I still don't understand is why we have to wait for the page to be fully loaded in order to control the webview. All we care about is that the webview is ready, not that the page is loaded. We'd prefer not to load a page at all!

sebv:
See the here, this is the error, case:
https://sebv.dev.saucelabs.net/tests/5ecd7433e6c24712a2b7b96c7c33d33e

Look at the end of the log for `readyState was complete`, this is the readyState for the Apple page.

sebv:
Can compare with this (with the fix): https://sebv.dev.saucelabs.net/tests/a937f3a1ffe049d5bc77fe71b5d9aeb5

jlipps:
It looks like what's happening in this case is that we fire off the URL request and then wait for the frame navigation message. When we receive that message, we assume the frame is navigating to the URL we requested. In reality, it has just navigated to the original URL (i.e., the webview is lagging behind and is just now getting around to navigating to apple.com). 

This still feels like something we can fix by say adding a check to the frame navigation code that ensures we navigated to the URL we indicated.

Or, what about 'priming the pump': adding a function like this:

```js
RemoteDebugger.prototype.navToUrlWithoutWaiting = function (url, cb) {
  var navToUrl = messages.setUrl(url, this.appIdKey, this.connId,
      this.senderId, this.pageIdKey, this.debuggerType);
  this.send(navToUrl, noop);
  setTimeout(function () {
    this.waitForFrameNavigated(cb);
  }.bind(this), 1000);
}
```

Then in ios.js we simply call this on any url. When we know that we've navigated (but without having had to wait), at that point we should be able to respond with the session id to the client.

I'm trying to think of any way to not have to wait on a web page to load for every Appium safari session. That is ridiculous.

sebv:
Ok I've figured out a way to skip the Apple page, will use a local welcome page instead, can also add a cap to go direct to the url.

```js
IOS.prototype.navToViewThroughFavorites = function (cb) {
  this.implicitWaitMs = 7000;
  function noArgsCb(cb) { return function(err) { cb(err); }; }
  var addressEl;
  async.waterfall([
    this.findElement.bind(this, 'name', 'URL'),
    function (res, cb) {
      this.nativeTap(res.value.ELEMENT, noArgsCb(cb));
    }.bind(this),
    this.findElements.bind(this, 'name', 'Address'),
    function (res, cb) {
      addressEl = res.value[1].ELEMENT;
      this.setValueImmediate(addressEl, 'http://127.0.0.1:4723/test/guinea-pig2.html', noArgsCb(cb));
    }.bind(this),
    this.findElement.bind(this, 'name', 'go'),
    function (res, cb) {
      this.nativeTap(res.value.ELEMENT, noArgsCb(cb));
    }.bind(this),
    function(cb) {
      setTimeout(cb, 3000);
    }
  ], function () {
    this.navToViewWithTitle(/another/i, function (err) {
      if (err) return cb(err);
      // Waits for the page to finish loading.
      this.remote.pageUnload(cb);
    }.bind(this));
  }.bind(this));
};
```

jlipps:
hooray!! :+1:

sebv:
Ready for review, people gonna love it, only enabled 8.1 for the moment, I suspect 8.0 and 7.1 will also work, but need to be tested.

jlipps:
LGTM

sebv:
@jlipps been tracking down more issues. They were down to these 2 timeouts : 
- A: https://github.com/appium/appium/blob/master/lib/devices/ios/remote-debugger.js#L371
- B: https://github.com/appium/appium/blob/master/lib/devices/ios/remote-debugger.js#L371

This is what I understand is happening:

Case 1 - fast requests:  
1 - navigate to apple page without waiting for page load 
2 - request new url (e.g. www.wikipedia.org)
3 - frame load notifications arrives from apple and wikipedia. They are all ignored because of A
4 - After 1 sec, start waiting aimlessly for wikipedia frames (arrived long ago).
5 - B trigger after 500ms, and trigger wikipedia frame load cbs.
Conclusion: It works, but it follows side use case, and between 3,4,5 we loose about 1.2 secs.

Case 2 - slow requests
1 - navigate to apple page without waiting for page load 
2 - request new url (e.g. www.wikipedia.org)
3 - frame load notifications arrives from apple they are ignored
4 - After 1 sec, start waiting for wikipedia frames, apple frame might still arive, wikipedia never arrive
(assuming it takes more than 1.5 secs)
5 - B trigger after 500ms, and trigger wikipedia frame load cbs, but we are still on the Apple page.
Conclusion: if the user is not spinning, it would look like the page is not loading.

I am going to push a version with some debug, and alternate timeout config for 8.1, so we understand more what is really happening.

jlipps:
Sebv I think you nailed it in terms of what's happening in both cases. I have no idea why we statically wait for 1000ms and 500ms respectively. Clearly it's a source of bugs and we should think about it more carefully. Also whoever wrote this failed to comment why we are doing this, which makes it hard for us in that situation.

Given that I wrote most of that early code, it's probably my fault, so I apologize. Something we should all think about from now on.

lib/devices/ios/ios.js
```diff
@@ -1551,19 +1551,57 @@ IOS.prototype.getLatestWebviewContextForTitle = function (titleRegex, cb) {
...
   var timeout = 0;
   if (this.args.udid) timeout = parseInt(this.iOSSDKVersion, 10) >= 8 ? 4000 : 6000;
   if (timeout > 0) logger.debug('Waiting for ' + timeout + ' ms before navigating to view.');
   setTimeout(function () {
-    if (parseInt(this.iOSSDKVersion, 10) >= 7 && !this.args.udid && this.capabilities.safari) {
-      this.navToViewThroughFavorites(cb);
+    if (this.useConservativePageLoadTimeouts()) {
+      return this.typeAndNavToUrl(cb);
+    } else if (parseInt(this.iOSSDKVersion, 10) >= 7 && !this.args.udid && this.capabilities.safari) {
+        this.navToViewThroughFavorites(cb);
```
jlipps: spacing here

sebv:
Commit for the 1000ms timeout: https://github.com/appium/appium/commit/e1040816b7e8a6e4366b263d4ab5ad692c5831c2

sebv:
B: https://github.com/appium/appium/commit/4b76ed0478163a1437593a6f8cf21555ddb127a8

B is more or less equivalent to the JsonWire PageLoadTimeout.

lib/devices/ios/remote-debugger.js
```diff
@@ -449,8 +459,21 @@ RemoteDebugger.prototype.waitForDom = function (cb) {
...
+    var args = [].slice.call(arguments);
+    _cb.apply(null, args);
+  }.bind(this);
+
   this.frameNavigatedCbs.push(cb);
-  this.navigatingTimeout = setTimeout(this.frameNavigated.bind(this), 500);
+  this.navigatingTimeout = setTimeout(function () {
+     this.logger.debug('frame navigated timeout triggered');
+     this.frameNavigated();
+  }.bind(this), this.useConservativePageLoadTimeouts ? 30000 : 500);
```
jlipps: wait so does this introduce a 30s delay on ios 8.1 before checking if the frame navigated?

sebv: No, the delay is the other timeout (line 371), this is the max time we wait. 

We should probably properly implement pageLoadTimeout properly, to be consistent with Selenium.

jlipps: I see so the frame can navigate itself, but if it doesn't, we do it 30s later

devquest3:
Hi, is this fixed? I cannot use appium for ios testing because of this issue. My organisation blocks external websites and since appium waits for apple website to load initially, cannot go any further.

jlipps:
this is fixed in master, yes. you'll need to wait for the next release if you don't want to run from source.

devquest3:
Great news. Thanks!!

devquest3:
Will the fix be available for android as well? Google website loads initially on android.

jlipps:
this fix is not for android. and when you use a chromedriver session, google loads up chrome without an initial url. not sure why you are seeing google load. maybe you're trying to automate chrome as an app rather than using our built-in chrome browser support?

devquest3:
Is it on a real device you are talking about? I am using the default android browser app on android emulator. I wanted to try Chrome as well, but looks like it is not possible to install a licensed chrome app on emulator at this point, can we? 
Also, quick question not related to this discussion - Using Appium, can we launch safari/android browsers on ios/android simulator/emulator with the desired settings like enabling/disabling javascript etc., at this point?

PR #4600, opened 2015-02-25
Re-install Selendroid server only on file changes.

paymand:
Currently, we look at the timestamp (mtime) of the Selendroid server
file to determine if we need to re-install it. Since we are creating it
every time, this means that we re-install it every single time!
This fix checks the MD5 hash of the server file, so that we re-install
only when the binary is different.

### DIFF
--- a/lib/devices/android/selendroid.js
+++ b/lib/devices/android/selendroid.js
@@ -15,6 +15,7 @@ var Device = require('../device.js')
   , androidCommon = require('./android-common.js')
   , androidHybrid = require('./android-hybrid.js')
   , path = require('path')
+  , md5 = require('md5calculator')
   , utf7 = require('utf7').imap;
 
 var Selendroid = function () {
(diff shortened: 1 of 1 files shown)


sebv:
when the hash change, you don't need to resave it?

lib/devices/android/selendroid.js
```diff
@@ -101,21 +102,29 @@ Selendroid.prototype.start = function (cb) {
...
   }.bind(this);
 
   var checkServerResigned = function (cb) {
     if (modServerExists) {
-      fs.stat(this.selendroidServerPath, function (err, stat) {
-        if (stat && stat.mtime.getTime() > modServerTimestamp) {
-          modServerExists = false;
-        }
+      if (!resignedServerMd5Hash) {
+        md5(this.selendroidServerPath, function (err, md5Hash) {
```
sebv: You can just do that in one place.

sebv:
test this please

sebv:
ok to test

sebv:
add to whitelist

lib/devices/android/selendroid.js
paymand: Fixed.

sebv:
retest this please

sebv:
retest this please

paymand:
@0x1mason: Could we please get this merged?

PR #4672, opened 2015-03-09
Add test for tapping an absolute position

imurchie:
In the course of working on an issue, I needed to write this. It's also an entirely different bit of code running server side than tapping an element or an element with an offset.


0x1mason:
lgtm

PR #4673, opened 2015-03-09
Fix reporting of url in logs

imurchie:
A few places where we get `[object Object]` in the logs, rather than something useful (in this case, the url from which we are downloading the app).


jlipps:
Interesting, I think that used to work. Are we using a new url object or something? Anyway, changes look good. Please merge when ready @0x1mason

PR #4679, opened 2015-03-09
Wrong helpers

imurchie:
Wrong helpers being accessed, so `safely` ends up undefined. Stops master from working. @0x1mason


jlipps:
not sure why the build failed. code looks good.

PR #4693, opened 2015-03-11
Fix for git pre-commit hook fail on Android.

paymand:
iOS-specific tests should only be run on iOS.

Committing on Android fails since the pre-commit hooks fails. Bug was introduced here: https://github.com/appium/appium/commit/54f9b3af226374329fbf0bec903de6fbad4550a6


paymand:
@0x1mason

0x1mason:
I'll commit my original changes that handle this.

paymand:
@0x1mason: Sure. Whatever works :)

PR #4697, opened 2015-03-11
Move Xcode tests to iOS functional tests

0x1mason:


0x1mason:
@sebv @paymand

paymand:
:+1:

paymand:
> Added logic to skip Xcode tests on non-OSX machines

What logic have you added here? To me it looks like you just moved the test from unit tests to iOS functests (which is the correct thing to do, but I guess the commit message should be updated).

0x1mason:
@paymand Good pt about the commit msg.  Will update.

0x1mason:
@paymand changed to "Fix for commit hook failure on non-OSX machines: moved Xcode path tests to functional tests"

paymand:
Very loooooong, but much better ;)

0x1mason:
hehe :+1:

paymand:
is what I think was enough: ^

PR #4712, opened 2015-03-13
Fix hasValue undefined

0x1mason:
Regression broke swipe tests.


imurchie:
https://github.com/appium/appium/commit/ab8f6e2f2b85dea3cc2ff0f05ab16a45dbdf5cd4

jlipps:
weird that this doesn't show a merge conflict, guess cause it's exactly the same change?

PR #4726, opened 2015-03-17
Update changelog.txt for 1.3.7-Beta

0x1mason:


jlipps:
i think we should just do the changelog when we make a release

Issue #4727, opened 2015-03-17
Issue with cookie encoding


SBoudrias:
Appium is automatically url encoding cookies values. This is inconsistent with other drivers and is causing issues when using generic selenium tools.

I could find this test that isn't clear about why this is happening: https://github.com/appium/appium/blob/23d1a76ed2dd52f0e0929d50dad9bfcaf232c51d/test/functional/common/webview/cookies-base.js#L27-L34 (There's a _todo_ about maybe removing, and a note about an android bug)

Just hoping to relaunch the discussion to figure if this is really necessary?

I feel like this might be something we want to relegate to the user to stay consistent with the other drivers.

SBoudrias:
Closing as the fix been merged.

PR #4729, opened 2015-03-18
Don't url encode cookie value

SBoudrias:
This change is to ensure we stay consistent with other selenium driver
implementation.

This is a follow PR on #4727

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -19,10 +19,6 @@ module.exports = function (cookieString) {
     return decodeURIComponent(s);
   }
 
-  function stringifyCookieValue(value) {
-    return encode(String(value));
-  }
-
   function parseCookieValue(s) {
     if (s.indexOf('"') === 0) {
       // This is a quoted cookie as according to RFC2068, unescape...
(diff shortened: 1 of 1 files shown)


SBoudrias:
Hey, any feedback on this change? We can also discuss it further on the issue tracker.

jlipps:
hi @SBoudrias i'm not sure what's at stake here. can @sebv chime in please?

SBoudrias:
On my side, the issue is that we'll run test in both desktop browsers and mobile browsers. So having a consistent behavior about how cookies are handled in both selenium drivers is very important.

jlipps:
sure i believe you, just wondering why we did this in the first place

SBoudrias:
It seems to me jquery.cookie encode the cookies, so the code encoding them have just been kept when porting the module to appium https://github.com/carhartl/jquery-cookie/blob/master/src/jquery.cookie.js#L23-L25

jlipps:
ok, merged. thank you!

SBoudrias:
Awesome, thanks!

PR #4752, opened 2015-03-23
create sessionRemain server flag and controller-specs unit test

VinceBarresi:
I have added a server flag called 'sessionRemain'. This allows anyone who sets it to instantiate multiple driver objects and switch between them in a single test without deleteSession killing the original session. This idea came about due to a use case at my company which involves multiple iOS devices communicating with each other in a single test. Without this inclusion, we simply cannot switch between iOS devices in the same test. I have also added a file called 'controller-specs.js' which includes the unit test for the change to the deleteSession function. I did not know exactly where in the project to place this file so as of now its just in the test/unit directory.

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -40,6 +40,7 @@ var Appium = function (args) {
   this.session = null;
   this.preLaunched = false;
   this.sessionOverride = this.args.sessionOverride;
+  this.sessionRemain = this.args.sessionRemain;
   this.resetting = false;
   this.defCommandTimeoutMs = this.args.defaultCommandTimeout * 1000;
   this.commandTimeoutMs = this.defCommandTimeoutMs;
--- a/lib/server/controller.js
+++ b/lib/server/controller.js
@@ -241,7 +241,13 @@ exports.background = function (req, res) {
 };
 
 exports.deleteSession = function (req, res) {
-  req.appium.stop(getResponseHandler(req, res));
+  if (req.appium.sessionRemain === true) {
+    safely(req, function () {
+      res.status(200).send("sessionRemain flag is set - Not killing session.");
+    });
+  } else {
+      req.appium.stop(getResponseHandler(req, res));
+  }
 };
 
 exports.equalsElement = function (req, res) {
--- a/lib/server/parser.js
+++ b/lib/server/parser.js
@@ -111,6 +111,17 @@ var args = [
   , nargs: 0
   }],
 
+  [['--session-remain'], {
+    defaultValue: false
+  , dest: 'sessionRemain'
+  , action: 'storeTrue'
+  , required: false
+  , help: 'Allows creation of a new session without deleting ' +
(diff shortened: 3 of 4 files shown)


jlipps:
not sure why you need a change in appium for this. what about just running different appium servers and creating different clients, one for each server?

VinceBarresi:
Hi Jonathan, we are using two different servers and we have one device delegated to each of the two. The issue is, when we switch between iOS device one to iOS device two within our test, the first session is getting killed.

jlipps:
Still not understanding here. What do you mean by "switch between iOS device on to iOS device two"? This should work (pseudocode):

```
d1 = remote.init('<appium-server-1>', caps1)
d2 = remote.init('<appium-server-2>', caps2)
# now I can do things with d1 and d2 however I want
```

VinceBarresi:
I apologize if I'm not being clear. So we're basically doing something similar to what you've provided (see code below). What I mean by switching between devices is within our test, if we set the current driver to driver2 for example, driver1 is getting killed.  We need to be able to switch between both devices and have them both stay in the state they were in prior to switching in each step in a single test. We have not been able to accomplish this otherwise. 

  Capybara.register_driver(:driver1) do |app|
    appium_lib_options = {
        server_url:     "http://127.0.0.1:4723/wd/hub"
    }
    all_options = {
        appium_lib:  appium_lib_options,
        caps:        desired_caps_ios_device1
    }
    Appium::Capybara::Driver.new(app, all_options)
 end
end

def register_appium_ios_device2
  Capybara.register_driver(:driver2) do |app|
    appium_lib_options = {
        server_url:     "http://10.59.2.114:4733/wd/hub"
    }
    all_options = {
        appium_lib:  appium_lib_options,
        caps:        desired_caps_ios_device2
    }
    Appium::Capybara::Driver.new(app, all_options)
  end
end

jlipps:
what does "set the current driver to driver2" look like in code? I'm wondering if this is a capybara issue.

VinceBarresi:
It looks like this: Capybara.current_driver = :driver2

jlipps:
Can you send in the logs for the first serve? I'm wondering if Capybara quits sessions when you switch `Capybara.current_driver`

VinceBarresi:
Sure. Here is a pastebin link: http://pastebin.com/gDVFY8Dv . I didn't want to paste the log on here as its fairly large. This ins't the full log, rather the contents from the last action on device1, to when device1 is killed.

bootstraponline:
The issue is the Ruby bindings quit the old session when a new session is started due to using a global driver. This is a great use case for eliminating the global driver in the [appium Ruby bindings](https://github.com/appium/ruby_lib/issues/312).

I'm closing the issue here as it's a Ruby specific issue and not a problem in the server.

Appium isn't working at all on my computer right now so I don't expect to get to this soon. Happy to review a PR if you'd like to submit a fix.

PR #4766, opened 2015-03-24
Fix #4763. Parse java version even when it does not come in the first line

### LINKED ISSUE #4763: Can't find Java version
Version 1.3.6

```
error: Failed to start an Appium session, err was: Error: Could not get the Java version. Is Java installed?
info: [debug] Error: Could not get the Java version. Is Java installed?
    at /Users/filipesperandio/petrofeed/mobile-smoke-tests/node_modules/appium/lib/devices/android/android-common.js:1040:17
    at ChildProcess.exithandler (child_process.js:646:7)
    at ChildProcess.emit (events.js:98:17)
    at maybeClose (child_process.js:756:16)
    at Socket.<anonymous> (child_process.js:969:11)
    at Socket.emit (events.js:95:17)
    at Pipe.close (net.js:465:12)
info: [debug] Responding to client with error: {"status":33,"value":{"message":"A new session could not be created. (Original error: Could not get the Java version. Is Java installed?)","origValue":"Could not get the Java version. Is Java installed?"},"sessionId":null}
info: <-- POST /wd/hub/session 500 175.016 ms - 222
^Cmake: *** [appium-debug] Error 130
```

Version 1.3.3 works with no Java problem with the same setup.

filipesperandio:
...of the command output


lib/devices/android/android-common.js
```diff
@@ -1025,16 +1025,25 @@ androidCommon.hideKeyboard = function () {
   }.bind(this));
 };
 
+androidCommon.parseJavaVersion = function (stderr) {
+  var lines = stderr.split("\n");
+  for (var i = 0; i < lines.length; ++i) {
```
jlipps: we do `i++`

lib/devices/android/android-common.js
```diff
@@ -1025,16 +1025,25 @@ androidCommon.hideKeyboard = function () {
...
+    var line = lines[i];
+    if (new RegExp("java version").test(line)) {
+      return line.split(" ")[2].replace(/"/g, '');
+    }
+  }
+  return null;
+};
+
 androidCommon.getJavaVersion = function (cb) {
+  var self = this;
```
jlipps: instead of this pattern, we use `.bind(this)` on the enclosing function

lib/devices/android/android-common.js
```diff
@@ -1025,16 +1025,25 @@ androidCommon.hideKeyboard = function () {
...
   exec("java -version", function (err, stdout, stderr) {
     var javaVersion = null;
     if (err) {
       return cb(new Error("'java -version' failed. " + err));
     } else if (stderr) {
-      var firstLine  = stderr.split("\n")[0];
-      if (new RegExp("java version").test(firstLine)) {
-        javaVersion = firstLine.split(" ")[2].replace(/"/g, '');
-      }
+      javaVersion = self.parseJavaVersion(stderr);
```
jlipps: so that `self` can be `this` here

test/unit/android-common-specs.js
```diff
@@ -0,0 +1,49 @@
...
+
+chai.should();
+
+var java_version_stderr = 'Picked up _JAVA_OPTIONS: -Djava.awt.headless=true\njava version "1.8.0_22"';
+
+var fakeExec = function (_, fn) {
+  fn(null, null, java_version_stderr);
+};
+
+var _test_double = function () {
```
jlipps: love the test! can we use camelCase throughout, though?

jlipps:
made a few small comments, otherwise :+1:

paymand:
:+1: for the fix. I didn't know that java version could end up on a line other than first.

lib/devices/android/android-common.js
filipesperandio: I usually do it too, but saw many other places in the code base doing so...

test/unit/android-common-specs.js
filipesperandio: too much ruby these days ;) fixing it.

filipesperandio: BTW, we can add a .jscsrc rule to enforce this, but currently it break too many places:
`"requireCamelCaseOrUpperCaseIdentifiers": true`

filipesperandio:
Adjustments done.

paymand:
Also please squash all commits into one when you're done.

filipesperandio:
Squashed.

jlipps:
awesome, thanks for the PR!

PR #4768, opened 2015-03-24
moved getAutomationTraceTemplate function from ios.js to xcode.js

Jonahss:
added a test. ooh tests!


jlipps:
unit tests failed because they can't run on linux; @sebv do we have this set up right?

@Jonahss this can all be one commit

lib/devices/ios/ios.js
```diff
@@ -340,23 +346,21 @@ IOS.prototype.startInstruments = function (cb) {
...
     justLoopInfinitely: this.shouldIgnoreInstrumentsExit(),
     autoAcceptAlerts: !(!this.args.autoAcceptAlerts || this.args.autoAcceptAlerts === 'false'),
     autoDismissAlerts: !(!this.args.autoDismissAlerts || this.args.autoDismissAlerts === 'false'),
     sendKeyStrategy: this.args.sendKeyStrategy || (this.args.udid ? 'grouped' : 'oneByOne')
-  }).then(
-    function (bootstrapPath) {
+  });
+
+  Q.all([bootstrap, this.automationTraceTemplatePath])
+  .spread(
```
jlipps: if we do chaining, let's line up periods

lib/devices/ios/ios.js
```diff
@@ -340,23 +346,21 @@ IOS.prototype.startInstruments = function (cb) {
...
+    function (bootstrapPath, automationTraceTemplatePath) {
```
jlipps: can you move this up a line so it's like `.spread(function (...`

lib/devices/ios/ios.js
```diff
@@ -376,8 +380,7 @@ IOS.prototype.makeInstruments = function (cb) {
       , traceDir: this.args.traceDir
       });
       cb(null, instruments);
-    }.bind(this), function (err) { cb(err); }
-  );
+    }.bind(this), cb ).fail(cb);
```
jlipps: extra space after `cb`.

lib/devices/ios/ios.js
```diff
@@ -1190,8 +1148,16 @@ IOS.prototype.prelaunchSimulator = function (cb) {
...
+  xcode.getPath(function (err, xcodePath) {
+    if (err) {
+      return cb(new Error('Could not find xcode folder. Needed to start simulator. ' + err.message));
+    }
+    xcodeFolder = xcodePath;
+  });
+
   logger.debug("Pre-launching simulator");
-  var iosSimPath = path.resolve(this.xcodeFolder,
+  var iosSimPath = path.resolve(xcodeFolder,
```
jlipps: xcodeFolder will almost always be incorrectly `null` here, right? since `xcode.getPath` is async?

test/unit/xcode-specs.js
```diff
@@ -0,0 +1,26 @@
...
+  it('should find the automation trace template', function (done) {
+    xcode.getAutomationTraceTemplatePath()
+    .then(function (traceTemplatePath) {
+      fileExists(traceTemplatePath).should.eventually.be.true;
+      var suffix = ".tracetemplate";
+      traceTemplatePath.slice(-suffix.length).should.equal(suffix);
+    })
+    .nodeify(done);
+  });
+
```
jlipps: can we add a test that we get an error when it can't find the tracetemplate? and in both these cases we could just mock xcode.getPath to return without actually needing to call xcode-select

jlipps:
made some comments, otherwise :+1:

lib/devices/ios/ios.js
Jonahss: Aye, that's a bad mistake. Correcting...

lib/devices/ios/ios.js
```diff
@@ -155,6 +157,12 @@ IOS.prototype.configure = function (args, caps, cb) {
     return cb(new Error(msg));
   }
 
+  if (!this.args.automationTraceTemplatePath) {
```
Jonahss: Turns out this block should be in configureIOS so that Safari.js gets it too.

Jonahss:
Added a test. Did change some stuff. Want to re-review @jlipps ?

jlipps:
@appium-ci retest this please

Jonahss:
will commit this later as part of a larger PR

PR #4778, opened 2015-03-25
add Xcode 6 iphonesim sdk location

0x1mason:
I reinstalled Xcode and discovered that Xcode 6.2 has changed the iphonesim sdk location.  Was getting a false positive because had existing sdks, I guess.

### DIFF
--- a/bin/publish.sh
+++ b/bin/publish.sh
@@ -8,7 +8,8 @@ else
 fi
 
 XCODE_PATH=$(xcode-select -print-path)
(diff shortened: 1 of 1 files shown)


0x1mason:
@jlipps @imurchie Please review.

imurchie:
On a related note: https://github.com/appium/appium/commit/695b7cd81916a89b054d70072694604f066ee0f0

0x1mason:
Ok, I had remembered you making a related change, but wasn't sure where it was.  Good to know as a sanity check (since I fail them regularly ;) )

imurchie:
I will open a PR with that commit. It should get in anyway.

jlipps:
:+1: we should probably update this to xcode 8.2 at some point as well

Issue #4708, opened 2015-03-13
Cannot test the old SDK with Xcode 6.2


bigfei:
Since the Xcode 6.2 is officially released in this Tuesday, I did a test for our app using appium. And find that the Xcode 6.2 change its old SDKs from "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator8.1.sdk/Applications" into "./Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 8.1.simruntime/Contents/Resources/RuntimeRoot/Applications/", so that I cannot do the Safari testing using ios Simulator anymore. 

Possible fix is on the Simulator.prototype.getBuiltInAppDir function with special treat for iOS SDK 8.2..

Damn, Apple sucks...

imurchie:
You are meaning, in order to use the iOS 8.1 simulator in Xcode 6.2?

jlipps:
interesting that this wouldn't have come across your radar @imurchie during testing

imurchie:
It _is_ interesting. Preliminary tests on my machine allow me to run 8.1, but not 7.1. Strange.

imurchie:
I've pushed [a commit](https://github.com/appium/appium/commit/695b7cd81916a89b054d70072694604f066ee0f0) that addresses this. But since on my machine 8.1 is found without problem, I can't really test it. 

It works, somewhat, for 7.1, though xcode 6.2 apparently [has a problem](http://stackoverflow.com/questions/28976694/unable-to-launch-ios7-1-simulator-in-xcode-6-2-instruments) with instrumenting iOS 7.1.

bigfei:
@imurchie Yes, testing the old SDKs Xcode 6.2, like 8.1/7.1. I will do a test for your patch if it works for our needs. :ok_hand:

imurchie:
Please let me know how it goes.

NozomiIto:
Hi.

I encountered the same problem in the following environment:

- appium 1.3.6
- Xcode 6.2
- simulator ios version 7.1
- simulator device iPhone 5

I manually fixed grunt-helpers.js and lib/devices/ios/simulator.js
according to this [commit](https://github.com/appium/appium/commit/695b7cd81916a89b054d70072694604f066ee0f0) and tried Safari test on simulator.

I found simulator for ios7.1 successfully started, but when starting Safari on simulator, the simulator unfortunately crashed with this console output message:

```
[...]

PR #4782, opened 2015-03-25
Check alternate sim location

imurchie:
Resolves partly #4708 (though because of an issue with Apple 7.1 sims in Xcode 6.2 fail (see http://stackoverflow.com/questions/28976694/unable-to-launch-ios7-1-simulator-in-xcode-6-2-instruments)).

### DIFF
--- a/lib/devices/ios/simulator.js
+++ b/lib/devices/ios/simulator.js
@@ -481,15 +481,31 @@ Simulator.prototype.getBuiltInAppDir = function (cb) {
     var appDir = path.resolve(xcodeDir, "Platforms/iPhoneSimulator.platform/" +
                               "Developer/SDKs/iPhoneSimulator" +
                               this.platformVer + ".sdk/Applications/");
-    fs.stat(appDir, function (err, s) {
+    var testDir = function (dir, ncb) {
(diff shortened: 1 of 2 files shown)


0x1mason:
Lgtm

lib/devices/ios/simulator.js
```diff
@@ -481,15 +481,31 @@ Simulator.prototype.getBuiltInAppDir = function (cb) {
...
+      });
+    };
+    testDir(appDir, function (err, dir) {
       if (err) {
-        cb(err);
-      } else if (!s.isDirectory()) {
-        cb(new Error("Could not load built in applications directory"));
+        // try out other location
+        logger.debug('Unable to find app dir ' + appDir);
+        appDir = path.resolve("/Library/Developer/CoreSimulator/" +
```
jlipps: assuming that the difference between this path.resolve and the other one changed in this commit is not significant? (above we escape the space, here we don't)

imurchie: Yes. The first one needs the escape because of the subsequent `chown` that will fail without it. The second is handled fine within Appium itself.

PR #4798, opened 2015-03-26
Add window/:windowhandle/maximize route as noop

SBoudrias:
Mobile browsers are always fullscreen. Make this endpoint a noop.

### DIFF
--- a/lib/server/controller.js
+++ b/lib/server/controller.js
@@ -701,6 +701,11 @@ exports.getWindowSize = function (req, res) {
   req.device.getWindowSize(windowHandle, getResponseHandler(req, res));
 };
 
+exports.maximizeWindow = function (req, res) {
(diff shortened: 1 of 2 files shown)


lib/server/controller.js
```diff
@@ -701,6 +701,11 @@ exports.getWindowSize = function (req, res) {
   req.device.getWindowSize(windowHandle, getResponseHandler(req, res));
 };
 
+exports.maximizeWindow = function (req, res) {
+  // noop
+  res.sendStatus(200);
```
jlipps: probably want to do something like `respondSuccess(req, res);` here instead to take care of jsonwp-ing everything

SBoudrias:
Addressed the review comment.

jlipps:
thanks!

PR #4801, opened 2015-03-27
Add basic test support for iOS 8.3

imurchie:
So that we can run `./bin/test.sh --ios83` and get Appium supporting the latest.


bin/test.sh
```diff
@@ -111,6 +115,10 @@ if $ios82_only || $all_tests; then
     run_ios_tests "8.2" "ios82" "@skip-ios82|@skip-ios81|@skip-ios8|@skip-ios-all|@skip-ios7up"
 fi
 
+if $ios83_only || $all_tests; then
+    run_ios_tests "8.3" "ios83" "@skip-ios83|@skip-ios82|@skip-ios81|@skip-ios8|@skip-ios-all|@skip-ios7up"
```
jlipps: looks good to me, though we probably want to run some of these tests just to see if something actually got fixed!

imurchie: I've run them, and there are relatively few errors. I haven't had time to go through and run the failing tests individually to see if it they are just flakes. I also ran 8.2 and 8.1 tests on Xcode 6.3 beta 4, with similar results. I'll look into the details at some point.

PR #4826, opened 2015-03-31
Check if a non default server argument is passed as a string True / Fals...

sangeethasriniv:
Check if a non default server argument is passed as a string True / False and convert it to boolean

This check is made for the desired server capabilities and not for the non default arguments, causing the checking of non default arguments to fail.

### DIFF
--- a/lib/server/helpers.js
+++ b/lib/server/helpers.js
@@ -191,6 +191,17 @@ var getNonDefaultArgs = function (parser, args) {
   _.each(parser.rawArgs, function (rawArg) {
     var arg = rawArg[1].dest;
     if (args[arg] !== rawArg[1].defaultValue) {
+      var value = args[arg];
+      if (typeof value === "string") {
+       // handle the case where users send in "true" or "false" as a string
+       // instead of a boolean
+       var tempValue = value.trim().toLowerCase();
+       if (tempValue === "false" || tempValue === "true") {
+         logger.warn("Converting cap " + arg  + " from string to boolean. " +
+                    "This might cause unexpected behavior.");
+          args[arg] = (tempValue === "true");
+        }
+      }
       nonDefaults[arg] = args[arg];
     }
   });


jlipps:
It looks like this code was copy/pasted from capabilities.js. I don't think it makes much sense as it stands. What server argument are you starting appium with that's doing the wrong thing at the moment?

sangeethasriniv:
This piece of code is in capabilities.js to convert the string 'true' or 'false' to boolen to convert the desired capabilities. But this check is missing for the non default server arguments , hence copied it over. I can add a helper method instead, before that, wanted to initiate a discussion on this, hence posted a review 

when argument stopAppOnReset is passed as 'false' from command line,(appium --stop-app-on-reset false) it always evaluates to true , as a string is passed down instead of a boolean value. (Refer adb.js line 1090 :  cmd += startAppOptions.stopApp && apiLevel >= 15 ? "-S" : "";  In this line startAppOptions.stopApp always evaluates to True)
With this patch, this flag is converted to boolean and all works well.

I noticed there are only 2 server arguments which has a default value 'true'  autoLaunch and stopAppOnReset. Since all other non default server arguments are by default set in the code as boolean 'false' , if a user passes in a 'true' for these arguments, the string always evaluates to boolean true, hence these arguments are evaluated correct.

jlipps:
in this case i think the misunderstanding is that we should never have added something with a default value of true to parser.js. It just doesn't make sense. We should have added a flag with the opposite case, e.g., `--no-stop-app-on-reset` which defaults to false, so that when it is present, the flag gets set to true.

sangeethasriniv:
This is the previous discussion about adding the flag
https://github.com/appium/appium/pull/4159

jlipps:
@sangeethasriniv yes it was my oversight not to catch the issue in parser before

jlipps:
@sangeethasriniv yes it was my oversight not to catch the issue in parser before

jlipps:
closing in favor of #4907

PR #4843, opened 2015-04-02
For iOS get home path from environment, not by guessing

amake:
In `simulator.js` there are some paths that are resolved by concatenating `"/Users"`, `username`, etc., but this fails when running under users with nonstandard home directories (like Jenkins).

The correct home directory is provided in the environment, so it would be better to use that.


jlipps:
:+1: definitely less brittle, thanks

Issue #4809, opened 2015-03-30
PlistBuddy fails to get Xcode version in some cases [v1.3.7] 


0x1mason:
From: https://discuss.appium.io/t/appium-1-3-7-release/3547

error: Could not determine Xcode version:Command failed: /bin/sh -c /usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' /Applications/ Xcode 6.1.1.app/Contents/Info.plist

error: Could not get the iOS SDK version because the Xcode version could not be determined.

error: Could not get the iOS SDK version because the Xcode version could not be determined.

error: Could not get the iOS SDK version because the Xcode version could not be determined.

error: Could not determine iOS SDK version

info: [debug] Cleaning up appium session

error: Could not pre-launch appium: Error: Could not get the iOS SDK version because the Xcode version could not be determined.

I'vent changed anything in my configurations , and they are equal to the previous configurations I've used with 1.3.6 Appium version

jlipps:
wondering if spaces need to be escaped

0x1mason:
@jlipps My thought exactly

0x1mason:
```js
var ESCAPE_SPACE_RE = /\s+/;
var SPACE_ESCAPE = '\ ';

function escapeSpace(str) {
  return str.split(ESCAPE_SPACE_RE).join(SPACE_ESCAPE);
}
```

0x1mason:
Add escapeSpace to lib/helpers.js. 
Will need a unit test for escapeSpace in test/unit/helper-specs.js.
Change needs to be made in lib/devices/ios/xcode.js. Call escapeSpace on plistPath on line 85.

jlipps:
resolving https://github.com/appium/appium/issues/2041 would also fix this issue (the wrapper would just always do the right thing, so we don't have to remember to escape every time we use exec

wensj:
I have the same problem:
error: Could not get the iOS SDK version because the Xcode version could not be determined.
error: Could not get the iOS SDK version because the Xcode version could not be determined.

info: [debug] Cleaning up appium session
info: [debug] Error: Could not get the iOS SDK version because the Xcode version could not be determined.
    at /Applications/Appium.app/Contents/Resourcesde_modules/appiumb/devices/ios/xcode.js:111:10
    at /Applications/Appium.app/Contents/Resourcesde_modules/appiumb/devices/ios/xcode.js:80:14
[...]

PR #4846, opened 2015-04-02
PlistBuddy fails to get Xcode version when there are spaces in path

tarunsujit:
#4809 added a method to handle spaces in path

### DIFF
--- a/lib/devices/ios/xcode.js
+++ b/lib/devices/ios/xcode.js
@@ -5,7 +5,9 @@ var logger = require('../../server/logger.js').get('appium')
   , async = require('async')
   , exec = require('child_process').exec
   , util = require('appium-support').util
-  , env = process.env;
+  , env = process.env
+  , helpers = require('../../helpers.js')
+  , escapeSpace = helpers.escapeSpace;
 
 var XCODE_SUBDIR = "/Contents/Developer";
 var xcode = {};
@@ -86,7 +88,12 @@ xcode.getVersion = function (cb) {
     if (!fs.existsSync(plistPath)) {
       return cb(new Error("Could not get Xcode version: " + plistPath + " does not exist on disk."), null);
     } else {
-      var cmd = "/usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' " + plistPath;
+      try {
+        var cmd = "/usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' " + 
+                  escapeSpace(plistPath);
+      } catch(e) {
+        return cb(e);
+      }
       exec(cmd, { maxBuffer: 524288, timeout: 3000 }, function (err, stdout) {
         if (err !== null) return cb(err, null);
         var versionPattern = /\d\.\d\.*\d*/;
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -10,7 +10,9 @@ var logger = require('./server/logger.js').get('appium')
   , isWindows = support.system.isWindows
   , isMac = support.system.isMac
   , tempDir = support.tempDir
-  , AdmZip = require('adm-zip');
+  , AdmZip = require('adm-zip')
+  , ESCAPE_SPACE_RE = '\\ '
+  , SPACE = / /;
 
 
 exports.downloadFile = function (fileUrl, suffix, cb) {
@@ -338,4 +340,7 @@ exports.truncateDecimals = function (number, digits) {
   return truncatedNum / multiplier;
 };
 
+exports.escapeSpace = function (str) {
+  return str.split(SPACE).join(ESCAPE_SPACE_RE);
+};  
 
--- a/test/unit/helper-specs.js
+++ b/test/unit/helper-specs.js
@@ -89,4 +89,11 @@ describe("Helpers", function () {
       logger.warn.called.should.equal(false);
     });
   });
+  describe('xcode version', function () {
+    it('parses xcode version with space', function () {
+      var actual = "/Applications/ Xcode 6.1.1.app/Contents/Developer";
+      var expected = "/Applications/\\ Xcode\\ 6.1.1.app/Contents/Developer"
+      helpers.escapeSpace(actual).should.equal(expected);
+    });
+  });
 });


lib/helpers.js
```diff
@@ -338,4 +340,11 @@ exports.truncateDecimals = function (number, digits) {
   return truncatedNum / multiplier;
 };
 
+exports.escapeSpace = function (str) {
+  if(support.util.hasValue(str)) {
+    return str.split(SPACE).join(ESCAPE_SPACE_RE);
+  } else {
+    throw new Error("Could not find the xcodePath");
```
jlipps: this error message doesn't make sense in this method; this method has nothing to do with xcodePath anymore. i think we don't even need to do a value check here at all. if we don't have a value in `str`, calling `.split()` on it will throw an error anyway. so this whole method can be one line.

lib/devices/ios/xcode.js
```diff
@@ -82,7 +84,11 @@ xcode.getVersion = function (cb) {
     }
     // we want to read the CFBundleShortVersionString from Xcode's plist.
     // It should be in /[root]/XCode.app/Contents/
-    var plistPath = xcodePath.replace(XCODE_SUBDIR, "/Contents/Info.plist");
+    try {
+      var plistPath = escapeSpace(xcodePath.replace(XCODE_SUBDIR, "/Contents/Info.plist"));
```
jlipps: i don't think we actually want to escape the path here. fs.existsSync doesn't have a problem with spaces in the path, and in fact it will return the wrong thing if the spaces are escaped. we just want to escape the plistPath in the `cmd` down below.

tarunsujit:
@jlipps updated as per your comments

test/unit/helper-specs.js
```diff
@@ -89,4 +89,15 @@ describe("Helpers", function () {
       logger.warn.called.should.equal(false);
     });
   });
-});
+  describe('xcode version', function () {
+    it('parses xcode version with space', function () {
+      var expected = "/Applications/ Xcode 6.1.1.app/Contents/Developer";
+      var actual = "/Applications/\\ Xcode\\ 6.1.1.app/Contents/Developer"
+      if(helpers.escapeSpace(expected).should.equal(actual)) {
```
jlipps: just looking at this test. I think there might be a small misunderstanding about how this works. when you do this:

```
helpers.escapeSpace(expected).should.equal(actual)
```

it will throw if they are not equal. So you don't need to have this conditional, and you don't need to throw anything yourself.

Also, you'll want to switch around your usage of 'actual' and 'expected'. You are escaping the 'actual' value, so that it matches the 'expected' value (i.e., the 'expected' value is the final value we are testing against, the one that is the result of the operation).

lib/devices/ios/xcode.js
```diff
@@ -86,7 +88,11 @@ xcode.getVersion = function (cb) {
     if (!fs.existsSync(plistPath)) {
       return cb(new Error("Could not get Xcode version: " + plistPath + " does not exist on disk."), null);
     } else {
-      var cmd = "/usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' " + plistPath;
+      try {
+        var cmd = "/usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' " + escapeSpace(plistPath);
```
jlipps: Style point: with a long line like this, please break into two:

```
var cmd = "/usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' " +
                 escapeSpace(plistPath);
```

lib/devices/ios/xcode.js
```diff
@@ -86,7 +88,11 @@ xcode.getVersion = function (cb) {
...
+      } catch(e) {
+        return cb(e, null);
```
jlipps: another style point: we do `cb(e)`, not `cb(e, null)`, making sure to just not pass any value at all into the callback in the failure case. (I see that someone did this two lines above, however we probably should have asked for that to be changed)

tarunsujit:
Thanks a lot for the corrections @jlipps . Updated

test/unit/helper-specs.js
```diff
@@ -89,4 +89,11 @@ describe("Helpers", function () {
       logger.warn.called.should.equal(false);
     });
   });
-});
+  describe('xcode version', function () {
+    it('parses xcode version with space', function () {
+      var expected = "/Applications/ Xcode 6.1.1.app/Contents/Developer";
+      var  = "/Applications/\\ Xcode\\ 6.1.1.app/Contents/Developer"
```
jlipps: looks like the variable name is missing here @tarunsujit. fix this and i think we're good to! and the first line (line 94) should be `actual`, this line is `expected`.

tarunsujit:
Something wrong with me today. My Bad :(

jlipps:
beautiful, thanks @tarunsujit!

jlipps:
fyi @tarunsujit looks like I was a bit hasty. I merged your commit when it was generating lint errors (see below). Did you run `./reset.sh --dev` before doing appium dev? That sets up git pre-commit hooks which ensures that you won't be able to make a commit with lint errors. Definitely a requirement for appium dev.

```
[14:15:12] Starting 'jscs'...

/home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/lib/devices/ios/xcode.js
  line 97  col 12  'cmd' used out of scope.
  line 92  col 13  'cmd' is defined but never used.

  ⚠  2 warnings


/home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/test/unit/helper-specs.js
  line 95  col 77  Missing semicolon.

  ⚠  1 warning

[14:15:35] 'jshint' errored after 23 s
[14:15:35] Error in plugin 'gulp-jshint'
Message:
    JSHint failed for: /home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/lib/devices/ios/xcode.js, /home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/test/unit/helper-specs.js
[14:15:35] 'default' errored after 24 s
[14:15:35] Error: [object Object]
    at formatError (/home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/gulp/bin/gulp.js:169:10)
    at Gulp.<anonymous> (/home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/gulp/bin/gulp.js:195:15)
    at Gulp.emit (events.js:117:20)
    at Gulp.Orchestrator._emitTaskDone (/home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/gulp/node_modules/orchestrator/index.js:264:8)
    at /home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/gulp/node_modules/orchestrator/index.js:275:23
    at finish (/home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:21:8)
    at /home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:45:4
    at _rejected (/home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/q/q.js:804:24)
    at /home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/q/q.js:830:30
    at Promise.when (/home/appium/Workspace/jenkins-remote-executor/workspace/Appium PRs Unit Tests/node_modules/q/q.js:1064:31)
[14:15:35] 'jscs' errored after 23 s
[14:15:35] Error in plugin 'gulp-jscs'
Message:
    Illegal trailing whitespace at devices/ios/xcode.js :
    90 |    } else {
    91 |      try {
    92 |        var cmd = "/usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' " + 
---------------------------------------------------------------------------------------------^
    93 |                  escapeSpace(plistPath);
    94 |      } catch(e) {

Missing space after "catch" keyword at devices/ios/xcode.js :
    92 |        var cmd = "/usr/libexec/PlistBuddy -c 'Print CFBundleShortVersionString' " + 
    93 |                  escapeSpace(plistPath);
    94 |      } catch(e) {
---------------------^
    95 |        return cb(e);
    96 |      }

Illegal trailing whitespace at helpers.js :
   343 |exports.escapeSpace = function (str) {
   344 |  return str.split(SPACE).join(ESCAPE_SPACE_RE);
   345 |};  
------------^
   346 |
   347 |
```

jlipps:
(i've gone ahead and fixed these in 839b9a8)

tarunsujit:
Thanks, will follow it

PR #4851, opened 2015-04-03
Add android actual capabilities

nwlunatic:


jlipps:
This seems like a reasonable addition to me. Can you add a test for it?

nwlunatic:
Of course. But going to need some time to dive into current tests layout. Will be back with tests

nwlunatic:
I've added test, but i'm not sure about:
+ Test setup. Should I move device and version capture to some kind of a beforeEach?
+ Testsuite naming label (- @skip-android-all). I'm quite a new to mocha, am I using it right?

test/functional/android/actual-capabilities-specs.js
```diff
@@ -0,0 +1,31 @@
...
+  , Q = require("q");
+
+
+describe("actual capabilities - session start - @skip-android-all", function () {
+    var adb = new ADB({}),
+        adbShell = Q.denodeify(adb.shell.bind(adb));
+
+    var device = adb.udid;
+    var version;
+    adbShell('getprop ro.build.version.release').then(function (res) {
```
jlipps: this is the right idea, but should probably go in a `before` block?

test/functional/android/actual-capabilities-specs.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+var setup = require("../common/setup-base")
+  , desired = require("./apidemos/desired")
+  , ADB = require('../../../lib/devices/android/adb')
+  , Q = require("q");
+
+
+describe("actual capabilities - session start - @skip-android-all", function () {
+    var adb = new ADB({}),
```
jlipps: we use 2 spaces in our testsuite as well

test/functional/android/actual-capabilities-specs.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+var setup = require("../common/setup-base")
+  , desired = require("./apidemos/desired")
+  , ADB = require('../../../lib/devices/android/adb')
+  , Q = require("q");
+
+
+describe("actual capabilities - session start - @skip-android-all", function () {
```
jlipps: I don't think we need the `@skip` here since we don't want to skip the test

lib/devices/android/android.js
```diff
@@ -557,6 +558,19 @@ Android.prototype.waitForActivityToStop = function (cb) {
   this.adb.waitForNotActivity(this.args.appWaitPackage, this.args.appWaitActivity, cb);
 };
 
+Android.prototype.setActualCapabilities = function (cb) {
+  this.capabilities.deviceName = this.adb.udid || this.adb.curDeviceId;
+  this.adb.shell("getprop ro.build.version.release", function (err, version) {
+    if (err) {
+      logger.warn(err);
+      cb(err);
```
jlipps: I'm not sure this is a serious enough error to halt the session start (which is what would happen). Maybe we just want to `cb()` in either case? (In which case, it could be pulled out of the conditional)

nwlunatic:
@jlipps agreed to all ) done

lib/devices/android/android.js
```diff
@@ -557,6 +558,18 @@ Android.prototype.waitForActivityToStop = function (cb) {
...
+  this.capabilities.deviceName = this.adb.udid || this.adb.curDeviceId;
+  this.adb.shell("getprop ro.build.version.release", function (err, version) {
+    if (err) {
+      logger.warn(err);
+    } else {
+      logger.debug("Device is at release version " + version);
+      this.capabilities.platformVersion = version;
+    }
+  }.bind(this));
+  return cb();
```
jlipps: last comment: if we don't need to short-circuit the logic, our style is to not use `return` here.

we can simply do `cb();`

jlipps:
OK one more tiny thing then we're ready to merge!

nwlunatic:
fixed

jlipps:
great, thanks @nwlunatic !

Fanch-:
Sorry to bother you, but is this PR supposed to set deviceName capability (from desiredCapabilities) to server udid arg ?

nwlunatic:
@Fanch-  yes, it supposed. Or the current device chosen in case no --udid argument passed

Fanch-:
It's strange because as you see below; I am up-to-date with last rev, but the deviceName isn't set : 

```javascript
info: Welcome to Appium v1.3.7 (REV 6c78716d967713259f081b93d5edcbb27df9eeee)
info: Appium REST http interface listener started on 0.0.0.0:4727
info: [debug] Non-default server args: {"udid":"0a9af1180bffc8ac","port":4727,"bootstrapPort":4728}
info: Console LogLevel: debug
info: --> POST /wd/hub/session {"desiredCapabilities":{"app":"/home/fanch/Downloads/LBC_apk/LBC.apk","platformName":"Android","browserName":"","appActivity":"lbc.ui.activities.SplashScreenActivity","device":"android","appPackage":"lbc.debug","version":"4.2"}}
info: Client User-Agent string: Apache-HttpClient/4.3.3 (java 1.5)
error: The following desired capabilities are required, but were not provided: deviceName
```

nwlunatic:
No, I miss understood you. What you are doing – is setting _desired_ capabilities.
And this commit is about _actual_ capabilities. https://code.google.com/p/selenium/wiki/JsonWireProtocol#Actual_Capabilities

So that when you are doing:
```bash
info: --> POST /wd/hub/session {"desiredCapabilities":{"app":"/home/fanch/Downloads/LBC_apk/LBC.apk","platformName":"Android","browserName":"","appActivity":"lbc.ui.activities.SplashScreenActivity","deviceName":"Android Emulator","appPackage":"lbc.debug","version":"4.2"}}
```

you are going to receive:
```bash
info: <--  {"deviceName":"emulator-4555","platformName":"Android", "platformVersion":"4.2", desired: {...}}}
```

Fanch-:
Oh yes, my bad, I see now. 

Didn't see deviceName capability is ignored on Android, so it sets the deviceName with udid after setting the desiredCapabilities. 

It works great, thank you.

Issue #4653, opened 2015-03-05
Restart instruments if AUT does not start


paymand:
Sometimes AUT doesn't start, instead we end up in **Springboard** (home screen of iOS)!
See the interesting part of the log when this happens: https://gist.github.com/paymand/cc4db3b418999dbfee6b
So as you see, the instruments hasn't crashed, but totally fails to start AUT and ``au.mainApp().getTreeForXML()`` got a dump of ``SpringBoard``.

So I tried with the following approach:
```python
    if self.driver.execute_script("au.mainApp()").text != "Opera Mini":
    if retry < 2:
        retry += 1
        self.log("'Opera Mini' did not start, retrying #" + str(retry))
        self.driver.quit()
        self.start(retry)
    else:
        raise Exception("Failed to start 'Opera Mini' app")
```
Here I check if the AUT was successfully started, otherwise I try restarting a couple of times before giving up. This has worked for me so far and now I get much more stable results.
We also have a similar thing in @bootstraponline's ruby test framework: restart the test session on failure up to 3 times.
I suggest adding a similar work around in the Appium server itself. There's no reason to continue automation if there's no app or main window

imurchie:
This sounds like a good idea to me. Adding to the next milestone so it gets looked at.

bootstraponline:
:+1:

jlipps:
@paymand have you tried using the `waitForAppScript` capability? it's designed so you can set custom logic for determining that your app has launched. would it work here?

paymand:
What happens in the script and hence ``waitForAppLaunched`` fails? Do we try to restart the session/app?

jlipps:
Yeah I think we re-run the whole instruments flow

paymand:
In that case it should totally work :)
Closing this issue...

PR #4855, opened 2015-04-03
Return false if SpringBoard loads instead of app

0x1mason:
Prophylactic for false-App-start positives reported in "Restart instruments if AUT does not start #4653".

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -429,6 +429,15 @@ IOS.prototype.setInitialOrientation = function (cb) {
   }
 };
 
+IOS.isSpringBoard = function (uiAppObj) {
+// Test for iOS homescreen (SpringBoard). AUT occassionally start the sim, but fails to load
+// the app. If that occurs, getSourceForElementFoXML will return a doc object that meets our
+// app-check conditions, resulting in a false positive. This function tests the UiApplication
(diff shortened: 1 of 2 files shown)


lib/devices/ios/ios.js
```diff
@@ -429,6 +429,17 @@ IOS.prototype.setInitialOrientation = function (cb) {
   }
 };
 
+IOS.isSpringBoard = function (uiAppObj) {
+  var isSpringBoard = false;
+  var metaData = uiAppObj['@'];
+
+  if (_.isObject(metaData) && _.has(metaData, "name")) {
```
jlipps: looks like `_.has` already checks that the object is OK for a check: http://underscorejs.org/docs/underscore.html#section-149

so this can just be `if (_.has(metaData, 'name')) {`

jlipps: actually, I think this whole function can just be:

```js
IOS.isSpringBoard = function (uiAppObj) {
  return _.propertyOf(uiAppObj['@'])('name') === 'SpringBoard';
}
```

Up to you if you still want to make it its own function. It is nice for the unit test that way.

0x1mason:
@jlipps updated

jlipps:
@0x1mason did you get my second comment? It's gone now...

0x1mason:
@jlipps Our messages crossed like chicks in a flight. Updated.

PR #4871, opened 2015-04-07
Refactor squashed

Jonahss:


lib/devices/ios/ios.js
```diff
@@ -339,23 +367,20 @@ IOS.prototype.startInstruments = function (cb) {
...
     interKeyDelay: this.args.interKeyDelay,
     justLoopInfinitely: this.shouldIgnoreInstrumentsExit(),
     autoAcceptAlerts: !(!this.args.autoAcceptAlerts || this.args.autoAcceptAlerts === 'false'),
     autoDismissAlerts: !(!this.args.autoDismissAlerts || this.args.autoDismissAlerts === 'false'),
     sendKeyStrategy: this.args.sendKeyStrategy || (this.args.udid ? 'grouped' : 'oneByOne')
-  }).then(
-    function (bootstrapPath) {
+  });
+
+  Q.all([bootstrap])
```
jlipps: `bootstrap().then(...);`

lib/devices/ios/ios-log.js
```diff
@@ -44,91 +40,112 @@ IosLog.prototype.startCapture = function (cb) {
...
-          return cb(new Error("iOS8 log capture requires a sim udid"));
-        }
-        logsPath = path.resolve(process.env.HOME, "Library", "Logs",
-                                "CoreSimulator", this.simUdid);
+    var spawnEnv = _.clone(process.env);
+
+    logger.debug("Attempting iOS device log capture via libimobiledevice idevicesyslog");
+    which("idevicesyslog", function (err) {
+      if (!err) {
+        this.proc = spawn("idevicesyslog", {env: spawnEnv});
```
jlipps: this was already a bug in this file, but it's worth pointing out that I think `spawn` can throw synchronously, so it may be a good idea for us to wrap this and other instances in try/catch.

lib/devices/ios/ios-log.js
```diff
@@ -44,91 +40,112 @@ IosLog.prototype.startCapture = function (cb) {
...
-          function (cb) {
-            require('fs').appendFile(
-              path.resolve(logsPath, "system.log"),
-              'A new simulator is about to start!',
-              errOnlyCb(cb));
+
+      this.finishStartingLogCapture(cb);
+    }.bind(this));
+
+  } else {
```
jlipps: this function is getting so long now that instead of having this massive conditional, I'd probably split it into two methods, `startCaptureOnRealDevice` and `startCaptureOnSim`

lib/devices/ios/ios-log.js
```diff
@@ -44,91 +40,112 @@ IosLog.prototype.startCapture = function (cb) {
...
+          logsPath = path.resolve(process.env.HOME, "Library", "Logs",
+                                  "iOS Simulator", "7.*");
+        }
+        var errOnlyCb = function (cb) { return function (err) { cb(err); }; }; // makes waterfall safer
+        var touchLog = function (cb) {
+          async.series([
+            function (cb) { mkdirp(logsPath, errOnlyCb(cb)); },
+            function (cb) { touch(path.resolve(logsPath, "system.log"), errOnlyCb(cb)); },
+            function (cb) {
+              require('fs').appendFile(
```
jlipps: we do `require`s at the top of the file (unless there's a strong reason not to)

lib/devices/ios/ios-log.js
```diff
@@ -44,91 +40,112 @@ IosLog.prototype.startCapture = function (cb) {
...
-        logger.warn('System log capture failed.');
-        // if system log fails this is no big deal
-        cb();
-      });
-    } else {
-      logger.debug("Starting iOS 6.* simulator log capture");
-      this.proc = spawn("tail", ["-f", "-n", "1", "/var/log/system.log"]);
-      this.finishStartingLogCapture(cb);
-    }
+          ], function (err) { cb(err); });
```
jlipps: this looks like the definition of `errOnlyCb`, maybe use it here again?

lib/devices/ios/ios-log.js
```diff
@@ -44,91 +40,112 @@ IosLog.prototype.startCapture = function (cb) {
...
+          logsPath = path.resolve(process.env.HOME, "Library", "Logs",
+                                  "CoreSimulator", this.simUdid);
+        } else {
+          logger.debug("Starting iOS 7.* simulator log capture");
+          logsPath = path.resolve(process.env.HOME, "Library", "Logs",
+                                  "iOS Simulator", "7.*");
+        }
+        var errOnlyCb = function (cb) { return function (err) { cb(err); }; }; // makes waterfall safer
+        var touchLog = function (cb) {
+          async.series([
```
jlipps: not sure we need `errOnlyCb` in `async.series`. I thought the default behavior for `series` was already that. It's `waterfall` we need to worry about, right?

lib/devices/ios/ios-log.js
```diff
@@ -44,91 +40,112 @@ IosLog.prototype.startCapture = function (cb) {
...
-      this.finishStartingLogCapture(cb);
-    }
+          ], function (err) { cb(err); });
+        };
+        async.waterfall([
+          function (cb) {
+            if (logsPath.indexOf('*') < 0) {
+              // there is no doubt on system.log location
+              touchLog(cb);
+            } else cb();
```
jlipps:
```js
} else {
  cb();
}
```

lib/devices/ios/ios-log.js
```diff
@@ -44,91 +40,112 @@ IosLog.prototype.startCapture = function (cb) {
...
-        cb();
-      });
-    } else {
-      logger.debug("Starting iOS 6.* simulator log capture");
-      this.proc = spawn("tail", ["-f", "-n", "1", "/var/log/system.log"]);
-      this.finishStartingLogCapture(cb);
-    }
+          ], function (err) { cb(err); });
+        };
+        async.waterfall([
```
jlipps: I'm really not a fan of big `async.series` or `async.waterfall` blocks that define the functions inline. It is really messy to read. I know you didn't initiate this pattern here in this file, but since we're refactoring, can we clean it up? I'd prefer to define each function ahead of time, e.g.:

```js
var prepLogPath = function (cb) {...};
var findSimLogs = function (cb) {...};
var tailSimLogs = function (files, cb) {...};
async.waterfall([prepLogPath, findSimLogs, tailSimLogs], function (err) {
  if (err) {
    logger.warn('System log capture failed');
  }
  cb();
});
```

(also note how I think I fixed a bug by only outputting the log warning when there was an error)

lib/devices/ios/ios-log.js
```diff
@@ -44,91 +40,112 @@ IosLog.prototype.startCapture = function (cb) {
...
+        this.proc = spawn("tail", ["-f", "-n", "1", "/var/log/system.log"]);
+        this.finishStartingLogCapture(cb);
+      }
+    }.bind(this));
   }
 };
 
 IosLog.prototype.finishStartingLogCapture = function (cb) {
+  if (!this.proc) {
+    logger.warn("Could not capture device log");
```
jlipps: I know it's a pain, but typically we do this:

```js
var msg = "Could not capture device log";
logger.warn(msg);
return cb(new Error(msg));
```

to keep it DRY

lib/devices/ios/ios.js
```diff
@@ -210,7 +212,7 @@ IOS.prototype.configureApp = function (cb) {
   }
 };
 
-IOS.prototype.preCleanup = function (cb) {
+IOS.prototype.removeInstrumentsSocket = function (cb) {
```
jlipps: :+1:

lib/devices/ios/ios-log.js
jlipps: no, you're right, i checked and we also need to worry about `series`

lib/devices/ios/ios.js
```diff
@@ -778,57 +784,33 @@ IOS.prototype.setSafariPrefs = function () {
...
-          "Automation.tracetemplate");
-        if (fs.existsSync(xcodeTraceTemplatePath)) {
-          this.args.automationTraceTemplatePath = xcodeTraceTemplatePath;
+IOS.prototype.detectUdid = function (cb) {
+  if (this.args.udid !== null && this.args.udid === "auto") {
+    which('idevice_id', function (err, cmdPath) {
+      logger.debug("Auto-detecting iOS udid...");
+
+      var udidetectPath = cmdPath + " -l";
+      if (err) {
```
jlipps: I like to do error handling first in the function

lib/devices/ios/ios.js
```diff
@@ -778,57 +784,33 @@ IOS.prototype.setSafariPrefs = function () {
...
+        udidetectPath = require.resolve('udidetect');
+      }
+
+      var udiddetectProc = exec(udidetectPath, { maxBuffer: 524288, timeout: 3000 }, function (err, stdout) {
```
jlipps: interesting. I thought if `exec` timed out, it passed it into the cb's `err` (i.e., so we wouldn't have to listen for a timeout event)

lib/devices/ios/ios.js
```diff
@@ -778,57 +784,33 @@ IOS.prototype.setSafariPrefs = function () {
   }
 };
 
-IOS.prototype.detectTraceTemplate = function (cb) {
```
jlipps: where did this method go, did it go into xcode.js? noticed you removed it from the main flow in both RD and sim paths

package.json
```diff
@@ -141,5 +144,8 @@
     "yargs": "^3.0.4",
     "yiewd": "~0.6.0",
     "rewire": "^2.3.1"
+  },
+  "optionalDependencies": {
+    "udidetect": "^1.0.7"
```
jlipps: how does someone make use of this optional dependency? how do they opt in?

jlipps:
OK, that's all I can think of in terms of review for now! Let me know when you're ready for another pass

lib/devices/ios/ios-log.js
Jonahss: makes sense. good move.

lib/devices/ios/ios-log.js
Jonahss: this is a 'require.resolve', it gives us the path to the function.
This is the same pattern I was going with with the npm modules for our test apps, where calling `require.resolve` gives us the path to a binary.

lib/devices/ios/ios-log.js
Jonahss: :+1:

lib/devices/ios/ios-log.js
Jonahss: I think I just copied that code from the previous implementation :)

lib/devices/ios/ios.js
Jonahss: uh again. that code is just from how this function was already implemented.

lib/devices/ios/ios.js
Jonahss: yup its gone! only xcode needed it, and now xcode get's it itself.

package.json
Jonahss: optionalDependencies get loaded along with all other dependencies, but if the install fails for any reason, installation resumes with just a warning.

lib/devices/ios/ios-log.js
jlipps: .... this is not a 'require.resolve'. it is a 'require'.

lib/devices/ios/ios.js
jlipps: I just ran a test, and it works as I expected. Do you mind cleaning this up while we're touching it?

package.json
jlipps: :+1:

Jonahss:
Ok @jlipps. All changes from you're comments added (except the one about `require.resolve`, what do you think about that one?)

Tests pass on iOS8.2
Tests pass on iOS7.1
Some basic TestApp tests pass on RealDevice

lib/devices/ios/ios-log.js
jlipps: still waiting for a response on this one

Jonahss: Oh gosh yeah. that one was just in the previous code. movin' on out

Jonahss: ha, fs was already defined anyways -_- 
boo copy-paste

Jonahss:
okeedoke. Ready for squashing and merging?

lib/devices/ios/ios.js
Jonahss: ah ok

Jonahss:
Just saw comment above. cleaning up uididetect code.

Jonahss:
squashed and rebased off master.

PR #4903, opened 2015-04-13
Use latest instruments

imurchie:
Master isn't usable for iOS tests that update app preferences, without the latest from appium/instruments.


jlipps:
has appium-instruments been published? usually a submodule ref and a package.json update go hand-in-hand.

imurchie:
Hmm.

The repo says 1.7.3, which is what is in package.json. But the submodule is not on the same commit.

jlipps:
ok (cc @Jonahss looks like this fell through the cracks)

Jonahss:
ah

Issue #4832, opened 2015-04-01
App download from HTTPS URL is not supported by Appium


rahulk06:
Hi,

Downloading the app from the URL with HTTPS  fails with the below error :

error: Unhandled error: Error: read ECONNRESET
    at errnoException (net.js:904:11)
    at TCP.onread (net.js:558:19) context: [POST /wd/hub/session {"desiredCapabilities":{"headers"
:{"Authorization":"apikey=prdakyresOf3EvFMkqSNMQKA6qYIyRmEO06Is4fS"},"platfor
mVersion":"4.4.2","app":"https://build.xxxxx.com/gala/android/release/trunk/xxx.com.4.2.0.2185.debug.apk]

Can someone please help me on this.

Thanks,
Rahul.

jlipps:
`headers` is not a capability appium supports. Is it necessary for connecting to your server?

rahulk06:
I know that headers is not a capability supported by Appium. We are passing is as a capability because our service is behind the gateway and therefore it is used for Authorization. (Here is the discussion link : https://github.com/SeleniumHQ/selenium/pull/276 ) 

Further I saw that Appium uses Request.js - that supports HTTPS but I think for that we may need to pass extra options which is not supported at present. I found this below function in Helper.js file which I think is used for downloading the apk.

exports.downloadFile = function (fileUrl, suffix, cb) {
  // We will be downloading the files to a directory, so make sure it's there
  // This step is not required if you have manually created the directory
  logger.debug('fileUrl ==='+fileUrl);
  tempDir.open({prefix: 'appium-app', suffix: suffix}).nodeify(function (err, info) {
    fs.close(info.fd);
    var file = fs.createWriteStream(info.path);
    request(fileUrl).pipe(file).on('close', function () {
      logger.debug(fileUrl + ' downloaded to ' + info.path);
      cb(info.path);
    });
  });
};

jlipps:
Yeah, what about just including the http auth in the url itself? e.g. the app capability would be `https://<user>:<pass>@<host>/<app>`

rahulk06:
Thanks Jonathan !

But the thing is when I hit the APP URL from a browser , I am able to download the app, there I don't have to specify any username/password. Therefore , I think , if I use use the same URL in app capability then appium should be able to download the app. Please refer to below screenshot :

![image](https://cloud.githubusercontent.com/assets/9517306/7017698/fbc4d268-dd18-11e4-9157-26f5d1c612c4.png)

[...]

PR #4905, opened 2015-04-13
if we are downloading an app, actually listen for the error event (add b...

jlipps:
...etter logs for #4832)

also make `downloadFile` conform to node cb pattern.

can i get a quick +1 @Jonahss

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -18,13 +18,20 @@ var logger = require('./server/logger.js').get('appium')
 exports.downloadFile = function (fileUrl, suffix, cb) {
   // We will be downloading the files to a directory, so make sure it's there
   // This step is not required if you have manually created the directory
+  cb = _.once(cb);
   tempDir.open({prefix: 'appium-app', suffix: suffix}).nodeify(function (err, info) {
     fs.close(info.fd);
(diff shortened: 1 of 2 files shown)


lib/helpers.js
```diff
@@ -18,13 +18,19 @@ var logger = require('./server/logger.js').get('appium')
 exports.downloadFile = function (fileUrl, suffix, cb) {
   // We will be downloading the files to a directory, so make sure it's there
   // This step is not required if you have manually created the directory
+  cb = _.once(cb);
```
Jonahss: is this just to be safe? Or is there a specific code path we know would end in this being called multiple times.

lib/helpers.js
```diff
@@ -18,13 +18,19 @@ var logger = require('./server/logger.js').get('appium')
...
-    request(fileUrl).pipe(file).on('close', function () {
-      logger.debug(fileUrl + ' downloaded to ' + info.path);
-      cb(info.path);
-    });
+    request(fileUrl)
+      .on('error', function (err) {
+        logger.error("Problem downloading file: " + err.message);
+        cb(err);
+      })
+      .pipe(file).on('close', function () {
```
Jonahss: newline for `.on()`?

PR #4914, opened 2015-04-14
Update tests to handle iOS 8.3 foibles

imurchie:
With these all our tests run on 8.3.

### DIFF
--- a/test/functional/ios/testapp/basics/calc-app-1-specs.js
+++ b/test/functional/ios/testapp/basics/calc-app-1-specs.js
@@ -123,9 +123,19 @@ describe('testapp - basics - calc app 1', function () {
     }).then(function (button) {
       return button.click()
         .elementByClassName('UIAAlert')
-        // maybe we could get alert body text too?
-        .elementByClassName('>', 'UIAStaticText').text().should.become("Cool title")
-        .dismissAlert();
+        .elementsByClassName('>', 'UIAStaticText')
+        .then(function (els) {
+          return Q.all(_.map(els, function (el) {
+            return el.text();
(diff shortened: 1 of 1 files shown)


test/functional/ios/testapp/basics/calc-app-2-specs.js
```diff
@@ -74,9 +75,16 @@ describe('testapp - basics - calc app 2', function () {
       .log('crashlog').then(function (logsBefore) {
         numBeforeLogs = logsBefore.length;
         fs.writeFileSync(dir + '/myApp_' + Date.parse(new Date()) + '_rocksauce.crash', msg);
-      }).log('crashlog').then(function (logsAfter) {
+      })
+      .log('crashlog').then(function (logsAfter) {
         logsAfter.length.should.be.above(0);
-        logsAfter.length.should.not.equal(numBeforeLogs);
+        if (env.IOS83) {
+          // getting the crash logs in 8.3 consumes them, so we will only get the one added above
```
jlipps: interesting, what mechanism retrieves the log for us under the hood? weird that the log is sensitive to something reading data from it.

imurchie: More interesting, if my reading of [this](https://github.com/appium/appium/blob/master/lib/devices/ios/ios-crash-log.js#L56) is correct then why were things cumulative before? It seems like it ought to only send logs not yet read. 

Investigating.

imurchie:
Ok. Removed the change to crash log tests. I'll make a PR with the change to how we handle crash logs, and the fixed test.

Issue #4915, opened 2015-04-14
`app_installed` endpoint is returning false for an installed app


redshirtrob:
I'm using the Python module to interact with the Appium web service.  I'm not 100% sure if this is a Python issue or an Appium issue, but my investigations indicate Appium.  I will try to reproduce with another library.

When I perform a simple `is_app_installed` on a real device request using the Python API, I get `false`.  I've tried explicitly setting `udid` (in device capabilities) to my device's UDID.  I've also used the `auto` value.  Neither works.

The issues seems to be the following line in `controller.js::isAppInstalled`.  This expression is false.
```
if ((req.appium.args.udid && req.appium.args.udid.length === 40) || (typeof stdout[0] !== "undefined")) {
```

Here's my Python code:

```
# This Python file uses the following encoding: utf-8

# iOS environment
import unittest
from appium import webdriver

desired_caps = {}
desired_caps['platformName'] = 'iOS'
desired_caps['deviceName'] = "...elided..."
#desired_caps['udid'] = '...elided...'
desired_caps['udid'] = 'auto'
desired_caps['bundleId'] = 'com.app.someapp'

driver = webdriver.Remote('http://localhost:4723/wd/hub', desired_caps)

is_app_installed = driver.is_app_installed("com.app.someapp")

if is_app_installed:
    print "Installed!"
else:
    print "Not Installed..."
```

The only way I can get this to work is by passing the `-U <UDID>` arguments to Appium on launch.  It seems like I should be able to request a device via the Python client, but I cannot figure out how to do so.

If this is a bug, I'm happy to attempt a fix and PR, but I need some direction regarding the correct approach.  That is, in this case, should `typeof stdout[0] !== "undefined"` be the trigger condition?

redshirtrob:
Hmmm, ok.  So, more investigation indicates that this is a problem down in `node-idevice`.  It's passing a boolean value as the second parameter in the callback (as can be seen below).  But the callback is expecting an array of file descriptors or something.  There's even a TODO message in the `controller.js` code to this effect.

So, I guess it seems like Appium is out of sync with node-idevice?  I've followed the setup instructions for running from source.

```
IDevice.prototype.isInstalled = function (appName, cb) {
    var self = this;

    self.listInstalled(function (err, apps) {
        if (err) {
            cb(err);
        } else {
            var i = 0,
                found = false;
            while (i < apps.length && !found) {
                if (apps[i]['name'].indexOf(appName) != -1 ||
[...]

PR #4918, opened 2015-04-15
Handle `node-idevice` callback parameter types for `isInstalled`.

redshirtrob:
This handles #4915.

Node package `node-idevice` has a method on the `IDevice` class
that checks to see if a given app is installed: `isInstalled`.
This method passes a boolean value as the second argument in the
callback function, but `isAppInstalled` expects an array.

The array support appears to be necessary for some other opaque
types so it cannot be removed.  Instead, we check the type of the
second parameter.  If it's `boolean` and `true` then we execute
the positive portion of the if block.

### DIFF
--- a/lib/server/controller.js
+++ b/lib/server/controller.js
@@ -108,15 +108,21 @@ exports.removeApp = function (req, res) {
   }
 };
 
-// TODO: fix this method so it expects a callback with a boolean value, not
-// some weird stdout thing
 exports.isAppInstalled = function (req, res) {
   if (checkMissingParams(req, res, {bundleId: req.body.bundleId}, true)) {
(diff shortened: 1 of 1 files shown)


imurchie:
LGTM.

jlipps:
:+1: for this change. Can do the following, please:

* remove the TODO (since you're fixing it!)
* add a brief comment about why we're looking for a strange variety of responses in our callback, so someone coming along later will know what this is about.

thanks for the investigation and fix.

redshirtrob:
@jlipps Yep, I'll take care of those items and squash them into the PR.

redshirtrob:
@jlipps @imurchie PR updated with requested changes.

jlipps:
thanks @redshirtrob

PR #4928, opened 2015-04-15
fix safariAllowPopups issue w/ ios 8+ (fix #4784)

### LINKED ISSUE #4784: safariAllowPopups not applied on iOS 8.2
For some reason `safariAllowPopups` is not set using iOS 8.2

Here's my desired capabilities:

```python
config = {
    'deviceName': 'iPhone Simulator',
    'browserName': 'Safari',
    'platformVersion': '8.2',
    'platformName': 'iOS',
    'safariAllowPopups': True,
    'safariIgnoreFraudWarning': True,
    'safariOpenLinksInBackground': True,
}
```

I've traced the actual appium code and `IOS.prototype.setSafariPrefs` is correctly called and the option for the popup blocker is passing through. But then when I inspect the booted device, the popup blocker is still on preventing us from opening new windows.

if it helps further, here's a link to the saucelabs run https://saucelabs.com/tests/bbbf5aa6a39a4cc0aa322af294172826

jlipps:
this was a pretty hairy combination of fixes:
- fix bad refactor in settings.js to do with the user settings object
- fix location of user settings plist files on disk
- fix use of `plist` to conform to usage in new version
- use the right key in mobile safari prefs
- update to use the new user settings plist format (now requires a `value` sub-object)

In addition, I added two tests that actually prove everything is working.


jlipps:
looks like ci failed because it was trying to check out 3ad32001fe5e3a632e68bbec1c601c887ed7694b, not sure why it was doing that

jlipps:
anyway, unit tests pass locally

SBoudrias:
Hey, just tested this fix and it seems to be working fine!

aaronbriel:
Hi there, I thought I would comment on this thread instead of starting a new one.  I'm running appium with webdriver and python with the following desiredcapabilities.  I've tried the values 'true' and True for safariAllowPopups but this doesn't toggle the setting.  With trace enabled I can see that Block Popups remains selected in the IOS Simulator session.

                        desired_capabilities={
                            'newCommandTimeout' : '600', 
                            'safariAllowPopups' : 'true', 
                            'javascriptEnabled' : 'true',
                            'platformName': 'iOS',
                            'deviceName': 'iPad Air',
                            'platformVersion': '8.3',
                            'app': app,
                            'browserName' : "Safari"
                        }

SBoudrias:
@aaronbriel the fix was first released as part of version 1.4.6-beta - the version you're using is broken.

aaronbriel:
@SBoudrias my apologies, I should have looked more closely at the milestone.  Thanks!

PR #4942, opened 2015-04-16
Fix ios crash log retrieval

imurchie:
Logs [should](https://code.google.com/p/selenium/wiki/JsonWireProtocol#/session/:sessionId/log) have buffer reset after each request. We were trying to do this, but not quite making it.


jlipps:
thanks

PR #4948, opened 2015-04-17
Added async-trace switch and longjohn and stack-trace dependencies.

0x1mason:


jlipps:
thanks! i checked it out. in general it seems like a useful feature, even though the traces themselves aren't what i'd really want to see (though because of node limitations, I think). once we move to async/await we may be able to write something that filters out async noise and presents something that looks more like a traditional stack trace

0x1mason:
@jlipps Yeah, there's something funky going on.  Uncaught errors will print with the entire async trace. I tried generating the trace with `new Error().stack`, but that broke Winston (surprise!).

0x1mason:
Oh apparently there are some differences is the stacks between 0.10 and 0.12, with the latter being not as useful.  I'm hoping one of the trace modules will get updated soon.

Issue #4835, opened 2015-04-01
implement driver.manage().logs().get("server") to get the appium server log 


hyeung-trulia:
currently for appium, there are two types of log you can get, one is "logcat" and the other is "client". logcat is getting the "adb logcat" log back and actually I am not sure what "client" log type is about since everytime, when I try it, it always returns nothing. 

One improvement I really want is to be able to get back the appium server log so I can see what the last error is and attached back to my exception error message.

TikhomirovSergey:
+1
I have just wanted to ask the same :)

jlipps:
Interesting, seems like an easy feature to implement. Might be memory concerns though if the log grows too large.

hyeung-trulia:
@jlipps  agree. so it would be even better if there is a way to control the amount of log I want to get back. I already did run into the problem when the logcat log size is too big and hung the test. Currently, to sort of work around this log size too big problem, I get the log once at the beginning of the test to "clear" it up.

TikhomirovSergey:
>Interesting, seems like an easy feature to implement. Might be memory concerns though if the log grows too large.

What is this action 
```java
driver.manage().logs().get("server");
```
will extract logs only for current session? If it will extract all logs then clients can throw exceptions. OutOfMemoryException of the Java side, for example.

TikhomirovSergey:
I think there is some work that has to be done on the client side...
@jlipps @hyeung-trulia 
What if there were an Appium-specific parameter for the log extracting? Like that 
```java
driver.manage().logs().get("appium_server");
```

hyeung-trulia:
@TikhomirovSergey Right now, appium server only supports one session at a time (@jlipps : correct me if I am wrong) so there should not be any session mixing problem.

TikhomirovSergey:
@hyeung-trulia I've meant that this action should extract logs of the CURRENT session, not all sessions that have been launched at the server :)

jlipps:
Since this is a command that will be run from a client for a particular session, we will always want this to be scoped to the session.

But I don't think we've addressed my concerns that storing all the logs for one session in Appium's memory might lead to issues. I think the only thing to do is implement the feature and run some memory tests during long sessions.

hyeung-trulia:
do we need to store all the log all in memory? can we store the log in a tmp file? and limit the size of the tmp file?

jlipps:
We could do that but at some point have to get the whole thing in memory to send to the client anyway

hyeung-trulia:
[...]

PR #4953, opened 2015-04-18
#4835 -- add logType server and retrive server log

hyeung-trulia:
add logType server and retrive server log


jlipps:
:+1: for the change. @sebv didn't we have a different tmp file thing we made?

sebv:
Here https://github.com/appium/appium-support/blob/master/lib/tempdir.js, instead of `tmp`.

sebv:
test this please

jlipps:
@hyeung-trulia yes please use the tempdir methods from our 'appium-support' package to keep things consistent.

hyeung-trulia:
@sebv and @jlipps where to put a unit test in? do you have a sample unit test on how test logging? We tested with our local grid and it's working fine.

jlipps:
@hyeung-trulia I would add a new -spec as a functional test for ios and android. we already have tests for different log types. see:

* test/functional/ios/testapp/basics/calc-app-2-specs.js
* test/functional/android/apidemos/basic-specs.js

(notice that for each we have a test that the log type is returned, and that we can get the log)

jlipps:
closing for now, @hyeung-trulia please resubmit with some tests. thanks!

PR #4964, opened 2015-04-20
On linux, don't reset iOS or build Gappium sample apps for iOS.

paymand:
This requires https://github.com/appium/io.appium.gappium.sampleapp/pull/4 to be merged as well.


jlipps:
thanks!

PR #4845, opened 2015-04-02
Updated docs with details on jchuong's iwdp repo

0x1mason:

PR #4990, opened 2015-04-22
Adding new document for instruction to install and run ios-webkit-debug-proxy and refactoring documentation relating to it.

moizjv:
- Added new document for instructions to install and run ios-webkit-debug-proxy.
- Added links for the same in hybrid and web tests on real ios devices.
- Refactoring #4845


jlipps:
thanks!

PR #5010, opened 2015-04-27
Safari fixes for Real Devices

Jonahss:
This depends on updating the submodule commit for SafariLauncher.
Reference PR: https://github.com/snevesbarros/SafariLauncher/pull/3

Tests pass on sims and my ios 8.3 iPhone 6.

Maybe if the owner of SafariLauncher doesn't respond we should fork?

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -328,48 +328,30 @@ IOS.prototype.startInstruments = function (cb) {
     });
   }.bind(this);
 
-  if (this.shouldIgnoreInstrumentsExit()) {
-    // for safari on a real device, instruments is going to crash, but we
-    // expect that and ignore it
-    logger.debug("Not starting command proxy since we're expecting that " +
-                 "Instruments won't be able to talk to it anyway");
-    this.instruments.start(
-      function (err) {
-        if (err) return treatError(err, cb);
-        // instruments will never check into the command proxy, so go ahead
-        // and just call back here once instruments is started, rather than
-        // waiting for it to make the first connection to the command proxy
-        cb();
-      }
-    , function () {
-        // ignore unexpected instruments exit since that's exactly what
-        // we're expecting anyway
-    });
-  } else {
-    logger.debug("Starting command proxy.");
-    this.commandProxy.start(
-      function (err) { // on first connection
-        // first let instruments know so that it does not restart itself
-        this.instruments.launchHandler(err);
-        // then we call the callback
-        cb(err);
-      }.bind(this)
-      //
-    , function (err) { // regular cb
-        if (err) return treatError(err, cb);
-        logger.debug("Starting instruments");
-        this.instruments.start(
-          function (err) {
-            if (err) return treatError(err, cb);
-            // we don't call cb here, waiting for first connection or error
-          }.bind(this)
-        , function (code) {
+  logger.debug("Starting command proxy.");
+  this.commandProxy.start(
+    function onFirstConnection(err) {
+      // first let instruments know so that it does not restart itself
+      this.instruments.launchHandler(err);
+      // then we call the callback
+      cb(err);
+    }.bind(this)
+  , function regularCallback(err) {
+      if (err) return treatError(err, cb);
+      logger.debug("Starting instruments");
+      this.instruments.start(
+        function (err) {
+          if (err) return treatError(err, cb);
+          // we don't call cb here, waiting for first connection or error
+        }.bind(this)
+      , function (code) {
+          if (!this.shouldIgnoreInstrumentsExit()) {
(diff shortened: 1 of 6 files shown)


lib/devices/ios/safari.js
```diff
@@ -69,13 +68,45 @@ Safari.prototype.start = function (cb, onDie) {
...
+      logger.debug("On Safari Launcher. Tapping button to launch Safari");
+
+      this.findElement('accessibility id', 'launch safari',
+                       function (err, res) {
+        if (err || res.status !== 0) {
+          var msg = "Error. Could not find button to launch Safari. " +
+                    "Make sure you are using the latest version of " +
+                    "SafariLauncher that appium is using";
+          logger.error(new Error(msg));
+          return cb(msg);
```
jlipps: think you got these swapped. you want to log msg, and callback with Error(msg)

lib/devices/ios/safari.js
```diff
@@ -69,13 +68,45 @@ Safari.prototype.start = function (cb, onDie) {
...
+        }
+
+        var buttonId = res.value.ELEMENT;
+
+        this.click(buttonId, function (err, res) {
+          if (err || res.status !== 0) {
+            var msg = "Error. For whatever reason, could not click " +
+                      "the 'launch safari' button of SafariLauncher";
+            logger.error(new Error(msg));
```
jlipps: same heree

lib/devices/ios/safari.js
```diff
@@ -69,13 +68,45 @@ Safari.prototype.start = function (cb, onDie) {
 
   this._start(function (err) {
     if (err) return cb(err);
-    this.navToInitialWebview(cb);
+
+    if (this.args.udid) {
+      logger.debug("On Safari Launcher. Tapping button to launch Safari");
```
jlipps: would prefer to wrap all of this up in its own method (maybe `startSafariFromLauncher` or something)

jlipps:
made a few comments, changes look good.

lib/devices/ios/safari.js
Jonahss: derp

lib/devices/ios/safari.js
Jonahss: double derp

lib/devices/ios/safari.js
Jonahss: mm agreed.

lib/devices/ios/ios.js
```diff
@@ -372,7 +354,7 @@ IOS.prototype.makeInstruments = function (cb) {
   var bootstrap = prepareBootstrap({
     sock: this.sock,
     interKeyDelay: this.args.interKeyDelay,
-    justLoopInfinitely: this.shouldIgnoreInstrumentsExit(),
+    justLoopInfinitely: false,
```
Jonahss: @jlipps I decided to leave this here, since bootstrap.js doesn't need to change its interface. Do you think we should remove it completely?

Jonahss:
@jlipps corrected based on comments.
What do you think about SafariLauncher PR mentioned above? Fork and update submodule?

lib/devices/ios/ios.js
jlipps: i think it's fine

lib/devices/ios/safari.js
```diff
@@ -69,13 +100,24 @@ Safari.prototype.start = function (cb, onDie) {
...
+
+      this._clickButtonToLaunchSafari(function (err) {
+        if (err) {
```
jlipps: our pattern is `if (err) return cb(err);` in one line

jlipps:
one last fix then this is good to merge pending the safarilauncher thing

jlipps:
great, this just needs the safarilauncher submodule update then, right?

Jonahss:
Added last bit; including new safariLauncher.
Will squash after :+1:

lib/devices/ios/ios.js
```diff
@@ -328,48 +328,30 @@ IOS.prototype.startInstruments = function (cb) {
...
-        logger.debug("Starting instruments");
-        this.instruments.start(
-          function (err) {
-            if (err) return treatError(err, cb);
-            // we don't call cb here, waiting for first connection or error
-          }.bind(this)
-        , function (code) {
+  logger.debug("Starting command proxy.");
+  this.commandProxy.start(
+    function (err) { // on first connection
```
sebv: Something we can do to improve the code (while we are still using callback), is naming anonymous function when it is not obvious what they do. 
```
this.commandProxy.start(
   function onFirstConnection(err) { 
```

sebv:
Otherwise looks good. (as much as it can be with cbs, this whole safari logic would be so much simpler if it was written with promise/async)

PR #5016, opened 2015-04-28
Fixing chrome driver to use correct ADB path

moizjv:
- Fixing chrome driver to use adb path from ANDROID_HOME


jlipps:
ok, will cherry pick into the branch

PR #5017, opened 2015-04-28
Update to allow safari on iOS 7.1 when xcode is 6+

imurchie:
Safari was not working for iOS 7.1 when using a version of Xcode above 5.1.1. Partly this was due to the use of SDK version rather than platform version, and partly because there was a problem with deleting the temporary mobile safari folder.

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -30,8 +30,7 @@ var path = require('path')
   , UnknownError = errors.UnknownError
   , binaryPlist = true
   , Args = require("vargs").Constructor
-  , logCustomDeprecationWarning = require('../../helpers.js').logCustomDeprecationWarning
-  ;
+  , logCustomDeprecationWarning = require('../../helpers.js').logCustomDeprecationWarning;
 
 // XML Plist library helper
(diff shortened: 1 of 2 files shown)


imurchie:
@jlipps rebased with your update.

lib/devices/ios/ios.js
```diff
@@ -1572,10 +1571,18 @@ IOS.prototype.getLatestWebviewContextForTitle = function (titleRegex, cb) {
...
     var matchingCtx;
     _(contexts).each(function (ctx) {
-      if (ctx.view &&
-          (parseFloat(this.iOSSDKVersion) < 7 || ctx.view.url !== "about:blank") &&
-          (ctx.view.title || "").match(titleRegex)) {
-        matchingCtx = ctx;
+      if (ctx.view && (ctx.view.title || "").match(titleRegex)) {
+        if (ctx.view.url === "about:blank") {
+          // in the case of Xcode  < 5 (i.e., iOS SDK Version less than 7)
+          // and in the case of iOS 7.1 with safari
```
jlipps: do you mean 'without' safari here? otherwise the logic below seems confusing

PR #5036, opened 2015-05-01
fix remote debugger so it doesn't use console.log

jlipps:
and truncate logs displayed


jlipps:
@Jonahss can you please do a quick real safari run with this to make sure i didn't blow anything up? i'm not set up to run on a real device ATM.

Jonahss:
simple tests pass. plus I'll be running all these tests alot, so will catch any small bugs.

PR #5062, opened 2015-05-05
Real device suite

### LINKED ISSUE #5061: Fix execute_async on Safari Real Device
The callback_url is set to localhost, needs to be local ip address for real devices.

Jonahss:
These changes establish our iOS real device test suite (via `bin.test.sh --real-device --ios83`), begins the task of marking tests with `@skip-real-device`.

This PR also changes the iOS test apps to be loaded by way of the `sample-apps` repo instead of through `submodules/sample-code`.

This PR also fixes #5061 execute_asynce on safari real device

### DIFF
--- a/lib/devices/ios/ios-controller.js
+++ b/lib/devices/ios/ios-controller.js
@@ -22,8 +22,11 @@ var uuid = require('uuid-js')
   , IOSPerfLog = require('./ios-perf-log')
   , errors = require('../../server/errors.js')
   , isWindows = require('appium-support').system.isWindows
+  , localIp = require('appium-support').util.localIp
   , NotImplementedError = errors.NotImplementedError
-  , NotYetImplementedError = errors.NotYetImplementedError;
+  , NotYetImplementedError = errors.NotYetImplementedError
+  , Parser = require('../../server/parser.js')
+  , url = require('url');
 
 var iOSController = {};
 var FLICK_MS = 3000;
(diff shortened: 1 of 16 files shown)


lib/devices/ios/ios-controller.js
```diff
@@ -41,6 +44,13 @@ var logTypesSupported = {
   'performance': 'Performance Logs - Debug Timelines on real devices and simulators'
 };
 
+var getDefaultCallbackAddress = function () {
+  var parser = new Parser();
+  return parser.rawArgs.filter(function (a) {
+      return a[0][1] === '--address';
```
jlipps: these two lines are indented too far.

also, is Array.filter implemented in node 0.10?

lib/devices/ios/ios-controller.js
```diff
@@ -22,8 +22,11 @@ var uuid = require('uuid-js')
   , IOSPerfLog = require('./ios-perf-log')
   , errors = require('../../server/errors.js')
   , isWindows = require('appium-support').system.isWindows
+  , localIp = require('appium-support').util.localIp
   , NotImplementedError = errors.NotImplementedError
-  , NotYetImplementedError = errors.NotYetImplementedError;
+  , NotYetImplementedError = errors.NotYetImplementedError
+  , Parser = require('../../server/parser.js')
+  , Url = require('url');
```
jlipps: why is this capitalized even though we're not using it as a class?

jlipps:
done with review; just those two comments!

jlipps:
(oh, and squash the first 4 commits please)

lib/devices/ios/ios-controller.js
Jonahss: Array.filter is implemented in 0.10.37, just tried it.

lib/devices/ios/ios-controller.js
Jonahss: well, we use its Class methods. I'ts kinda funky, we're not calling `new Url()` anywhere, but we are using it as a urlObject factory and then calling class functions like `Url.resolve()` and such, just felt right. But if you think that's inconsistent, no problem!

Jonahss:
All squashed tight. Ready for mergin ^.^

Issue #4981, opened 2015-04-21
Verify and fix send_text on android


rubixthegreat:
Appium Version: 1.3.5
Android Version: 4.4.2

I'm using automation to click on an element via X,Y coordinates. In this scenario I do not have the id of the object I want to type text in, just the X, Y coordinates that are inside of it. After performing the tap action on it, I attempt to send text like this:

Actions builder = new Actions(driver);
builder.sendKeys('test').perform();

The above code works fine on iOS but it doesn't work on Android. Here is the log statement I received:
info: --> POST /wd/hub/session/e01ceded-ca24-4d22-adc3-59daf2ccdc0f/keys {"value":["test"]}
info: [debug] Pushing command to appium work queue: ["element:setText",{"elementId":"test"}]
info: [debug] [BOOTSTRAP] [debug] Got data from client: {"cmd":"action","action":"element:setText","params":{"elementId":"test"}}
info: [debug] [BOOTSTRAP] [debug] Got command of type ACTION
info: [debug] [BOOTSTRAP] [debug] Got command action: setText
info: [debug] [BOOTSTRAP] [debug] Returning result: {"value":"Unknown error","status":13}
info: --> GET /wd/hub/status {}
info: [debug] Responding to client with success: {"status":0,"value":{"build":{"version":"1.3.5","revision":"a124a15677e26b33db16e81c4b3b34d9c6b8cac9"}},"sessionId":"e01ceded-ca24-4d22-adc3-59daf2ccdc0f"}

The weird thing is that I'm not sending over an element id with this action but it's attempting to set the text on the element id. Why is this?

I did some research into how to get this to work and I ran accross this issue:
https://github.com/appium/appium/issues/856

I don't think that will work for me but is there something similar to that? Or is there a way for me to send text to the element that is currently in focus? Any inputs would be helpful!

imurchie:
Yes, the `keys` function is broken. Not sure if it is broken in iOS or Android.

`lib/server/controller.js` takes the `value` received and passes it along. In iOS this is fine. In Android the method expects an element and a text value, and so we get the above error.

Controller: https://github.com/appium/appium/blob/master/lib/server/controller.js#L762
Android: https://github.com/appium/appium/blob/master/lib/devices/android/android-controller.js#L205
iOS: https://github.com/appium/appium/blob/master/lib/devices/ios/ios-controller.js#L1019

robert-langenfeld:
[...]

PR #5067, opened 2015-05-06
Fix Android keys by finding focused element

imurchie:
For Android we had the incorrect `keys` method. This fixes it by leveraging the ability to set text, but using the currently focused element rather than sending in an element.

Addresses #4981


jlipps:
awesome, thanks!

PR #5163, opened 2015-05-26
Add tests for config

imurchie:
Refactor config so that it is easier to test (e.g., no exiting directly from config, rather do that in main). Add tests for each method.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import path from 'path';
+import fs from 'fs';
 import _mkdirp from 'mkdirp';
 import B from 'bluebird';
 import { exec } from 'teen_process';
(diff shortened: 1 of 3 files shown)


lib/config.js
```diff
@@ -21,37 +28,35 @@ async function getGitRev () {
 }
 
 async function getAppiumConfig () {
+  let stat = await B.promisify(fs.stat)('./build');
```
jlipps: I think we should probably not assume a particular cwd? Maybe this should be `path.resolve(__dirname, "..")` instead of `'./build'`?

test/config-specs.js
```diff
@@ -0,0 +1,282 @@
...
+        checkNodeOk.should.throw;
+      });
+      it('should succeed if node is 0.10+', () => {
+        process.version = 'v0.10.0';
+        checkNodeOk.should.not.throw;
+      });
+    });
+
+    describe('warnNodeDeprecations', () => {
+      let called = false;
```
jlipps: instead of setting up our own `called` logic, why not use `sinon`?

test/config-specs.js
```diff
@@ -0,0 +1,282 @@
...
+    });
+  });
+
+  describe('server arguments', () => {
+    let parser = getParser();
+    describe('getNonDefaultArgs', () => {
+      let args = {};
+      before(() => {
+        // give all the defaults
+        _.each(parser.rawArgs, (rawArg) => {args[rawArg[1].dest] = rawArg[1].defaultValue;});
```
jlipps: i've been moving away from `_.each` to actual `for...of` loops. i do like that this is on one line though.

lib/config.js
imurchie: Good point.

test/config-specs.js
imurchie: Well! That makes things easier, doesn't it.

test/config-specs.js
imurchie: There is much to be said for consistency. I'll go to `for...of`.

imurchie:
Updated as per the comments.

PR #5168, opened 2015-05-27
logger + logsink

sebv:
- All logging call redirected to npmlog.
- logsink listens to all npm log call and redirect it to one winston instance.

requires https://github.com/appium/logger/pull/1

ping @jlipps for review.

### DIFF
--- a/lib/logsink.js
+++ b/lib/logsink.js
@@ -1,0 +1,189 @@
+import npmlog from 'npmlog';
+import { patchLogger } from 'appium-logger';
+import winston  from 'winston';
+import fs from 'fs';
+import 'date-utils'; // TODO: make --localtimezone work without it
+
+// set up distributed logging before everything else
+patchLogger(npmlog);
+global._global_npmlog = npmlog;
+
+// npmlog is used only for emitting, we use winston for output
+npmlog.level = "silent";
+let levels = {
+  debug: 1
+, info: 2
+, warn: 3
+, error: 4
+};
+
+let colors = {
+  info: 'cyan'
+, debug: 'grey'
+, warn: 'yellow'
+, error: 'red'
+};
+
+let npmToWinstonLevels = {
+  silly: 'debug'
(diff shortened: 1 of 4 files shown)


lib/logsink.js
```diff
@@ -0,0 +1,189 @@
...
+    date = new Date(date.valueOf() + date.getTimezoneOffset() * 60000);
+  }
+  return date.toFormat("YYYY-MM-DD HH24:MI:SS:LL");
+};
+
+// Strip the color marking within messages.
+// We need to patch the transports, because the stripColor functionality in
+// Winston is wrongly implemented at the logger level, and we want to avoid
+// having to create 2 loggers.
+function applyStripColorPatch(transport) {
```
jlipps: space before `(`

lib/logsink.js
```diff
@@ -0,0 +1,189 @@
...
+    let code = /\u001b\[(\d+(;\d+)*)?m/g;
+    msg = ('' + msg).replace(code, '');
+    _log(level, msg, meta, callback);
+  };
+}
+
+let _createConsoleTransport = function (args, logLvl) {
+  let transport = new (winston.transports.Console)({
+    name: "console"
+    , timestamp: args.logTimestamp ? timestamp : undefined
```
jlipps: I think we should standardize on comma-last, now that trailing commas are supported

jlipps:
:+1: for the approach, made a few style comments

lib/logsink.js
sebv: We don't usually put space after function name. https://github.com/appium/appium/blob/master/lib/server/helpers.js#L290

lib/logsink.js
sebv: yes comma last is better.

lib/logsink.js
jlipps: i thought we did. we have been doing it in all our new projects as well

PR #5174, opened 2015-05-27
delegate ios simulator app installation/removal/launching to simctl

whoward:
This PR introduces the installApp, removeApp and startActivity endpoint for the iOS simulator.  I haven't gotten startActivity on a real device working yet because I'm not even sure if libidevice can support that (I will look into it soon).

This work depends upon appium/node-simctl#4 so there will probably need to be some updates done to ensure that dependency is properly pulled in once it gets merged.

Please help me bring this PR up to spec - I'm happy to write some tests but I think I might need some direction here.

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -1494,7 +1494,7 @@ IOS.prototype.removeApp = function (bundleId, cb) {
   if (this.args.udid) {
     this.realDevice.remove(bundleId, cb);
   } else {
-    cb(new Error("You can not call removeApp for the iOS simulator!"));
+    this.sim.remove(bundleId, cb);
   }
 };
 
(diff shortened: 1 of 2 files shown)


Jonahss:
We might have to add checks to make sure that these methods can't be called on older versions of xcode.

Roping in @jlipps to review.

jlipps:
:+1: great patch. My only concern is what @Jonahss mentioned. Do we know which versions of Xcode/iOS support this? I want to make sure that the versions we officially support all work with this, otherwise we'll need to add conditional logic. The ones we support currently are: iOS 7.1 and up, and Xcode 6.3 and up (except for iOS 8.0 where we have to support Xcode 6.0)

whoward:
from what reading i've been doing it appears that simctl was added in Xcode 6 - I haven't been able to track down much of a history of this tool, let alone official documentation, so the easiest approach is probably to find a way of downgrading Xcode to 6.0 and see what it looks like.

I would imagine the version of the simulator is irrelevant but again it should be tested manually i think.

jlipps:
@whoward you should be able to download 6.0 from the apple dev portal (it's a bit complex, you have to scroll down to a link that says "looking for older versions of xcode?")

whoward:
I'll have a look on the weekend or possibly monday :) sadly I left my mac at the office and I'm out until tomorrow evening.

whoward:
Alright that was more work than I expected - I needed to 'downgrade' to Mavericks so I just ended up making a VM in VMWare.  Thankfully it looks like 6.0.1 supports all of the 'install', 'uninstall' and 'launch' commands from simctl so we shouldn't need any conditional logic.

I haven't tested 6.1.1, 6.2, 6.3, 6.3.1 yet but I will soon - I wouldn't expect these to have any issues however.

whoward:
Okay I've tested all remaining versions of Xcode 6.x - the simctl install/uninstall/launch works on all versions.

jlipps:
awesome, thanks @whoward !

PR #5179, opened 2015-05-28
Allow for non-boolean deprecated server arguments

imurchie:
Our deprecation code only handled boolean arguments that were by default `false`. Any other deprecated arguments would be reported even if they were not used.


jlipps:
:+1:

PR #5183, opened 2015-05-29
fixing iOS real device tests

Jonahss:


imurchie:
All these `@skip-real-device` lines make me sad :( .

test/functional/ios/safari/windows-frame-specs.js
```diff
@@ -50,9 +50,12 @@ describe('safari - windows and frames (' + env.DEVICE + ') @skip-ios6"', functio
         .elementById('only_on_page_2')
         .nodeify(done);
     });
-    it("should be able to open js popup windows with safariAllowPopups set to true", function (done) {
+
+    // broken on real devices, see https://github.com/appium/appium/issues/5167
+    it("should be able to open js popup windows with safariAllowPopups set to true @skip-real-device", function (done) {
```
jlipps: yeah, not surprising. any capability that works by modifying simulator plist files isn't going to be supported on real devices. we probably need a doc about this (or to put "simulator-only" in caps.md)

test/functional/ios/testapp/device-specs.js
```diff
@@ -60,23 +60,32 @@ describe('testapp - device', function () {
     });
   });
 
-  _.each(['iPhone', 'iPad'], function (device) {
-    describe('generic ' + device + ' deviceName @skip-ios6', function () {
+  describe('generic deviceName @skip-ios6', function () {
```
jlipps: not sure why you undid the DRY-ness here?

test/functional/ios/safari/windows-frame-specs.js
Jonahss: so "safariAllowPopups" cap doesn't do anything on real devices of course. Caps.md already says 'sim only' for a lot of these.

The github issue is actually a separate bug which causes the test to fail anyways....

test/functional/ios/testapp/device-specs.js
Jonahss: Ah, there was a failing test around here somewhere and I was suspicious if this contained a weird race condition (or dropped a callback). It wasn't the issue but I liked the more-declarative wetter version better after I was done.

I'd go back to the dry way if there were three tests, but for just two it seemed too complex for just test code.

jlipps:
yeah our test code isn't typically aiming to be the DRYest anyway

Issue #5142, opened 2015-05-20
[Android][Appium 1.4.0] Need to use get_attribute("name") but not get_attribute("content-desc") to get content-desc attribute value


chenhengjie123:
I have try to get an element's content-desc attribute by using codes below:
```python
# python
self.driver.find_element_by_name("name").get_attribute("content-desc")
```
But it raises error in appium server:
```
2015-05-20 23:27:22:328 - info: --> GET /wd/hub/session/829284c9-d8fc-4df4-88b3-15b8d088e6cd/element/3/attribute/content-desc {}
2015-05-20 23:27:22:329 - info: [debug] Pushing command to appium work queue: ["element:getAttribute",{"elementId":"3","attribute":"content-desc"}]
2015-05-20 23:27:22:331 - info: [debug] [BOOTSTRAP] [debug] Got data from client: {"cmd":"action","action":"element:getAttribute","params":{"elementId":"3","attribute":"content-desc"}}
2015-05-20 23:27:22:332 - info: [debug] [BOOTSTRAP] [debug] Got command of type ACTION
2015-05-20 23:27:22:332 - info: [debug] Responding to client with error: {"status":7,"value":{"message":"An element could not be located on the page using the given search parameters.","origValue":"This element does not have the 'content-desc' attribute"},"sessionId":"829284c9-d8fc-4df4-88b3-15b8d088e6cd"}
2015-05-20 23:27:22:336 - info: <-- GET /wd/hub/session/829284c9-d8fc-4df4-88b3-15b8d088e6cd/element/3/attribute/content-desc 500 8.390 ms - 234 
```

I look into the source codes, found that it is caused by some codes:
Codes in lib/devices/android/bootstrap/src/io/appium/android/bootstrap/handler/GetAttribute.java
```
        final AndroidElement el = command.getElement();
        final String attr = params.get("attribute").toString();
        if (attr.equals("name") || attr.equals("text")
            || attr.equals("className") || attr.equals("resourceId")) {
          return getSuccessResult(el.getStringAttribute(attr));
        } else {
          return getSuccessResult(String.valueOf(el.getBoolAttribute(attr)));
        }
```
And the codes for el.getStringAttribute:
```
public String getStringAttribute(final String attr)
      throws UiObjectNotFoundException, NoAttributeFoundException {
    String res;
    if (attr.equals("name")) {
      res = getContentDesc();
      if (res.equals("")) {
        res = getText();
      }
    } else if (attr.equals("text")) {
      res = getText();
[...]

PR #5189, opened 2015-06-01
Allow access to element content description attribute directly

imurchie:
It is confusing to users (see #5142) to not be able to get the content description directly, but to have to get the `name` attribute. Add `contentDescription` attribute.


jlipps:
:+1:

PR #5205, opened 2015-06-03
Add a few dependencies back

imurchie:
Some dependencies were removed that are still needed.


jlipps:
uh, that's odd... i had made this same change myself but it seems not to have been applied and i can't find it in my git history.

PR #5217, opened 2015-06-05
Deprecate server args that are also desired caps

imurchie:
Appium has a number of server arguments that are also desired capabilities. They should be deprecated in favor of a `--default-capabilities` argument, which takes a JSON string of desired capabilities to be used as defaults for each new session.

Also, the deprecated server arguments are mapped to the desired capabilities, so that even if we pass in `--show-ios-log` we will get, internally, `defaultCapabilities = {showIOSLog: true}`.

Also, finally remove the deprecated `--show-sim-log` in favor of `--show-ios-log`.

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -32,12 +32,13 @@ class AppiumDriver extends BaseDriver {
   async getSessions () {
     let sessions = [];
     for (let [id, driver] of _.pairs(this.sessions)) {
-      sessions.push(Object.assign({id}, driver.caps));
+      sessions.push({id: id, capabilities: driver.caps});
     }
     return sessions;
   }
 
   async createSession (caps, reqCaps) {
+    caps = _.defaults(_.clone(caps), this.args.defaultCapabilities);
     let InnerDriver = this.getDriverForCaps(caps);
     let curSessions;
     log.info(`Creating new ${InnerDriver.name} session`);
--- a/lib/main.js
+++ b/lib/main.js
@@ -32,6 +32,13 @@ async function preflightChecks (parser, args) {
   }
 }
 
+function logDeprecationWarning (deprecatedArgs) {
+  logger.warn('Deprecated server args: ');
+  for (let [arg, realArg] of _.pairs(deprecatedArgs)) {
+    logger.warn(`        ${arg} => ${realArg}`);
+  }
+}
+
 async function logStartupInfo (parser, args) {
   let welcome = `Welcome to Appium v${APPIUM_VER}`;
   let appiumRev = await getGitRev();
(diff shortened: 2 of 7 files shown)


lib/main.js
```diff
@@ -32,6 +32,13 @@ async function preflightChecks (parser, args) {
   }
 }
 
+function logDeprecationWarning (deprecatedArgs) {
+  logger.warn('Deprecated server args: ');
+  for (let [arg, realArg] of _.pairs(deprecatedArgs)) {
+    logger.warn(`        ${arg} => ${realArg}`);
```
moizjv: is this space intentional?

lib/parser.js
```diff
@@ -1,666 +1,736 @@
...
-  , example: "/Users/me/Automation.tracetemplate"
-  , help: "(IOS-only) .tracetemplate file to use with Instruments"
+const deprecatedArgs = [
+  [['-k', '--keep-artifacts'], {
+    defaultValue: false,
+    dest: 'keepArtifacts',
+    action: 'storeTrue',
+    required: false,
+    help: '[DEPRECATED] no effect, trace is now in tmp dir by default and is ' +
+          ' cleared before each run. Please also refer to the --trace-dir flag.',
```
moizjv: extra space here i guess

lib/parser.js
```diff
@@ -1,666 +1,736 @@
...
+    parsedArgs.defaultCapabilities = parsedArgs.defaultCapabilities || {};
+    for (let argEntry of deprecatedArgs) {
+      let arg = argEntry[1].dest;
+      if (argEntry[1].deprecatedFor === '--default-capabilities') {
+        if (arg in parsedArgs && parsedArgs[arg] !== argEntry[1].defaultValue) {
+          parsedArgs.defaultCapabilities[arg] = parsedArgs[arg];
+          // j s h i n t can't handle complex interpolated strings
+          let capDict = {[arg]: parsedArgs[arg]};
+          argEntry[1].deprecatedFor = `--default-capabilities ` +
+                                      `'${JSON.stringify(capDict)}'`;
```
moizjv: i wonder if this would fit in one line

lib/main.js
imurchie: It is. This way we get logs like

```shell
warn: [Appium] Deprecated server args:
warn: [Appium]         --show-ios-log => --default-capabilities '{"showIOSLog":true}'
```

Which I happen to like. :)

lib/parser.js
imurchie: Yes indeed! Thanks.

lib/parser.js
imurchie: Not one 80 char line, unfortunately.

lib/main.js
moizjv: :+1:  :)

lib/appium.js
```diff
@@ -69,6 +70,11 @@ class AppiumDriver extends BaseDriver {
    }
 
   async deleteSession (sessionId) {
+    if (!this.sessions[sessionId]) {
```
jlipps: we should never run into this case, because `deleteSession` should be guarded by `sessionExists`. if this is not the case, it's either a bug here, in base-driver, or mjsonwp

test/driver-specs.js
```diff
@@ -0,0 +1,131 @@
...
+      let appium
+        , mockFakeDriver;
+      beforeEach(() => {
+        [appium, mockFakeDriver] = getDriverAndFakeDriver();
+      });
+      afterEach(() => {
+        mockFakeDriver.restore();
+        appium.args.defaultCapabilities = {};
+      });
+      it('should do nothing if there is no session', async () =>{
```
jlipps: i might be wrong but shouldn't we get a 404 here according to the spec? (see note above as well)

jlipps:
made a few comments about the deleteSession logic, otherwise everything looks outstanding :+1:

test/driver-specs.js
imurchie: Yes. Both are handled elsewhere. There appears to be no need to unit test `deleteSession` with a non-existent session.

imurchie:
Removed the unit test for `driver.deleteSession` with no session, and accompanying logic for handling no session. It should be good to go now.

jlipps:
yay. just think of how many lines we'll get to delete in 1.6 ^_^

PR #5238, opened 2015-06-10
Add ability to configure remote debugger connection retries

imurchie:
Add an iOS-only `webviewConnectRetries` capability, which will control the number of times that we resend the connection request to remote debugger. Also change some logging so we get more pertinent information during the connection process.

Pertains to #5222.


docs/en/writing-running-appium/caps.md
```diff
@@ -86,3 +86,4 @@
 |`sendKeyStrategy`| strategy to use to type test into a test field. Simulator default: `oneByOne`. Real device default: `grouped` |`oneByOne`, `grouped` or `setValue`|
 |`screenshotWaitTimeout`| Max timeout in sec to wait for a screenshot to be generated. default: 10 |e.g., `5`|
 |`waitForAppScript`| The ios automation script used to determined if the app has been launched, by default the system wait for the page source not to be empty. The result must be a boolean |e.g. `true;`, `target.elements().length > 0;`, `$.delay(5000); true;` |
+|`webviewConnectRetries`| Number of times to send connection message to remote debugger, to get webview. |e.g., `12`|
```
jlipps: can we say what the default is?

jlipps:
one comment otherwise :+1:

docs/en/writing-running-appium/caps.md
imurchie: We can!

imurchie:
Added the default for the capability to the docs.

PR #5242, opened 2015-06-11
add help information for command-timeout option

WKPlus:
add help information for command-timeout option: 
- the option is in seconds 
- the value should be less than 2147483 since setTimeout using a 32 bit int to store the delay


jlipps:
thanks. @imurchie want to pull into 1.5 parser?

imurchie:
Will do.

PR #5247, opened 2015-06-11
copyLocalZip doesn't check all possible errors

icecrystal23:
One error wasn't reported, which made for a very confusing failure.  Adding this line helped us dig deeper into the actual issue we have.


jlipps:
:+1: nice catch

Issue #5269, opened 2015-06-18
Appium failed to create driver for native browser


nikhil969:
Here is the code snippet 

 @Before
    public void setUp() throws Exception {
        DesiredCapabilities capabilities = new DesiredCapabilities();
        capabilities.setCapability("deviceName", "emulator-5554");
        capabilities.setCapability("platformName", "Android");
        capabilities.setCapability("platformVersion", "4.4");
        capabilities.setCapability("browserName", "Browser");
        driver = new AndroidDriver(new URL("http://127.0.0.1:4723/wd/hub"),
                capabilities);
    }

Appium was not able to complete setup and waited for the execution of

       driver = new AndroidDriver(new URL("http://127.0.0.1:4723/wd/hub"),
                capabilities);
 
Here is the server stack trace:

info: Welcome to Appium v1.4.3 (REV a357c7fa73222315dd85c3d2dd8334767cca1b0d)
info: Appium REST http interface listener started on 0.0.0.0:4723
info: Console LogLevel: debug
info: --> POST /wd/hub/session {"desiredCapabilities":{"platformVersion":"4.4","deviceName":"emulator-5554","platformName":"Android","browserName":"Browser"}}
info: Client User-Agent string: Apache-HttpClient/4.3.3 (java 1.5)
info: Set mode: Proxying straight through to Chromedriver
info: [debug] Looks like we want chrome on android
info: [debug] Creating new appium session b8fb6fac-e3de-41cb-8d0d-f045c503195e
info: [debug] Checking whether adb is present
info: [debug] Using adb from /home/spartans/.linuxbrew/Cellar/android-sdk/24.2/platform-tools/adb
info: [debug] Using fast reset? true
info: [debug] Preparing device for session
info: [debug] Not checking whether app is present since we are assuming it's already on the device
info: Retrieving device
info: [debug] Trying to find a connected android device
info: [debug] Getting connected devices...
info: [debug] executing cmd: /home/spartans/.linuxbrew/Cellar/android-sdk/24.2/platform-tools/adb devices
info: [debug] 1 device(s) connected
info: Found device emulator-5554
info: [debug] Setting device id to emulator-5554
info: [debug] Waiting for device to be ready and to respond to shell commands (timeout = 5)
info: [debug] executing cmd: /home/spartans/.linuxbrew/Cellar/android-sdk/24.2/platform-tools/adb -s emulator-5554 wait-for-device
[...]

Issue #5266, opened 2015-06-17
"TypeError: undefined is not a function" when creating a new session


diana-zaharia:
Trying to run tests for mobile web app on Android Browser, using following capabilities:

```
capabilities.setCapability(MobileCapabilityType.BROWSER_NAME, "Browser");
capabilities.setCapability(MobileCapabilityType.PLATFORM_VERSION, "5.1.1");
capabilities.setCapability(MobileCapabilityType.PLATFORM_NAME, MobilePlatform.ANDROID);
capabilities.setCapability(MobileCapabilityType.DEVICE_NAME, "Android Emulator");
```

appium 1.4.3, running on linux x64 with chromedriver 2.16, following error is logged when creating new session:

```
info: [debug] [BOOTSTRAP] [debug] Got command action: getDataDir
info: [debug] [BOOTSTRAP] [debug] Returning result: {"status":0,"value":"\/data"}
info: [debug] dataDir set to: /data
error: Unhandled error: TypeError: undefined is not a function
    at [object Object].ChromeAndroid.createSession (/home/emulator/nodejs/lib/node_modules/appium/lib/devices/android/chrome.js:154:33)
    at /home/emulator/nodejs/lib/node_modules/appium/node_modules/async/lib/async.js:607:21
    at /home/emulator/nodejs/lib/node_modules/appium/node_modules/async/lib/async.js:246:17
    at iterate (/home/emulator/nodejs/lib/node_modules/appium/node_modules/async/lib/async.js:146:13)
    at /home/emulator/nodejs/lib/node_modules/appium/node_modules/async/lib/async.js:157:25
    at /home/emulator/nodejs/lib/node_modules/appium/node_modules/async/lib/async.js:248:21
    at /home/emulator/nodejs/lib/node_modules/appium/node_modules/async/lib/async.js:612:34
    at [object Object].<anonymous> (/home/emulator/nodejs/lib/node_modules/appium/lib/devices/android/android.js:541:5)
    at next (/home/emulator/nodejs/lib/node_modules/appium/node_modules/async/lib/async.js:798:43)
    at /home/emulator/nodejs/lib/node_modules/appium/node_modules/async/lib/async.js:32:16
    at [object Object].exports.respond (/home/emulator/nodejs/lib/node_modules/appium/lib/devices/common.js:28:9)
    at [object Object].<anonymous> (/home/emulator/nodejs/lib/node_modules/appium/lib/devices/android/android.js:507:16)
[...]

PR #5273, opened 2015-06-18
Fixing chromedriver nodeify issue

moizjv:
- @jlipps @sebv please review 
- #5266 #5269


jlipps:
@moizjv aren't there other instances of `.nodeify` with appium-chromedriver in both chrome.js and android-hybrid.js? i'm thinking of the stop methods too.

moizjv:
yes, I did not consider those instances of Promises. So all `async` functions are needed to be nodeifyied?

jlipps:
yes. for now probably the way sebv is doing it is safest. in new appium we'll just `await` everything.

PR #5279, opened 2015-06-19
Better promise handling

sebv:
rewrote those Q style.


sebv:
ping @jlipps @moizjv

moizjv:
why not use `nodeify`? we can completely avoid using `then` in that case

sebv:
@moizjv,  there is some logic in there.

jlipps:
yeah, i like it better

PR #5282, opened 2015-06-19
Publish and Shrinkwrap improvements

sebv:


sebv:
- moved shrinkwrap to publish
- added `--no-npmlink` to reset
- making sure shrinkwrap run against production packages (there is a bug if not)
- fixed version compute script

I've published 1.4.6-beta2 with it, seem to work.

sebv:
ping @jlipps

jlipps:
not sure I like this approach. First of all, I think I already worked around the bugs by using `npm-shrinkwrap` rather than `npm shrinkwrap`. Secondly, having it automatically run as part of publish means there's no testing step. This means that whenever we publish we'll get the most recent versions of packages, whether or not they work. So we'll run into the same bug that we want to use shrinkwrap to fix. I believe shrinkwrap should only be used _after_ intentionally re-`npm install`ing packages, i.e., shrinkwrap should be done and then tests should be run, then publish should happen.

Jonahss:
to address @jlipps' concerns, how about a full `npm-install` then the publish script deletes the dev dependencies from node_modules, then runs shrinkwrap?

sebv:
I don't know any way to just delete dev dependencies.

Jonahss:
You can just iterate over the dependencies in package.json and remove those files from node_modules ;)

sebv:
Doesn't work, because some dev dependencies are prod dependencies for  subpackages.

It is where shrinkwrap doesn't work.

PR #5308, opened 2015-06-26
Concurrent android tests throw race condition error

d0lfin:
We run our android test on powerful server with 8 android emulators and 8 appium instances. When tests start execution at one time, and all tests work with one application, they try to write and read one file at 'tmpDir' + 'package name'. Sometimes one thread doesn't finish to write file, but another thread try to read it and throw JSON parse error. I think we need to use unique path for 'strings.json' for each appium instance.


lib/devices/android/android-common.js
```diff
@@ -12,7 +12,8 @@ var logger = require('../../server/logger.js').get('appium')
   , Device = require('../device.js')
   , NotYetImplementedError = errors.NotYetImplementedError
   , Args = require("vargs").Constructor
-  , ADB = require('./adb.js');
+  , ADB = require('./adb.js')
+  , temp = require('temp');
```
jlipps: instead of this, can you use the tempDir support from the `appium-support` package? we're trying to standardize on that

lib/devices/android/android-controller.js
```diff
@@ -1060,7 +1059,7 @@ androidController.parseTouch = function (gestures, multi, cb) {
             var touchStateObject = {
               action: gesture.action,
               options: options,
-              timeOffset: 0.005,
+              timeOffset: 0.005
```
jlipps: we actually intentionally include trailing commas in these lists

jlipps:
:+1: for this change, just made a few comments. thanks!

d0lfin:
updated

jlipps:
thanks!

Issue #5295, opened 2015-06-23
runAppInBackground() actually restarts the app


Jonahss:
From @exstan

runAppInBackground() method actually restarts the app (after the specified timeout), because it brings the app to foreground using adb -S, which "force stop the target app before starting the activity".
Here is the log snippet:

info: [debug] executing cmd: /.../android-sdk_r24.0.2/platform-tools/adb -s emulator-5554 shell "input keyevent 3"
info: [debug] Getting device API level
info: [debug] executing cmd: /.../android-sdk_r24.0.2/platform-tools/adb -s emulator-5554 shell "getprop ro.build.version.sdk"
info: [debug] Device is at API Level 19
info: [debug] executing cmd: /.../android-sdk_r24.0.2/platform-tools/adb -s emulator-5554 shell "am start -S -a android.intent.action.MAIN -c android.intent.category.LAUNCHER -f 0x10200000 -n com.test.bla.bla/com.test.bla.bla.MyTestApp"
info: [debug] Waiting for pkg "com.test.bla.bla" and activity ".MyTestApp" to be focused

see https://github.com/appium/java-client/issues/209

Jonahss:
@exstan if you want to submit a pull request to fix this bit of code. Shouldn't be too bad :)

mayureshshirodkar:
@jlipps @imurchie I see this issue happening with the current build version 1.6.4-beta.
On Android, when i do a runAppInBackground, it actually restarts the app instead of just bringing the app to the foreground.
The issue exists in the current build.

jlipps:
The API has been updated so that if a seconds value lower than 0 is passed, the app never returns. So this work has been done.

PR #5315, opened 2015-06-30
Prevent force stopping the app when bringing back to foreground

stkomitov:
Resolving issue #5295


jlipps:
i'm a bit confused why we simple stop caring about the value of this parameter

exstan:
I found this [discussion](https://discuss.appium.io/t/runappinbackground-in-appium-1-3-5/2831), where that value was changed from `false` to this parameter. 
But I still don't understand why `runAppInBackground()` method should care about the reset state on starting the app. I think the idea of `androidCommon.background` is not to start or reset the app, but to brings it back to foreground (saving app session).
All that happens in one call to the Appium server and the client can test how his app behaves in background, and I think that app reset is unwanted behaviour in any situation, there is a separate method for that purpose.

imurchie:

PR #5327, opened 2015-07-01
Adding socket.io-client as a dependency instead of dev dependency

moizjv:
- updated package.json
- regenerated shrinkwrap


npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appium",
-  "version": "1.4.6",
+  "version": "1.4.7-beta",
```
moizjv: should be beta2

PR #5362, opened 2015-07-10
have calls to /status endpoint hit the appium server and NOT be proxi…

### LINKED ISSUE #5335: Selenium Grid hangs on /wd/hub/status
I've noticed two different behaviour on appium 1.3.7 instead of appium 1.4.6; when using appium as a selenium node, i've launched a test using 

capabilities = {"version":"5.0", "deviceName":"ANDROID","browserName":"chrome","platformName":"ANDROID","id":"5bf1e9de2372193b"}

and, during the android execution test, when the /wd/hub/status hangs up and this behaviour freeze the selenium hub for the 1.4.6 version, instead of 1.3.7 (with a 500 response from chromedriver)

Jonahss:
…ed to chromedriver when testing in chrome on android. fixes #5335

### DIFF
--- a/lib/devices/android/chrome.js
+++ b/lib/devices/android/chrome.js
@@ -34,6 +34,7 @@ ChromeAndroid.prototype.setChromedriverMode = function () {
   // when proxying to chromedriver, we need to make sure the context endpoints
   // are trapped by appium for its own purposes
   this.avoidProxy = [
+    ['GET', new RegExp('^/wd/hub/status.*')],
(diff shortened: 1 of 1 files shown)


Jonahss:
This would need to be cherrypicked into 1.4 branch. and then can we release v1.4.8? @jlipps

jlipps:
cherry-picked into 1.4 so it's ready if someone wants to do the publish. think that should be you since you're on triage duty :-) discuss with moiz since he's been releasing recently and knows the drill.

also, yay, i just encoded this fix into MJSONWP yesterday!

PR #5365, opened 2015-07-11
adding support for importing user defined javascript libraries

fernyb:
Depends on https://github.com/appium/appium-uiauto/pull/73


jlipps:
yeah, I think just sending all the custom stuff you need as an initial executeScript at the beginning of your test should be enough, right?

fernyb:
@jlipps I was doing just that. Reading the file and using execute_script to load the javascript code, all was fine until it got large enough and started to fail. I tried  "#import" within execute_script and that fails when it does evals.
And "#import" works fine within Instruments and when it's in dynamic-bootstrap.js

jlipps:
@fernyb, interesting, '#import' broke at one point. i guess they added it back in.

if sending via execute_script doesn't work, i guess having the ability to add your own library files via an import makes sense.

imurchie:

PR #5366, opened 2015-07-11
Fix simulator launch for Xcode 7

bayandin:
Fix error with Xcode 7 (beta 3).

I specified caps:

```
platformName = "iOS"
deviceName = "iPhone 5s"
```

But got error:

```
Could not find a device to launch. You requested 'iPhone (8.1 Simulator)', but the available devices were: [...]
```

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -941,9 +941,10 @@ IOS.getDeviceStringFromOpts = function (opts) {
     deviceName: opts.deviceName,
     platformVersion: opts.platformVersion
   }));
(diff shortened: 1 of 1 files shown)


jlipps:
@bayandin xcode7/ios9 should work on the 'ios9' branch of appium. can you try that? we'll be merging it soon once @Jonahss has done some more verification

bayandin:
Thanks @jlipps.
As I see it should be fixed in 2cb3fd1265d6f9cb2e098e51f88072e40e91fb9c.
I can only suggest to use comparison with numbers instead of comparison with strings.

jlipps:
closing in favor of #5395

bayandin:
Thanks!

PR #5395, opened 2015-07-17
Initial iOS 9 Support

jlipps:
This gets at least the basic tests running. Note that Xcode 7 beta 3 is required to work on 10.10

### DIFF
--- a/test/functional/ios/crash-specs.js
+++ b/test/functional/ios/crash-specs.js
@@ -17,7 +17,7 @@ describe('crash recovery @skip-real-device', function () {
       .elementByAccessibilityId("Crash")
       .click()
       .then(function () {
-        return driver.sleep(10000);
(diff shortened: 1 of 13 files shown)


test/functional/ios/crash-specs.js
```diff
@@ -17,7 +17,7 @@ describe('crash recovery @skip-real-device', function () {
       .elementByAccessibilityId("Crash")
       .click()
       .then(function () {
```
jlipps: @moizjv this can just be `.sleep(3000)`, and we always include a comment about why we're using a static wait

moizjv: @jlipps yeah i can do that i had to increase from `10000` to `30000` to make test pass i can try gradually with 15 sec to 20 sec. Comment is a good idea

PR #5399, opened 2015-07-19
Fix appium-doctor for OS X 10.11

bayandin:
```
$ bin/appium-doctor.js
Running iOS Checks
✖ Could not detect Mac OS X Version
Appium-Doctor detected problems. Please fix and rerun Appium-Doctor.
$
$ sw_vers -productVersion
10.11
```


jlipps:
thanks @bayandin! @sebv: we just merged this change into the old version of appium-doctor that was still in appium. Can you replicate this fix in the new appium-doctor? (And ideally get rid of the horrible duplicated code?)

PR #5422, opened 2015-07-22
Fix removeApp and installApp not providing deviceId to ADB.

mjdetullio:
When multiple devices are attached to ADB, `adb uninstall <package>` will hang waiting for only a single device to remain connected.  `adb install <apk>` will fail immediately.

Always provide the current device ID for ADB using the `-s <deviceId>` argument.

`this.args.udid` was always null, for any circumstance.

### DIFF
--- a/lib/devices/android/android-controller.js
+++ b/lib/devices/android/android-controller.js
@@ -779,23 +779,21 @@ androidController.isAppInstalled = function (appPackage, cb) {
 };
 
 androidController.removeApp = function (appPackage, cb) {
-  var removeCommand = null;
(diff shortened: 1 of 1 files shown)


jlipps:
thanks @mjdetullio! FWIW, this logic is going away in appium 1.5 and we'll be using consolidated logic that won't have this drawback

mjdetullio:
Yeah when I was digging I noticed normal app under test installs and installApp was different.  I was hoping to eventually be able to resign APKs and reset data as part of the installApp process just like the other logic

PR #5428, opened 2015-07-23
Fix #5408: Prep device and set actual caps during Android noLaunchSetup.

### LINKED ISSUE #5408: Getting UDID from Actual Capabilities not working with autoLaunch:False
https://github.com/appium/appium/pull/4851 ("Add android actual capabilities")

With autoLaunch:True  > "deviceName" -> "#UDID#" 
With autoLaunch:False > "deviceName" -> "Android" 

Result is the same after using "driver.install" and then "driver.launchApp"

Appium 1.4.0
Android 4.4.2
io.appium:java-client:3.0.0

mjdetullio:
When setting the `autoLaunch` capability to false, the device was not prepared and actual capabilities was not set.  Therefore, the driver was not able to retrieve the device ID or install/remove apps prior to invoking launchApp.

Add `prepareDevice` and `setActualCapabilities` to `noLaunchSetup`.  Add check to prepareDevice for initialized ADB Logcat to prevent `start` (launchApp) from attempting to prepare the device a second time when using `autoLaunch=false`.

### DIFF
--- a/lib/devices/android/android-common.js
+++ b/lib/devices/android/android-common.js
@@ -319,6 +319,10 @@ androidCommon.toggleLocationServices = function (ocb) {
 };
 
 androidCommon.prepareDevice = function (onReady) {
+  if (this.adb.logcat !== null) {
+    logger.debug("Device already prepared for session");
+    return onReady();
+  }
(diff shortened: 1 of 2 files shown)


jlipps:
thanks @mjdetullio!

PR #5462, opened 2015-08-03
set Executable and Registered on bundle instead of root

theyelllowdart:
fixes locationd[50056]: type IDs do not match - 7 vs 18 error
reverts bundle scope logic to pre 03ae1ab


theyelllowdart:
fyi, I also experimented with completely omitting Executable and Registered which worked on a 8.4 simulator. Are the Executable and Registered keys required for older iOS versions?

jlipps:
@Jonahss should have a look at this. it's possible both ways will be broken for certain xcodes and certain ios versions. we might need to keep both versions of the code around depending. and it should be in the new ios-sim package. thanks for looking into this!

Jonahss:
Should I be looking into this? I think @imurchie should since they're now working directly with this.

jlipps:
sure, whoever's working on it now

imurchie:
I'll take a look.

imurchie:

Issue #4195, opened 2014-11-28
Implement new touch action spec


sebv:
I might be misreading it, but it looks to me as if our touch implementation has become out of date. 

https://dvcs.w3.org/hg/webdriver/raw-file/default/webdriver-spec.html#parallel-actions-1

actions endpoint now start with `actions`, and name are like pointerDown/pointerUp instead of press/release. 

It seems that implementation has also started within the official drivers.

See this thread, new protocol is being implemented within selendroid:
https://github.com/selendroid/selendroid/pull/601#issuecomment-64906565

PR #5485, opened 2015-08-07
Fix to redirect appium to call the selendroid 'actions'

arunnat:
Fix to redirect appium to call the selendroid 'actions' instead of 'touch/multi/perform' which throws 404 undefined exception  Related to issue https://github.com/appium/appium/issues/4195

### DIFF
--- a/lib/devices/android/selendroid.js
+++ b/lib/devices/android/selendroid.js
@@ -54,6 +54,7 @@ Selendroid.prototype.init = function () {
     , 'toggleWiFi'
     , 'toggleLocationServices'
     , 'getStrings'
+    , 'performMultiAction'
   ];
   this.proxyHost = this.selendroidHost;
   this.avoidProxy = [
@@ -71,6 +72,7 @@ Selendroid.prototype.init = function () {
     , ['POST', new RegExp('^/wd/hub/session/[^/]+/ime')]
     , ['GET', new RegExp('^/wd/hub/session/[^/]+/ime')]
     , ['POST', new RegExp('^/wd/hub/session/[^/]+/keys')]
+    , ['POST', new RegExp('^/wd/hub/session/[^/]+/touch/multi/perform')]
   ];
   this.curContext = this.defaultContext();
 };
@@ -577,6 +579,25 @@ Selendroid.prototype.keys = function (value, cb) {
 
 Selendroid.prototype.back = function (cb) {
   this.keyevent(4, null, cb);
+};
+
+Selendroid.prototype.performMultiAction = function (elementId, actions, cb) {
+  logger.debug("Sending multi touch action to Selendroid"+ actions);
+  var reqUrl = this.proxyHost + ':' + this.proxyPort +
+      '/wd/hub/session/' + this.proxySessionId +
+      '/actions';
+  //selendroid expects the actions to be in 'payload' object
+  var payload = {
+    "payload" : actions
+  };
+
+  doRequest(reqUrl, 'POST', payload, null, function (err) {
+    if (err) return cb(err);
+    cb(null, {
+      status: status.codes.Success.code,
+      value: ''
+    });
+  });
 };
 
 module.exports = Selendroid;


arunnat:
This is just a server side fix, the client will still need to send the right actions. Here is the logs before and after for comparison

info: --> POST /wd/hub/session/5d8b6328-a367-84ca-18f6-24849281726e/touch/multi/perform {"actions":[[{"options":{"x":534,"y":755},"action":"press"},{"options":{"x":-106,"y":0},"action":"moveTo"},{"options":{},"action":"release"}],[{"options":{"x":534,"y":755},"action":"press"},{"options":{"x":106,"y":0},"action":"moveTo"},{"options":{},"action":"release"}]]}
info: [debug] Proxying command to localhost:8080
info: [debug] Making http request with opts: {"url":"http://localhost:8080/wd/hub/session/5d8b6328-a367-84ca-18f6-24849281726e/touch/multi/perform","method":"POST","json":{"actions":[[{"options":{"x":534,"y":755},"action":"press"},{"options":{"x":-106,"y":0},"action":"moveTo"},{"options":{},"action":"release"}],[{"options":{"x":534,"y":755},"action":"press"},{"options":{"x":106,"y":0},"action":"moveTo"},{"options":{},"action":"release"}]]}}
info: [debug] Proxied response received with status 404: undefined
info: <-- POST /wd/hub/session/5d8b6328-a367-84ca-18f6-24849281726e/touch/

arunnat:
After the fix (ignore the logging)
info: <-- POST /wd/hub/session/6e8cf227-9c7a-ed37-e99e-417a3084f36e/touch/multi/perform 200 54164.943 ms - 74 {"status":0,"value":"","sessionId":"6e8cf227-9c7a-ed37-e99e-417a3084f36e"}

info: --> POST /wd/hub/session/6e8cf227-9c7a-ed37-e99e-417a3084f36e/touch/multi/perform {"actions":[[{"action":"press","options":{"x":300,"y":300}},{"action":"moveTo","options":{"x":300,"y":100}},{"action":"release","options":{}}],[{"action":"press","options":{"x":300,"y":300}},{"action":"moveTo","options":{"x":300,"y":900}},{"action":"release","options":{}}]]}

info: [debug] SENDING MULTI TOUCH

info: [debug] Zooming in and out

info: [debug] Making http request with opts: {"url":"http://localhost:8080/wd/hub/session/6e8cf227-9c7a-ed37-e99e-417a3084f36e/actions","method":"POST","json":{"payload":[{"inputDevice":"touch","id":"1","actions":[{"x":500,"name":"pointerDown","y":500},{"ms":100,"name":"pause"},{"x":500,"name":"pointerMove","y":100},{"name":"pointerUp"}]},{"inputDevice":"touch","id":"2","actions":[{"x":500,"name":"pointerDown","y":500},{"ms":100,"name":"pause"},{"x":500,"name":"pointerMove","y":900},{"name":"pointerUp"}]}]}}

jlipps:
so all that's happening is we'd now be wrapping the same json object in a 'payload' key?

arunnat:
Also i am no longer just doing a proxy call directly, since Selendroid does not handle /touch/multi/perform and hence the undefined 404, instead I am wrapping it in payload redirecting it to /actions

jlipps:
ah I see. but otherwise it uses the same payload data structure? i'm surprised that it would work, and i'm confused why the endpoint is different. @lukeis are you guys implementing the updated touch api? and if so, isn't the data structure going to differ from what appium expects currently?

arunnat:
No the structure is also different, the client would need a separate fix. I mimic'd what the Selendroid client is sending using my own client to verify the fix. If you look at the above logs, you can see the Json object are different. There is no more press or moveTo, it uses the webdriver specs mentioned here

https://w3c.github.io/webdriver/webdriver-spec.html#parallel-actions

Current
{"actions":[[{"options":{"x":534,"y":755},"action":"press"},{"options":{"x":-106,"y":0},"action":"moveTo"},{"options":{},"action":"release"}],[{"options":{"x":534,"y":755},"action":"press"},{"options":{"x":106,"y":0},"action":"moveTo"},{"options":{},"action":"release"}]]}

Expected
{{"inputDevice":"touch","id":"1","actions":[{"x":500,"name":"pointerDown","y":500},{"ms":100,"name":"pause"},{"x":500,"name":"pointerMove","y":100},{"name":"pointerUp"}]},{"inputDevice":"touch","id":"2","actions":[{"x":500,"name":"pointerDown","y":500},{"ms":100,"name":"pause"},{"x":500,"name":"pointerMove","y":900},{"name":"pointerUp"}]}]

jlipps:
OK, that makes sense. I missed the part about the client. Right now I'm wondering if it would be better to hold off and just implement the new spec throughout the server that way we don't need to special-case selendroid.

arunnat:
Hmm this change is fairly small and non-intrusive(it only affects selendroid mode), so if clients can write their own override they can unblock themselves. Not sure if adopting the new spec could be done in a short time, the issue has been open for a fairly long time. Perhaps you can keep the original issue open and incorporate these changes as a temporary fix? Let me know

jlipps:
@arunnat ok I agree with you. the reason i'm waffling is we're in the process of rewriting selendroid (https://github.com/appium/appium-selendroid-driver) so we'll have to port the fix there. but i think it's a good idea

jlipps:
ok, i ported it to selendroid driver. it's not tested yet but seems pretty easy. https://github.com/appium/appium-selendroid-driver/commit/e137da3cb2381df82642092eeafe9b72ded7067b

arunnat:
awesome, thanks for the update

PR #5515, opened 2015-08-19
package upgrade

sebv:
Upgrading package and using `latest` for 0.x packages, so they also get upgraded automatically. (When we move to 1.x package, we'll need to switch them to ^1...)

ping @jlipps @Jonahss @moizjv @imurchie.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,23 +35,22 @@
     "doc": "./docs"
   },
   "dependencies": {
-    "appium-android-driver": "^0.0.1",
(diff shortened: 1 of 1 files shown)


Jonahss:
:+1:

moizjv:
@sebv are we going to do shrink-wrap magic here too?

jlipps:
:+1:  we'll also want to shrinkwrap appium 1.5 before we actually push it, yes

PR #5532, opened 2015-08-25
Allow Java version parsing to match OpenJDK

michaelpreston:
When an Android device is connected to Appium, the `android-common` class attempts to get the version of Java installed on the system by calling `java -version`. If a user has an alternative version of the JDK installed on their machine (e.g. OpenJDK), the Java version parsing in android-common will fail due to the RegExp expecting 'java version' as opposed to what will be output ('openjdk version').

This pull request solves the issue by matching on java and openjdk in the RegExp matcher inside `androidCommon.parseJavaVersion`.

### DIFF
--- a/lib/devices/android/android-common.js
+++ b/lib/devices/android/android-common.js
@@ -1048,7 +1048,7 @@ androidCommon.parseJavaVersion = function (stderr) {
   var lines = stderr.split("\n");
   for (var i = 0; i < lines.length; i++) {
     var line = lines[i];
-    if (new RegExp("java version").test(line)) {
+    if (new RegExp("java|openjdk version").test(line)) {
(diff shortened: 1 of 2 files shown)


jlipps:
thanks for the explanation, fix, and test!

moizjv:
this fails 2 tests 
`devices/android/android-common.js java version parsing parses multiple line output:`

michaelpreston:
@moizjv that's odd. I ran the whole spec suite before submitting this and all passed, if you clone my fork of appium and run `grunt` you'll see it's all green. Trying to figure out why it's failing now.

michaelpreston:
@moizjv https://github.com/appium/appium/pull/5612

PR #5535, opened 2015-08-26
Fixes sessionExists and command timeout issues.

sebv:
Handles the following cases:
- When we reset sessionId is null in the session object, but not in the sessionCache.
- There is a command timeout running after the session starts, which is not cleared by the internal drivers. 

ping @jlipps.

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -6,6 +6,7 @@ import { AndroidDriver } from 'appium-android-driver';
 import { IosDriver } from 'appium-ios-driver';
 import { routeConfiguringFunction, errors,
          isSessionCommand } from 'mobile-json-wire-protocol';
+import B from 'bluebird';
 
 class AppiumDriver extends BaseDriver {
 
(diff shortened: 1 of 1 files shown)


sebv:
+ onUnexpectedShutdown logic

lib/appium.js
```diff
@@ -60,6 +62,18 @@ class AppiumDriver extends BaseDriver {
     let d = new InnerDriver(this.args);
     let [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, curSessions);
     this.sessions[innerSessionId] = d;
+
+    // Remove the session on unexpected shutdown, so that we are in a position
+    // to open another session later on.
+    // TODO: this should be removed and replaced by a onShutdown callback.
+    d.onUnexpectedShutdown
```
jlipps: why not make use of the unexpected shutdown facility baked into basedriver?

```js
d.onUnexpectedShutdown.catch(err => {
  await this.startUnexpectedShutdown(err);
});
```

Is it too different?

sebv: We don't want to call `await this.sessions[sessionId].deleteSession();`. Btw I think long term it's a bad idea for AppiumDriver to extends BaseDriver , the sesson cache should be handled outside a driver.

sebv: I don't want to change this for the beta, It is going to be a mess to chain `onUnexpectedShutdown` method or go into crazy class method override mode. This reveals a bigger issue, which we will need to tackle later.  https://github.com/appium/appium/issues/5537

sebv:
Merging, we can revisit the shutdown later, it is good enough for beta.

PR #5568, opened 2015-09-04
newCommandTimeout is not set when autLaunch is set to false

yaphatak:
newCommandTimeout is not set when autLaunch is set to false.


jlipps:
This probably needs to be integrated into Appium 1.5 instead (where I don't think we will have the autoLaunch anyway), cc @imurchie @sebv

imsardine:
Currently a workaround for this issue is to explicitly specify default command timeout (`--command-timeout`) in the command line.

BTW, @jlipps I'm not sure what you mean by "I don't think we will have the autoLaunch anyway", and I'm afraid of the removal of the cap. We do need this cap to support "multiple test cases in a session", where each test case opens the app (setup) and closes it later (teardown).

jlipps:
@sebv had a good proposal of how to provide the equivalent of autoLaunch without appium having to actually implement

imurchie:

PR #5612, opened 2015-09-15
Wrap java versions in brackets for Regex matching

michaelpreston:
Previously, we were matching the java/openjdk version without wrapping
the words in brackets, meaning it would only match 'java' and not 'java
version'.

This commit addresses the issue.


jlipps:
I believe @moizjv fixed this with another commit. Thanks for the fix though @michaelpreston!

PR #5635, opened 2015-09-22
Initial iOS 9.1 support

scottdixon-zz:
Also for 9.1 support, [uiauto has been updated](https://github.com/appium/appium-uiauto/commit/41d99a2ac799d7ddd730efc65f955646b287665f) to fix delete key issue.

@jlipps @moizjv @imurchie

### DIFF
--- a/test/helpers/env.js
+++ b/test/helpers/env.js
@@ -74,6 +74,8 @@ function iphoneOrIpadSimulator(device, version) {
       return isIpad ? 'iPad 2' : 'iPhone 6';
     case '9.0':
       return isIpad ? 'iPad 2' : 'iPhone 6';
+    case '9.1':
+      return isIpad ? 'iPad 2' : 'iPhone 6';
     default:
       throw new Error("invalid version");
   }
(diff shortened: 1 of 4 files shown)


test/helpers/env.js
```diff
@@ -203,6 +214,7 @@ env.IOS82 = env.DEVICE.match(/ios82/i);
 env.IOS83 = env.DEVICE.match(/ios83/i);
 env.IOS84 = env.DEVICE.match(/ios84/i);
 env.IOS9 = env.DEVICE.match(/ios9/i);
+env.IOS91 = env.DEVICE.match(/ios91/i);
```
imurchie: Are there any operations in the codebase that only affect iOS 9? If not, this is ok. If there are, env.iOS9 will be turned on when it is iOS 9.1 also (i.e., the regular expression for iOS 9 here will also match for iOS 9.1).

scottdixon: The same functionality is in place for iOS7 & iOS8. @moiz can you confirm there are no _iOS9.0 only_ features?

imurchie: As you can see, there is `env.IOS8` and `env.IOS80` to distinguish the two, since there are things that iOS 8.0 does that 8.1, etc., do not. I don't know that that is the case here, just bringing it up to be on the lookout.

scottdixon: Good to know. Thanks

Issue #5619, opened 2015-09-17
Ambiguous device name/identifier ios9 xcode 7


hardik-d:
get following error:

 Failure/Error: Unable to find matching line from backtrace
 Selenium::WebDriver::Error::WebDriverError:
   Instruments crashed on startup

appium server logs:
info: [debug] [INST STDERR] Instruments Usage Error : Ambiguous device name/identifier 'iPhone 6 (9.0)':
iPhone 6 (9.0) [705F77E8-654F-405A-B9BB-549B727A8609]
iPhone 6 (9.0) + Apple Watch - 38mm (2.0) [9B3EE868-D843-4495-BD1F-7BDBA4B7EAA8]
info: [debug] [INSTSERVER] Instruments exited with code 255
info: [debug] Killall instruments
info: [debug] Instruments crashed on startup
.
..
....

my caps:

{caps: {
platformName: "iOS",
platformVersion: "9.0",
deviceName: "iPhone 6",
autoDismissAlerts: "true",
app: "#{ENV["APP_PATH"]}"
}

NOTE: I am running appium 1.4.11

fangmobile:
same here.
change to iphone 5s will work.

Jonahss:
Are you using appium version 1.4.11?

cc @moizjv

moizjv:
@scottdixon also had this issue
```
iPhone 6 (9.0) [705F77E8-654F-405A-B9BB-549B727A8609]
iPhone 6 (9.0) + Apple Watch - 38mm (2.0) [9B3EE868-D843-4495-BD1F-7BDBA4B7EAA8]
```
is the problem

moizjv:
@hardik-d @fangmobile can you use udid instead `705F77E8-654F-405A-B9BB-549B727A8609`, we have fixed this issue in appium 1.5?

Jonahss:
Ok, as a workaround you can specify the `udid` in the desired capabilities.

this should be fixed in appium v1.5 so you can install the beta if you want and help us find errors :)
https://discuss.appium.io/t/appium-1-5-beta-release/6058

hardik-d:
I will get beta

fangmobile:
@moizjv I also noticed apple watch was the problem. that is why I switch to 5s and things are fine.
Another way is delete the simulator I guess.
I never used --udid before and getting ideviceinstaller problem.
Shouldn't ideviceinstaller be built already when I install appium js ?

hardik-d:
[...]

PR #5636, opened 2015-09-22
Fixing ambiguous device issue for iPhone 6 (9.0) #5619

moizjv:
- @imurchie I realize that my earlier fix only fixes issue when caps have simulator in them, now this fix code change will also fix for iPhone 6 cap


imurchie:
LGTM

PR #5639, opened 2015-09-22
Added option to specify webkit debug proxy port

michalsarna:
Additional option to run appium with specified webkit debug proxy port. Removed use of system env REMOTE_DEBUGGER_PORT for hybrid iOS apps in webkit-remote-debugger.js.

### DIFF
--- a/lib/devices/ios/webkit-remote-debugger.js
+++ b/lib/devices/ios/webkit-remote-debugger.js
@@ -10,10 +10,11 @@ var _ = require('underscore')
 // CONFIG
 // ====================================
 
-var WebKitRemoteDebugger = function (onDisconnect) {
+var WebKitRemoteDebugger = function (onDisconnect, webkitDebugProxyPort) {
   this.dataMethods = {};
   this.host = 'localhost';
-  this.port = process.env.REMOTE_DEBUGGER_PORT || 27753;
+  //this.port = process.env.REMOTE_DEBUGGER_PORT || 27753;
+  this.port = webkitDebugProxyPort || 27753;
   this.socketDisconnectCb = null;
   this.init(1, onDisconnect);
 };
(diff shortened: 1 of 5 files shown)


jlipps:
i like this change however it should probably be raised against the 1.5 branch (and therefore appium/appium-ios-driver). cc @imurchie

lib/server/capabilities.js
```diff
@@ -93,6 +93,7 @@ var iosCaps = [
 , 'sendKeyStrategy'
 , 'waitForAppScript'
 , 'webviewConnectRetries'
+, 'webkitDebugProxyPort'
```
imurchie: I'm confused as to why this is in with capabilities. Is this not a server argument?

lib/devices/ios/webkit-remote-debugger.js
```diff
@@ -10,10 +10,11 @@ var _ = require('underscore')
 // CONFIG
 // ====================================
 
-var WebKitRemoteDebugger = function (onDisconnect) {
+var WebKitRemoteDebugger = function (onDisconnect, webkitDebugProxyPort) {
   this.dataMethods = {};
   this.host = 'localhost';
-  this.port = process.env.REMOTE_DEBUGGER_PORT || 27753;
+  //this.port = process.env.REMOTE_DEBUGGER_PORT || 27753;
```
imurchie: If this is no longer needed, it should be deleted, not commented out. 

I am also wary of summarily removing functionality. There needs to be a way to allow people, for now, to use the former functionality.

```js
if (process.env.REMOTE_DEBUGGER_PORT) {
  this.logger.debug('Found remote debugger port through REMOTE_DEBUGGER_PORT environment variable. This is deprecated, use `--webkit-debug-proxy-port` server argument');
  webkitDebugProxyPort = webkitDebugProxyPort || process.env.REMOTE_DEBUGGER_PORT;
}
```

michalsarna:
Closing this one.

PR #5644, opened 2015-09-23
appium-uiauto 1.10.9 / iOS9.1 support

scottdixon-zz:


package.json
```diff
@@ -46,7 +46,7 @@
     "appium-chromedriver": "^2.1.3",
     "appium-instruments": "^2.0.5",
     "appium-support": "=1.1.2",
-    "appium-uiauto": "=1.10.8",
+    "appium-uiauto": "=1.10.9",
```
imurchie: Is there a reason that this is locked to `1.10.9` and not just the highest in this version?

scottdixon: Seems like it's been locked for a while.

https://github.com/appium/appium/commit/9764d482cbecf67389878550f7d2bbdafa537288
https://github.com/appium/appium/commit/15ef928f8077b7b90a17561984e0994476940fc6

@moiz @bootstraponline do you know if there's any reason it's locked?

bootstraponline: I think it's a mistake.

PR #5648, opened 2015-09-24
Update appium-bindings.md

nicolasavenza:
Saw in the javadoc that PushesFiles is only implemented by AndroidDriver currently. Should update this doc to reflect.


jlipps:
pushFile is supported on iOS (for simulators). so it sounds like the javadoc is incorrect.

PR #5651, opened 2015-09-24
Added xcode.getVersion to pref tests

scottdixon-zz:


test/functional/ios/file-movement-specs.js
```diff
@@ -102,9 +102,12 @@ describe('file movements - pullFile and pushFile @skip-real-device', function ()
           });
         };
         if (parseFloat(sdk) >= 8) {
-          getSimUdid('6', sdk, env.CAPS, function (err, udid) {
+          xcode.getVersion(function(err, xcodeVersion){
```
jlipps: sounds like this will get superseded by 1.5 soon anyway, but: our style guide is spaces around function definition parentheses. `function (err, xcodeVersion) {`

test/functional/ios/prefs/check-safari-settings.js
```diff
@@ -62,9 +62,12 @@ var ios7up = function (version, udid, setting, expected, cb) {
 exports.ios7up = function (desired, setting, expected, cb) {
   xcode.getMaxIOSSDK(function (err, sdk) {
     if (parseFloat(sdk) >= 8) {
-      getSimUdid('6', sdk, desired, function (err, udid) {
+      xcode.getVersion(function(err, xcodeVersion){
```
jlipps: here too.

test/functional/ios/file-movement-specs.js
imurchie: `function (err, ...`

test/functional/ios/prefs/check-safari-settings.js
imurchie: Ditto on space after `function`.

imurchie:
If this were staying as we move forward, we'd want to refactor so there is only one version, but this is ok for now.

PR #5652, opened 2015-09-24
Fix to use udid option with idevicesyslog

srautiai:
Without -u option idevicesyslog shows logs from any connected iOS device.


jlipps:
this looks like a great change @srautiai however it probably should be incorporated into the 1.5 branch instead. cc @imurchie

srautiai:
Here the same for v1.5 => appium/appium-ios-log#4 - Feel free to close this pull request for this repo

PR #5653, opened 2015-09-25
iOS9+ about:blank fix

scottdixon-zz:
When a page can't load, `getContextsAndViews` returns different information depending on SDK version.

8.4 returns:

```
{ id: 'WEBVIEW_1',
  view:
   { id: 1,
     title: 'Cannot Open Page',
     url: 'x-appbundle:///StandardError.html',
     isKey: false } }`
```

Where as 9.0 & 9.1 return:

```
{ id: 'WEBVIEW_1',
  view:
   { id: 1,
     title: 'Cannot Open Page',
     url: 'about:blank',
     isKey: false } }
```

In 9.0 & 9.1 this results in the refresh button being repeatedly tapped. The bug can be reproduced by running prelaunch specs on iOS9.0 or iOS9.1.

The same fix was used for iOS7.1 (I think @sebv created the original fix)

@moizjv @imurchie


imurchie:
Looks good to me.

PR #5672, opened 2015-09-30
Reverting selendroid submodule and appium-chromedriver to stable versions

moizjv:
- Reverting selendroid submodule for stable version
- Reverting appium-chromedriver to 2.3.2 because of issues in node version 0.10


scottdixon:
LGTM

Issue #5691, opened 2015-10-06
Tested Hybrid apps on Android real device - only get NATIVE_APP


greedypinky:
Tested with Appium version 1.4.13 (lastest as of today)
[steps]
I tried to load a Hybrid apps on Android real device for testing (Android OS 4.4.2).
The apps starts the splashscreen first and when the webview is loaded, it is switched to the
webview activity. I tried put some sleep to wait for the webview content to get loaded but 
the appium server can only give me back the "NATIVE_APP" context.
anyone knows how to fix this?

[test by wd client using Mocha]
it("should login successfully", function () {
  
 return driver
    .sleep(30000)
    .contexts()
      .then(function (ctxs) {
        console.log("What is the context after waiting for webview to load?" + ctxs);
        return driver.context(ctxs[ctxs.length - 1]);
      })
      .sleep(20000)
      .elementById("login-username").sendKeys("test@testcom");
     
  });

[log]
info: [debug] Waiting for pkg "xxx" and activity "com.xxx.activity
.WebviewActivity" to be focused
....
info: [debug] Device is at release version 4.4.2
info: [debug] Device launched! Ready for commands
info: [debug] Setting command timeout to 120 secs
info: [debug] Appium session started with sessionId 9afe0355-2f84-4ad5-ac09-6b74770678fa
info: <-- POST /wd/hub/session 303 10858.429 ms - 74
info: --> POST /wd/hub/session/9afe0355-2f84-4ad5-ac09-6b74770678fa/timeouts/implicit_wait {"ms":100
00}
info: [debug] Set Android implicit wait to 10000ms
info: [debug] Responding to client with success: {"status":0,"value":null,"sessionId":"9afe0355-2f84
-4ad5-ac09-6b74770678fa"}
info: <-- POST /wd/hub/session/9afe0355-2f84-4ad5-ac09-6b74770678fa/timeouts/implicit_wait 200 7.190
 ms - 76 {"status":0,"value":null,"sessionId":"9afe0355-2f84-4ad5-ac09-6b74770678fa"}
info: --> GET /wd/hub/session/9afe0355-2f84-4ad5-ac09-6b74770678fa/contexts {}
info: [debug] Getting a list of available webviews
info: [debug] executing cmd: C:\andriod\adt-bundle-windows-x86_64-20140702\sdk\platfor
m-tools\adb.exe -s 551bb0db shell "cat /proc/net/unix"
info: [debug] Available contexts:
info: [debug] []
[...]

PR #5726, opened 2015-10-14
Readme correction

scottdixon-zz:
https://github.com/appium/appium/issues/5691


jlipps:
actually it's correct as-is. Selendroid is used for native apps < 4.2, and for webview-based apps < 4.4. there was a period where uiautomator works for automating native apps (4.2 and above), but webviews are not chromedriver-backed until 4.4, so we had no way to automate them.

PR #5743, opened 2015-10-19
Using appium-adb for installApp and removeApp

Footpad:
When trying to use install_app or remove_app, Appium fails to execute the adb commands on OS X:

```
info: --> POST /wd/hub/session/04ba7d06-65c1-4167-8ecf-ecf3adfb8012/appium/device/install_app {"appPath":"/var/folders/8n/wksz2lp943q043yf9nxr2y9x2qkb7w/T/AndroidChromeAppFetcher2354111942119848933.apk"}

info: [debug] Installing app using cmd: adb install /var/folders/8n/wksz2lp943q043yf9nxr2y9x2qkb7w/T/AndroidChromeAppFetcher2354111942119848933.apk

info: [debug] Responding to client with error: {"status":1,"value":{"message":"Unable to install [/var/folders/8n/wksz2lp943q043yf9nxr2y9x2qkb7w/T/AndroidChromeAppFetcher2354111942119848933.apk] to device with id [null]. Error [Error: Command failed: /bin/sh -c adb install /var/folders/8n/wksz2lp943q043yf9nxr2y9x2qkb7w/T/AndroidChromeAppFetcher2354111942119848933.apk\n/bin/sh: adb: command not found\n]"},"sessionId":"04ba7d06-65c1-4167-8ecf-ecf3adfb8012"}
info: <-- POST /wd/hub/session/04ba7d06-65c1-4167-8ecf-ecf3adfb8012/appium/device/install_app 500 13.649 ms - 412 
```

Appium appears to be assuming that adb is present in the PATH on OS X, however this will not work (it is supposed to find it from $ANDROID_HOME). This change is analogous to [this one](https://github.com/appium/appium/commit/f14c5c9108d5c600d373d37388c3984d2531950e) previously made.

I noticed [this commit](https://github.com/appium/appium/commit/5e5f5ded366c5e8f69d5896abb74e88c9d13e6bb) recently alleged that these methods did work correctly, however that may have been on a non-POSIX platform and appium-adb appears to support using adb on all platforms. Note that this PR preserves the device ID behavior of the aforementioned change since appium-adb takes care of that in a more sophisticated fashion.

I also removed the now-uncalled methods from `lib/device/common.js` since they didn't work for me and couldn't find any other references to those methods.

### DIFF
--- a/lib/devices/android/android-controller.js
+++ b/lib/devices/android/android-controller.js
@@ -779,21 +779,25 @@ androidController.isAppInstalled = function (appPackage, cb) {
 };
 
 androidController.removeApp = function (appPackage, cb) {
-  if (this.adb.curDeviceId) {
-    var removeCommand = 'adb -s ' + this.adb.curDeviceId + ' uninstall ' + appPackage;
-    deviceCommon.removeApp(removeCommand, this.adb.curDeviceId, appPackage, cb);
-  } else {
-    cb(new Error('Unable to un-install [' + appPackage + '] from undefined device id'));
-  }
+  var deviceId = this.adb.curDeviceId;
+  this.adb.uninstallApk(appPackage, function (err) {
+    if (err) {
+      cb(err, 'Unable to un-install [' + appPackage + '] from device with id [' + deviceId + ']. Error [' + err + ']');
+    } else {
+      cb(null, 'Successfully un-installed [' + appPackage + '] from device with id [' + deviceId + ']');
+    }
+  });
 };
 
 androidController.installApp = function (appPath, cb) {
-  if (this.adb.curDeviceId) {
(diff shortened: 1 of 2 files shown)


jlipps:
thanks for this change @Footpad. couple notes:

1) your git user doesn't seem to be the same as your github user
2) we're just about to do away with appium/master in favor of appium/1.5, which is a complete rewrite. i'm pretty sure this issue has already been resolved for android in the rewrite (check out https://github.com/appium/appium-android-driver)

cc @imurchie

imurchie:
Yes, `removeApp` is fixed in 1.5 (https://github.com/appium/appium-android-driver/blob/master/lib/commands/general.js#L82), as is `installApp` (https://github.com/appium/appium-android-driver/blob/master/lib/commands/general.js#L86) though with more stuff for remote installation.

imurchie:
Thanks!

PR #5770, opened 2015-10-28
don't let getDriverForCaps blow up mysteriously on forgotten platformName

jlipps:
ping @imurchie


imurchie:
Good catch.

jlipps:
one question i had was whether we want cap validation on the general appium level as well, using the same strategy as for other drivers. i did it without since it didn't exist but if we need to validate other caps we might consider adding that to appium general as well.

imurchie:
I think it is probably a good idea. Probably good to get the scaffolding in now so we don't _ad hoc_ it in the future. It's pretty easy to wire up. I'll add it to the tasks for the cleanup of the Appium package, which is coming soon!

PR #5782, opened 2015-10-29
Add helpers for checking proxy status

imurchie:
Proxying currently requires much too much knowledge of the internals of our drivers. This PR adds some helpers so that the dependency can be reduced.

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -127,9 +127,34 @@ class AppiumDriver extends BaseDriver {
     // since we don't call super.executeCommand, we need
     // to clear the appium driver timeout manually
     this.clearNewCommandTimeout();
-
     let sessionId = args[args.length - 1];
     return this.sessions[sessionId].executeCommand(cmd, ...args);
+  }
+
+  proxyActive (sessionId) {
+    let driver = this.sessions[sessionId];
+
+    // drivers need to explicitly say when the proxy is active
+    if (!driver || !driver.jwpProxyActive) {
+      return false;
+    }
+
+    return true;
+  }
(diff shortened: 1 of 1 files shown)


imurchie:
These methods also need to be added to `appium-base-driver`, with reasonable defaults.

lib/appium.js
```diff
@@ -127,10 +127,35 @@ class AppiumDriver extends BaseDriver {
...
-
     let sessionId = args[args.length - 1];
     return this.sessions[sessionId].executeCommand(cmd, ...args);
   }
+
+  proxyActive (sessionId) {
+    let driver = this.sessions[sessionId];
+
+    // drivers need to explicitly say when the proxy is active
+    if (!driver || !driver.jwpProxyActive) {
```
jlipps: i'm confused. given that all the drivers implement MJSONWP, shouldn't _all_ the drivers (especially the ones that currently countenance `driver.jwpProxyActive`) implement `driver.proxyActive()` instead?

in other words, I'm wondering why you inspect `driver.jwpProxyActive` here instead of the sub-driver's `proxyActive()` method

lib/appium.js
```diff
@@ -127,10 +127,35 @@ class AppiumDriver extends BaseDriver {
...
+      return false;
+    }
+
+    return true;
+  }
+
+  getProxyAvoidList (sessionId) {
+    let driver = this.sessions[sessionId];
+    let proxyAvoid = driver.jwpProxyAvoid || [];
```
jlipps: same q here

PR #5800, opened 2015-11-05
Keep track of updated remote debugger app

imurchie:
We have had some issues with iOS 9+ and webviews connecting then hanging. Fix this by remembering the actually connected app.


imurchie:
@scottdixon @jlipps

scottdixon:
:+1:

PR #5817, opened 2015-11-10
Update package.json

### LINKED ISSUE #5700: Appium 1.4 optional dependencies fail to compile with Node 4.
Appium 1.4.13 fails to compile `utf-8-validate@1.1.0` and `bufferutil@1.1.0` via npm. Appium 1.5 beta compiles compiles all dependencies with Node 4, but appears to be far from ready.

Could we please have an interim 1.4.14 release with newer versions of these dependencies for Node 4 compatibility?

imurchie:
1. Fix license so no longer using deprecated format.
2. Update `ws` so that we get buildable `bufferutils` and `utf-8-validate`. Fixes #5700


uipoet:
Thank you, @imurchie!

PR #5821, opened 2015-11-11
Stop recommending use of ios-webkit-debug-proxy fork

gempesaw:
The docs currently recommend using a fork of ios-webkit-debug-proxy since the main project was outdated, but this no longer seems to be the case based on a perusal of the most recent commit dates. As of Nov 11, 2015, the latest commit on the [fork's master branch](https://github.com/jchuong/ios-webkit-debug-proxy) is from June 29th and appears to be 24 commits behind google:master; the latest commit from [google:master is from October 7th.](https://github.com/google/ios-webkit-debug-proxy)

But, I haven't used either the main project or the fork to verify that either of them work; I'm just looking at the commit dates. Thanks either way! :)


imurchie:
This will require trying out the base version and seeing how it works with Appium. Until then we can't change the documentation.

gempesaw:
That makes sense. I'm short a provisioning profile but I'm planning to go through the rest of the instructions; I'll update if/when I ever manage the provisioning profile bit. Thanks!

imurchie:
Cool! And thanks for letting us know where the docs are lacking. We really appreciate it!

gempesaw:
I've verified that [google's repo](https://github.com/google/ios-webkit-debug-proxy) of the webkit debug proxy works as expected with Appium for real device operation for Safari. To be specific, I used `brew install ios-webkit-debug-proxy` which I double checked does [install from google's repo](https://github.com/Homebrew/homebrew/blob/c81d819d8c3672341e12edf2f0f8ce4993cd26cc/Library/Formula/ios-webkit-debug-proxy.rb). It works fine - I ended up setting it up on three separate macs, all running Yosemite 10.10.5 and XCode 7.1 & 7.2.

With regards to the JSON Wire commands, I was successfully able to execute "get this page" and "click on these items" and "get the contents of the current page," - I didn't try any others, but I assume that simply being able to install, set up, and get connected to the real device with google's repo was the sticking point, not a particular JSON Wire command.

I didn't test the fork mentioned in the docs, since I was able to get it working with google's repo.

imurchie:
Done.

PR #5836, opened 2015-11-16
Add check for particular url in safari app selection

imurchie:
Fix issue where we find the wrong app through the remote debugger.

### DIFF
--- a/lib/devices/ios/remote-debugger.js
+++ b/lib/devices/ios/remote-debugger.js
@@ -18,7 +18,7 @@ var net = require('net')
 // CONFIG
 // ====================================
 
-var CONNECTION_RETRIES = 8;
+var CONNECTION_RETRIES = 16;
 
 var RemoteDebugger = function (onDisconnect) {
   this.init(2, onDisconnect);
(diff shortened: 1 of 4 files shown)


lib/devices/ios/remote-debugger.js
```diff
@@ -148,9 +148,23 @@ RemoteDebugger.prototype.selectApp = function (appIdKey, cb, maxTries, tries) {
     // in iOS 8.2 the connect logic happens, but with an empty dictionary
     // which leads to the remote debugger getting disconnected, and into a loop
     if (_.isEmpty(pageDict)) {
-      var msg = "Empty page dictionary received. Trying again.";
-      return onRequiresRetry(this.appIdKey, msg);
+      return onRequiresRetry(this.appIdKey, "Empty page dictionary received. Trying again.");
     }
+
+    if (currentUrl) {
```
jlipps: what if currentUrl is not truthy? seems like this will loop infinitely since found is only set to true in this block. if currentUrl is required, we should `assert.ok(currentUrl)` above and get rid of this `if` statement.

unless i'm reading this wrong

imurchie: If `currentUrl` is not truthy then we proceed as we did before, connecting to whatever app it is that the remote debugger has sent along.

jlipps: aha, I misread the bracket, and thought it was closing the function when it was closing the `if` block

moizjv:
@jlipps @imurchie is this change good to go?

imurchie:
LGTM

PR #5876, opened 2015-11-26
Truly Disable Android Watchers, if disableAndroidWatchers capability …

rseek:
…is set


jlipps:
thanks for this! at this point given the imminent release of appium 1.5, this should be committed to https://github.com/appium/appium-android-bootstrap instead

PR #5890, opened 2015-12-01
Sanitize the deviceOrientation

ngsankha:
Throughout the code we check if `deviceOrientation` is either `LANDSCAPE` or `PORTRAIT` (case sensitive), but the user can pass those strings in any case. While the iOS Simulators have the correct orientation, it returns the orientation in the original case as it was passed by user. This causes our screenshot rotation code to break as it checks for `LANDSCAPE` in uppercase.

It's better to sanitize the orientation whenever we get it.

### DIFF
--- a/lib/devices/ios/ios.js
+++ b/lib/devices/ios/ios.js
@@ -168,9 +168,9 @@ IOS.prototype.configure = function (args, caps, cb) {
 IOS.prototype.setIOSArgs = function () {
   this.args.withoutDelay = !this.args.nativeInstrumentsLib;
   this.args.reset = !this.args.noReset;
-  this.args.initialOrientation = this.capabilities.deviceOrientation ||
-                                 this.args.orientation ||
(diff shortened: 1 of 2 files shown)


jlipps:
thanks for this change @sankha93 however with the imminent release of appium 1.5, this change should probably be made in a different place, if it still needs to (perhaps node-mobile-json-wire-protocol, @imurchie?)

imurchie:
This should indeed come into 1.5. We don't have a generalized way to transform inputs, so I think this will have to go into ech driver :(

sankha93:
@imurchie Thanks for merging this. Btw, I think having a generalized way to sanitize JSONWire params and capabilities would be good, and maybe it can be implemented in node-mobile-json-wire-protocol?

imurchie:
I agree. It will not make it in this version, as we need to think of the best way to do it. I'll add it to the roadmap as a feature.

Issue #5396, opened 2015-07-17
Error getting device property persist.sys.language: -> more than one device and emulator 


nikolmarku1:
Hi,

Appium v1.4.1

I have 2 real devices attached :
1. IPad 1 Mini - OS 8.3
2. Nexus 10 Android version 5.1.1

I have registered both devices on Selenium grid

I tried to run the test and got the error below, see logs
This error is not shown if I remove the connection cable from Ipad

Thanks for your help
# Logs

autotestmac:~ qa$ /Users/qa/automation/selenium/appium/2_start-appium-android-tablet.csh ; exit;
2015-07-17 10:24:53:785 - info: Welcome to Appium v1.4.1 (REV ce6220829063679bcccf5724515c29569c4a16d9)
2015-07-17 10:24:53:789 - info: Appium REST http interface listener started on 10.24.208.226:4723
2015-07-17 10:24:53:793 - info: [debug] Non-default server args: {"address":"10.24.208.226","fullReset":true,"log":"/tmp/appium1.log","logTimestamp":true,"localTimezone":true,"androidDeviceReadyTimeout":"25","deviceName":"Nexus 10","platformName":"Android","platformVersion":"5.0.2","automationName":"Appium","language":"en","locale":"GB","nodeconfig":"/Users/qa/automation/selenium/appium/configAndroid.json","keystorePassword":"","keyAlias":"","keyPassword":"","defaultCommandTimeout":150}
2015-07-17 10:24:53:793 - info: Console LogLevel: debug
2015-07-17 10:24:53:793 - info: File LogLevel: debug
2015-07-17 10:24:53:797 - info: [debug] starting auto register thread for grid. Will try to register every 20000 ms.
2015-07-17 10:25:14:116 - info: [debug] Appium successfully registered with the grid on usb-etbuild5:4444
2015-07-17 10:26:34:316 - info: [debug] Appium successfully registered with the grid on usb-etbuild5:4444
2015-07-17 10:27:54:333 - info: [debug] Appium successfully registered with the grid on usb-etbuild5:4444
[...]

PR #5908, opened 2015-12-10
Fix for getting android device property for running multiple appium i…

pearlselvan:
Fix for https://github.com/appium/appium/issues/5396


imurchie:
Commented on https://github.com/appium/appium-adb/pull/126, which is a prerequisite.

imurchie:
Thanks for this. The problem has been fixed at the ADB level.

PR #5950, opened 2015-12-23
Add a new mobile command for iOS background issue

wuhao5:
The workaround script, (written in AppleScript so it needs Accessiblity permission), to communicate Simulator using Apple Events to mimic background and foreground by clicking the home button and finding the app icon to open it

this depends on #5971 so it would be able to get the correct appName

### DIFF
--- a/lib/devices/ios/ios-controller.js
+++ b/lib/devices/ios/ios-controller.js
@@ -26,6 +26,7 @@ var uuid = require('uuid-js')
   , NotImplementedError = errors.NotImplementedError
   , NotYetImplementedError = errors.NotYetImplementedError
   , Parser = require('../../server/parser.js')
+  , backgroundOSA = path.resolve(__dirname, "./osa/background.scpt")
   , url = require('url');
 
 var iOSController = {};
@@ -1231,6 +1232,17 @@ iOSController.lock = function (secs, cb) {
 iOSController.isLocked = function (cb) {
   cb(new NotYetImplementedError(), null);
 };
+
+iOSController.osaBackground = function (appName, secs, cb) {
+  logger.debug("Execute background osa: '" + appName + "'," + secs);
+  exec ('osascript ' + backgroundOSA + ' \'' + appName + '\' \'' + secs + '\'', function (error, stdout, stderr) {
+    if (error !== null) {
+      cb (error);
+    } else {
+      cb (null, {status:0, result: "" + stderr + ";" + stdout});
+    }
+  })
+}
 
 iOSController.background = function (secs, cb) {
   this.proxy(["au.background(", secs, ")"].join(''), cb);
--- a/lib/devices/ios/ios-controller.js
+++ b/lib/devices/ios/ios-controller.js
@@ -1233,7 +1233,7 @@ iOSController.isLocked = function (cb) {
   cb(new NotYetImplementedError(), null);
 };
 
-iOSController.osaBackground = function (appName, secs, cb) {
+var osaBackground = function (appName, secs, cb) {
   logger.debug("Execute background osa: '" + appName + "'," + secs);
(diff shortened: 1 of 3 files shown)


imurchie:
Awesome! This works well, though the terminal needs to have access to control the computer. We are, however, moving away from the `"mobile: xxx` commands as much as possible. 

It would be preferable to have the same command, which would switch based on the iOS version being used. We can add a new desired capability to pass in the `appName`, and if it is not passed in get it through the app's `Info.plist` (I think the app name is stored in `CFBundleDisplayName`).

If you want, I can open a PR to add the capability and the required code to get the app name, which can then be wired into this PR.

imurchie:
Then we can port it over into Appium 1.5.

wuhao5:
@imurchie sounds like a great plan. To inject to "mobile: " commands is just for me the easiest way to wire this as well as that it needs to specify the appName explicitly. If we can just grab the appName from the Info.plist and hook it up to "background()", that would be a better approach. let me know what I can help.

imurchie:
I've opened a PR with the code to get the appName from a desired capability, or from the plist.

wuhao5:
I patched your PR locally and changed this to use your appName cap. it works very well. once your PR is merged in, this one is good to go too (or I can pull in your PR here too)

wuhao5:
@imurchie I also updated this gist [here](https://gist.github.com/wuhao5/b62ea6f997ce87792817), to remove osaBackground mobile command.

lib/devices/ios/ios-controller.js
```diff
@@ -1232,8 +1233,34 @@ iOSController.isLocked = function (cb) {
...
 
+var osaBackground = function (appName, secs, cb) {
+  logger.debug("Execute background osa: '" + appName + "'," + secs);
+  exec ('osascript ' + backgroundOSA + ' \'' + appName + '\' \'' + secs + '\'', function (error, stdout, stderr) {
+    if (error !== null) {
+      cb (error);
+    } else {
+      cb (null, {status:0, result: "" + stderr + ";" + stdout});
+    }
+  })
```
imurchie: Need semicolon.

lib/devices/ios/ios-controller.js
```diff
@@ -1232,8 +1233,34 @@ iOSController.isLocked = function (cb) {
...
+}
```
imurchie: Need semicolon.

lib/devices/ios/ios-controller.js
```diff
@@ -1232,8 +1233,34 @@ iOSController.isLocked = function (cb) {
...
+
 iOSController.background = function (secs, cb) {
-  this.proxy(["au.background(", secs, ")"].join(''), cb);
+  var v = (this.args.platformVersion || this.iOSSDKVersion);
+  if (v >= 9) {
+    if (!this.args.appName) {
+      this.getAppNameFromApp (function(err, appName) {
```
imurchie: No space after function name in a call.

lib/devices/ios/ios-controller.js
```diff
@@ -1232,8 +1233,34 @@ iOSController.isLocked = function (cb) {
...
+        if (err) {
```
imurchie: This would be better as `if (err) return cb(err);`. That would be in line with our style. The call to `osaBackground` then does not need to be in an `else` block.

imurchie:
A few style fixes, then it should be good to go!

wuhao5:
@imurchie fixed!

wuhao5:
rebased to the latest master branch

jlipps:
LGTM, thanks @wuhao5!

imurchie:
Thanks for this fix!

I'm working on a port into 1.5.

wuhao5:
Thanks!

PR #5963, opened 2015-12-28
Add appium-doctor and update bins

imurchie:
Use the packages themselves.

### DIFF
--- a/bin/authorize-ios.js
+++ b/bin/authorize-ios.js
@@ -1,6 +1,0 @@
-#!/usr/bin/env node
-'use strict';
-
-var authorize = require('authorize-ios');
-
-authorize();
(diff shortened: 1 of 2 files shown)


jlipps:
hmm, earlier I intentionally removed appium-doctor as a dependency because it depends on socket.io or some such that has problems compiling on all platforms. i also updated the docs to have instructions on how to use appium-doctor that don't involve installing it with appium. if we want to bring appium-doctor back in i think we need to resolve the issue with the npm+node versions that were not able to install socket.io on a first go

imurchie:
Ah. Didn't know that. If you can remember the details, can you create an issue in `appium-doctor`?

imurchie:
Opened PR to remove: https://github.com/appium/appium/pull/5979

jlipps:
shoot i don't remember now but I remember @scottdixon originally brought it up, so maybe he remembers?

scottdixon:
@imurchie https://github.com/appium/appium-doctor/issues/10

PR #5974, opened 2016-01-05
Adding capability to change Accept cookie preference for iOS devices

akhillb:
This PR adds an extra capability called `acceptCookies`, which will set the NSHTTPAcceptCookie value for iOS devices. It accepts a string, which could be either of `always`, `never`, and `current page`.


jlipps:
looks good to me @akhillb, thank you. however it will need to be retooled for appium 1.5 (cc @imurchie). also there are no tests for this feature which we'd prefer to have in order to accept it. likewise with some documentation (see caps.md).

akhillb:
So i had a look at user settings and effective user settings files. There seems to be a plist entry called safariAcceptCookies. Should that be changed too, in order for the acceptCookies value to be in effect properly?

jlipps:
@akhillb yeah probably a good idea to change all values that could contribute to this.

akhillb:
Also, can you give me some directions on what should the tests include?

jlipps:
if you look at the tests for appium-ios-driver (where I think this code will need to go for appium 1.5) you should see a bunch of examples of tests. Ultimately I think we want an e2e test that proves this is doing something. @imurchie can provide more direction on how to set that up

imurchie:
The tests should be something along the same lines as the `"should be able to set a cookie for a page"` tests in https://github.com/appium/appium/blob/master/test/functional/common/webview/cookies-base.js#L49

Essentially, if the capability is set to `true`, you should be able to send a new cookie and then retrieve it. If not set, or set to `false`, it should not be possible to send a new cookie (or it should be sent but not saved, I'm not sure which).

imurchie:
Once this PR lands, with all attendant features, we can figure out how it should go to get into 1.5.

imurchie:

PR #5977, opened 2016-01-05
updated node version

SrinivasanTarget:
Updated the supported node version


jlipps:
This isn't quite true yet---0.12 will be required as of appium 1.5 I believe. This change could be made as a PR into the 1.5 branch instead.

SrinivasanTarget:
Do i need to close this and raise a new one? @jlipps

PR #5987, opened 2016-01-07
rm old.apk--> rm -f old.apk

lucyking:
Some stricter phone's /local/data/ dir is not accessible to generally user.(Moto XT1085 etc.)

```
rm /data/local/tmp/a60..1de.apk
```

will lead to:

```
override r--r--r-- shell:shell for '/data/local/tmp/a60...1de.apk'? ^C
```
- appium cannot  deal with this **interactive prompt** then will get freezed forever<br>
  So **rm -f** may be better.

### DIFF
--- a/lib/devices/android/android-common.js
+++ b/lib/devices/android/android-common.js
@@ -609,7 +609,7 @@ androidCommon.removeTempApks = function (exceptMd5s, cb) {
           }
         });
         if (noMd5Matched) {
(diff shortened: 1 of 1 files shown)


lucyking:
update:I try to reproduce this bug today but failed.
Even though, rm -f may be safer.

imurchie:
Seems reasonable. Thanks for the fix!

PR #5993, opened 2016-01-08
Update chromedriver version to fix web testing with customer adb port

aluedeke:
See https://github.com/appium/appium-chromedriver/issues/11 for details

@jlipps

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "adm-zip": "~0.4.7",
     "appium-adb": "=1.7.5",
     "appium-atoms": "=0.0.5",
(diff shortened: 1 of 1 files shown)


SrinivasanTarget:
@aluedeke @jlipps I see 2.5.1v is latest.Can't we use that? Is there any reason behind that?

jlipps:
This seems to say 2.20 is the latest: https://sites.google.com/a/chromium.org/chromedriver/downloads

jlipps:
@aluedeke given the impending release of appium 1.5 which I believe has chromedriver 2.20, I'm closing this in favor of that.

PR #6005, opened 2016-01-11
add npm-shrinkwrap

Jonahss:
created on node v5.4.0


Jonahss:
travis will check on different versions of node for me. Thanks travis :)

Issue #6001, opened 2016-01-11
Interacting with app level alert view fails on iOS when 'autoAcceptAlerts=true' capability is set


JaniJegoroff:
Interacting with app level alert views fails with unknown server-side error on iOS when `autoAcceptAlerts = true` capability is set. App level alert view is accepted through Ruby bindings using `alert_accept` method which calls `acceptAlert` method. App level alert view is actually dismissed correctly but error is still raised by `instruments`. Without `autoAcceptAlerts` capability everything is working fine. I just need to use `autoAcceptAlerts` because I need to dismiss system level push notification alert view before interacting with app level alert views.

Steps for more info about the use case
```
1. App is launched
2. System level push notification alert view is shown and dismissed correctly by Appium
3. User selects log out in the app
4. App level alert view is displayed
5. Accept app level alert view with acceptAlert method
6. App level alert view is dismissed but instruments raises an error "An unknown server-side error occurred while processing the command"
```

Environment
```
OS X 10.11.2
iOS 9.2
Xcode 7.2
Appium server v1.4.16
iOS simulator iPhone 6 (9.2)
Ruby 2.1.5p273
ruby_lib 7.0.0
```

Appium server log snippet
```
info: --> POST /wd/hub/session/dce5cd98-b7b5-4aa8-9f2c-7b10eace9028/accept_alert {}
info: [debug] Pushing command to appium work queue: "au.acceptAlert()"
info: [debug] Sending command to instruments: au.acceptAlert()
info: [debug] [INST] 2016-01-11 13:22:57 +0000 Debug: target.frontMostApp().alert().buttons()[1].tap()
info: [debug] [INST] 2016-01-11 13:22:57 +0000 Debug: Waiting for alert to close...
info: [debug] [INST] 2016-01-11 13:22:57 +0000 Debug: Waiting for alert to close...
info: [debug] [INST] 2016-01-11 13:22:58 +0000 Debug: Got new command 18 from instruments: au.acceptAlert()
info: [debug] [INST] 2016-01-11 13:22:58 +0000 Debug: evaluating au.acceptAlert()
info: [debug] Socket data received (93 bytes)
info: [debug] Socket data being routed.
info: [debug] Got result from instruments: {"status":13,"value":"An unknown server-side error occurred while processing the command."}
[...]

PR #6014, opened 2016-01-12
Clarified auto accept/dismiss alert docs

### LINKED ISSUE #4623: Clarify autoAcceptAlerts capability description
According to the documentation http://appium.io/slate/en/master/?java#ios-only the autoAcceptAlerts  being used to automatically handle " iOS privacy access permission alerts" only, but from what I see it actually closes each iOS alert that comes up. I am using java client 2.1 and appium version 1.3.5

scottdixon-zz:
cc @imurchie 

Resolves #4623 #6001


imurchie:
LGTM.

Issue #6048, opened 2016-01-20
autoLaunch=False doesn't works on 1.5.0-beta11


cthorell:
I am trying out 1.5.0-beta11 with iOS and noticed that capability "autoLaunch":False no longer works. When autoLaunch is set to False, Appium is still trying to launch the app when creating a session (as if autoLaunch was set to True).

$ appium --session-override --no-reset --address 127.0.0.1  --command-timeout 99999 
[Appium] Welcome to Appium v1.5.0-beta11 (REV 0ca777677c46bad4429f548706096d859130178f)
[Appium] Non-default server args:
[Appium]   address: '127.0.0.1'
[Appium]   sessionOverride: true
[Appium]   defaultCommandTimeout: 99999
[Appium]   noReset: true
[Appium] Deprecated server args:
[Appium]   --no-reset => --default-capabilities '{"noReset":true}'
[Appium] Default capabilities, which will be added to each request unless overridden by desired capabilities:
[Appium]   noReset: true
[Appium] Appium REST http interface listener started on 127.0.0.1:4723
[HTTP] --> POST /wd/hub/session 
[MJSONWP] Calling AppiumDriver.createSession() with args: [{"deviceName":"iPhone","udid":"b3a6c616f49783d9455f0e1fbf124704a7f08482","autoAcceptAlerts":true,"app":"/Users/clatho/Library/Developer/Xcode/Deri...
[Appium] Creating new IosDriver session
[Appium] Capabilities:
[Appium]   deviceName: 'iPhone'
[Appium]   udid: 'b3a6c616f49783d9455f0e1fbf124704a7f08482'
[Appium]   autoAcceptAlerts: true
[Appium]   app: '/Users/clatho/Library/Developer/Xcode/DerivedData/Test-etihnmpbvpfvcuexozvpioqnylyu/Build/Products/Debug-iphoneos/Test.app'
[Appium]   browserName: ''
[Appium]   autoLaunch: false
[Appium]   newCommandTimeout: 99999
[Appium]   platformName: 'iOS'
[Appium]   noReset: true
[BaseDriver] Session created with session id: a1e37df2-aa3d-400b-aebf-03547fd8dc72
[debug] [iOS] Not auto-detecting udid.
[debug] [Support] Attempting to parse plist file '/Applications/Xcode.app/Contents/Info.plist'
[debug] [Support] Parsed plist file as binary
[...]

PR #6060, opened 2016-01-23
explain that we are not supporting autoLaunch=false anymore

jlipps:


moizjv:
https://github.com/appium/appium/issues/6048

imurchie:
:+1:

Issue #4309, opened 2014-12-23
driver.setLocation() method not working with android device/simulator


hungl:
Hi,

I am trying to set geo location for my app every time it runs a test. I use driver.setLocation(). It works perfectly for iOS, but does not work for Android. It does not work on genymotion as well as Android device. The app starts and shut down right away, and the test fails. I'm running tests on Appium 1.3.4
Here are the logs:

IDE log:

org.openqa.selenium.WebDriverException: An unknown server-side error occurred while processing the command. (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 7 milliseconds
Build info: version: '2.43.1', revision: '5163bceef1bc36d43f3dc0b83c88998168a363a0', time: '2014-09-10 09:43:55'
System info: host: 'Hungs-MacBook-Pro.local', ip: '192.168.1.14', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.9.5', java.version: '1.7.0_71'
Session ID: 09069989-3dd7-4182-8b06-73f304b29344
Driver info: io.appium.java_client.android.AndroidDriver
Capabilities [{automationName=Appium, platform=LINUX, app=/Users/hunglac/Documents/tests-mobile-ui/xxxxx-android/xxxxx.apk, javascriptEnabled=true, appActivity=.ui.splash.SplashActivity, browserName=Android, networkConnectionEnabled=true, desired={automationName=Appium, app=/Users/hunglac/Documents/tests-mobile-ui/xxxxxx-android/xxxxxx.apk, fullReset=true, deviceName=Android Emulator, platformName=Android, appActivity=.ui.splash.SplashActivity, appWaitActivity=.ui.login.LoginActivity,.ui.splash.SplashActivity, appPackage=com.xxxxxx.ordering, noReset=true}, locationContextEnabled=false, appPackage=com.xxx.ordering, noReset=true, platformVersion=4.1, databaseEnabled=false, fullReset=true, deviceName=Android Emulator, platformName=Android, webStorageEnabled=false, appWaitActivity=.ui.login.LoginActivity,.ui.splash.SplashActivity, warnings={}, takesScreenshot=true}]
    at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
[...]

PR #6094, opened 2016-02-02
Set mock location on device - invokes the Android Service from the settings apk

anandsadu:
#### Fix :

The code fix addresses the issue [https://github.com/appium/appium/issues/4309] which is to set the Location on the Device.It is linked with the [https://github.com/appium/io.appium.settings/pull/6] This feature will work once the settings apk is published to npm.
#### Commit summary:

set Location uses telnet for Android emulator and Android Service (in the settings apk )is invoked in the case of device.


moizjv:

anandsadu:
@aluedeke 
> scenario: imagine  emulator and a real device connected to the same host. 

As it is possible to run the test on one connected device(or emulator) with one appium instance,We can have a better check to distinguish between emulator and device something like below code

```
var deviceId = this.adb.curDeviceId;
var portPattern = /emulator-(\d+)/;
var isDevice = !(portPattern.test(deviceId));

    if (isDevice) {
	// Invoke Android Service
	} else {
	// do telnet
	}
```

bootstraponline:
:+1: for checking the current device.

jlipps:
@anandsadu I like this change however it should be done in appium 1.5, which means the change really should live in http://github.com/appium/appium-android-driver. mind porting it there for us to review? we'll be happy to help walk you through the changes.

aluedeke:
@anandsadu i just double checked how we do it @TestObject and we are setting MockLocation in two ways:

first via android.location.LocationManager
second via com.google.android.gms.location.LocationClient from PlayServices

Actually both are required. As users can get GPS Positions from either one or the other. (Actually PlayServices is more common from my experience)

The LocationManager approach should actually work on all devices (emulator, genymotion, real devices -- not all real devices do support this e.g. amazon)

Some sample code:
```
	private void prepareLocationManager() {
		locationManager = (LocationManager) getSystemService(LOCATION_SERVICE);
		try {
			locationManager.addTestProvider(LocationManager.GPS_PROVIDER, false, true, false, false, true, true, true, Criteria.POWER_HIGH,
					Criteria.ACCURACY_FINE);
			locationManager.setTestProviderEnabled(LocationManager.GPS_PROVIDER, true);
		} catch (SecurityException e) {
			Log.e(LOG_TAG, "setting mocklocations is not supported by this device", e);
		}
	}
```

````
private void setLocationManagerLocation(String networkProvider, double latitude, double longitude, double altitude, float bearing,
			float speed, float accuracy) {
		try {
			Location loc = new Location(networkProvider);
			loc.setLatitude(latitude);
			loc.setLongitude(longitude);
			loc.setAccuracy(accuracy);
			loc.setAltitude(altitude);
			loc.setBearing(bearing);
			loc.setSpeed(speed);
			loc.setTime(System.currentTimeMillis());
			setElapsedTime(loc);

			Log.i(LOG_TAG, "setting mock location for location manager " + loc);
			locationManager.setTestProviderLocation(networkProvider, loc);
		} catch (java.lang.SecurityException e) {
			Log.i(LOG_TAG, ACCESS_MOCK_LOCATION + " permission denied. skipping.");
		}
		Util.sleep(LOCATION_UPDATE_INTERVAL_MS);
	}
```

The approach via play services of course works only on devices with play services installed. You can actually include play services into an apk an check if they are installed. So apk is still installable on non playservices devices like the emulator, genymotion, whatever.

```
private void prepareLocationClient() {
		if (GooglePlayServicesUtil.isGooglePlayServicesAvailable(this) != ConnectionResult.SUCCESS) {
			return;
		}

		locationClient = new LocationClient(this, connectionCallback, new OnConnectionFailedListenerStub());
		locationClient.connect();
	}
```
```
private void setLocationClientLocation(String provider, double latitude, double longitude, double altitude, float bearing,
			float speed, float accuracy) {
		if (GooglePlayServicesUtil.isGooglePlayServicesAvailable(this) != ConnectionResult.SUCCESS) {
			return;
		}

		if (connectionCallback.connected == false) {
			return;
		}

		locationClient.setMockMode(true);

		try {
			Location loc = new Location(provider);
			loc.setLatitude(latitude);
			loc.setLongitude(longitude);
			loc.setAccuracy(accuracy);
			loc.setAltitude(altitude);
			loc.setBearing(bearing);
			loc.setSpeed(speed);
			loc.setTime(System.currentTimeMillis());
			setElapsedTime(loc);

			Log.i(LOG_TAG, "setting mock location for location client " + loc);
			locationClient.setMockLocation(loc);
		} catch (java.lang.SecurityException e) {
			Log.i(LOG_TAG, ACCESS_MOCK_LOCATION + " permission denied. skipping.");
		}
		Util.sleep(LOCATION_UPDATE_INTERVAL_MS);
	}
```

We are actually setting mock locations always for both services. So its transparent for the enduser. I think with that approach you can get rid of the telnet stuff.

What do you think?

jlipps:
i'm very much in favor of getting rid of the telnet code

bootstraponline:
> What do you think?

:+1:

anandsadu:
@aluedeke 
Yes, setting the Mock location with `LocationManager` works well on Android Emulators and Genymotion as well except for Android Emulator API 23 since it doesn't have `Allow Mock Location Checkbox`. Tested with the following cases

Galaxy Nexus Android 4.3.1 - API 18 : Able to set the location with the `LocationManager`
Geny Motion 5.0 : Able to set the location with the `LocationManager`

Nexus Android 6  API 23 : This  has  `set mock location app` instead of  `Allow mock location option`.

The LocationManager Implementation works well on devices,android emulators and genymotion emulators irrespective whether PlayServices is installed or not. Hence we can get rid of the telnet code.

> Actually both are required. As users can get GPS Positions from either one or the other.

Do you think we still need to include the play services implementation ?

anandsadu:
@jlipps 
Sure..we'll embed the code changes into appium 1.5. If required we may expect your help.

jlipps:
@anandsadu thank you, if you require help please ping me and @imurchie

anandsadu:
@jlipps @imurchie 
Can you please review the codes changes for invoking the android service from appium 1.5 to set the mock location [https://github.com/appium/appium-android-driver/pull/95] .

imurchie:
Closing in lieu of the `appium-android-driver` PR. This was messed up by the shift to 1.5.

PR #6100, opened 2016-02-04
if the shrinkwrap doesn't exist, don't try to fix it

jlipps:
this was a hiccup i found during our publish process after rming the shrinkwrap in order to generate a new one. i think it's safe to skip fixing the shrinkwrap if it doesn't exist.

please review @imurchie @Jonahss


jlipps:
not sure why we're running cloudbees on this

imurchie:
LGTM.

I'm not sure about CloudBees either.

PR #6138, opened 2016-02-14
sudo necessary? / clearer example path

rompic:
- i had to run it with sudo in order to have the file changed.
- changed the example path to make it clearer


moizjv:

imurchie:
I would rather this remain as it was, and add a note that, depending on where your Appium server is installed, you may need to run the command as superuser. Using `sudo`, though, should _not_ be the default.

rompic:
Fine with me. Maybe the script could also check whether it ran successfully, as there is currently no error if it fails.

imurchie:

ntulele:
Issue still exists:  The plist file is not changed until the script is executed using sudo
Evidence: Executing twice without sudo didn't change plist file's timestamp; and output of the script indicated the EnvVars entry still did not exist.
After execution using sudo, timestamp was updated; and entry reported as existing upon 2nd execution.

PR #6176, opened 2016-02-23
Enable support for xpath find commands on elements

guysmoilov:
Will work in combination with https://github.com/appium/appium-android-bootstrap/pull/17


moizjv:

guysmoilov:
No longer relevant after the big 1.5 merge, please see https://github.com/appium/appium-android-driver/pull/111

PR #6195, opened 2016-02-26
updated ISSUE TEMPLATE

vikramvi:
## Proposed changes

Objective of PR change is encouraging appium users to submit issue with one of the code samples of appium and in case they don't find scenario there, asking them to submit PR for sample-code. This way we will have active sample-code rep and easy to reproduce steps as well.
## Types of changes

updated .md file
## Checklist

NA
## Further comments

NA


moizjv:

imurchie:
Is this really something we want users to do? It seems like a major hurdle that could stop people from submitting issues. While it places more work on those triaging and fixing bugs, what is the real problem with us reproducing?

vikramvi:
@imurchie This can be optional but good to have , also had spoke about this on slack channel with @jlipps and he was happy with this idea.

vikramvi:
Also it will make users more disciplined with bug reproducing steps and help to keep "demo projects" upto date with latest scenarios. IMHO

SrinivasanTarget:
Is Appium logs is not enough? I don't think giving more exercises to users(
reproduce the issue in our app) will add more values. Also I think not
everything is possible to be reproduced in our sample app. As an user I
don't spend time in reproducing it on our sample apps.

SrinivasanTarget:
Also when it comes to iOS there are lot more things like app should be
debug signed and profiles to be installed in our device, which user may not
prefer doing it.

.github/ISSUE_TEMPLATE.md
```diff
@@ -19,3 +19,11 @@ Describe in more detail the problem you have been experiencing, if necessary.
 ## Link to Appium logs
 
 Create a [gist](https://gist.github.com) which is a paste of your _full_ Appium logs, and link them here. Do _not_ paste your full Appium logs here, as it will make this issue very long and hard to read! If you are reporting a bug, _always_ include Appium logs!
+
+
+## Code To Reproduce Issue
+
+Please git clone https://github.com/appium/sample-code and use one of your favourite languages and sample apps to reproduce issue.
+In case similar scenario is missing in sample-code , please submit PR with one of the sample apps provided.
```
jlipps: * I would say `In case a similar scenario...` (add 'a')
* And likewise, `please submit a PR`

.github/ISSUE_TEMPLATE.md
```diff
@@ -19,3 +19,11 @@ Describe in more detail the problem you have been experiencing, if necessary.
...
+
+PLEASE remember with sample code, it's easier to reproduce bug and much faster to fix it.
```
jlipps: `Please remember that, with sample code, ...`

jlipps:
Made a couple language comments. However in agreement with others here I think I would make this an optional encouragement rather than a requirement. I think we'll miss valid bugs if we have strict requirement for this. But we can certainly say that we will be able to investigate faster if they do this!

vikramvi:
@jlipps Please review and merge.  

Thanks , have a great week ahead

PR #6215, opened 2016-03-01
Added You.i Engine support

simongranger:
## Proposed changes

Adding support to You.i Engine into Appium
http://www.youi.tv/youi-engine/
## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._
- [X] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc
- [X] I have signed the [CLA](https://docs.google.com/forms/d/1lOfXRw_0VCk7gYzjj4WLetGu7yelDVo5LWh0z7pGftE/viewform)
- [X] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)
- [X] Any dependent changes have been merged and published in downstream modules
## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #6259, opened 2016-03-09
Remove npm-shrinkwrap from master branch

imurchie:
## Proposed changes

I’m beginning to question why we have `npm-shrinkwrap.json` in our master branch. It makes sense to lock the dependencies down in the release branch, but `master` is a de facto development branch, so we want to be able to test out packages as they are updated. Having `npm-shrinkwrap.json` there means that our local branches are always dirty, since we have to delete it to do any work.

Further, for contributors trying to do stuff in `master` that involves new dependencies (see, for instance, #6215), it means the burden of dealing with shrinkwrap and making sure things are actually sane.

Removing `npm-shrinkwrap.json` in `master` has no consequences for normal users. Release will be locked in their respective branches, and master will remain truly the "source" install location.
## Types of changes

## Checklist

## Further comments

There are currently tests to make sure that we have a shrinkwrap file, and that the optional dependencies are removed from it. I've skipped them here, but in general I'm not sure of their utility. On release we run our tests without it, so they will fail anyway. So as long as we follow the release process it should be there. Or perhaps the test ought to be, somehow, tied to the `npm publish` task, so that we cannot publish without it, but we can run the regular test suite.


scottdixon:
I like this. Can't think of any repercussions.

jlipps:
I like this. However I'd like to make sure that the developers-overview.md is updated if necessary to clarify why this is would only be on the release branch, and instructions about re-shrinkwrapping and pushing that change _only_ on the release branch

imurchie:
Added a little blurb in the release documentation.

jlipps:
LGTM then!

PR #6283, opened 2016-03-15
Update grid-register.js

lucyking:
If register fail, print the failure details friendly to we newbee.


moizjv:

imurchie:
Thanks! This looks great!

PR #6449, opened 2016-04-29
Remove an only-restrictor from tests

imurchie:
Branch says it all.


jlipps:
not stupid, just uncharacteristically hasty :-)

PR #6451, opened 2016-04-30
Allows --desired-caps to be passed a filename

usmonster:
## Proposed changes

Today, we are expected to pass a long string of raw JSON to the appium command line to specify the default desired capabilities of the server. This can be cumbersome, error-prone, and difficult to maintain.

This PR proposes a way to pass either raw JSON or the name of a file containing the JSON that describes the default desired capabilities of the server. (Following the example of the `-dd` alias for `--default-device`, it also adds a `-dc` alias for `--default-capabilities`.)
## Types of changes

## Checklist
- [ ] _Any dependent changes have been merged and published in downstream modules_ **(not applicable)**
## Further comments

Some open questions/concerns about the changes I made:
1. I added some synchronous filesystem access logic to this codebase--is this OK? It's got some minimal checks, but it's mostly "fail-fast" like the existing codebase. Asynchronous access would also be possible, but it would require a bit more refactoring in other areas of the codebase.
2. Is the [filename-vs-JSON test](/appium/appium/pull/6451/files#diff-3fab227e34d65fe9bb2d6ea53eec41cfR737) sufficient? Too complex? Too naive? Too fragile?
3. Since `42` and `true` and `null` are also valid JSON, I added another check to make sure that any valid JSON that parses to something other than a plain object ends up being silently converted to the default empty object. Please let me know if you'd rather not have this check.
4. Should I export the new function I added to the parser module?
5. I added a JSON file to the test directory for the test I added, though not sure if you prefer to put fixtures in a "fixtures" subdirectory, or directly in the [`appium-fake-driver`](/appium/appium-fake-driver/tree/master/test/fixtures) project? (Also, is the filename OK? Should I prefix it by "parser-" or something?)
6. I added a new assertion to an existing test case instead of making an entirely new test case, though it seems to me like the assertions could belong in the same test case. Sill, I would understand if you preferred one assertion per test case. Please let me know.

That's all, for now--thanks!

The --desired-caps flag can now be passed the name of a JSON file as an
argument. This also adds an alias for the flag: `-dc`.


moizjv:

imurchie:
I like this a lot! Thank you!

With regard to your questions:

1. I am ok with the synchronous file access at this point. It will not block any other processing since it is at startup time. Changing it to async would require mucking with `argparse`.
1. Filename-vs-JSON-object: As long as the user does not want a filename that fulfills those conditions, it ought to be ok! I would also not be against just trying everything for a file, and then moving on if it is not.
1. 👍 
1. No. The function is not needed anywhere but in this file.
1. I'd move the json file into a `fixtures` directory, yes. But the name is fine.
1. I'd prefer a different test case for the file test. Perhaps also a test with a string that looks like a file but does not exist, so we make sure we get `{}` out in that case.

Again, thanks!

usmonster:
Thanks for the response @imurchie! I'll make the changes, though one question related to always checking for the file first--should I also change the order of the documentation to make passing the filename be the preferred way (i.e. show the filename examples before the JSON string examples)? Thanks!

imurchie:
I don't think that is necessary. It's not so much that it is preferred, it is just checked first. If they pass in a JSON string, it is probably not likely that they have a file with the same path.

usmonster:
Thanks again @imurchie. Just one more question before I push, this time about this:

> Perhaps also a test with a string that looks like a file but does not exist, so we make sure we get {} out in that case.

In order to keep the behavior "fail-fast" I never added anything to catch the error that `JSON.parse` might throw, which was intentional. It might be worth keeping it as-is, since the user should probably notice when there's a problem. In which case, I should probably actually explicitly throw an error if the valid JSON string-or-file parses to a non-object... What do you think?

imurchie:
Looks good to me! Throwing an error seems when it comes out not an object seems reasonable.

imurchie:
Thanks for this!

jlipps:
yeah, thanks a ton @usmonster! One belated comment:

> I am ok with the synchronous file access at this point. It will not block any other processing since it is at startup time. Changing it to async would require mucking with argparse.

I agree it's OK in this case, but given that it is usually a no-no in our codebase, I'd have just asked for a comment explaining the motivation for using a sync method (and that it happens during startup so can't block sessions, etc...)

imurchie:
I agree.  https://github.com/appium/appium/pull/6462

PR #6455, opened 2016-05-01
Fixes tests' incorrect use of .throw() assertion

usmonster:
## Proposed changes

This fixes some bad test cases. (Some minor refactoring was required for the tests that were failing.)
## Types of changes

## Checklist
- [ ] _I have added tests that prove my fix is effective or that my feature works_ **(not applicable)**
- [ ] _I have added necessary documentation (if appropriate)_ **(not applicable)**
- [ ] _Any dependent changes have been merged and published in downstream modules_ **(not applicable)**
## Further comments

A couple open questions/concerns about the changes I made:
1. I needed to put the parser instances in debug mode to make sure they don't quit the entire process when checking for a thrown exception. I do this in a kind of hacky way, i.e. by directly setting `p.debug = true;`... Unless we can live with that, a better way might be to add an argument to `getParser()` that takes a boolean (or an options hash), which could allow us to pass along the debug flag in a cleaner way. What do you think?
2. I only touched one file in this PR, since I made the same fix in another file as part of #6451. Should I still include the fixes from the other file in this PR, since one never really knows whether or not a PR will be approved, and it logically fits with this one as well?

As pointed out in #6451, there are many test cases that do
`should[.not].throw` instead of `should[.not].throw()`, which is
incorrect and will never fail.

This fixes those incorrect uses of the method as well as refactors/fixes
the tests that turned out to be actually failing.

### DIFF
--- a/test/config-specs.js
+++ b/test/config-specs.js
@@ -64,17 +64,21 @@ describe('Config', () => {
       process = _process;
     });
     describe('checkNodeOk', () => {
-      it('should fail if node is below 0.10', () => {
+      it('should fail if node is below 0.12', () => {
         process.version = 'v0.9.12';
-        checkNodeOk.should.throw;
+        checkNodeOk.should.throw();
+        process.version = 'v0.1';
+        checkNodeOk.should.throw();
+        process.version = 'v0.10.36';
+        checkNodeOk.should.throw();
       });
-      it('should succeed if node is 0.10+', () => {
-        process.version = 'v0.10.0';
-        checkNodeOk.should.not.throw;
+      it('should succeed if node is 0.12+', () => {
+        process.version = 'v0.12.0';
+        checkNodeOk.should.not.throw();
       });
       it('should succeed if node is 1.x', () => {
         process.version = 'v1.0.0';
-        checkNodeOk.should.not.throw;
+        checkNodeOk.should.not.throw();
       });
     });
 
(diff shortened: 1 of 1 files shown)


moizjv:

test/config-specs.js
```diff
@@ -64,17 +64,21 @@ describe('Config', () => {
       process = _process;
     });
     describe('checkNodeOk', () => {
-      it('should fail if node is below 0.10', () => {
+      it('should fail if node is below 0.12', () => {
         process.version = 'v0.9.12';
-        checkNodeOk.should.throw;
+        checkNodeOk.should.throw();
+        process.version = 'v0.1';
```
usmonster: Please let me know if you'd prefer to separate these new assertions into separate test cases.

imurchie:
I'm fine with just setting the debug flag in the test, but I also wouldn't mind if the `getParser` method were updated.

As for the parts in a different PR: I would not repeat them. It would make merging more difficult in the long run.

test/config-specs.js
imurchie: This is fine.

imurchie:
I'm good with this as-is. Thanks for taking this on!

jlipps:
👍  thanks @usmonster. Guess we forgot to make sure our tests failed before they passed!

imurchie:
Yup. The problem with post hoc tdd.

PR #6650, opened 2016-07-11
little fix to Zh trans

lucyking:
_predicate_ means "the element should be" here,"断言" may be better.

> 谓词： In some systems of grammar, the predicate of a clause is the part of it that is not the subject. For example, in "I decided what to do," "decided what to do" is the predicate.
> 
> 断言：If you say that one situation is predicated on another, you mean that the first situation can be true or real only if the second one is true or real.

Reviewers:
 @jlipps, @lihuazhang

### DIFF
--- a/docs/cn/writing-running-appium/ios_predicate.cn.md
+++ b/docs/cn/writing-running-appium/ios_predicate.cn.md
@@ -1,6 +1,6 @@
-## iOS 谓词(Predicate)
+## iOS 断言(Predicate)
 
-在查看 *'-ios uiautomation'* 搜索策略时了解 **谓词(Predicate)** 十分必要。 [UIAutomation JavaScript API](https://developer.apple.com/library/ios/documentation/DeveloperTools/Reference/UIAutomationRef/)有下列几种非常有用的方法:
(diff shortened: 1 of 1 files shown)


triager:

imurchie:
I don't know any Chinese, so I can't say. But if this makes it more clear, I support it. :)

lucyking:
What fllowing echo word is the English explanation.

lihuazhang:
I prefer 断言.

chenhengjie123:
Hi, I'm the person who translate this doc to Chinese. 

Firstly, I agree that "谓词" is not suitable for the word "predicate". Apple do not mean this in Chinese

However, for most IT Engineers in China, "断言" is equal to the word "assert". It may confuse some of them if it means something else.

I prefer "断定" is better. Or may be just leave it as "predicate" but add some words to explain it.

jlipps:
I also can't opine on this; IMO the way to understand 'ios predicate' in English is the same concept as a 'locator' in Selenium. It is a string which tells you where to find something. Does that help?

jlipps:
@chenhengjie123 @lihuazhang @lucyking we defer to your judgment as we are not chinese speakers. do you all have a consensus around this yet?

jlipps:
closing since no consensus

PR #6745, opened 2016-08-18
make XCUITest driver the default for ios tests with platformVersion >= 10

jlipps:
cc @imurchie


imurchie:
Looks good! And good idea!

SrinivasanTarget:
@jlipps @imurchie I like this idea 👍 But one query, if driver automatically switches to XCUITestDriver based on platformVersion then why do we need to specify the ```automationName``` in desired capabilities?

imurchie:
You can use XCUITest for 9.3 above, so that would be one reason. Also
explicit is better than implicit.

PR #6822, opened 2016-09-09
add reboot option for android emulator

alizelzele:
## Proposed changes

If reboot option activated, it will always wipe data of emulator on start, and close emulator at the end. it is required for [support reboot parameter for android](https://github.com/appium/appium-android-driver/pull/180)
## Types of changes
## Checklist

### DIFF
--- a/docs/en/writing-running-appium/default-capabilities-arg.md
+++ b/docs/en/writing-running-appium/default-capabilities-arg.md
@@ -52,3 +52,4 @@ Appium 1.5 does away with most CLI flags; the remainder can be converted into JS
 | --keep-keychains          | keepKeyChains           |
 | --localizable-strings-dir | localizableStringsDir   |
 | --show-ios-log            | showIOSLog              |
+| --reboot                  | reboot                  |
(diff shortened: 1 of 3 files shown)


triager:

triager:

jlipps:
@alizelzele we're moving away from having command line parameters in addition to capabilities. Things which are relevant for individual sessions should be capabilities only, not command line flags. So I think we should just document it in caps and avoid adding a `--reboot` flag.

alizelzele:
This flag is for appium to kill emulator at end of each session and open a new wiped emulator at start of each session. It is much better if it is defined as command line parameter, instead of individual sessions.

If we define it as a cap, one test can have it and kill the Emulator at the end, so all the test should have this parameter set to be able to open the emulator again (setting the `avd name` parameter will work too, but this way it will not close the emulator at the end, and of course will not wipe it).

I think setting it as a flag make sense in this case.

jlipps:
OK, makes sense to me then. Ready to merge whenever the upstream functionality is available.

alizelzele:
@jlipps every other PR required for this functionality is already merged, but because of versions defined in `package.json` it needs `npm update` to fetch the currect version..

jlipps:
great, thanks!

SrinivasanTarget:
@jlipps @imurchie I didn't understand why ```--reboot``` is listed in both Server flags and default capabilities flag? Isn't it a duplicate?

alizelzele:
@SrinivasanTarget This parameter should not be used as a capability, but as a Server flag. You are right, it should not be added as default capabilities. 😞 

We even discussed it here, but it some how end up in documents.

PR #6898, opened 2016-09-27
Add server parameter for WebDriverAgent local port

domuro:
## Proposed changes

Enable specifying the WebDriverAgent port as a server parameter, so multiple Appium XCUI tests can be run concurrently.
## Types of changes

## Checklist

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -351,6 +351,15 @@ const args = [
     help: '(IOS-only) Local port used for communication with ios-webkit-debug-proxy'
   }],
 
+  [['--webdriveragent-port'], {
+    defaultValue: 8100,
+    dest: 'wdaLocalPort',
+    required: false,
+    type: 'int',
+    example: "8100",
+    help: '(IOS-only, XCUITest-only) Local port used for communication with WebDriverAgent'
+  }],
+
   [['-dc', '--default-capabilities'], {
     dest: 'defaultCapabilities',
     defaultValue: {},


triager:

triager:

imurchie:
This is a desired capability. Why do you want it to be a server flag?

yaphatak:
@imurchie I think it will be better to have this so that people can launch a node per device with different port. This is something similar to what we have for `--chromedriver-port` or `selendroid`

lib/parser.js
```diff
@@ -351,6 +351,15 @@ const args = [
     help: '(IOS-only) Local port used for communication with ios-webkit-debug-proxy'
   }],
 
+  [['--web-driver-agent-port'], {
```
imurchie: I would prefer `webdriveragent-port`.

lib/parser.js
```diff
@@ -351,6 +351,15 @@ const args = [
...
+    defaultValue: 8100,
+    dest: 'wdaLocalPort',
+    required: false,
+    type: 'int',
+    example: "8100",
+    help: '(IOS-only) Local port used for communication with WebDriverAgent'
```
imurchie: This should include info that it is XCUITest only, not just iOS only.

imurchie (changes requested):
Thanks for this! A couple of changes, but otherwise good.

imurchie (approved):

imurchie:
Awesome! Thanks for this!

PR #7015, opened 2016-10-14
graceful handling of SIGTERM and SIGINT such that return codes are mo…

montdidier:
## Proposed changes

Upon receiving SIGTERM appium exits with code 143 which is line with the documentation presented [here](https://nodejs.org/api/process.html#process_signal_events). However it is not an expected result from the point of view of many existing system tools. A SIGTERM would usually result in a graceful shutdown and an exit code of 0. 

I've installed basic signal handers for SIGTERM and SIGINT that shutdown express when they are met. The code then exits gracefully with an exit code of 0.

The specific problem I encountered was met when managing appium using launchctl on OSX which I use to manage appium process life. Shutdown is performed via SIGTERM and launchctl expects an exit code of 0 as is typical. 
## Types of changes

## Checklist
## Further comments

NA

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -132,6 +132,17 @@ async function main (args = null) {
     await server.close();
     throw err;
   }
+
+  process.on('SIGINT', async function () {
+    logger.info(`Received SIGINT - shutting down`);
+    await server.close();
+  });
+
+  process.on('SIGTERM', async function () {
+    logger.info(`Received SIGTERM - shutting down`);
+    await server.close();
+  });
+
   logServerPort(args.address, args.port);
 
   return server;


triager:

triager:

lib/main.js
```diff
@@ -132,6 +132,17 @@ async function main (args = null) {
     await server.close();
     throw err;
   }
+
+  process.on('SIGINT', function() {
+    logger.info(`Received SIGINT - shutting down`);
+    server.close();
+  });
+
+  process.on('SIGTERM', function() {
```
imurchie: Same.

imurchie (changes requested):
Two small style changes (I need to update our eslint config to make that an error, not a warning).

But thi is great! Thank you for taking the time to improve the system!

lib/main.js
```diff
@@ -132,6 +132,17 @@ async function main (args = null) {
     await server.close();
     throw err;
   }
+
+  process.on('SIGINT', function() {
+    logger.info(`Received SIGINT - shutting down`);
+    server.close();
```
jlipps: we `await server.close()` earlier, so we should probably convert this into an async function for consistency.

jlipps (changes requested):

lib/main.js
montdidier: No problem will address all comments.

imurchie:
Looks good to me! Could you, as a last step, squash the two commits? Then we can get this in!

imurchie (approved):

montdidier:
squash done. anything else?

imurchie:
I'm good with this. @jlipps?

jlipps (approved):

jlipps:
LGTM now, thanks

PR #7154, opened 2016-11-02
update license and add note about appium agile process

jlipps:
what do you think @imurchie @dpgraham ?


imurchie:
Looks good to me!

dpgraham:
Looks good

PR #7547, opened 2016-12-15
How Will Appium Support iOS 10?

ShirleyPerkins:
[](http://blog.testproject.io/2016/12/12/appium-support-ios-10/
)


triager:

triager:

PR #7767, opened 2017-01-27
Add swiping documentations

dpgraham:
I kept it simple and just instructed the reader to replace 'mobile: scroll' with 'mobile: swipe'. Since the two have the exact same API, no sense in maintaining both sets of instructions.

### DIFF
--- a/docs/en/writing-running-appium/touch-actions.md
+++ b/docs/en/writing-running-appium/touch-actions.md
@@ -191,7 +191,7 @@ $driver->executeScript("mobile: scroll", $params);
 
 **Swiping**
 
-This is an XCTest specific method that is similar to scrolling (for reference, see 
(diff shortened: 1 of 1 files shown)


triager:

docs/en/writing-running-appium/touch-actions.md
```diff
@@ -189,6 +189,14 @@ $params = array(array('direction' => 'down', 'element' => element.GetAttribute("
 $driver->executeScript("mobile: scroll", $params);
 ```
 
+**Swiping**
+
+This is an XCTest specific method that is similar to scrolling (for reference, see
```
jlipps: Maybe instead of `XCTest` say `XCUITest driver` to tie it more strongly to how we call things on the Appium side? Otherwise LGTM

triager:

PR #8018, opened 2017-03-10
Document 'startIWDP' desired capability

dpgraham:


triager:

triager:

jlipps (commented):
can we give a little more context? Something like:

> Run an instance of ios_webkit_debug_proxy automatically inside of Appium rather than requiring that the user run it on their own outside of Appium.

dpgraham:
I updated it. I agree it's better to give the context of why you'd use `startIWDP`.

PR #8479, opened 2017-05-17
add event timing docs

jlipps:
figured we should have this documented somewhere!


imurchie (approved):
Excellent!

Do you want to add anything about the parser?

jlipps:
yes, good point. added!

imurchie (approved):

Issue #8564, opened 2017-05-30
Feature request for Android driver: promote remoteAdbHost option


serggl:
From the `appium-adb` code I see that there is an option called `remoteAdbHost` ([lib/adb.js](https://github.com/appium/appium-adb/blob/fb52f6fe12a2f1a8e0e96d631ffdfe5ef403151c/lib/adb.js#L41)) but it is not currently promoted in base driver ([lib/desired-caps.js](https://github.com/appium/appium-android-driver/blob/b9f6a9501c6d0cbc8b7919f4aebb89e21aa0a9f6/lib/desired-caps.js)). Same time `adbPort` option is available... 

`remoteAdbHost` option could be very helpfull in docker environment where emulator could be running in a container separate from the appium server (Which is what I currently trying to setup), but  appium now is filtering out this option.
And `adb tcpip / adb connect` with a `deviceName` capability  trick does not really works for an emulator running an a docker container.

Is there a specific reason why appium has a ADB's port option configurable but not the host? If no, can we make it available?

Thank you.

dpgraham:
@serggl This is a good idea. The use-case you described is probably a fairly common one. 

I'll add this to our triaged pipeline. Not certain when we'll get to it, but we do accept PR's.

serggl:
Here is the PR (https://github.com/appium/appium-android-driver/pull/241), relatively simple.
However I've found that just having this option set is not enough to make appium communicate to the emulator betweeen the docker containers.
The problem is in `AppiumBootstrap` app. It still needs an appium container's port 4724 (bootstrap port) to be forwarded to the emulator container. Solved that with the `iptables` setup.

Here's my compose file for the once who interested:
```yaml
version: "2"

services:
  selenium:
    container_name: selenium
    image: selenium/standalone-chrome

  emulator:
    container_name: emulator
    image: jenkins-appium-emulator
    command: ./run_emulator.sh

  appium:
    container_name: appium
    image: jenkins-appium-emulator
    privileged: true
    command: ./run_appium.sh

  fuse-box:
    container_name: fuse-box
    image: ...# image running my specs
```
`run_appium.sh` script:
```bash
iptables -t nat -A OUTPUT -m addrtype --src-type LOCAL --dst-type LOCAL -p tcp --dport 4700 -j DNAT --to-destination $(getent hosts emulator | awk '{ print $1 }')
iptables -t nat -A POSTROUTING -m addrtype --src-type LOCAL --dst-type UNICAST -j MASQUERADE
[...]

PR #8613, opened 2017-06-06
Advanced concept document: running android tests with docker

serggl:
inspired by #8564

### DIFF
--- a/docs/en/advanced-concepts/android-testing-with-docker.md
+++ b/docs/en/advanced-concepts/android-testing-with-docker.md
@@ -1,0 +1,66 @@
+## Android Tests With Docker
+
+When time comes to run appium tests on your own CI (like Jenkins) you would probably think about using docker.
+In order to run Android tests you will also need to have an emulator. Fortunately you can find official docker image that has both Appium and Android Emulator bundled inside: 
+https://hub.docker.com/r/appium/appium-emulator/
+ 
+All you need to know is just 2 simple tricks:
+1. Run ADB server using `-a` switch in the docker container running emulator. ADB will listen to all connected network interfaces, so you can connect emulator from another running container
+2. You will need to setup port forwarding from the container running Appium to the container running emulator. 
+Appium uses a `bootstrap-port` to setup communication to the emulator and it will bind to the localhost by default.
+But since emulator running in another container, you will need to bind that local `bootstrap-port` to the same port on the emulator container
+
+Here is a sample docker compose file that does all these tricks:
+```yaml
+version: "2"
+
+services:
+  emulator:
+    container_name: emulator
+    image: appium/appium-emulator
+    command: >
+          bash -c "adb -a nodaemon server &
+                   emulator64-arm -avd Nexus -no-boot-anim -no-window -noaudio -gpu off"
+
+  appium:
+    container_name: appium
+    image: appium/appium-emulator
+    privileged: true
+    command: >
+      bash -c "iptables -t nat -A OUTPUT -m addrtype --src-type LOCAL --dst-type LOCAL -p tcp --dport 4700 -j DNAT --to-destination $(getent hosts emulator | awk '{ print $1 }') && 
+               iptables -t nat -A POSTROUTING -m addrtype --src-type LOCAL --dst-type UNICAST -j MASQUERADE &&
+               sysctl -w net.ipv4.conf.all.route_localnet=1 && 
+               echo "net.ipv4.conf.all.route_localnet=1" >> /etc/sysctl.conf &&
+               appium --bootstrap-port 4700"
+               
+  app:
+    container_name: app
+    image: [YOUR APP IMAGE WITH TESTS]      
+```
+
+Here is a caps file for your app's image that uses this setup (notice the `remoteAdbHost` option and `server_url` which refer docker containers): 
+
+```
+[caps]
+platformName = 'Android'
+deviceName = 'Android Emulator'
+appPackage = '[YOUR APP PACKAGE]'
+appActivity = '[YOUR APP ACTIVITY]'
+remoteAdbHost = 'emulator'
+
+[appium_lib] 
+server_url = "http://appium:4723/wd/hub"
+```
+
+Finally, here is how a CI script may look like:
+```bash
+# run emulator and appium
+docker-compose up -d
+
+# copy APK to the emulator and install
+docker cp YOUR_APK emulator:/sdcard/YOUR_APK
+docker exec emulator bash -c 'adb wait-for-device && adb install -r /sdcard/YOUR_APK'
+
+# run tests (in this sample it is rspec)
+docker exec -t app bash -c 'rspec'
+```


triager:

triager:

imurchie:
@SrinivasanTarget might be a good reviewer.

docs/en/advanced-concepts/android-testing-with-docker.md
```diff
@@ -0,0 +1,66 @@
...
+  emulator:
+    container_name: emulator
+    image: appium/appium-emulator
+    command: >
+          bash -c "adb -a nodaemon server &
+                   emulator64-arm -avd Nexus -no-boot-anim -no-window -noaudio -gpu off"
+
+  appium:
+    container_name: appium
+    image: appium/appium-emulator
```
SrinivasanTarget: @serggl I didn't understand why are you using ```appium/appium-emulator``` here and what is the idea behind running appium in one container and emulator in another container? Is it to run tests parallely in a single device/emulator? If you have looked at ```appium/appium-emulator``` it runs both appium and android emulator in a single container. Also FYI, if you look at recent release of https://github.com/appium/appium-docker-android/releases/tag/2.0-barcelona, we have deprecated official emulator image in favor of https://github.com/butomo1989/docker-android.  We will be adding a detailed readme links to advanced concepts section once we use docker-appium as a base image in docker-android. Probably you can have a look at [this](https://github.com/butomo1989/docker-android/blob/master/docker-compose.yml) and [this](https://github.com/SrinivasanTarget/DockerAppium/blob/master/docker-compose.yml). 
Ping @butomo1989

docs/en/advanced-concepts/android-testing-with-docker.md
```diff
@@ -0,0 +1,66 @@
...
+
+Here is a caps file for your app's image that uses this setup (notice the `remoteAdbHost` option and `server_url` which refer docker containers): 
+
+```
+[caps]
+platformName = 'Android'
+deviceName = 'Android Emulator'
+appPackage = '[YOUR APP PACKAGE]'
+appActivity = '[YOUR APP ACTIVITY]'
+remoteAdbHost = 'emulator'
```
SrinivasanTarget: As per doc i see ```remoteAdbHost``` should be an ip address.

docs/en/advanced-concepts/android-testing-with-docker.md
```diff
@@ -0,0 +1,66 @@
...
+
+[appium_lib] 
+server_url = "http://appium:4723/wd/hub"
```
SrinivasanTarget: same here. Should be ip address of container where appium is running.

docs/en/advanced-concepts/android-testing-with-docker.md
serggl: @SrinivasanTarget the idea behind running emulator and appium separately is very simple. Emulator takes a lot of time to start and become ready when running in docker (tens of minutes on our CI), so it make sense to start this container only once and then connect to it on each CI run. Also I found that appium server can often became non-responsive if for some reason I decide to cancel running rspec test session and start a new one. I have to restart appium server quite often, so I would not like to run it as a daemon service in docker. Unlike emulator it is better to start this service on each CI test run.
As for the docker image - I took it here just for a sample, since it contains both appium and emulator. Speaking honestly, I run a custom image on my own. So we can change that part of a doc if you like

docs/en/advanced-concepts/android-testing-with-docker.md
serggl: Not especially. `adb` accepts both IP and host name. So it works very well here, since it is running in docker

docs/en/advanced-concepts/android-testing-with-docker.md
serggl: Docke resolves this to IP on its own. `appium` is a network alias from the compose file

docs/en/advanced-concepts/android-testing-with-docker.md
serggl: `emulator ` is actually a network alias from the compose file

docs/en/advanced-concepts/android-testing-with-docker.md
butomo1989: @serggl You can check https://github.com/butomo1989/docker-android. If your CI is linux machine, you can use x86 system image for the fastest performance. for x86 system image it takes around 10-40 sec and arm system image it takes around 10 min like you mentioned (all depends on computer specification).  
and we run appium server in separate thread, so you don't need to run appium server separately in different container. In case you want to combine all containers in one hub, you can use selenium hub.

docs/en/advanced-concepts/android-testing-with-docker.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+Finally, here is how a CI script may look like:
+```bash
+# run emulator and appium
+docker-compose up -d
+
+# copy APK to the emulator and install
+docker cp YOUR_APK emulator:/sdcard/YOUR_APK
+docker exec emulator bash -c 'adb wait-for-device && adb install -r /sdcard/YOUR_APK'
```
butomo1989: You don't need to install the apk manually. use capability ''app", and the appium will install the apk for your before run the tests. example: 'app': '/sdcard/YOUR_APK'

docs/en/advanced-concepts/android-testing-with-docker.md
serggl: @butomo1989 thats probably a subject for another thread, but we tried x86 images from your repository but they do not work on our Jenkins running on Ubuntu 14.04. Emulator does not start for some reason 🤷‍♂️ 

Again, my main concern here is that appium and emulator are better to be isolated from each other. Which exact docker images we use - that's kind of a personal decision, all up to developers.

docs/en/advanced-concepts/android-testing-with-docker.md
serggl: agree. That was copy&pasted from our CI, not to good for the documentation

docs/en/advanced-concepts/android-testing-with-docker.md
butomo1989: If you want, you can open issue and discuss there. And regarding run appium and emulator separately, I don't feel it is necessary, because it seems you have appium server in one machine and have emulator on different machine and connect both of them which make more complex. but again it is my opinion. :)

docs/en/advanced-concepts/android-testing-with-docker.md
SrinivasanTarget: Ok, if that is the case that needs to be explicitly documented for our users.

docs/en/advanced-concepts/android-testing-with-docker.md
SrinivasanTarget: @serggl I also go with @butomo1989 thoughts here, i would prefer to use appium and emulator in same container. Running it separately and hooking it up every time with another container would be an overhead which i don't prefer or suggest to docker-appium users, but based on your solution posted here i've got few ideas,
1) Passing server arguments to appium server using a json
2) Checking the stability of spinning more emulator containers for a long run as like i did for docker-selenium [here](https://github.com/zalando/zalenium/issues/116#issuecomment-305246957)
3) Adding more samples/tests to suggest users.
 
@butomo1989 @serggl if you have any other ideas from here, please add on.

andrcuns: Does it actually make sense to use emulator in container with arm image in case one actually intends to leave emulator running? I think it's a bit counter productive to use docker wrapper in such case and for example something like [this](https://wiki.jenkins-ci.org/display/JENKINS/Android+Emulator+Plugin) would be IMHO better suited for such use case. Also I'm very skeptical about using arm based emulator images in any sort of CI setup since it's not only slow to boot, it is slow as hell in general as well and not sure how fast/stable tests would be. As for x86 images running on Jenkins slave, if it's AWS host, make sure KVM support is there on instance, maybe that is why it doesn't start.

jlipps:
Hi @serggl are you planning on making any updates based on the reviewers' comments?

serggl:
@jlipps I have a feeling that reviewers are not very happy with this concept, so you can probably close this one

jlipps:
OK, I will do so for now. Thanks!

PR #8640, opened 2017-06-08
Espresso Driver added to appium

dpgraham:
I tested this in Appium Desktop and confirmed it work.

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -12,6 +12,7 @@ import { XCUITestDriver } from 'appium-xcuitest-driver';
 import { YouiEngineDriver } from 'appium-youiengine-driver';
 import { WindowsDriver } from 'appium-windows-driver';
 import { MacDriver } from 'appium-mac-driver';
+import { EspressoDriver } from 'appium-espresso-driver';
 import B from 'bluebird';
 import util from 'util';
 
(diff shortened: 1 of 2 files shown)


triager:

triager:

lib/appium.js
```diff
@@ -58,6 +59,9 @@ class AppiumDriver extends BaseDriver {
       } else if (caps.automationName.toLowerCase() === 'youiengine') {
         // but if we do and it is 'YouiEngine', act on it
         return YouiEngineDriver;
+      } else if (caps.automationName.toLowerCase() === 'espresso') {
+        log.warn('The Appium Espresso Driver is currently in beta and meant for experimental usage. Do not use in production.');
```
jlipps: Since people don't typically use Appium "in production" in a meaningful sense anyway, maybe:

> The Appium Espresso driver is currently in early beta and meant only for experimental usage. Its API is not yet complete or guaranteed to work. Please report bugs to the Appium team on GitHub.

Or something?

jlipps (approved):
good idea to include a beta warning

package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "appium-android-driver": "1.x",
     "appium-base-driver": "2.x",
+    "appium-espresso-driver": "^1.0.0-beta.1",
```
imurchie: How does npm semver work with weird version "numbers". We could, instead, just specify "beta" to get whatever is most recently tagged as beta.

lib/appium.js
dpgraham: Yeah that sounds good

package.json
dpgraham: Not sure. I'll have to look into that. I would like it if it always was the latest beta.

dpgraham:
Changes made.

PR #8907, opened 2017-07-24
Pass the list of running sessions to deleteSession

mykola-mokhnach:
## Proposed changes

We already provide the list of active sessions for the current driver class in _createSession_ method. This list will be also useful for _deleteSession_ method if we want parallel sessions handling to work properly.

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -239,9 +239,11 @@ class AppiumDriver extends BaseDriver {
   async deleteSession (sessionId) {
     try {
       if (this.sessions[sessionId]) {
-        const curSessionsData = this.curSessionDataForDriver(this.sessions[sessionId].constructor);
-        const curDriverData = this.sessions[sessionId].driverData;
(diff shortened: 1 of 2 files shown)


lib/appium.js
```diff
@@ -239,7 +239,11 @@ class AppiumDriver extends BaseDriver {
   async deleteSession (sessionId) {
     try {
       if (this.sessions[sessionId]) {
-        await this.sessions[sessionId].deleteSession();
+        const curConstructorName = this.sessions[sessionId].constructor.name;
+        const otherSessionsData = _.toPairs(this.sessions)
```
jlipps: yes, this is better than testing on object equality

jlipps (approved):

jlipps:
(should probably figure out a way to add a test as well)

mykola-mokhnach:
I've updated the existing unit test. Unfortunately, FakeDriver does not allow to create parallel sessions in the current implementation. But this is another issue, that we will need to address later.

imurchie (approved):

PR #8925, opened 2017-07-27
Include the information about pending drivers being currently at creation stage into createSession argument

mykola-mokhnach:
## Proposed changes

Current algorithm only passes the information for drivers that already have sessions, but this does not work as expected for the case when several drivers are being created at the same time and we are trying to figure out which of them is the first one in the pipeline. That is why I also include the data of other pending drivers, whose state is 'session creation'.

## Types of changes

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -27,6 +27,8 @@ class AppiumDriver extends BaseDriver {
     this.args = args;
 
     this.sessions = {};
+
+    this.pendingDrivers = {};
   }
 
   sessionExists (sessionId) {
@@ -176,15 +178,26 @@ class AppiumDriver extends BaseDriver {
       }
     }
 
-    let curSessions;
+    let runningDriversData;
     try {
-      curSessions = this.curSessionDataForDriver(InnerDriver);
+      runningDriversData = this.curSessionDataForDriver(InnerDriver);
     } catch (e) {
       throw new errors.SessionNotCreatedError(e.message);
     }
 
+    let innerSessionId, dCaps;
     let d = new InnerDriver(this.args);
-    let [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, curSessions);
+    try {
+      if (_.isArray(this.pendingDrivers[InnerDriver.name])) {
+        this.pendingDrivers[InnerDriver.name].push(d);
+      } else {
+        this.pendingDrivers[InnerDriver.name] = [d];
+      }
+      const otherPendingDriversData = this.pendingDrivers[InnerDriver.name].filter((drv) => d !== drv).map((drv) => drv.driverData);
+      [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, [...runningDriversData, ...otherPendingDriversData]);
+    } finally {
+      _.pull(this.pendingDrivers[InnerDriver.name], d);
+    }
     this.sessions[innerSessionId] = d;
 
     // this is an async function but we don't await it because it handles
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -188,12 +188,9 @@ class AppiumDriver extends BaseDriver {
     let innerSessionId, dCaps;
     let d = new InnerDriver(this.args);
     try {
-      if (_.isArray(this.pendingDrivers[InnerDriver.name])) {
-        this.pendingDrivers[InnerDriver.name].push(d);
-      } else {
-        this.pendingDrivers[InnerDriver.name] = [d];
-      }
-      const otherPendingDriversData = this.pendingDrivers[InnerDriver.name].filter((drv) => d !== drv).map((drv) => drv.driverData);
+      this.pendingDrivers[InnerDriver.name] = this.pendingDrivers[InnerDriver.name] || [];
+      const otherPendingDriversData = this.pendingDrivers[InnerDriver.name].map((drv) => drv.driverData);
+      this.pendingDrivers[InnerDriver.name].push(d);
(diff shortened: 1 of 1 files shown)


lib/appium.js
```diff
@@ -176,15 +178,26 @@ class AppiumDriver extends BaseDriver {
...
+      runningDriversData = this.curSessionDataForDriver(InnerDriver);
     } catch (e) {
       throw new errors.SessionNotCreatedError(e.message);
     }
 
+    let innerSessionId, dCaps;
     let d = new InnerDriver(this.args);
-    let [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, curSessions);
+    try {
+      if (_.isArray(this.pendingDrivers[InnerDriver.name])) {
```
imurchie: Will `this.pendingDrivers[InnerDriver.name]` ever be anything other than `undefined` or an `Array`? This could just be
```js
this.pendingDrivers[InnerDriver.name] = this.pendingDrivers[InnerDriver.name] || [];
```

imurchie (commented):
One small comment. Otherwise seems like a good idea.

lib/appium.js
```diff
@@ -27,6 +27,8 @@ class AppiumDriver extends BaseDriver {
     this.args = args;
 
     this.sessions = {};
+
+    this.pendingDrivers = {};
```
jlipps: can we standardize on `sessions` rather than `drivers`? unless we really mean a singleton driver class rather than a driver instance (corresponding to a session).

willing to discuss other options too.

lib/appium.js
```diff
@@ -176,15 +178,26 @@ class AppiumDriver extends BaseDriver {
...
+    let innerSessionId, dCaps;
     let d = new InnerDriver(this.args);
-    let [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, curSessions);
+    try {
+      if (_.isArray(this.pendingDrivers[InnerDriver.name])) {
+        this.pendingDrivers[InnerDriver.name].push(d);
+      } else {
+        this.pendingDrivers[InnerDriver.name] = [d];
+      }
+      const otherPendingDriversData = this.pendingDrivers[InnerDriver.name].filter((drv) => d !== drv).map((drv) => drv.driverData);
```
jlipps: would I be correct in assuming that this line of code is attempting to find all the pending sessions other than the one we just added to the hash?

if so, why not just discard the `filter` and do the `map` _before_ we push the current driver? i.e.,

```js
try {
  if (!this.pendingSessions[InnerDriver.name]) {
    this.pendingSessions[InnerDriver.name] = [];
  } 
  const otherPendingSessionsData = this.pendingSessions[InnerDriver.name].map(d => d.driverData);
  this.pendingSessions[InnerDriver.name].push(d);
  ...
}
```

etc

lib/appium.js
```diff
@@ -176,15 +178,26 @@ class AppiumDriver extends BaseDriver {
...
+      [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, [...runningDriversData, ...otherPendingDriversData]);
```
jlipps: since we're just combining pending drivers and running drivers in the call to `createSession`, why even have a different container for pending drivers? why not just temporarily add them to the running drivers set?

lib/appium.js
mykola-mokhnach: but we do keep driver instances mapping here rather than sessions

lib/appium.js
mykola-mokhnach: yep, thanks for the idea. I cannot get used to node async model after Java

lib/appium.js
mykola-mokhnach: running drivers set is mapped by session ids and we don't have these for pending drivers yet

jlipps: ok that makes sense. however we're pushing them all into one array; is that the intention? how will the driver itself separate out these different types of sessiondata, or does it need to?

lib/appium.js
jlipps: hmm hmm ok that makes sense.

lib/appium.js
```diff
@@ -176,15 +178,23 @@ class AppiumDriver extends BaseDriver {
...
     }
 
+    let innerSessionId, dCaps;
     let d = new InnerDriver(this.args);
-    let [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, curSessions);
+    try {
+      this.pendingDrivers[InnerDriver.name] = this.pendingDrivers[InnerDriver.name] || [];
+      const otherPendingDriversData = this.pendingDrivers[InnerDriver.name].map((drv) => drv.driverData);
+      this.pendingDrivers[InnerDriver.name].push(d);
+      [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, [...runningDriversData, ...otherPendingDriversData]);
```
jlipps: rewriting this comment in case it got lost in new diff:  we're pushing them all the driver data sets into one array; is that the intention? how will the driver itself separate out these different types of driver data, or does it need to?

lib/appium.js
mykola-mokhnach: For now we only return an empty object as session data and what we really care about is count of items in this array.
Nevertheless, if we really need to know more details for each driver type then we could simply include this information into _driverData_ structure itself

lib/appium.js
jlipps: 👍 the idea behind driver data was for drivers to put the kinds of resources they are using in their driverData field, like ports, etc..., so that other drivers would know not to use that port or this other tmpDir, or whatever.

imurchie (approved):

PR #8955, opened 2017-08-01
Update custom WDA server initialization script

mykola-mokhnach:
## Proposed changes

Update the example script to reflect the recent changes in the integration with WDA

### DIFF
--- a/docs/en/advanced-concepts/wda-custom-server.md
+++ b/docs/en/advanced-concepts/wda-custom-server.md
@@ -3,9 +3,9 @@
 Appium for iOS uses [Facebook's WebDriverAgent](https://github.com/facebook/WebDriverAgent)
 as the automation backend. This backend is based on Apple's XCTest framework and shares all the
 known problem that are present in XCTest. For some of them we have workarounds, but there
-are some that are hardly possible to workaround. The approach described in this article
-enables you to have full control over how WDA is built, managed, and run on the device. This way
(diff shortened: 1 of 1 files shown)


imurchie (approved):
Looks good.

docs/en/advanced-concepts/wda-custom-server.md
```diff
@@ -28,6 +28,12 @@ cd /usr/local/lib/node_modules/appium/node_modules/appium-xcuitest-driver/WebDri
 ./Scripts/bootstrap.sh -d
 ```
 
+Also, it might be necessary to create an empty folder for WDA resources:
+
+```bash
+mkdir -p /usr/local/lib/node_modules/appium/node_modules/appium-xcuitest-driver/WebDriverAgent/Resources/WebDriverAgent.bundle
```
jlipps: why don't we just do this in appium rather than make the user do it?

jlipps (approved):

docs/en/advanced-concepts/wda-custom-server.md
imurchie: We do, if they're allowing Appium to automate the process. But we can't if they are managing things themselves.

jlipps: Right, these are the instructions for rolling your own. 👍

messianic:
What about the import statements?? Couldn't find the time delta class

PR #8983, opened 2017-08-07
Add basic synchonization to session creation and deletion

mykola-mokhnach:
## Proposed changes

Our functional tests demonstrated that session creation is done in parallel, which means that each driver does not know about other driver executed at the same time, since the list of pending sessions is being created exactly at the same time for each instance. Synchronization solves this problem, so they wait for each other to update the shared list of pending and active sessions. Same thing with session deletion.

## Types of changes

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -19,7 +19,7 @@ import AsyncLock from 'async-lock';
 
 
 const sessionsListGuard = new AsyncLock();
-const pendindDriversGuard = new AsyncLock();
+const pendingDriversGuard = new AsyncLock();
 
 class AppiumDriver extends BaseDriver {
   constructor (args) {
@@ -30,12 +30,12 @@ class AppiumDriver extends BaseDriver {
 
     this.args = args;
 
-    // Access to sessions list must be guarded with a Sepamphore, because
+    // Access to sessions list must be guarded with a Semaphore, because
     // it might be changed by other async calls at any time
     // It is not recommended to access this property directly from the outside
     this.sessions = {};
 
-    // Access to pending drivers list must be guarded with a Sepamphore, because
+    // Access to pending drivers list must be guarded with a Semaphore, because
     // it might be changed by other async calls at any time
     // It is not recommended to access this property directly from the outside
     this.pendingDrivers = {};
@@ -198,7 +198,7 @@ class AppiumDriver extends BaseDriver {
     } catch (e) {
       throw new errors.SessionNotCreatedError(e.message);
     }
-    await pendindDriversGuard.acquire(AppiumDriver.name, () => {
+    await pendingDriversGuard.acquire(AppiumDriver.name, () => {
       this.pendingDrivers[InnerDriver.name] = this.pendingDrivers[InnerDriver.name] || [];
       otherPendingDriversData = this.pendingDrivers[InnerDriver.name].map((drv) => drv.driverData);
       this.pendingDrivers[InnerDriver.name].push(d);
@@ -206,14 +206,12 @@ class AppiumDriver extends BaseDriver {
     let innerSessionId, dCaps;
     try {
       [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, [...runningDriversData, ...otherPendingDriversData]);
-    } finally {
-      await pendindDriversGuard.acquire(AppiumDriver.name, () => {
-        _.pull(this.pendingDrivers[InnerDriver.name], d);
-      });
       await sessionsListGuard.acquire(AppiumDriver.name, () => {
-        if (innerSessionId) {
-          this.sessions[innerSessionId] = d;
-        }
+        this.sessions[innerSessionId] = d;
+      });
+    } finally {
+      await pendingDriversGuard.acquire(AppiumDriver.name, () => {
+        _.pull(this.pendingDrivers[InnerDriver.name], d);
       });
     }
 
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -15,7 +15,11 @@ import { MacDriver } from 'appium-mac-driver';
 import { EspressoDriver } from 'appium-espresso-driver';
 import B from 'bluebird';
 import util from 'util';
+import AsyncLock from 'async-lock';
 
+
+const sessionsListGuard = new AsyncLock();
+const pendindDriversGuard = new AsyncLock();
 
 class AppiumDriver extends BaseDriver {
   constructor (args) {
@@ -26,14 +30,20 @@ class AppiumDriver extends BaseDriver {
 
     this.args = args;
 
+    // Access to sessions list must be guarded with a Sepamphore, because
+    // it might be changed by other async calls at any time
+    // It is not recommended to access this property directly from the outside
     this.sessions = {};
 
+    // Access to pending drivers list must be guarded with a Sepamphore, because
+    // it might be changed by other async calls at any time
+    // It is not recommended to access this property directly from the outside
     this.pendingDrivers = {};
   }
 
   sessionExists (sessionId) {
-    return _.includes(_.keys(this.sessions), sessionId) &&
-           this.sessions[sessionId].sessionId !== null;
+    const dstSession = this.sessions[sessionId];
+    return dstSession && dstSession.sessionId !== null;
   }
 
   driverForSession (sessionId) {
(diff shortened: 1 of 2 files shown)


lib/appium.js
```diff
@@ -178,24 +183,30 @@ class AppiumDriver extends BaseDriver {
...
-      runningDriversData = this.curSessionDataForDriver(InnerDriver);
-    } catch (e) {
-      throw new errors.SessionNotCreatedError(e.message);
-    }
-
-    let innerSessionId, dCaps;
+    let runningDriversData, otherPendingDriversData;
     let d = new InnerDriver(this.args);
-    try {
+    await sessionCreationLock.acquire(AppiumDriver.name, async () => {
```
jlipps: i'm a little confused why we need to lock this block; we're not doing anything asynchronous in here right?

lib/appium.js
```diff
@@ -178,24 +183,30 @@ class AppiumDriver extends BaseDriver {
...
-      const otherPendingDriversData = this.pendingDrivers[InnerDriver.name].map((drv) => drv.driverData);
+      otherPendingDriversData = this.pendingDrivers[InnerDriver.name].map((drv) => drv.driverData);
       this.pendingDrivers[InnerDriver.name].push(d);
+    });
+    let innerSessionId, dCaps;
+    try {
       [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, [...runningDriversData, ...otherPendingDriversData]);
     } finally {
-      _.pull(this.pendingDrivers[InnerDriver.name], d);
+      await sessionCreationLock.acquire(AppiumDriver.name, async () => {
```
jlipps: same for this. or is it that we want to lock both sides of the `await d.createSession` call?

lib/appium.js
```diff
@@ -250,9 +261,12 @@ class AppiumDriver extends BaseDriver {
     try {
       if (this.sessions[sessionId]) {
         const curConstructorName = this.sessions[sessionId].constructor.name;
-        const otherSessionsData = _.toPairs(this.sessions)
+        let otherSessionsData;
+        await sessionDeletionLock.acquire(AppiumDriver.name, async () => {
```
jlipps: same question here. what could possibly cause a race condition here? certainly not a bunch of lodash methods?

lib/appium.js
mykola-mokhnach: I see it as the first step, that prepares AppiumDriver to by truly async and allow to execute multiple commands in parallel, without scheduling.

For now these changes don't make any difference, because each createSession (and other belonging to AppiumDriver) command is put into commands queue and I'd like to change this behaviour.

lib/appium.js
```diff
@@ -26,14 +30,20 @@ class AppiumDriver extends BaseDriver {
 
     this.args = args;
 
+    // Access to sessions list must be guarded with a Sepamphore, because
```
jlipps: I think you mean `semaphore`

lib/appium.js
```diff
@@ -26,14 +30,20 @@ class AppiumDriver extends BaseDriver {
...
+    // it might be changed by other async calls at any time
+    // It is not recommended to access this property directly from the outside
     this.sessions = {};
 
+    // Access to pending drivers list must be guarded with a Sepamphore, because
```
jlipps: here too

lib/appium.js
```diff
@@ -165,37 +175,47 @@ class AppiumDriver extends BaseDriver {
...
-      runningDriversData = this.curSessionDataForDriver(InnerDriver);
+      runningDriversData = await this.curSessionDataForDriver(InnerDriver);
     } catch (e) {
       throw new errors.SessionNotCreatedError(e.message);
     }
-
-    let innerSessionId, dCaps;
-    let d = new InnerDriver(this.args);
-    try {
+    await pendindDriversGuard.acquire(AppiumDriver.name, () => {
```
jlipps: where is this defined? also it should be `pendingDriversGuard`

lib/appium.js
```diff
@@ -165,37 +175,47 @@ class AppiumDriver extends BaseDriver {
...
-      const otherPendingDriversData = this.pendingDrivers[InnerDriver.name].map((drv) => drv.driverData);
+      otherPendingDriversData = this.pendingDrivers[InnerDriver.name].map((drv) => drv.driverData);
       this.pendingDrivers[InnerDriver.name].push(d);
+    });
+    let innerSessionId, dCaps;
+    try {
       [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, [...runningDriversData, ...otherPendingDriversData]);
     } finally {
-      _.pull(this.pendingDrivers[InnerDriver.name], d);
+      await pendindDriversGuard.acquire(AppiumDriver.name, () => {
```
jlipps: `pending`

lib/appium.js
```diff
@@ -15,8 +15,12 @@ import { MacDriver } from 'appium-mac-driver';
 import { EspressoDriver } from 'appium-espresso-driver';
 import B from 'bluebird';
 import util from 'util';
+import AsyncLock from 'async-lock';
 
 
+const sessionsListGuard = new AsyncLock();
+const pendindDriversGuard = new AsyncLock();
```
jlipps: oh yes it's defined here. `pendingDriversGuard`

lib/appium.js
```diff
@@ -165,37 +175,47 @@ class AppiumDriver extends BaseDriver {
...
-      log.info('Session override is on. Deleting other sessions.');
-      for (let id of _.keys(this.sessions)) {
-        log.info(`    Deleting session '${id}'`);
-        try {
-          await this.deleteSession(id);
-        } catch (ign) {
-          // the error has already been logged in AppiumDriver.deleteSession
-          // continue
+    if (this.args.sessionOverride) {
+      const sessionIdsToDelete = await sessionsListGuard.acquire(AppiumDriver.name, () => _.keys(this.sessions));
```
jlipps: we see a lot of repetition of naming the lock here. could create a simple helper function:

```js
async guardedSessions (fn) {
  await sessionsListGuard.acquire(AppiumDriver.name, () => {
    return fn(this.sessions);
  });
}
```

then we could change lines like this to:

```js
const sessionIdsToDelete = await this.guardedSessions(_.keys);
```

lib/appium.js
```diff
@@ -165,37 +175,47 @@ class AppiumDriver extends BaseDriver {
...
+    let innerSessionId, dCaps;
+    try {
       [innerSessionId, dCaps] = await d.createSession(caps, reqCaps, [...runningDriversData, ...otherPendingDriversData]);
     } finally {
-      _.pull(this.pendingDrivers[InnerDriver.name], d);
+      await pendindDriversGuard.acquire(AppiumDriver.name, () => {
+        _.pull(this.pendingDrivers[InnerDriver.name], d);
+      });
+      await sessionsListGuard.acquire(AppiumDriver.name, () => {
+        if (innerSessionId) {
```
jlipps: why not put the `if` outside the `acquire`? if we have no session, no need to wait on the lock

jlipps: actually now that i'm looking at this, why wouldn't we want this to be inside the `try` block? then we don't need an `if` statement at all because we know we have a session.

jlipps (commented):
made some comments. looks good.

lib/appium.js
mykola-mokhnach: fixed

lib/appium.js
mykola-mokhnach: fixed

lib/appium.js
mykola-mokhnach: there are only two places where this.sessions itself is guarded. Wouldn't it be odd?

lib/appium.js
mykola-mokhnach: fixed

lib/appium.js
mykola-mokhnach: fixed

lib/appium.js
jlipps: just because we pass `this.sessions` to a method doesn't mean it has to be used. the other instances would be simplified because they wouldn't require remembering the lock name:

```js
await this.guardedSessions(() => {
  // mutate this.sessions in here or do whatever
});
```

is still a bit nicer looking IMO.

this isn't a big deal, i just habitually look for ways to remove duplication :-)

jlipps (approved):

lib/appium.js
mykola-mokhnach: Yep, it would probably look nicer, however there is a probability one can create a deadlock by calling this wrapper method inside some block of code which is already protected by the same semaphore.

jlipps: OK makes sense. Probably not worth it in that case.

imurchie (approved):

PR #9016, opened 2017-08-11
Add MacOS to Travis build

imurchie:
## Proposed changes

Expand our testing on Travis to test both on Linux and MacOS.

## Types of changes

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
   "scripts": {
     "prepublish": "gulp prepublish",
     "test": "gulp once",
+    "e2e-test": "gulp e2e-test",
     "watch": "gulp watch",
     "build": "gulp transpile",
     "mocha": "mocha",
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ addons:
     packages:
       - g++-4.8
 script:
-  - npm run generate-docs && npm run test
+  - npm run generate-docs && npm run test && npm run e2e-test


lib/config.js
```diff
@@ -16,7 +13,8 @@ function getNodeVersion () {
 }
 
 async function getGitRev () {
-  let cwd = path.resolve(__dirname, "..", "..");
+  let cwd = __filename.indexOf('build/lib/config') !== -1 ?
```
mykola-mokhnach: this, probably, won't work as expected in Windows

jlipps (commented):
looks great, had a few comments

lib/config.js
jlipps: agree with this comment, also this logic could probably use a code comment as well (why could it be in one of two places?)

lib/config.js
```diff
@@ -28,10 +26,15 @@ async function getGitRev () {
 async function getAppiumConfig () {
   let stat = await fs.stat(path.resolve(__dirname, '..'));
   let built = stat.mtime.getTime();
+
+  let pkg = __filename.indexOf('build/lib/config') !== -1 ?
```
jlipps: `let {version} = ...` is more succinct

package.json
```diff
@@ -44,6 +44,7 @@
     "appium-support": "2.x",
     "appium-uiautomator2-driver": "0.x",
     "appium-windows-driver": "0.x",
+    "appium-xcode": "^3.3.0",
```
jlipps: why wouldn't this be a devDep?

test/functional/ios/parallel-e2e-specs.js
```diff
@@ -0,0 +1,54 @@
...
+      server = await appiumServer(args);
+    }
+  });
+  after(async function () {
+    if (server) {
+      await server.close();
+    }
+  });
+
+  it('should start a session and get status', async function () {
```
jlipps: do we not want to actually try to run multiple sessions in this test? otherwise i'm not sure what it gets us

imurchie: I'm not going to write that test. This is just a placeholder with the plumbing to restrict to the right platform. The Mykola can do his thing.

lib/config.js
```diff
@@ -13,7 +13,9 @@ function getNodeVersion () {
 }
 
 async function getGitRev () {
-  let cwd = __filename.indexOf('build/lib/config') !== -1 ?
+  // this code can be run pre-transpiled (so, code is in 'build')
+  // or transpiled on the fly (so, code is in root directory)
+  let cwd = __filename.indexOf(path.join('build', 'lib', 'config')) !== -1 ?
```
mykola-mokhnach: ES6 has _includes_ method available for strings

lib/config.js
```diff
@@ -28,10 +28,16 @@ async function getGitRev () {
 async function getAppiumConfig () {
   let stat = await fs.stat(path.resolve(__dirname, '..'));
   let built = stat.mtime.getTime();
+
+  // this code can be run pre-transpiled (so, code is in 'build')
+  // or transpiled on the fly (so, code is in root directory)
+  let {version} = __filename.indexOf(path.join('build', 'lib', 'config')) !== -1 ?
```
mykola-mokhnach: probably, it would be better to extract it into some common method to avoid unnecessary duplication?

lib/parser.js
```diff
@@ -788,6 +787,8 @@ function parseDefaultCaps (caps) {
 }
 
 function getParser () {
+  let pkgObj = __filename.indexOf('build/lib/parser') !== -1 ?
```
mykola-mokhnach: should be fixed here as well

test/helpers.js
```diff
@@ -2,11 +2,13 @@ import path from 'path';
...
 import B from 'bluebird';
 
+
 const TEST_HOST = 'localhost';
 const TEST_PORT = 4723;
-const TEST_FAKE_APP = path.resolve(__dirname, "..", "..", "node_modules",
-                                   "appium-fake-driver", "test", "fixtures",
-                                   "app.xml");
+const MOCHA_TIMEOUT = 60 * 1000 * (process.env.TRAVIS ? 8 : 4);
+const TEST_FAKE_APP = __filename.indexOf('build/test/helpers') !== -1 ?
```
mykola-mokhnach: ..and here

lib/config.js
imurchie: Yeah. But we still technically support Node 4 and it is not available there.

mykola-mokhnach: do we? I thought for this version the min node version should be 6+

imurchie: Nothing has ever been decided. So we support 4.

imurchie:
I guess, but to what end?

jlipps:
should we do something with this?

jlipps:
@imurchie this is now conflicted. do we care about this PR anymore?

imurchie:
Yes. I think this should go in. But I also think we should wait until we've released 1.7.0. I'll fix the conflicts then.

dpgraham:
@imurchie Now that 1.7 is out, should we revisit this?

Issue #9138, opened 2017-08-31
create a main doc for each driver


jlipps:
Info each doc should have:
    * prose explanation of how the driver works and what technology it's based on
    * automationName to use
    * basic platform support (i.e., XCUITest is iOS 9.3+)
    * any specific setup instructions (this is where most of the driver-specific setup will go) or guides (could be sub-docs)
    * links to driver repos and pointers for contribution

Drivers:
* [X] XCUITest
* [X] UiAutomator2
* [X] Instruments (big deprecation notice)
* [X] UiAutomator (deprecation notice)
* [X] Windows (info exists)
* [ ] Youi.tv (get help from the Youi folks)
* [X] Mac
* [X] Espresso
* [X] Selendroid

PR #9146, opened 2017-08-31
write doc for xcuitest driver

jlipps:
- consolidate information about it into one place
- part of #9138


mykola-mokhnach (approved):

imurchie (approved):

docs/en/drivers/ios-xcuitest-real-devices.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+In addition to the dependency on `libimobiledevice`, Appium support for real
+devices running iOS 9.3 and above using Xcode 8+ also depends on `ios-deploy`,
+which is easily available through [npm](https://www.npmjs.com/package/ios-deploy):
+
+```
+npm install -g ios-deploy
+```
+
```
jamesvanhorn: ios-deploy is available through homebrew as well now. 
```
brew install ios-deploy
```

jlipps:
@jamesvanhorn thanks! given that this has been merged, would you like to propose a pull request with the change?

PR #9218, opened 2017-09-15
Add commands template and example commands

dpgraham:


docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
...
+## Description
+
+Clicks element at it's center point. If the element’s center point is obscured by another element, an element click intercepted error is returned. If the element is outside the viewport, an element not interactable error is returned.
+
+Not all drivers automatically scroll the element into view and may need to be scrolled to in order to interact with it.
+
+## Client Docs
+
+* [Java](https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/WebElement.html#click--)
+* [Python](http://selenium-python.readthedocs.io/api.html#selenium.webdriver.common.action_chains.ActionChains.click)
```
imurchie: This is the link to the ActionChains `click`, which is different.

http://selenium-python.readthedocs.io/api.html#selenium.webdriver.remote.webelement.WebElement.click

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
...
+el = self.driver.find_element_by_accessibility_id('SomeId')
+el.click();
+```
+```javascript
+// Javascript
+// webdriver.io
+driver.click('#SomeId');
+
+// wd (async/await)
+var element = await driver.elementByAccessibilityId('id', 'SomeId');
```
jlipps: come on, `var`? :-)

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
...
+await element.click();
+```
+```ruby
+find_element(:accessibility_id, 'SomeId').click
+```
+
+## Description
+
+Clicks element at it's center point. If the element’s center point is obscured by another element, an element click intercepted error is returned. If the element is outside the viewport, an element not interactable error is returned.
```
jlipps: `its`

docs/en/commands/template.md
```diff
@@ -0,0 +1,84 @@
...
+
+```javascript
+{
+  value: "A value"
+}
+```
+
+### Example Response
+
+```javascript
```
imurchie: Is there no `json` language for code blocks?

docs/en/commands/element/click.md
jlipps: is the `'` in `element's` a normal quote? if not we might have problems later on.

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
...
+## Compatibility
+
+|Platform|Automation Name|Supported|
+| ------------- |-------------| -----|
+|iOS|[XCUITest](docs/en/drivers/ios-xcuitest.md)|(yes)|
+||[iOS](docs/en/drivers/ios-xcuitest.md)|(yes)|(yes)|
+|Android|[Espresso](docs/en/drivers/android-espresso.md)|(yes)|(yes)|
+||[UiAutomator2](docs/en/drivers/android-uiautomator2.md)|(yes)|(yes)|
+||[UiAutomator](docs/en/drivers/android-uiautomator.md)|(yes)|(yes)|
+|Windows|[WinAppDriver](https://github.com/Microsoft/WinAppDriver)|(yes)|(yes)|
```
jlipps: link for this driver will be `docs/en/drivers/windows.md`

also, we will need a `/` in front of docs, if we want linking to work on github.

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
...
+|Mac|[Mac](https://github.com/appium/appium-mac-driver)|(yes)|(yes)|
```
jlipps: `/docs/en/drivers/mac.md`

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
...
+
+## HTTP API Specifications
+
+### Endpoint
+
+`GET /wd/hub/status`
+
+### Parameters
+
+`none`
```
jlipps: I think we should list two kinds of parameters here: `URL Parameters` and `JSON Parameters`. For example `POST /session/:sessionid/element` has a URL parameter as well as JSON parameters.

Also, if there are none, I think we should just say "None" rather than `none`. Putting it in backticks makes it look like an actual value rather than a description of  nothingness.

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Example Response
```
imurchie: I think I'd rather see this as actual response (generic) than example response. For this case this is nothing, but say it returns a string, something like 
```json
{
  "status": <integer code>,
  "value": <string>
}
```

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
...
+||[UiAutomator2](docs/en/drivers/android-uiautomator2.md)|(yes)|(yes)|
+||[UiAutomator](docs/en/drivers/android-uiautomator.md)|(yes)|(yes)|
+|Windows|[WinAppDriver](https://github.com/Microsoft/WinAppDriver)|(yes)|(yes)|
+|Mac|[Mac](https://github.com/appium/appium-mac-driver)|(yes)|(yes)|
+
+## HTTP API Specifications
+
+### Endpoint
+
+`GET /wd/hub/status`
```
jlipps: this is the wrong endpoint for this command

imurchie (commented):
What does this look like when run through the mkdocs and served?

docs/en/commands/status.md
```diff
@@ -0,0 +1,75 @@
+# Status
+
+Retrieve the server’s current status.
+
+## Example Usage
```
jlipps: we should add examples in all languages. if we don't know what they look like, we should add a "TODO". but all should show up, I think.

jlipps (commented):
looks great! made some comments

docs/en/commands/element/click.md
jlipps: one downside of this is it means we can't serve different responses for W3C vs JSONWP. or we have to notate what they would look like for each.

docs/en/commands/template.md
dpgraham: There is, I just like using Javascript because you can add comment `//` blocks to add descriptions

docs/en/commands/element/click.md
dpgraham: Yeah, I did it because it's the lowest common denominator. But I suppose if we're using async/await, we can use let/const as well.

docs/en/commands/element/click.md
dpgraham: @jlipps What's your opinion?

jlipps: I like raw JSON responses as samples, but then we run into the problem I mentioned. the w3c response will always be different than the jsonwp response. so do we just include what's inside? or list both? seems like a lot of work.

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,72 @@
+# Click
+
+Retrieve the server’s current status.
```
mykola-mokhnach: Click retrieves server status? it's something new %)

dpgraham:
Here's a screenshot of what it looks like:

2017-09-15-11-50-localhost-4000

jlipps:
one other thought @dpgraham: i think we will want information about what appium version, driver version, platform version, and client library version this was first supported in. might require a few different tables to show that eh?

dpgraham:
Yeah that's a good idea.

dpgraham:
I think what we should do is have a whole sub-section dedicated to 'support' which includes appium version, driver version, platform version, automation version and client library version.

dpgraham:
@jlipps What do you think of this? I figure Appium Version and Driver Version can be the same table. I can add platform version as well. 

And then we can have a separate clients table.

screen shot 2017-09-15 at 1 27 12 pm

docs/en/commands/status.md
```diff
@@ -0,0 +1,79 @@
...
+
+Implementations may optionally include additional meta information as part of the body, but the top-level properties ready and message are reserved and must not be overwritten.
+
+## Client Docs
+
+* [Java](http://seleniumhq.github.io/selenium/docs/api/java/index.html)
+* [Python](http://selenium-python.readthedocs.io/api.html#selenium.webdriver.common.utils.is_url_connectable)
+* [Javascript (WebdriverIO)](http://webdriver.io/api/protocol/status.html)
+* [Javascript (WD)](https://github.com/admc/wd/blob/master/lib/commands.js#L44)
+
```
KazuCocoa: * [Ruby](http://www.rubydoc.info/gems/selenium-webdriver/Selenium/WebDriver/DriverExtensions/HasRemoteStatus#remote_status-instance_method)

---

^ is for now.(because it's marked `private`. So, I'll define in appium client side soon.)

docs/en/commands/status.md
```diff
@@ -0,0 +1,79 @@
...
+`none`
+
+### JSON Parameters
+
+`none`
+
+### Example Response
+
+```javascript
```
KazuCocoa: I got the following response. Which is expected against `/wd/hub/status`?
Or is it an example yet? :P

```
[HTTP] <-- GET /wd/hub/status 200 22 ms - 126
[HTTP] --> GET /wd/hub/status {}
[debug] [MJSONWP] Calling AppiumDriver.getStatus() with args: []
[debug] [MJSONWP] Responding to client with driver.getStatus() result: {"build":{"version":"1.7.0-beta","revision":"cf24a80809309fb5467099e570cddd256cacbb28"}}
```

KazuCocoa (commented):
I put some comments for the ruby client.

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,87 @@
...
+* [Ruby](http://www.rubydoc.info/gems/selenium-webdriver/0.0.28/Selenium/WebDriver/Element#click-instance_method)
+* PHP (TODO)
+* C# (TODO)
+
+## Compatibility
+
+|Platform|Automation Name|Supported|
+| ------------- |-------------| -----|
+|iOS|[XCUITest](/docs/en/drivers/ios-xcuitest.md)|(yes)|
+||[iOS](/docs/en/drivers/ios-xcuitest.md)|(yes)|(yes)|
```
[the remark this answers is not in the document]
dpgraham: Yes. Good eye.

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,87 @@
...
+// Javascript
+// webdriver.io
+driver.click('#SomeId');
+
+// wd (async/await)
+let element = await driver.elementByAccessibilityId('id', 'SomeId');
+await element.click();
+```
+```ruby
+find_element(:accessibility_id, 'SomeId').click
```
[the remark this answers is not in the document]
dpgraham: Sure, thanks @KazuCocoa. Do you want us to add you as a code reviewer whenever there is a ruby code example?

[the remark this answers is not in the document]
KazuCocoa: Yes, I can help the example. 👍

jlipps:
@dpgraham I like that new idea with separate tables for server + client. A couple more ideas:

* i think the server table should have a column called "Support", which just says "Yes" or "No". Then the "version" cells can just be blank if it's "No"
* in the client table, do the same thing: have a "Support" column, then also a "Version" column
* change "Automation Name" to "Driver"
* I think we can just say "Yes" or "No" or "All" without parens, but that's a nit.

docs/en/commands/status.md
jlipps: yes, looks like the example should be updated

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,87 @@
...
+||[UiAutomator2](/docs/en/drivers/android-uiautomator2.md)|(yes)|(yes)|
+||[UiAutomator](/docs/en/drivers/android-uiautomator.md)|(yes)|(yes)|
+|Windows|[Windows](/docs/en/drivers/windows.md)|(yes)|(yes)|
+|Mac|[Mac](/docs/en/drivers/mac.md)|(yes)|(yes)|
+
+## HTTP API Specifications
+
+### Endpoint
+
+`GET /wd/hub/session/:session_id/element/:element_id/click`
```
jlipps: click is `POST`

docs/en/commands/element/click.md
```diff
@@ -0,0 +1,87 @@
...
+
+(none)
+
+### Example Response
+
+Returns `null` if successful.
+
+## See Also
+
+[W3C Specification](https://www.w3.org/TR/webdriver/#element-click)
```
jlipps: given that we're in transition between JSONWP and W3C, maybe we should just link to both. the JSONWP link for this command is https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidelementidclick

docs/toc.json
```diff
@@ -0,0 +1,69 @@
+{
```
jlipps: we shouldn't be adding a toc.json anymore, right? we're doing toc.js, and it already exists

dpgraham: Yeah, I'll take that out.

dpgraham:
@jlipps What do you think about this for compatibility tables? I think two tables is good enough.

screen shot 2017-09-18 at 11 51 26 am

imurchie:
For some commands there will be a minimum version that each client supports. Also, "Automation Name" is a bit weird, since sometimes it is `automationName` that dictates what the driver is, and sometimes it is other things (like `platformVersion`).

dpgraham:
@imurchie For clients that have a minimum version, the version will be listed in the 'Supported' cell

dpgraham:
@jlipps I think the support columns should go:

`Platform -> Supported -> Driver -> Driver Version -> Appium Version`

If it's not supported on a platform, let's say iOS, then the row should be

`iOS -> No` and the rest of the columns should be blank, like what you suggested

If it's supported on all versions:

`iOS -> Yes`

If it has a minimum version support

`iOS -> 9.0 ->` followed by the version details

imurchie:
Ah! Ok.

dpgraham:
Changes pushed

dpgraham:
Changes made

dpgraham:
@jlipps @imurchie Do either of you have any more comments for this?

jlipps (commented):
ok, made my final review comments

docs/en/commands/status.md
```diff
@@ -0,0 +1,96 @@
...
+
+## HTTP API Specifications
+
+### Endpoint
+
+`GET /wd/hub/status`
+
+### URL Parameters
+
+`none`
```
jlipps: I thought we decided that None should just be a description, not an encoded value as it appears here.

docs/en/commands/template.md
```diff
@@ -0,0 +1,114 @@
...
+### Response
+
+|name|type|description|
+|----|----|-----------|
+|field_name_1|number|Description of field_name_1|
+|field_name_2|number|Description of field_name_2|
+|object.key1|string|Description of key1|
+|object.key2|string|Description of key2|
+
+## Specifications
```
jlipps: looks like Specifications and See Also are doing the same thing here?

docs/toc.js
```diff
@@ -13,7 +13,23 @@ module.exports = {
...
+      ["Mac", "mac.md"],
+      ["UIAutomation (iOS)", "ios-uiautomation.md"]]],
+    ["Commands", ["commands",
+      ["Status", "status.md"],
+      ["Template", "template.md"],
+      ["Element", ["element",
+        ["Click", "click.md"]
+      ]]
+    ]],
+    /*["Setup", ["appium-setup",
```
jlipps: why add commented stuff?

docs/toc.js
```diff
@@ -13,7 +13,23 @@ module.exports = {
       ["Espresso (Android)", "android-espresso.md"],
       ["Selendroid (Android)", "android-selendroid.md"],
       ["Windows", "windows.md"],
-      ["Mac", "mac.md"]]],
+      ["Mac", "mac.md"],
+      ["UIAutomation (iOS)", "ios-uiautomation.md"]]],
+    ["Commands", ["commands",
+      ["Status", "status.md"],
+      ["Template", "template.md"],
```
jlipps: we probably don't want the command template in our user-facing docs, right?

dpgraham: Good catch.

jlipps (approved):
this looks good. i am starting to think about auto-generating the markdown now. keeping the template in sync everywhere is going to be a nightmare. it might be worth seeing if we could use yaml or even a markdown -> markdown method for generating the boilerplate for the template.

dpgraham:
Yeah. We could even have generated docs that are committed to GitHub.

PR #9220, opened 2017-09-15
Move sitemap from appium.io to docs

dpgraham:


jlipps:
this has all the commands stuff in it, can you move to a different branch so we can review it independently? the commands stuff should be in its own branch IMO

dpgraham:
Done.

jlipps (commented):
looks good. to be honest i’d prefer that this were a JS file that exports an object, just in case we ever want to use variables or add a comment.

dpgraham:
I would also prefer .js but the limitation is that 'appium' gets pulled down as a zip file that gets extracted and we have no way of importing the .js file.

PR #9260, opened 2017-09-21
Render commands markdown using YML

dpgraham:
* Made directory 'commands-yml' which contains:
  * .babelrc to transpile the scripts
  * parse.js which renders the markdown
  * template.md which is a handlebars template that the yml gets passed to
  * validator.js which validates the yml data
* The rendering works by translating the yml to js, validating the js and then passing the data to template.md (which uses handlebars)
* The directory `/commands-yml/commands` gets rendered and then copied to `/docs/en/commands` as markdown
* Uses babel-node to run the script
* Uses validatejs to validate the javascript
* Template.yml is just a sample


dpgraham:
I used .babelrc with babel-node for the script, but if need be I could change it to a Gulp task.

Next steps after this is merged will be to port over `status` and `click` to YML

imurchie (approved):
I'm fine with keeping the running as it is, until such time as we need to do this outside the appium repo.

commands-yml/commands/template.yml
```diff
@@ -0,0 +1,99 @@
...
+  ruby: "https://github.com/appium/ruby_lib/releases/latest"
+  php: "https://github.com/appium/php-client/releases/latest-"
+  csharp: "https://github.com/appium/appium" # TODO
+
+# Driver support by platform
+driver_support:
+  ios:
+    xcuitest:
+      platform_min: '1.2'
+      platform_max: '9.3'
```
jlipps: i know this is the template so we're showing everything, but is it possible to omit mins and maxes and have the renderer do the right thing?

commands-yml/commands/template.yml
```diff
@@ -0,0 +1,99 @@
...
+      driver_max: '9.9'
+    uiautomation:
+      platform_min: '1.0'
+  android:
+    uiautomator2:
+      platform_min: '8.0'
+  mac:
+    mac:
+      platform_min: '?' # TODO
+client_support:
```
jlipps: yes, this is what i'm talking about here. great.

commands-yml/parse.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  if (object === true) {
+    return 'All';
+  }
+
+  const min = object[name ? `${name}_min` : 'min'];
+  const max = object[name ? `${name}_max` : 'max'];
+
+  if (!min && !max) {
```
jlipps: beautiful

commands-yml/parse.js
```diff
@@ -0,0 +1,87 @@
...
+    case 'uiautomator2':
+      return 'UiAutomator2';
+    case 'uiautomator':
+      return 'UiAutomator';
+    default:
+      return driverName[0].toUpperCase() + driverName.substr(1);
+  }
+});
+
+(async function () {
```
jlipps: rather than lose the whitespace due to the try/catch, you could just do this:

```js
async function main ( ) {
  // all your code, but not in a try catch
}

asyncify(main); // asyncbox's asyncify takes care of error logging
```

commands-yml/parse.js
```diff
@@ -0,0 +1,87 @@
...
+      const validationErrors = validate(inputJS, validator);
+      if (validationErrors) {
+        throw new Error(validationErrors);
+      }
+
+      // Pass the inputJS into our Handlebars template
+      const template = Handlebars.compile(await fs.readFile(path.resolve(__dirname, 'template.md'), 'utf8'), {noEscape: true, strict: true});
+      const markdown = template(inputJS);
+
+      // Write the markdown to it's right place
```
jlipps: its

commands-yml/validator.js
```diff
@@ -0,0 +1,51 @@
...
+    for (let option of options) {
+      if (!item[option]) {
+        return `must have attributes: ${options}`;
+      }
+    }
+  }
+};
+
+export default {
+  'name': { 'presence': true },
```
jlipps: according to our style, we don't want spaces around braces here, and we also don't want string quotes unless necessary:

```js
name: {presence: true},
...
'client_docs.ruby': {presence: true, url: true},
...
```

docs/en/commands/template.md
```diff
@@ -1,114 +1,115 @@
-# Template <Name of Command>
+# Template
 
-A brief description of the command.
+An indepth description of what this command does
```
jlipps: in-depth

docs/en/commands/template.md
```diff
@@ -1,114 +1,115 @@
-# Template <Name of Command>
+# Template
```
jlipps: we probably don't want to output the template as markdown? the template is just for people who will be writing the code, right?

jlipps (changes requested):
looks mostly good, just a couple style changes! and i think rather than merging this it would be great to have click and status as part of this PR, though that can be done after we've approved what's here

commands-yml/commands/template.yml
dpgraham: Yeah. I wrote a handlebars helper that covers that.

commands-yml/parse.js
dpgraham: Haha my achilles heel. Is there an eslint rule for that?

docs/en/commands/template.md
dpgraham: Yeah, I can take that out. I just wanted the PR to show the actual product of parsing.

commands-yml/parse.js
```diff
@@ -0,0 +1,87 @@
...
+  const min = object[name ? `${name}_min` : 'min'];
+  const max = object[name ? `${name}_max` : 'max'];
+
+  if (!min && !max) {
+    return 'All';
+  } else if (!max) {
+    return `${min}+`;
+  } else if (!min) {
+    return `<= ${max}`;
+  } else {
```
mykola-mokhnach: else is not needed

commands-yml/parse.js
```diff
@@ -0,0 +1,87 @@
...
+    return `${min} to ${max}`;
+  }
+});
+
+Handlebars.registerHelper('hyphenate', (str) => {
+  return str.replace('_', '-');
```
mykola-mokhnach: can be also `(str) => str.replace('_', '-')`

commands-yml/parse.js
```diff
@@ -0,0 +1,87 @@
...
+    case 'xcuitest':
+      return 'XCUITest';
+    case 'uiautomation':
+      return 'UIAutomation';
+    case 'uiautomator2':
+      return 'UiAutomator2';
+    case 'uiautomator':
+      return 'UiAutomator';
+    default:
+      return driverName[0].toUpperCase() + driverName.substr(1);
```
mykola-mokhnach: should we check if driver name length is greater than 1?

commands-yml/parse.js
```diff
@@ -0,0 +1,87 @@
...
+Handlebars.registerHelper('hyphenate', (str) => {
+  return str.replace('_', '-');
+});
+
+Handlebars.registerHelper('uppercase', (str) => {
+  return str.toUpperCase();
+});
+
+Handlebars.registerHelper('capitalize', (driverName) => {
+  switch (driverName) {
```
mykola-mokhnach: `driverName.toLowerCase()` ?

commands-yml/validator.js
```diff
@@ -0,0 +1,51 @@
...
+    return `must be an array`;
+  }
+};
+
+validate.validators.hasAttributes = function (value, options) {
+  if (!value) {
+    return;
+  }
+
+  if (typeof value === 'object' && !Array.isArray(value)) {
```
mykola-mokhnach: lodash lib might be useful here

commands-yml/validator.js
```diff
@@ -0,0 +1,51 @@
...
+    value = [value];
+  }
+
+  for (let item of value) {
+    for (let option of options) {
+      if (!item[option]) {
```
mykola-mokhnach: should we explicitly check if this is _undefined_?

commands-yml/parse.js
dpgraham: Good call. Should be case insensitive.

dpgraham:
Changes made.

Made the PR fixes.

Also I added an enhancement to the versioning helper that infers the appium version support and the platform version support. So for example, if you set `xcuitest: true` the platform support will be 9.3+ and the Appium support will be 1.6.0+

dpgraham:
I'll port over `status` and `click` on a new pull request so as to not clutter this one.

commands-yml/parse.js
```diff
@@ -88,38 +83,35 @@ Handlebars.registerHelper('capitalize', (driverName) => {
     case 'uiautomator':
       return 'UiAutomator';
     default:
-      return driverName[0].toUpperCase() + driverName.substr(1);
+      return driverName.length === 0 ? driverName : driverName[0].toUpperCase() + driverName.substr(1);
```
mykola-mokhnach: -> `length < 2 ? `

commands-yml/validator.js
```diff
@@ -11,41 +12,41 @@ validate.validators.hasAttributes = function (value, options) {
     return;
   }
 
-  if (typeof value === 'object' && !Array.isArray(value)) {
+  if (_.isObject(value) && !_.isArray(value)) {
```
mykola-mokhnach: what should be checked here? maybe you want to apply _isPlainObject_ check?

commands-yml/parse.js
dpgraham: If the length is '1' it won't be a problem because `driverName.substr(1)` will just return an empty string.

commands-yml/validator.js
dpgraham: We're expecting either an object or an array of objects. If it's just an object, I want to change it to an array of one. 

Does `isPlainObject` do the same thing as `_.isObject(value) && !_.isArray(value)`?

mykola-mokhnach: `!_.isArray(value)` should suffice

PR #9261, opened 2017-09-21
Update committer list

bootstraponline:


jlipps (approved):
looks good. we may want to revisit some of these memberships but we can do that in the UI and then run your handy script.

bootstraponline:
> looks good. we may want to revisit some of these memberships but we can do that in the UI and then run your handy script.

I think we should open an issue and ping them. My guess is we have a lot of committers who are no longer interested in appium. Removing members who aren't responsive on GitHub probably makes sense.

PR #9268, opened 2017-09-21
Added status and click YML

dpgraham:
* Renamed template.yml to example.yml
* Made some minor revisions (e.g.: made endpoint.response not required and have it show as 'null' when undefined)

### DIFF
--- a/commands-yml/commands/element/click.yml
+++ b/commands-yml/commands/element/click.yml
@@ -1,6 +1,6 @@
 ---
 name: Click
-short_description: Click element at it's center point.
+short_description: Click element at its center point.
 description:
   >
     Clicks element at its center point. If the element's center point is obscured by another element, an element click intercepted error is returned. If the element is outside the viewport, an element not interactable error is returned.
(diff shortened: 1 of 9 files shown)


commands-yml/commands/element/click.yml
```diff
@@ -0,0 +1,79 @@
+---
+name: Click
+short_description: Click element at it's center point.
```
jlipps: its

commands-yml/commands/status.yml
```diff
@@ -0,0 +1,77 @@
...
+      await driver.status();
+  javascript_wdio:
+    |
+      driver.status();
+  ruby:
+    |
+      # TODO
+  php:
+    |
+      # TODO
```
jlipps: php uses c-style comments

docs/en/commands/status.md
```diff
@@ -77,20 +90,22 @@ Implementations may optionally include additional meta information as part of th
 
 ### URL Parameters
 
-None
+|name|description|
```
jlipps: probably don't want a blank table, right?

jlipps (commented):
looks good, just a few remarks

docs/en/commands/status.md
dpgraham: Right. I need to move the if block.

dpgraham:
Fixes pushed

jlipps (approved):

PR #9351, opened 2017-10-02
Document several attributes commands

dpgraham:


imurchie (approved):

commands-yml/commands/element/attributes/replace-value.yml
```diff
@@ -0,0 +1,78 @@
+---
+name: Get Element Text
```
vmaxim: It seems different commands are mixed here: `/session/:sessionId/element/:elementId/value`, `/session/:sessionId/element/:elementId/text`, `/session/:sessionId/element/:elementId/submit`

commands-yml/commands/session/create.yml
```diff
@@ -0,0 +1,91 @@
...
+    The server should attempt to create a session that most closely matches the desired and required capabilities. 
+    
+    * [JSONWP Spec](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#session-1) Required capabilities have higher priority than desired capabilities and must be set for the session to be created
+    * [W3C specification])(https://www.w3.org/TR/webdriver/#dfn-new-session) capabilities.alwaysMatch must be set for session to be created; capabilities.firstMatch must match at least one (the first one to match will be used)
+    
+example_usage:
+  java:
+    |
+      MobileElement element = (MobileElement) driver.findElementByAccessibilityId("SomeAccessibilityID");
+      element.clear();
```
vmaxim: Here should be example of `Create session` command.

PR #9354, opened 2017-10-02
Implemented mouse and session commands

dpgraham:


imurchie (approved):

SrinivasanTarget (changes requested):
Do we have mouse interactions in Mobile? @dpgraham @imurchie @jlipps 

I think we should document endpoints only related(only that works) in our project and leave rest to Selenium to handle the doc's. Thoughts please.

imurchie:
Yeah, that makes sense.

dpgraham:
I already have done a bunch of the documentation for non-mobile endpoints.

I think we should include web/mouse interactions because it's relevant to Appium users when working within a web context and also, the Mac and Windows drivers may support these features.

PR #9398, opened 2017-10-06
Add generated docs

dpgraham:
These docs are all generated from the .yml files so no need to review the contents of the actual markdown files.

Once we publish the docs to a staging site, we'll do another round of verification before we publish to https://appium.io


imurchie:
Do you mind if I push a commit to this to fix some things?

PR #9501, opened 2017-10-22
Moved /basics up to root folder

dpgraham:
* Moving markdown files to basics folder caused some links to break (including on Appium Desktop)
* Moving them up one place instead


dpgraham:
Travis seems to show tests that were cancelled due to newer tests as 'failed'.

imurchie:
Yes. It can only report pass/fail.

imurchie (approved):
Ok. But it is also ok to have broken links out there if it makes going forward more organized.

PR #9550, opened 2017-10-28
Fix synchrous issue: appium may call session check even after session…

gu-wei:


gu-wei:
@imurchie

imurchie:
This is good to get into the style that we recommend. However, if there is a problem with the code, this is not actually going to change anything. There is no functional difference between the old code (not awaiting, just returning) and the new.

wei-gu:
@imurchie, I got error like during daily run:
: A session is either terminated or not started TestCleanup method VideoBuddy.Tests.OnDeviceTest.Search_BVTest.TestCleanup threw exception. System.InvalidOperationException: System.InvalidOperationException: Session [441d8bf6-48de-4162-a703-95368075ebd9] was terminated due to BROWSER_TIMEOUT

after I add the **await**, no more such error appears, not sure how it works

imurchie:
Alright. Well, I'm happy to merge this, whatever the case. :)

wei-gu:
thanks

dpgraham:
That's surprising that adding 'await' fixed that, because `return promise` and `return await promise` ought to be functionally equivalent.

imurchie (approved):

imurchie:
The first thing that comes to mind is that it is not being awaited somewhere else. I looked and couldn't find any such place, though. :(

PR #9790, opened 2017-12-13
Add relaxed security command line flag

mykola-mokhnach:
This flag, if provided to the server via command line, will allow to perform potentially insecure actions in client drivers, like direct access to adb shell in Android driver or command line access in OSX driver.


lib/appium.js
```diff
@@ -28,7 +28,11 @@ class AppiumDriver extends BaseDriver {
     // the main Appium Driver has no new command timeout
     this.newCommandTimeoutMs = 0;
 
-    this.args = args;
+    this.args = Object.assign({}, args);
+    if (this.args.relaxedSecurityEnabled) {
+      this.relaxedSecurityEnabled = true;
+      delete this.args.relaxedSecurityEnabled;
```
imurchie: Why this extra step?

imurchie (commented):
One question about the implementation. I would also like to see some extra logging of the fact that this flag has been turned on and its implications, at startup time.

lib/appium.js
mykola-mokhnach: because this option should not be in capabilities. One anyway cannot change it from there

imurchie: But only the default capabilities from the command line arguments make it into the caps sent to the driver (https://github.com/appium/appium/blob/master/lib/appium.js#L176-L177)

mykola-mokhnach: ok, makes sense

mykola-mokhnach:
> One question about the implementation. I would also like to see some extra logging of the fact that this flag has been turned on and its implications, at startup time.

Yes, added more logging

imurchie (approved):

PR #9791, opened 2017-12-13
Support w3c createSession

dpgraham:
* In 'createSession', process w3c caps using 'processCapabilities' method exposed by base-driver
* Add tests that run W3C requests on server to verify that it's accepting w3c format

### DIFF
--- a/test/driver-e2e-specs.js
+++ b/test/driver-e2e-specs.js
@@ -5,6 +5,7 @@ import B from 'bluebird';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import wd from 'wd';
+import request from 'request-promise';
 import { main as appiumServer } from '../lib/main';
 import { TEST_FAKE_APP, TEST_HOST, TEST_PORT } from './helpers';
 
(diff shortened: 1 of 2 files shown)


test/driver-e2e-specs.js
```diff
@@ -75,6 +76,43 @@ describe('FakeDriver - via HTTP', () => {
...
+      sessionId.should.be.a.string;
+      value.should.exist;
+
+      // Now use that sessionId to call /screenshot
+      const { status:screenshotStatus, value:screenshotValue } = await request({url: `http://${TEST_HOST}:${TEST_PORT}/wd/hub/session/${sessionId}/screenshot`, json: true});
+      screenshotValue.should.equal('hahahanotreallyascreenshot');
+      screenshotStatus.should.equal(0);
+
+      // Now use that sessionID to call an arbitrary W3C-only endpoint that isn't implemented to see if it throws correct error
+      await request.post({url: `http://${TEST_HOST}:${TEST_PORT}/wd/hub/session/${sessionId}/execute/async`, form: {script: '', args: ['a']}, json: true}).should.eventually.be.rejectedWith(/not yet been implemented/);
```
jlipps: not sure but i think you can just set the request body via the `json` param, rather than setting it via `form` and then also setting `json` to `true`?

jlipps (approved):

PR #9837, opened 2017-12-21
ensure we get 1.x versions of windows driver

jlipps:
cc @yodurr

@imurchie will need to merge this and cherry-pick into the 1.7 branch before the next publish if it is to make it into the next version of appiupm.


imurchie:
Yes. I pull everything over on each release. This will get in.

PR #9869, opened 2017-12-27
Force MJSONWP protocol

dpgraham:
* The 'executeCommand' on appium-base-driver checks the protocol by looking for the w3c 'capabilities' object and is called before 'createSession' is
* Right now it's finding the 'capabilities' object and then appium is merging it into desiredCapabilities and running an MJSONWP session
* Because of this, session is being misidentified as W3C
* Fix for this upcoming release is just to monkey-patch the method that determines the protocol to return MJSONWP
* Next release we'll remove this and start fully accepting W3C sessions

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -17,6 +17,8 @@ import B from 'bluebird';
 import AsyncLock from 'async-lock';
 import { inspectObject } from './utils';
 
+// Force protocol to be MJSONWP until we start accepting W3C capabilities
+BaseDriver.determineProtocol = () => 'MJSONWP';
 
 const sessionsListGuard = new AsyncLock();
 const pendingDriversGuard = new AsyncLock();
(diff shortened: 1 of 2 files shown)


imurchie (approved):

lib/appium.js
```diff
@@ -17,6 +17,8 @@ import B from 'bluebird';
 import AsyncLock from 'async-lock';
 import { inspectObject } from './utils';
 
+// Force protocol to be MJSONWP until we start accepting W3C capabilities
+BaseDriver.determineProtocol = () => 'MJSONWP';
```
mykola-mokhnach: Don't we have a constant for the protocol name?

mykola-mokhnach:
> Next release we'll remove this and start fully accepting W3C sessions

Do you mean the next beta?

dpgraham:
@mykola-mokhnach we were thinking that the next release would actually be 1.8 to mark the addition of W3C

PR #9887, opened 2018-01-02
Sample code stub

dpgraham:
* For some reason, `sample-code/` was being ignored. Must have been there before. Unignored it.
* Added `sample-code/` and subdirectories for all of the supported languages
* Added README.md stubs for all of them
* Added dummy test for WD
* For WD I'm using async/await and running mocha tests by using `babel`. For WebdriverIO it may be better to use less modern Javascript
* I also am using the chai assert library instead of `chai-as-promised` because I figured it would be more accessible to a broader audience


jlipps (commented):
why do we want this in appium/appium? isn't this what the sample-code repo is for?

dpgraham:
@jlippa: @imurchie  and I weighed the pros and cons of both. We figured that having the sample-code in appium/appium would have benefits of:

* We can run `sample-code` via continuous integration and be made aware of problems with the clients sooner
* Having it in the same repo as appium would make it more likely to be maintained it would consolidate pull requests (e.g.: a new feature you can update the documentation and the sample code in one PR)
* One less repo to deal maintain

@jlipps What are your thoughts? Do you think there's benefits to having it in a separate repo?

.npmignore
```diff
@@ -14,3 +14,4 @@ _vimrc_local.vim
 *.prefs
 ./.npmignore
 coverage
+sample-code/javascript-wd/node_modules
```
imurchie: This should just be `sample-code`. There is no need to distribute any of the sample code with the package.

sample-code/javascript-wd/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "javascript-wd",
+  "version": "1.0.0",
+  "description": "Appium sample code using 'wd'",
+  "scripts": {
+    "test": "mocha test/**/*.test.js"
+  },
+  "author": "Dan Graham",
+  "license": "ISC",
```
imurchie: Make this `Apache-2.0`

jlipps:
@dpgraham ah I see. OK I'm in favor of dearchitecting sample-code into appium. I appreciate the benefits you mentioned. being able to test clients along with server releases is a good idea.

imurchie:
Seems like this needs to be rebased with master. The errors are the ones fixed yesterday.

PR #9907, opened 2018-01-03
Create session and basic interactions samples

dpgraham:
* Added the sample apps to `sample-code/apps`. Cut-and-paste from old sample-code
* Added session creation and basic interaction examples under `sample-code/javascript-wd/`


imurchie (commented):
Do we need to have all these apps in order to test the functionality? I would prefer to pare it down to one iOS app and one Android app.

jlipps:
I also hate that we'd be adding all these files to the appium repo. What about adding just `.app.zip` and `.apk` files? We can point people to the source if they want to tinker with the apps themselves.

I also agree with @imurchie; if this is the first step in our "rosetta stone" sample code project, it'd be nice to have just one app for each platform. (ideally one app for all platforms, like a react native one, but hey).

imurchie:
I agree with a zipped file. The main README for the sample code can explain where they are from.

dpgraham:
Pushed fixes. That is a lot better.

saikrishna321:
@jlipps 
> (ideally one app for all platforms, like a react native one)

We already have a react-native app for both iOS and Android

https://github.com/appium/sample-code/blob/master/NOTICE#L4-L11

Can we consider this app for all client sample-code?

imurchie:
That app looks like it could be very good for these purposes. 

@saikrishna321 would you consider moving it into the Appium organization on GitHub?

sample-code/javascript-wd/test/basic/android-create-session.test.js
```diff
@@ -0,0 +1,30 @@
...
+
+    // We haven't started a session yet, so we shouldn't see any sessions running
+    assert.equal((await driver.sessions()).length, 0);
+
+    // Start the session
+    await driver.init({
+      ...androidCaps,
+      app: require('../helpers/apps').androidApiDemos
+    });
+
```
imurchie: Might be useful to check that the App running is the app we expect (e.g., on Android, get the running package and activity; on iOS just check that the `XCUIElementTypeApplication` element is what we expect).

sample-code/javascript-wd/test/helpers/apps.js
```diff
@@ -0,0 +1,8 @@
+if (process.env.DEV) {
```
imurchie: This condition should be switched. I think the idea is that we can run on Sauce? In that case we should be forced to explicitly ask for Sauce stuff (e.g., an environment variable like `SAUCE` set) rather than forcing _all_ local usage to set the `DEV` variable.

sample-code/javascript-wd/test/helpers/caps.js
```diff
@@ -0,0 +1,23 @@
+import path from 'path';
+
+const iosCaps = {
+  platformName: 'iOS',
+  automationName: 'XCUITest',
+  deviceName: 'iPhone 6s' || process.env.IOS_DEVICE_NAME,
```
imurchie: This will _always_ set `deviceName` to `'iPhone 6s'`, since that string is always truthy. These need to be switched around:
```js
deviceName: process.env.IOS_DEVICE_NAME ||  'iPhone 6s',
platformVersion: process.env.IOS_PLATFORM_VERSION || '11.2',
```

sample-code/javascript-wd/test/helpers/caps.js
```diff
@@ -0,0 +1,23 @@
...
+  platformVersion: '11.2' || process.env.IOS_PLATFORM_VERSION,
+  app: undefined, // Will be added in tests
+};
+
+const androidCaps = {
+  platformName: 'Android',
```
imurchie: No `platformVersion`?

sample-code/javascript-wd/README.md
```diff
@@ -0,0 +1,5 @@
+# C# Sample Code
```
imurchie: Not C#

imurchie (changes requested):
Tests seem reasonable. A couple of things to be looked at. And I think we should consider @saikrishna321 's app.

saikrishna321:
@imurchie for sure, let's move it under appium repo..

imurchie:
That would be great!

saikrishna321:
@imurchie as we have moved the sample app, should we update this PR w.r.t to new app ?

imurchie:
We should. I'll let @dpgraham do that.

In the meantime, how does one get the build products for that app? I realize we have the zip files in the current sample-code repo, but if changes are made, how does the process go?

SrinivasanTarget:
> how does one get the build products for that app? I realize we have the zip files in the current sample-code repo, but if changes are made, how does the process go?

Hmm did it in my local and updated there yesterday. May be a readme out there will help out? Also we can integrate it to our CI in future to build automatically.

imurchie:
I can update the readme. What commands are used, and where do the products go?

SrinivasanTarget:
> I can update the readme. What commands are used, and where do the products go?

For Android:
` $ cd android`
`$ gradle clean assembleRelease`

Release build is generated at: `VodQAReactNative/android/app/build/output`

To run on device: `react-native run-android`

dpgraham:
Fixes pushed @imurchie 

I'll look into the React Native app too. That'll be a separate PR from this though, if we do decide to use it.

PR #9972, opened 2018-01-12
W3C Full Implementation

dpgraham:
* W3C compatibility 
    * Refactored code from createSession into new `util.js` method `parseCapsForInnerDriver()`
    * `parseCapsForInnerDriver()` takes jsonwp and w3c capabilities and translates them into objects that can be accepted by the "inner driver" that it proxies to
      * It adds defaultCapabilities to `capabilities.alwaysMatch` and jsonwp `desiredCaps`
      * It creates a new w3cCaps object that only has `alwaysMatch` which only contains the matching caps (because we're proxying through to an inner driver, there's no sense in passing along any other non-matching first match capabilities)
    * After `parseCapsForInnerDriver()` parses the JSONWP and W3C it passes the results along to the matching inner driver

* Insert Appium prefixes into W3C caps 
    * Added util method `insertAppiumPrefixes()`
    * Previously, Appium just took the W3C capabilities, processed them into a caps object and passed them into the inner driver as
    
    ```
    {
      alwaysMatch: parsedCaps,
      firstMatch: [{}],
    }
    ```
    
    * The problem with this is that all of the parsedCaps became unprefixed because `appium-base-driver` strips out the `appium:` and the InnerDriver complains about non-standard, non-prefixed caps
    * Now, it takes the parsedCaps and calls a new method called 'insertAppiumPrefixes' on the parsed caps like this:
    
    ```
    {
      alwaysMatch: {...insertAppiumPrefixes(parsedCaps)},
      firstMatch: [{}],
    }
    ```

* Add validation to W3C session creation  
    * Call to 'processCapabilities' does validation now so that it will match an object that passes Appium's validation rules (e.g.: platformName, deviceName required; automation name must be valid, etc...)

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -17,9 +17,6 @@ import B from 'bluebird';
 import AsyncLock from 'async-lock';
 import { inspectObject } from './utils';
 
-// Force protocol to be MJSONWP until we start accepting W3C capabilities
-BaseDriver.determineProtocol = () => 'MJSONWP';
-
 const sessionsListGuard = new AsyncLock();
 const pendingDriversGuard = new AsyncLock();
 
@@ -268,7 +265,7 @@ class AppiumDriver extends BaseDriver {
     let {defaultCapabilities} = this.args;
 
     // Get copies of the capabilities that reconcile any discrepancies between JSONWP and W3C
-    let {desiredCaps, newJsonwpCaps, newW3CCapabilities} = this.getCaps(jsonwpCaps, w3cCapabilities, {}, defaultCapabilities); // TODO: Set the constraints
+    let {desiredCaps, newJsonwpCaps, newW3CCapabilities} = this.getCaps(jsonwpCaps, w3cCapabilities, this.desiredCapConstraints, defaultCapabilities);
     let InnerDriver = this.getDriverForCaps(desiredCaps);
     this.printNewSessionAnnouncement(InnerDriver, desiredCaps);
 
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -182,7 +182,11 @@ class AppiumDriver extends BaseDriver {
     let {defaultCapabilities} = this.args;
 
     // Parse the caps into a format that the InnerDriver will accept
-    let {desiredCaps, newJsonwpCaps, newW3CCapabilities} = parseCapsForInnerDriver(jsonwpCaps, w3cCapabilities, this.desiredCapConstraints, defaultCapabilities);
+    let {desiredCaps, processedJsonwpCaps, processedW3CCapabilities} = parseCapsForInnerDriver(
+      jsonwpCaps,
+      w3cCapabilities,
+      this.desiredCapConstraints, defaultCapabilities
+    );
     let InnerDriver = this.getDriverForCaps(desiredCaps);
     this.printNewSessionAnnouncement(InnerDriver, desiredCaps);
 
@@ -215,7 +219,12 @@ class AppiumDriver extends BaseDriver {
 
     let innerSessionId, dCaps;
     try {
-      [innerSessionId, dCaps] = await d.createSession(newJsonwpCaps, reqCaps, newW3CCapabilities, [...runningDriversData, ...otherPendingDriversData]);
+      [innerSessionId, dCaps] = await d.createSession(
+        processedJsonwpCaps,
+        reqCaps,
+        processedW3CCapabilities,
+        [...runningDriversData, ...otherPendingDriversData]
+      );
       await sessionsListGuard.acquire(AppiumDriver.name, () => {
         this.sessions[innerSessionId] = d;
       });
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -2,7 +2,7 @@ import _ from 'lodash';
 import log from './logger';
 import { getAppiumConfig } from './config';
 import { BaseDriver, routeConfiguringFunction, errors,
-         isSessionCommand, processCapabilities } from 'appium-base-driver';
+         isSessionCommand } from 'appium-base-driver';
 import { FakeDriver } from 'appium-fake-driver';
 import { AndroidDriver } from 'appium-android-driver';
 import { IosDriver } from 'appium-ios-driver';
@@ -15,7 +15,7 @@ import { MacDriver } from 'appium-mac-driver';
 import { EspressoDriver } from 'appium-espresso-driver';
 import B from 'bluebird';
 import AsyncLock from 'async-lock';
-import { inspectObject } from './utils';
+import { inspectObject, parseCapsForInnerDriver } from './utils';
 
 const sessionsListGuard = new AsyncLock();
 const pendingDriversGuard = new AsyncLock();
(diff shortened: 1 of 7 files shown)


lib/appium.js
```diff
@@ -176,19 +173,16 @@ class AppiumDriver extends BaseDriver {
...
-      log.info(`Merged W3C capabilities ${_.truncate(JSON.stringify(capabilities), {length: 50})} into desiredCapabilities object ${_.truncate(JSON.stringify(desiredCaps), {length: 50})}`);
-      let w3cCaps = processCapabilities(capabilities, null, false);
-      desiredCaps = _.merge(desiredCaps, w3cCaps);
-    }
-    desiredCaps = _.defaults(_.clone(desiredCaps), this.args.defaultCapabilities);
+  async createSession (jsonwpCaps, reqCaps, w3cCapabilities) {
+    let {defaultCapabilities} = this.args;
+
+    // Parse the caps into a format that the InnerDriver will accept
+    let {desiredCaps, newJsonwpCaps, newW3CCapabilities} = parseCapsForInnerDriver(jsonwpCaps, w3cCapabilities, this.desiredCapConstraints, defaultCapabilities);
```
mykola-mokhnach: this line is too long

lib/appium.js
```diff
@@ -218,10 +212,10 @@ class AppiumDriver extends BaseDriver {
       otherPendingDriversData = this.pendingDrivers[InnerDriver.name].map((drv) => drv.driverData);
       this.pendingDrivers[InnerDriver.name].push(d);
     });
+
     let innerSessionId, dCaps;
     try {
-      // TODO: When we support W3C pass in capabilities object
-      [innerSessionId, dCaps] = await d.createSession(desiredCaps, reqCaps, [...runningDriversData, ...otherPendingDriversData]);
+      [innerSessionId, dCaps] = await d.createSession(newJsonwpCaps, reqCaps, newW3CCapabilities, [...runningDriversData, ...otherPendingDriversData]);
```
mykola-mokhnach: the line is too long

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
   }
 }
 
-export { inspectObject };
+/**
+ * Takes the caps that were provided in the request and translates them
+ * into caps that can be used by the inner drivers.
+ * @param {*} jsonwpCaps
```
mykola-mokhnach: we can say the basic type of these are Object

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+ * @param {*} w3cCapabilities
+ * @param {*} constraints
+ * @param {*} defaultCapabilities
+ */
+function parseCapsForInnerDriver (jsonwpCaps, w3cCapabilities, constraints={}, defaultCapabilities={}) {
+  // Check if the caller sent JSONWP caps, W3C caps, or both
+  const hasW3CCaps = _.isObject(w3cCapabilities);
```
mykola-mokhnach: isPlainObject

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+  const hasJSONWPCaps = _.isObject(jsonwpCaps);
+
+  // Make copies of the capabilities that include the default capabilities
+  if (hasW3CCaps) {
+    w3cCapabilities = {
+      ...w3cCapabilities,
```
mykola-mokhnach: does ellipsis do the same as _.toPairs?

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+    let matchingW3CCaps = processCapabilities(w3cCapabilities, constraints, true);
+
+    if (hasJSONWPCaps) {
+      let jsonwpKeys = _.keys(desiredCaps);
+      let matchingW3CCapsKeys = _.keys(matchingW3CCaps);
+
+      // If JSONWP has keys that W3C doesn't have, log a warning message
+      let differingKeys = _.differenceWith(jsonwpKeys, matchingW3CCapsKeys);
+      if (differingKeys.length > 0) {
+        logger.warn(`The following capabilities were provided in the JSONWP desired capabilities but not the W3C capabilities: ${JSON.stringify(differingKeys)}`);
```
mykola-mokhnach: the line is too long

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+    // Call the process capabilities algorithm to find matching caps on the W3C (https://github.com/jlipps/simple-wd-spec#processing-capabilities)
+    let matchingW3CCaps = processCapabilities(w3cCapabilities, constraints, true);
+
+    if (hasJSONWPCaps) {
+      let jsonwpKeys = _.keys(desiredCaps);
+      let matchingW3CCapsKeys = _.keys(matchingW3CCaps);
+
+      // If JSONWP has keys that W3C doesn't have, log a warning message
+      let differingKeys = _.differenceWith(jsonwpKeys, matchingW3CCapsKeys);
+      if (differingKeys.length > 0) {
```
mykola-mokhnach: this can also be !_.isEmpty

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+  // Create new JSONWP desired capabilities and W3C 'capabilities' objects that have the reconciled attributes
+  if (hasJSONWPCaps) {
+    newJsonwpCaps = {...desiredCaps};
+  }
+
+  // Translate the matched W3C caps back into W3C capabilities object so that it
+  // can be proxied to the inner driver
+  if (hasW3CCaps) {
+    newW3CCapabilities = {
+      alwaysMatch: {...insertAppiumPrefixes(desiredCaps)}, // Insert 'appium:' prefix to non-prefixed caps so that the Inner Driver doesn't complain about bad W3C prefixes
```
mykola-mokhnach: this line is too long

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+    // If JSONWP caps and W3C caps were provided and there are discrepancies between them
+    // we'll be forgiving for now and just merge the two. But in the future, the clients
+    // should not be sending discrepant caps and Appium should ignore the extraneous MJSONWP caps
+    desiredCaps = {
+      ...desiredCaps, // TODO: This line should be removed once we decide to not allow discrepent caps
+      ...matchingW3CCaps,
+    };
+  }
+
+  let newJsonwpCaps, newW3CCapabilities;
```
mykola-mokhnach: can we have better names for these vars? For example, `processedCaps`

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+  // Translate the matched W3C caps back into W3C capabilities object so that it
+  // can be proxied to the inner driver
+  if (hasW3CCaps) {
+    newW3CCapabilities = {
+      alwaysMatch: {...insertAppiumPrefixes(desiredCaps)}, // Insert 'appium:' prefix to non-prefixed caps so that the Inner Driver doesn't complain about bad W3C prefixes
+      firstMatch: [{}],
+    };
+  }
+
+  return {desiredCaps, newJsonwpCaps, newW3CCapabilities};
```
mykola-mokhnach: so it's ok if some of these are undefined?

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+ * @param {*} jsonwpCaps
+ * @param {*} w3cCapabilities
+ * @param {*} constraints
+ * @param {*} defaultCapabilities
+ */
+function parseCapsForInnerDriver (jsonwpCaps, w3cCapabilities, constraints={}, defaultCapabilities={}) {
+  // Check if the caller sent JSONWP caps, W3C caps, or both
+  const hasW3CCaps = _.isObject(w3cCapabilities);
+  const hasJSONWPCaps = _.isObject(jsonwpCaps);
+
```
mykola-mokhnach: should we verify, that at least any of these caps is set

lib/utils.js
```diff
@@ -30,4 +31,116 @@ function inspectObject (args) {
...
+    };
+  }
+
+  return {desiredCaps, newJsonwpCaps, newW3CCapabilities};
+}
+
+
+/**
+ * Takes a capabilities objects and prefixes capabilities with `appium:`
+ * @param {*} caps {Object} Desired capabilities object
```
mykola-mokhnach: * -> Object

lib/utils.js
dpgraham: It is yeah. That's how drivers distinguish between JSONWP vs w3c

lib/utils.js
dpgraham: It does what Object.assign does

lib/utils.js
dpgraham: We don't need to. The route handler validates the parameters so it's impossible for none of them to be set.

lib/utils.js
mykola-mokhnach: I'd rather pass null instead of undefined

lib/utils.js
dpgraham:
```javascript
w3cCapabilities = {
    ...w3cCapabilities,
    alwaysMatch: {
      ...defaultCapabilities,
      ...w3cCapabilities.alwaysMatch,
    },
  };
```

is equivalent to:

```javascript
w3cCapabilities = Object.assign(
  {},
  w3cCapabilities,
  {
    alwaysMatch: {
      Object.assign(
        {}, 
        defaultCapabilities, 
        w3cCapabilities.alwaysMatch,
    }
  },
```

mykola-mokhnach (approved):

test/driver-e2e-specs.js
```diff
@@ -17,9 +17,11 @@ const caps = {platformName: "Fake", deviceName: "Fake", app: TEST_FAKE_APP};
 
 describe('FakeDriver - via HTTP', () => {
   let server = null;
+  let baseUrl;
   before(async () => {
     if (shouldStartServer) {
       let args = {port: TEST_PORT, host: TEST_HOST};
+      baseUrl = `http://${TEST_HOST}:${TEST_PORT}/wd/hub/session`;
```
imurchie: If we use a running server this means that `baseUrl` will remain undefined. On line 20 it should just be 
```
const baseUrl = `http://${TEST_HOST}:${TEST_PORT}/wd/hub/session`;
```

imurchie (changes requested):

jlipps:
by the way, now that we're all w3c-y, shouldn't we think about removing the legacy `/wd/hub` base? i bet we could just re-register all the endpoints at `/` and they could both live happily for some time.

dpgraham:
@jlipps Would that be something that would also need to be refactored on the clients?

Issue #8305, opened 2017-04-21
Need Capability/Feature to specify the Android user profile for driver initialisation/App Launch Activity


vinodbenure:
## The problem

For Android OS Version 5.0 and above, An Android device can have multiple user profiles (For eg. Android For Work). However, When Appium driver is initialised , appium driver by default points to the default user profile and there is no way in Appium to mention which user profile to pick before driver initialisation.

## Environment

* Appium version (or git revision) that exhibits the issue: 1.6.3
* Last Appium version that did not exhibit the issue (if applicable):NA
* Desktop OS/version used to run Appium: Windows 7
* Node.js version (unless using Appium.app|exe): 6.9.4
* Mobile platform/version under test: Android 6.0
* Real device or emulator/simulator: Real Device (Nexus 5)
* Appium CLI or Appium.app|exe: Appium CLI

## Details

When Appium driver is initialised , appium driver by default points to the default user and there is no way in Appium to mention which user profile to pick before driver initialisation.

For e.g. If there are two users (user A[default user] and user B). If a user wants to initialise  driver for an app installed in the user profile B then currently there is no way to select the user B or any other user.

We can get a list of available profiles in the device using below command:
**`adb shell pm list users`**

However, We can mention the user if we need to launch the app through command line as below:
**`adb shell am start --user USER_PROFILE_VALUE_IN_INT -n com.package./ActivityName`**

_But, When the present appium tries to launch the app or tries to initialise the app, it does not pass the **--user** parameter as above mentioned. Due to this appium driver always points to the default user profile._

If we can find a way to mention user profile name then we can easily get access to app which is installed any user profile.

## Link to Appium logs

You can clearly see in the logs that appium does not mention user profile name while launching the app

[Link to Logs](https://gist.github.com/vinodbenure/aad575fabec4de4cba6915a8fbbe6b95)

mykola-mokhnach:
Thanks for the suggestion @vinodbenure. However, you could create a PR yourself. We will be happy to review it.

dpgraham:
This sounds pretty sensible to me. I don't think the Appium core team will be able to get around to this any time in the near future but if @vinodbenure or anyone else in the community wants to tackle this, then we'd be happy to accept a pull request for this. It doesn't like it would be too difficult to implement.

jlipps:
Hi @vinodbenure the appium team would be happy to help walk you through a PR. Please take whatever idea you have and start the PR process and we will help make sure it is good.

[...]

PR #10005, opened 2018-01-17
Added new Capabilty for Android user profile

vinodbenure:
## Proposed changes
Added new capability as per #8305 

I am sure that we might need to make changes in many other places to get this working. I am very new to Appium. If @mykola-mokhnach  or @jlipps  or anybody from appium dev team helps me to solve this issue, I will be happy to follow the instructions and make changes accordingly.

Idea is to make this capability optional and if user adds this capability then the Android driver should be initialized with the app which installed the user profile id which is mentioned in the '**user**' as new capability. Also, I guess we need to pass this argument to adb command wherever the adb command is used to interact with app which is installed in the profile which is mentioned in the capability.

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [ x] I have signed the [CLA](https://cla.js.foundation/appium/appium)

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -483,6 +483,16 @@ const deprecatedArgs = [
     help: '[DEPRECATED] - Unique device identifier of the connected physical device',
   }],
 
+  [['--user'], {
+    dest: 'user',
+    required: false,
+    defaultValue: 0,
+	type: 'int',
+    example: '10',
+    deprecatedFor: '--default-capabilities',
+    help: '[DEPRECATED] - (Android-only) Unique user profile id if the device has more than one user profiles',
+  }],
+  
   [['--orientation'], {
     dest: 'orientation',
     defaultValue: null,


mykola-mokhnach:
should the capability be global? I'd rather add it to the Android driver

jlipps (commented):
Hi @vinodbenure thanks for this. A few things:

1. I think @mykola-mokhnach is right that we should add this to the appium-android-driver package, not the main appium package

2. In that package, we can simply add it to the list of capabilities; there is no need for it to be a server flag

3. I'd prefer the capability name `userProfile`, since `user` is a bit generic and might end up meaning different things.

Why don't you check out the appium-android-driver project and see if you can find out how to add this capability there, and we'll help you along.

vinodbenure:
@jlipps  and @mykola-mokhnach Thanks for helping me. I will create a new pull request for android-driver.

PR #10019, opened 2018-01-18
Use MJSONWP if dcaps have extraneous params

dpgraham:
* Update `parseCapsForInnerDriver` so that now if it finds extraneous keys in `jsonwpCaps`, fall back to using MJSONWP instead of W3C
* After getting processed caps, check the protocol again because if `processedW3CCapabilities` was set to null, it should be JWP
* Removed tests that tested for capabilities being merged together (those are no longer needed)
* Added tests (unit test and e2e test) that check that creation of a new session will fall back to JSONWP when there are extraneous keys

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -64,54 +64,37 @@ function parseCapsForInnerDriver (jsonwpCaps, w3cCapabilities, constraints={}, d
 
   // Get MJSONWP caps
   let desiredCaps = {};
+  let processedJsonwpCaps = null;
   if (hasJSONWPCaps) {
     desiredCaps = jsonwpCaps;
+    processedJsonwpCaps = {...desiredCaps};
   }
 
   // Get W3C caps
+  let processedW3CCapabilities = null;
   if (hasW3CCaps) {
     // Call the process capabilities algorithm to find matching caps on the W3C
     // (see: https://github.com/jlipps/simple-wd-spec#processing-capabilities)
     let matchingW3CCaps = processCapabilities(w3cCapabilities, constraints, true);
+    desiredCaps = matchingW3CCaps;
 
-    if (hasJSONWPCaps) {
-      let jsonwpKeys = _.keys(desiredCaps);
-      let matchingW3CCapsKeys = _.keys(matchingW3CCaps);
-
-      // If JSONWP has keys that W3C doesn't have, log a warning message
-      let differingKeys = _.differenceWith(jsonwpKeys, matchingW3CCapsKeys);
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -64,54 +64,37 @@ function parseCapsForInnerDriver (jsonwpCaps, w3cCapabilities, constraints={}, d
...
 
-      // If JSONWP has keys that W3C doesn't have, log a warning message
-      let differingKeys = _.differenceWith(jsonwpKeys, matchingW3CCapsKeys);
+    // If we found extraneuous keys in JSONWP caps, fall back to JSONWP
+    if (hasJSONWPCaps) {
+      let differingKeys = _.differenceWith(_.keys(jsonwpCaps), _.keys(matchingW3CCaps));
       if (!_.isEmpty(differingKeys)) {
-        logger.warn(`The following capabilities were provided in the JSONWP desired capabilities but
-          not the W3C capabilities: ${JSON.stringify(differingKeys)}`);
+        logger.warn(`The following capabilities were provided in the JSONWP desired capabilities that are missing
```
jlipps: this will look weird in the log with a newline; i would just use concatenation to put the string across lines

jlipps (approved):
LGTM, one style comment

lib/utils.js
```diff
@@ -64,54 +64,37 @@ function parseCapsForInnerDriver (jsonwpCaps, w3cCapabilities, constraints={}, d
 
   // Get MJSONWP caps
   let desiredCaps = {};
+  let processedJsonwpCaps = null;
```
mykola-mokhnach: Caps or Capabilities?

lib/utils.js
```diff
@@ -64,54 +64,37 @@ function parseCapsForInnerDriver (jsonwpCaps, w3cCapabilities, constraints={}, d
...
+    processedW3CCapabilities = {
+      alwaysMatch: {...insertAppiumPrefixes(desiredCaps)},
+      firstMatch: [{}],
+    };
 
-      // If JSONWP has keys that W3C doesn't have, log a warning message
-      let differingKeys = _.differenceWith(jsonwpKeys, matchingW3CCapsKeys);
+    // If we found extraneuous keys in JSONWP caps, fall back to JSONWP
+    if (hasJSONWPCaps) {
+      let differingKeys = _.differenceWith(_.keys(jsonwpCaps), _.keys(matchingW3CCaps));
```
mykola-mokhnach: why not to use just _.difference?

imurchie (approved):
I agree with Mykola and Jonathan's comments.

PR #10073, opened 2018-01-25
Use uia2 backend by default if platform version is greater or equal to 6

mykola-mokhnach:
## Proposed changes

Use uia2 backend by default if platform version is greater or equal to 6

## Types of changes

## Further comments

We had to do this a while ago. UIObject is considered as obsolete and is not supported anymore by Google since Android 6 (or even 5).

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -90,15 +90,11 @@ class AppiumDriver extends BaseDriver {
 
     if (caps.platformName.toLowerCase() === 'android') {
       const platformVersion = semver.valid(semver.coerce(caps.platformVersion));
-      if ((!_.isString(caps.automationName) || caps.automationName.toLowerCase() !== 'uiautomator')
-          && platformVersion && semver.satisfies(platformVersion, '>=6.0.0')) {
-        log.info("Requested Android support with version >= 6, " +
-                 `using ${this.validAutomations.UIAUTOMATOR2.toLowerCase()} ` +
-                 "driver instead of UIAutomation-based driver, since the " +
-                 "latter is considered as obsolete on Android 6 and up. " +
-                 "Assign 'automationName' capability to 'uiautomator' " +
-                 "if this is an undesired behavior.");
-        return AndroidUiautomator2Driver;
+      if (platformVersion && semver.satisfies(platformVersion, '>=6.0.0')) {
+        log.warn("Consider setting 'automationName' capability to " +
+                 `${this.validAutomations.UIAUTOMATOR2.toLowerCase()} ` +
+                 "on Android >= 6, since UIAutomation1 framework " +
+                 "is neither supported nor maintaied by the OS vendor.");
       }
 
       return AndroidDriver;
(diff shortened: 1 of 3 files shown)


imurchie (commented):
While in theory I'm ok with this. it would necessitate publishing Appium 2.0, since it is a breaking change.

At this point I would suggest just logging the fact that the results might be inconsistent, telling them to set the `automationName`, and allowing them to go on their merry way.

jlipps (commented):
i agree with isaac, i think we can't push this now but should log a warning to the user instead saying they're using a deprecated driver

mykola-mokhnach:
Ok, lets put a warning message instead

dpgraham (approved):

mykola-mokhnach:
@imurchie are you ok with these changes?

lib/appium.js
```diff
@@ -88,18 +89,25 @@ class AppiumDriver extends BaseDriver {
     }
 
     if (caps.platformName.toLowerCase() === 'android') {
+      const platformVersion = semver.valid(semver.coerce(caps.platformVersion));
+      if (platformVersion && semver.satisfies(platformVersion, '>=6.0.0')) {
+        log.warn("Consider setting 'automationName' capability to " +
+                 `${this.validAutomations.UIAUTOMATOR2.toLowerCase()} ` +
```
imurchie: The name should be in quotes.

lib/appium.js
```diff
@@ -88,18 +89,25 @@ class AppiumDriver extends BaseDriver {
...
+                 "on Android >= 6, since UIAutomation1 framework " +
```
imurchie: UIAutomator

lib/appium.js
```diff
@@ -88,18 +89,25 @@ class AppiumDriver extends BaseDriver {
...
-        if (parseInt(majorVer, 10) >= 10) {
-          log.info("Requested iOS support with version >= 10, using XCUITest " +
-                   "driver instead of UIAutomation-based driver, since the " +
-                   "latter is unsupported on iOS 10 and up.");
-          return XCUITestDriver;
-        }
+      const platformVersion = semver.valid(semver.coerce(caps.platformVersion));
+      if (platformVersion && semver.satisfies(platformVersion, '>=10.0.0')) {
+        log.info("Requested iOS support with version >= 10, " +
+                 `using ${this.validAutomations.XCUITEST.toLowerCase()} ` +
```
imurchie: Quotes around name.

imurchie (changes requested):

lib/appium.js
mykola-mokhnach: fixed

lib/appium.js
mykola-mokhnach: added quotes

lib/appium.js
mykola-mokhnach: added quotes

imurchie (approved):

PR #10076, opened 2018-01-25
Remove unnecessary 'finally' block

dpgraham:
* Shouldn't have had finally block because then it does two checks to fall back to MJSONWP
* Also fixed outdated tests


imurchie:
Any idea why the tests modified would be flakey? Were they always like that?

dpgraham:
@imurchie They weren't flakey, they were outdated. Previously, `appium-base-driver` was incorrectly just returning the error as a message. Now, it returns errors as:

```javasript
{
  value: {message: "...", stacktrace: "...", stacktrace: "..."}
}
```

imurchie:
Oh. but in one run it only failed once out of four in Travis, I thought.

dpgraham:
I think the failing ones had been using base driver version 2.20.1 which we would expect to fail. 2.20.2 it should pass.

mykola-mokhnach (approved):

PR #10155, opened 2018-02-08
Fix customSSLCert documentation error

dpgraham:
Should be Sim only, not Sim/Emu


imurchie (approved):
CI error has been fixed, I think.

PR #10205, opened 2018-02-15
Go back to using nodeca/argparse

dpgraham:
* nodeca/argparse was published with the `--help` flag fix
* going back to using that instead of self-published `appium-argparse`


imurchie:
Haha! Jumped the gun!

imurchie (approved):

PR #10231, opened 2018-02-21
docs update

trix0:
## Proposed changes

I think correct should be closeApp()
 http://webdriver.io/api/mobile/closeApp.html

.


imurchie (approved):
Yes. It's right in the link, but not here. Thanks for this!

PR #10362, opened 2018-03-16
Pass protocol in execute command

dpgraham:
* Previous implementations of W3C assumed that there was only one session running at a time
* We should be assuming that there are _multiple_ sessions that may have different protocols
* The fix is to make it so executeCommand returns protocol and value/error (instead of just throwing the value/error)
    * This is so that base driver can read the protocol and serve the response (base driver doesn't know what the protocol of the inner drivers are)
    * Added tests to confirm that it falls back to MJSONWP if an inner driver is not ready and handles concurrent MJSONWP and W3C sessions
    * Updated unit tests to match new format

* e.g.)

  * Previously executeCommand `getScreenshot` would have returned

   ```javascript
    "somescreenshotdata...."
    ```

  * Now it returns

    ```javascript
    { protocol: "W3C | MJSONWP", value: "somescreenshotdata...." }
    ```

* The new base driver changes (https://github.com/appium/appium-base-driver/compare/dpgraham-parse-protocol-from-execute-command?expand=1) looks for this new format and sends the appropriate HTTP response

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -335,10 +335,17 @@ class AppiumDriver extends BaseDriver {
         // be in otherwise
         delete this.sessions[sessionId];
       });
-      await dstSession.deleteSession(sessionId, otherSessionsData);
+      return {
+        protocol: this.protocol,
+        value: await dstSession.deleteSession(sessionId, otherSessionsData),
+      };
     } catch (e) {
       log.error(`Had trouble ending session ${sessionId}: ${e.message}`);
-      throw e;
+      //throw e;
+      return {
+        protocol: this.protocol,
+        error: e,
+      };
     }
   }
 
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -187,81 +187,96 @@ class AppiumDriver extends BaseDriver {
    */
   async createSession (jsonwpCaps, reqCaps, w3cCapabilities) {
     let {defaultCapabilities} = this.args;
-
-    // Parse the caps into a format that the InnerDriver will accept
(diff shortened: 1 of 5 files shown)


mykola-mokhnach:
Such change might require many changes in all underlying methods. I'd rather add `X-Protocol` HTTP header to all responses and put the actual protocol value there instead of changing the response format

dpgraham:
@mykola-mokhnach The `executeCommand` function doesn't return an HTTP response though. This solution was made to be backwards compatible so that it still accepts the original form of `executeCommand`.

imurchie (commented):
I'm ok with this as it is. Tests need to be investigated, though.

dpgraham:
It just needed a new appium base driver version.

jlipps (commented):
also OK with this. it does mean that only appium will be able to do the translation (i.e., we're not W3C-ifying all the inner drivers)

dpgraham:
@jlipps The inner drivers are W3C compliant too.

The way this works is that:

* A session is created and protocol of the session is determined based on session creation payload (`{alwaysMatch: {...}, firstMatch: [{...}...]}` vs `{desiredCapabilities: {...}}`
* When a session is created on an inner driver, the inner driver may not be W3C compliant yet
* Made it so that when appium calls `executeCommand` on the inner driver, it checks the inner driver's protocol, and tells `buildHandler` to serve a response using the protocol that the inner driver is using
* So for example, if `appium/appium` is running a W3C session and `appium/appium-youiengine-driver` is running an MJSONWP session, it will return a result in `executeCommand` that indicates that it's MJSONWP and `buildHandler` returns MJSONWP instead of W3C

Issue #6098, opened 2016-02-02
React Native test apps


scottdixon:
I propose we use [React Native](https://facebook.github.io/react-native/) to generate future test apps.

Some of the benefits:
- Create both iOS & Android applications using JS.
- Appium developers will be familiar with the ES6 syntax.

webzepter:
:+1:

scottdixon:
hmm this might not work.

`...a button is not implemented as a native UIButton, and a ListView doesn't use UITableView. Instead they are composed of native UIViews with a lot of JavaScript glue..`

GeoffreyPlitt:
Does Appium work for React-Native Android yet?

saikrishna321:
yah it does work . I run my test against react-native Android and iOS

zachrnolan:
@saikrishna321 are you on xcode 8? It looks like Appium doesn't support it yet.

saikrishna321:
yah I'm using xcode 8and test run on iOS 10

Do you have difficulty with locating the elements ?

stereodenis:
@saikrishna321 could you get link to your tests?

saikrishna321:
@stereodenis https://github.com/saikrishna321/VodQaAdvancedAppium

KazuCocoa:
https://github.com/appium/sample-code/blob/master/sample-code/apps/VodQAReactNative.zip have been added by @saikrishna321

saikrishna321:
Closing as we have React Native app in place 
github.com/appium/VodQAReactNative

PR #10427, opened 2018-03-27
Java sample code

dpgraham:
* Written in IntelliJ and uses TestNG framework (which is basically an advanced version of JUnit)

### DIFF
--- a/sample-code/java/src/AndroidBasicInteractionsTest.java
+++ b/sample-code/java/src/AndroidBasicInteractionsTest.java
@@ -1,48 +1,45 @@
 import io.appium.java_client.android.Activity;
 import io.appium.java_client.android.AndroidDriver;
 import io.appium.java_client.android.AndroidElement;
-import io.appium.java_client.service.local.AppiumDriverLocalService;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.remote.DesiredCapabilities;
 import org.openqa.selenium.support.ui.ExpectedConditions;
 import org.openqa.selenium.support.ui.WebDriverWait;
 import org.testng.Assert;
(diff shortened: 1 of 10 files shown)


sample-code/java/src/AndroidBasicInteractionsTest.java
```diff
@@ -0,0 +1,74 @@
...
+        capabilities.setCapability("app", app.getAbsolutePath());
+        capabilities.setCapability("appPackage", PACKAGE);
+        capabilities.setCapability("appActivity", SEARCH_ACTIVITY);
+        driver = new AndroidDriver<WebElement>(service.getUrl(), capabilities);
+    }
+
+    @AfterSuite
+    public void tearDown() {
+        driver.quit();
+    }
```
SrinivasanTarget: Ideally we should stop the service too. @dpgraham Have another query, are we running these tests in Travis? Or it's just available as sample for reference?

SrinivasanTarget (changes requested):

saikrishna321:
@dpgraham @jlipps as we are getting new sample code in place for all clients, can we start using the RN app as test application (https://github.com/appium/VodQAReactNative)

https://github.com/appium/appium/issues/6098

sample-code/java/src/AndroidBasicInteractionsTest.java
dpgraham: This is just for sample but if we could run it in Travis that would be a bonus.

SrinivasanTarget: @dpgraham we can do that too.

dpgraham:
Fixes pushed.

SrinivasanTarget (approved):

Issue #6257, opened 2016-03-09
[Appium 1.5] IOS coordinate reversed, vertical is x, horizontal is y.


HailongZhang:
## The problem

IOS coordinate is reversed, vertical is x, horizontal is y, so swipe and tap coordinate not working after upgrade. Android platform works well, no change.

## Environment

* Appium version (or git revision) that exhibits the issue: 1.5.0
* Last Appium version that did not exhibit the issue (if applicable): 1.4.16
* Desktop OS/version used to run Appium: Mac 10.11.3
* Node.js version (unless using Appium.app|exe):
* Mobile platform/version under test: iphone 6s
* Real device or emulator/simulator:
* Appium CLI or Appium.app|exe:

## Details

After upgrade appium from 1.4.16 to 1.5.0, Appium::TouchAction.new.press(x: start_x, y: start_y).wait(duration).move_to(x: end_x, y:  end_y).release.perform not working. I found start_x changed to start_y, start_y changed to start_x. 

## Code To Reproduce Issue [ Good To Have ]
 Not working: Appium::TouchAction.new.press(x: 300, y: 100).wait(500).move_to(x: 50, y:  100).release.perform.  # swipe from right to left, ever worked in 1.4.16
If switch start_x and start_y, it works:
Appium::TouchAction.new.press(x: 100, y: 300).wait(500).move_to(x: 100, y:  50).release.perform
![screen shot 2016-03-08 at 10 33 50 am](https://cloud.githubusercontent.com/assets/16573179/13624449/375064da-e5e9-11e5-8e47-78656a584448.png)

scottdixon:
Thanks for reporting @HailongZhang. Looking into it. Can you please include your Appium logs as a gist?

HailongZhang:
@scottdixon 
I think the log is not much useful. If the coordinate is with screen, the action is success, no error message, if the coordinate is out of screen , it shows as the attached screenshot.

Thanks

![screen shot 2016-03-10 at 5 00 32 pm](https://cloud.githubusercontent.com/assets/16573179/13664377/d828abd0-e6e1-11e5-92ae-6a09522a2bae.png)

scottdixon:
The logs would be helpful in determining where the problem lies (ruby client, Appium or instruments). Please post your 'success' logs.

shermanericSF:
Hey Scott.  I work with Zhe who originally posted this.  I'll put some
gists up in addition to whatever Zhe may have.
Specifically:

[...]

PR #10445, opened 2018-03-30
Update move examples

wswebcreation:
I've been struggling the last days with the `move` command because the `moveTo` implies an absolute position, but it's just a relative offset, see also https://github.com/appium/appium/issues/6257#issuecomment-196949576

This PR hopefully makes it more clear

### DIFF
--- a/docs/en/commands/interactions/touch/move.md
+++ b/docs/en/commands/interactions/touch/move.md
@@ -1,6 +1,6 @@
 # Tap
 
-Finger move on the screen
+Finger move on the screen where the movement is relative to the initial press
 ## Example Usage
 
 ```java
@@ -24,9 +24,17 @@ actions.perform()
 ```javascript
 // Javascript
 // webdriver.io example
+// Move to 200,200 (as in, an additional 100 in each direction)
 driver.touchMultiPerform([
-  { action: 'press', options: { x: 100, y: 250 }},
-  { action: 'moveTo', options: { x: 300, y: 100 }},
+  { action: 'press', options: { x: 100, y: 100 }},
+  { action: 'moveTo', options: { x: 100, y: 100 }},
+  { action: 'release' }
+]);
+
+// Move to // Move to 0,0 (as in, an additional -100 in each direction)
+driver.touchMultiPerform([
+  { action: 'press', options: { x: 100, y: 100 }},
+  { action: 'moveTo', options: { x: -100, y: -100 }},
   { action: 'release' }
 ]);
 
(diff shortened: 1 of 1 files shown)


docs/en/commands/interactions/touch/move.md
```diff
@@ -24,9 +24,17 @@ actions.perform()
...
+// Move to 200,200 (as in, an additional 100 in each direction)
 driver.touchMultiPerform([
-  { action: 'press', options: { x: 100, y: 250 }},
-  { action: 'moveTo', options: { x: 300, y: 100 }},
+  { action: 'press', options: { x: 100, y: 100 }},
+  { action: 'moveTo', options: { x: 100, y: 100 }},
+  { action: 'release' }
+]);
+
+// Move to // Move to 0,0 (as in, an additional -100 in each direction)
```
jeaye: We've a double comment here.

jeaye:
This should save people a lot of time. Thank you, @wswebcreation!

jeaye (changes requested):

docs/en/commands/interactions/touch/move.md
wswebcreation: Fixed it

tfcardoso:
@wswebcreation you sure? I was under the impression that behavior had been changed by @mykola-mokhnach in https://github.com/appium/appium-android-driver/pull/292

mykola-mokhnach:
Yes the moveTo coordinates should be absolute according to the spec. However, previous appium version had a bug, where these coordinates were considered as relative.

wswebcreation:
Ok, so if I understand correctly this will be fixed in 1.8.0?

Meaning that this PR is only for the versions < 1.8.0

Maybe we should still tell something about it in the docs, I've seen a lot of questions in the webdriver.io channel concerning this part

jeaye (approved):

dpgraham:
Sorry for the late review on this one.

This .md file is actually auto-generated and the actual file that needs to be edited is: https://github.com/appium/appium/blob/master/commands-yml/commands/interactions/touch/move.yml

@wswebcreation Do you want to transfer your changes over to the .yml file or do you want me to do that?

wswebcreation:
Hi @dpgraham

I’d like to but is this still valid. Because when I look at the remark of @mykola-mokhnach I do wonder.

jlipps:
@dpgraham we should put a big invisible comment at the top of these files that they are auto-generated and shouldn't be edited by hand.

dpgraham:
@jlipps Yeah, that's a good idea

PR #10498, opened 2018-04-08
ignore lib

Bamieh:
no need to send the unbuilt files to npm

## Proposed changes

Ignore `lib` folder for npm releases since the `build` folder is the entry file of the project.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -14,3 +14,4 @@ _vimrc_local.vim
 *.prefs
 ./.npmignore
 coverage
+./lib


jlipps:
I think we should still send the source files to NPM, that way it's easier to debug from within an install. exceptions show line numbers in the lib files, so if we don't include those in the module, I think it might be confusing for people who are looking at an npm install rather than a git clone.

Bamieh:
@jlipps sadly my errors throw at babel level, `_generatorAsync` or such.  Sourcemaps are used solve this issue. Digging through source code is a very niche use case IMHO, people who dig through the sourcecode will also dig it through github. Its a good practice to keep the npm package size at a minimum.

imurchie:
I'm good with this. I thought we had done it a long time ago, actually.

_However_, this will have to be done in all the packages to have any particular affect on package size. The size of the `lib` directory in this repo is negligible in comparison to other factors. Right now it's about 84k while the `release.pdf` file is 80k (and node_modules is 850M).

mykola-mokhnach:
I'd rather keep the `lib` folder, at least for beta releases. It is sometimes quite handy if it is needed to check a quick fix there.

jlipps:
Given the difficulty in making this change project-wide, and the relatively small gain in ignoring lib in npm releases, my feeling is that it's not worth it. If you feel like this should be done though @imurchie I'm OK with it.

imurchie:
I'm ok with not doing it. There are bigger fish to fry in terms of stuff to include/not include. 

Once Babel 7 comes out I have to go through every package to update, so I can adjust things then.

imurchie:
Thanks for the work, though, @Bamieh , and the impulse to get it smaller. Hopefully the next release will be much smaller.

PR #10499, opened 2018-04-08
Collapsable widget instead of gists

Bamieh:
Use collapsable widget instead of asking submitters to submit a gist. This is easier for the issue submitter without compromising readability.

The logs in the issue will look like this:

## Appium logs
<details><summary>Appium Logs</summary>
<p>
  
  ```
  [ PASTE YOUR APPIUM LOGS HERE ]
  ```
  
</p>
</details>


jlipps (commented):
interesting, what do you think @imurchie ?

imurchie:
I personally would still like to have the logs in a gist, so they are openable in another tab. I guess the issue could be open twice. But I am partial to the gist.

PR #10614, opened 2018-04-26
Document `executeMobile` commands

dpgraham:
* We have several `mobile:` commands and many of them aren't documented
* Some of them are documented under `writing-running-appium/ios`
* I'd like to have this document be the central place where you document `mobile:` commands
* If the command needs a heavy amount of documentation though, we can still write separate docs but then link to them from here


commands-yml/commands/mobile-command.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      List of available commands:
+
+      ### IOS
+      | Command | Description | Argument | Argument Example |
+      | ------- | ----------- | -------- | ---------------- |
+      | mobile:viewportScreenshot | Like [screenshot](/commands/session/screenshot/) but only includes contents of viewport | <none> | <none> |
+
+      ### IOS (XCUITest Only)
+      * [IOS XCTest Mobile Gestures](/writing-running-appium/ios/ios-xctest-mobile-gestures)
```
jlipps: still think we should list the `mobile: xxx` name here for reference, and _theN_ link to it

jlipps (approved):
i like this. i think we should either do this, or have the "executeScript 'mobile' commands" actually be a category, and give every command its own page (just like we do for other commands)

commands-yml/commands/mobile-command.yml
```diff
@@ -0,0 +1,129 @@
...
+      | mobile:scrollBackTo | Scroll from one element to another | `{elementId, elementToId}` | `{elementId: 2, elementToId: 1}` |
+      | mobile:viewportScreenshot | Like [screenshot](/commands/session/screenshot/) but only includes contents of viewport | <none> | <none> |
+      | mobile:deepLink | Opens a deep-link URL for testing [Instant Apps](https://support.google.com/googleplay/answer/7240211?hl=en) | `{url, package}` | `{url: "https://www.site.com/", package: "com.site.SomeAndroidPackage"}` |
+      
+  
+  example_usage:
+    java:
+      |
+        // java
+        JavascriptExecutor js = (JavascriptExecutor) driver;
```
mykola-mokhnach: AppiumDriver driver anyway implements JavascriptExecutor interface, so this can be simplified as `driver.executeScript`. There is no need to do any typecast

docs/en/commands/mobile-command.md
```diff
@@ -0,0 +1,149 @@
...
+|[Javascript (WD)](https://github.com/admc/wd/releases/latest)| All |  [github.com](https://github.com/admc/wd/blob/master/lib/commands.js#L102)  |
+|[Ruby](https://github.com/appium/ruby_lib/releases/latest)| All |  [www.rubydoc.info](http://www.rubydoc.info/gems/selenium-webdriver/Selenium/WebDriver/Driver:execute_script)  |
+|[PHP](https://github.com/appium/php-client/releases/latest)| All |  [github.com](https://github.com/appium/php-client/)  |
+|[C#](https://github.com/appium/appium-dotnet-driver/releases/latest)| All |  [github.com](https://github.com/appium/appium-dotnet-driver/)  |
+
+## HTTP API Specifications
+
+### Endpoint
+
+`GET /wd/hub/session/:session_id/execute`
```
mykola-mokhnach: I assume the endpoint looks different in W3C mode

dpgraham: Yeah, this was done before W3C. I'll need to update it in the future to distinguish JWP from W3C

dpgraham:
Fixes pushed

mykola-mokhnach (approved):

PR #10660, opened 2018-05-07
Re-enable package-lock.json

dpgraham:
## Proposed changes

It turns out that disabling package-lock.json _also_ disables npm-shrinkwrap.json, which means that users who have been installing Appium using npm >= 5 have been getting a non-shrinkwrapped version of Appium.

## Types of changes

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,2 @@
 # Do not disable package-lock. Disabling package-lock also disables npm-shrinkwrap
-# package-lock=false
+package-lock=true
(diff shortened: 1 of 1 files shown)


jlipps:
should we do this or just run npm shrinkwrap as part of the publish step?

dpgraham:
The issue is that when `package-lock=false`, if you install Appium it will ignore `npm-shrinkwrap.json`

.npmrc
```diff
@@ -1 +1,2 @@
-package-lock=false
\ No newline at end of file
+# Do not disable package-lock. Disabling package-lock also disables npm-shrinkwrap
+# package-lock=false
```
imurchie: I think this should be `package-lock=true` rather than commented out.

imurchie (approved):
Nice of npm to have these undocumented effects.

PR #10683, opened 2018-05-11
Add VSCode launch configurations

dpgraham:
This was a spur-of-the-moment thing. I did this in Appium Desktop earlier today to make it easier to troubleshoot an issue that was reported and figured I'd do the same in `appium/appium`.

* Added debug, attach-to-debug, test-all, and test-current-file
* The debug and attach to debug allow you to run Appium server and set breakpoints in the _source_ code (not just build code)
* `test-all` and `test-current-file` lets you run mocha tests from VSCode and also set up breakpoints in both the test files and source files
* Had to make some changes to how package.json is located to support being able to run Mocha tests directly from 'test/' instead of from 'build/test/'
* Doesn't change how tests are run in CI so if there's any inconsistencies between how VSCode runs tests and how it's run on terminal, errors will still be caught. This is just for the convenience of the developer.


lib/config.js
```diff
@@ -17,6 +18,12 @@ function getNodeVersion () {
 
 async function getGitRev () {
   let cwd = path.resolve(__dirname, "..", "..");
+
+  // Check if we're in the right folder, relative to package.json
+  if (!await fs.exists(path.resolve(cwd, 'package.json'))) {
```
jlipps: is there a way to know based on some external criteria, rather than just guessing and trying one and then another?

lib/utils.js
```diff
@@ -149,4 +149,17 @@ function insertAppiumPrefixes (caps) {
   return prefixedCaps;
 }
 
-export { inspectObject, parseCapsForInnerDriver, insertAppiumPrefixes };
+/**
+ * Get package.json object
+ */
+function getPkgObject () {
```
jlipps: this is duplicated in several places, why not make it part of something more global?

lib/config.js
dpgraham: There's a package called `find-root`. Would that be good?

dpgraham:
Fixes pushed

package.json
```diff
@@ -53,6 +53,7 @@
     "bluebird": "3.x",
     "continuation-local-storage": "3.x",
     "dateformat": "^3.0.3",
+    "find-root": "^1.1.0",
```
jlipps: does this need to be a real dep rather than a dev dep? isn't debugging for dev?

jlipps (approved):

package.json
dpgraham: Yeah good catch.

Down the road, we may need to make something like this a real dependency if we ever want to be able to run Appium on non-compiled code.

PR #10742, opened 2018-05-23
Dynamically figure out line numbers to routes file when necessary

imurchie:
## Proposed changes

The problem: We have links to `https://github.com/appium/appium-base-driver/blob/master/lib/protocol/routes.js` in our docs, linking directly to the line for the endpoint for the command. However, this breaks whenever we change something in the routes file, which is relatively often.

Rather than require one to remember to go into `appium/appium` and update the correct command yaml file, dynamically parse those files and add the correct line number at the time of publishing

Most of the work is in `commands-yml/parse.js`, as a Handlebars helper function.

Doing this work also showed a number of places where we had the wrong endpoint, so those files are fixed here.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/commands-yml/commands/device/activity/current-activity.yml
+++ b/commands-yml/commands/device/activity/current-activity.yml
@@ -58,7 +58,7 @@ client_support:
 
 # Information about the HTTP endpoints
 endpoint:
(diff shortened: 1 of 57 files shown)


jlipps (approved):
good idea!

PR #10933, opened 2018-06-22
Fix: Travis YML formatted incorrectly

dpgraham:
Last PR had messed up YML tabbing and that broke the Travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ jobs:
       node_js: 9
       branches:
         only:
-	        - master
+          - master
       os: osx # Need to use Mac OSX build to build XCUITestDriver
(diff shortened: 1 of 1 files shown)


imurchie:
Good. Also... stupid yaml.

dpgraham:
@imurchie Agreed.

imurchie:
Since you're mucking with the travis config, can you change the node versions? There is no need to test on _either_ of these, as both are EOLed (in general, we shouldn't bother with odd node versions anyway).

We support node 6, so we should test on that. Then perhaps node 8.

That all being said, the new changes to the config ignores the node version directive anyway. Only node 9 is being tested for the actual tests.

PR #10934, opened 2018-06-22
Update appium/appium node versions

dpgraham:


imurchie:
Something weird is up. The second job is not starting (see https://travis-ci.org/appium/appium/builds/395554146 as well as this PR).

imurchie:
An alternative arrangement: https://github.com/appium/appium/blob/isaac-travis/.travis.yml (in action: https://travis-ci.org/appium/appium/builds/396079454)

dpgraham:
@imurchie The Travis stages run sequentially, so that's probably why the 2nd one took time to start up.

PR #10960, opened 2018-06-27
Add option to use a remote debugger proxy

ThomasFettSauce:
## Proposed changes

The changes make it possible to let Appium's remote debugger connect to a proxy instead of the iOS simulator directly when running tests on iOS simulators.
This enables sending additional Web Inspector protocol messages to the simulator and also logging all Web Inspector protocol messages that are sent from Appium to the simulator and vice versa.

The actual implementation and message handling of the proxy is completely up to the developers who want to use this feature.

The pull requests for the other affected Appium packages of this feature are:
* https://github.com/appium/appium-xcuitest-driver/pull/710
* https://github.com/appium/appium-remote-debugger/pull/81

## Types of changes

## Checklist

### DIFF
--- a/docs/en/writing-running-appium/caps.md
+++ b/docs/en/writing-running-appium/caps.md
@@ -144,8 +144,7 @@ Driver](/docs/en/drivers/ios-uiautomation.md).
 |`appName`| The display name of the application under test. Used to automate backgrounding the app in iOS 9+.|e.g., `UICatalog`|
 |`customSSLCert`|(Sim only) Add an SSL certificate to IOS Simulator. | e.g. <br/>`-----BEGIN CERTIFICATE-----MIIFWjCCBEKg...`<br/>`-----END CERTIFICATE-----`|
 |`webkitResponseTimeout`|(Real device only) Set the time, in ms, to wait for a response from WebKit in a Safari session. Defaults to `5000`|e.g., `10000`|
-|`remoteDebugProxyPort`| (Sim only, <= 11.2) If set, Appium sends and receives remote debugging messages through a proxy on this local port instead of communicating with the iOS remote debugger directly. |e.g. `12000`|
-|`remoteDebugProxySocketPath`| (Sim only >= 11.3) If set, Appium sends and receives remote debugging messages through a proxy on this unix socket instead of communicating with the iOS remote debugger directly. |e.g. `"/tmp/my.proxy.socket"`|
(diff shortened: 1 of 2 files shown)


lib/parser.js
```diff
@@ -350,6 +350,27 @@ const args = [
     help: '(IOS-only) Local port used for communication with ios-webkit-debug-proxy'
   }],
 
+  [['--remote-debug-proxy-port'], {
+    defaultValue: false,
+    dest: 'remoteDebugProxyPort',
+    required: false,
+    type: 'int',
+    example: "12000",
+    help: '(IOS-only, <= 11.2) If set, Appium sends and receives remote debugging messages' +
```
imurchie: As implemented in the `appium-remote-debugger`, this will only work for simulators. That should be mentioned here and below.

imurchie (commented):
Why server flags rather than desired capabilities?

lib/parser.js
ThomasFettSauce: Thanks for the feedback! I wasn't aware that desired capabilities just work out of the box. I removed the server flags and added them as desired capabilities.

KazuCocoa:
Is it good to add them to https://github.com/appium/appium-ios-driver/blob/f6e335adf3d1d0c450df378aee9e1b85c90ee1e8/lib/desired-caps.js#L99 to validate them?

imurchie:
Yes, the final step for the desired capability is to add them as @KazuCocoa noted above. Without that, there will be a line in the logs saying that the cap is not recognized, and also the type of value would be unknown.

docs/en/writing-running-appium/caps.md
```diff
@@ -144,6 +144,8 @@ Driver](/docs/en/drivers/ios-uiautomation.md).
 |`appName`| The display name of the application under test. Used to automate backgrounding the app in iOS 9+.|e.g., `UICatalog`|
 |`customSSLCert`|(Sim only) Add an SSL certificate to IOS Simulator. | e.g. <br/>`-----BEGIN CERTIFICATE-----MIIFWjCCBEKg...`<br/>`-----END CERTIFICATE-----`|
 |`webkitResponseTimeout`|(Real device only) Set the time, in ms, to wait for a response from WebKit in a Safari session. Defaults to `5000`|e.g., `10000`|
+|`remoteDebugProxyPort`| (Sim only, <= 11.2) If set, Appium sends and receives remote debugging messages through a proxy on this local port instead of communicating with the iOS remote debugger directly. |e.g. `12000`|
```
jlipps: what about having just one capability `remoteDebugProxy` which could take either a port or a socket?

imurchie:
Everything is available in the drivers now, so merging. Thanks!

PR #10966, opened 2018-06-28
Refactor driver creation logic in the umbrella driver

mykola-mokhnach:
## Proposed changes

The previous implementation has unnecessary code duplication.

## Types of changes

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -5,16 +5,12 @@ import { BaseDriver, errors, isSessionCommand } from 'appium-base-driver';
 import { FakeDriver } from 'appium-fake-driver';
 import { AndroidDriver } from 'appium-android-driver';
 import { IosDriver } from 'appium-ios-driver';
-// eslint-disable-next-line no-unused-vars
 import { AndroidUiautomator2Driver } from 'appium-uiautomator2-driver';
-// eslint-disable-next-line no-unused-vars
 import { SelendroidDriver } from 'appium-selendroid-driver';
 import { XCUITestDriver } from 'appium-xcuitest-driver';
-// eslint-disable-next-line no-unused-vars
 import { YouiEngineDriver } from 'appium-youiengine-driver';
 import { WindowsDriver } from 'appium-windows-driver';
 import { MacDriver } from 'appium-mac-driver';
-// eslint-disable-next-line no-unused-vars
 import { EspressoDriver } from 'appium-espresso-driver';
 import { TizenDriver } from 'appium-tizen-driver';
 import B from 'bluebird';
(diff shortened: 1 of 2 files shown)


lib/appium.js
```diff
@@ -57,110 +60,155 @@ class AppiumDriver extends BaseDriver {
...
-        // but if we do and it is 'XCUITest', act on it
-        return XCUITestDriver;
-      } else if (caps.automationName.toLowerCase() === this.validAutomations.YOUIENGINE.toLowerCase()) {
-        // but if we do and it is 'YouiEngine', act on it
-        return YouiEngineDriver;
-      } else if (caps.automationName.toLowerCase() === this.validAutomations.ESPRESSO.toLowerCase()) {
-        log.warn('The Appium Espresso driver is currently in early beta and meant only for experimental usage. Its API is not yet complete or guaranteed to work. Please report bugs to the Appium team on GitHub.');
-        return EspressoDriver;
-      }
+    this._driversMapping = {
```
imurchie: All this can be a constant, at the top, rather than buried in here.

test/driver-specs.js
```diff
@@ -263,70 +263,70 @@ describe('AppiumDriver', function () {
         _.keys(appium.sessions).should.contain(sessionId);
       });
     });
-    describe('getDriverForCaps', function () {
-      it('should not blow up if user does not provide platformName', function () {
+    describe('getDriverForCaps', async function () {
```
imurchie: This does not need to be async.

lib/appium.js
mykola-mokhnach: we still need `this.validAutomations` to be available inside it, so I cannot move it out of the class. Perhaps, we could make it to a property

test/driver-specs.js
mykola-mokhnach: ok

lib/appium.js
imurchie: Oh. Why in the world is that information set on the driver instance? That is pointless. It should just be exported as a constant from the base driver.

mykola-mokhnach: I agree it is not very useful, but this is the fact we have now

imurchie: It seems to be there so that we can validate capabilities. But the normal drivers (outside AppiumDriver) will _never_ see an automation name that is invalid, since the code here on session start, _before_ the inner driver is picked, should fail if a bad one is sent.

Would you mind pulling that stuff (https://github.com/appium/appium-base-driver/blob/master/lib/basedriver/automation-names.js) over into here and just making a constant that this uses? I can make a PR to remove it from the base driver.

mykola-mokhnach: I'd rather let you design it inside base driver.

imurchie: Base driver should know nothing of this. This is something specific to AppiumDriver. I'm happy to do it, though.

imurchie (approved):
The second step in CI should be skipped, so this can be taken as passing. I presume the recently merged conditional wasn't in this branch.

lib/appium.js
mykola-mokhnach: In which file it would be better to define these constants for Appium driver? Right here, in appium.js or somewhere else?

mykola-mokhnach: or feel free to alert the stuff to this PR

mykola-mokhnach:
Ok, I'll merge this one, so you could refactor `this.validAutomations` stuff

jlipps (commented):
👍

PR #10967, opened 2018-06-28
Move maps into constants

imurchie:
## Proposed changes

Moce all driver-related specifications to constants. There should be no need to muck about in the code, just add to the mapping.

This also moves the list of valid automation names out of `appium-base-driver`, where it has no place. It seemed to be there so that the caps could be validated, but the individual drivers should never get an invalid automation name capability, since AppiumDriver ought to weed those out and throw an error before ever getting the particular driver.

## Types of changes

## Checklist

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -15,7 +15,7 @@ import { EspressoDriver } from 'appium-espresso-driver';
 import { TizenDriver } from 'appium-tizen-driver';
 import B from 'bluebird';
 import AsyncLock from 'async-lock';
-import { inspectObject, parseCapsForInnerDriver } from './utils';
+import { inspectObject, parseCapsForInnerDriver, getPackageVersion } from './utils';
 import semver from 'semver';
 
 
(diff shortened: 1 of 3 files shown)


dpgraham (approved):

lib/appium.js
```diff
@@ -18,6 +18,104 @@ import AsyncLock from 'async-lock';
...
+    automationName: AUTOMATION_NAMES.ESPRESSO,
+  },
+  TizenDriver: {
+    driverClass: TizenDriver,
+    moduleName: 'appium-tizen-driver',
+    automationName: AUTOMATION_NAMES.TIZEN,
+  },
+};
+// load the version for each driver
+for (const driverProps of _.values(DRIVER_MAP)) {
```
jlipps: doesn't feel super `const`y to be setting properties on this, and mutating what's inside `DRIVER_MAP` /shrug

jlipps (approved):

lib/appium.js
imurchie: Adjusted for Moar Constantiness.

lib/appium.js
```diff
@@ -60,125 +150,23 @@ class AppiumDriver extends BaseDriver {
...
           return driverClass;
         }
       }
     }
 
-    const driverSelector = this.platformsMapping[caps.platformName.toLowerCase()];
+    const driverSelector = PLATFORMS_MAP[caps.platformName.toLowerCase()];
     if (driverSelector) {
-      return await driverSelector(caps);
+      return driverSelector(caps);
```
mykola-mokhnach: I'd rather keep it sync just in case it will be necessary to add some more advanced driver selection algorithm

PR #11039, opened 2018-07-12
Update validation to allow multiple client urls

imurchie:
## Proposed changes

A recent doc update (09b80e2e625909b2e09c69d8e6581ddf56816b26) highlighted the need for updated validation on client urls, to allow for different urls for different devices.

This PR has three commits, in order of importance:
1. [3ed552dfb2f5a3b300c07475fce037d74842c736] Change to parsing to allow
      * bare url (the current situation), which is parsed to display `[hostname](url)` as link
      * list with three possibilities: 
        * `url` - same as above
        * `ios` - link becomes `[iOS](url)`
        * `android` - link becomes `[Android](url)`
      * this list will have to be expanded, or some mechanism for allowing more types, should the need arise.
1. [eb3c50325c522f71dfc4e3cf47569eeb80850ba9] Add the new pages to the table of contents, so they show up in appium.io
1. [b95e283063837df4563fce56a60f8d21b1ef08d5] Generate the documentation with these parsing changes (i.e., run `npm run generate-docs`)

## Types of changes

## Checklist

Cc: @KazuCocoa


commands-yml/parse.js
```diff
@@ -98,9 +98,39 @@ Handlebars.registerHelper('if_eq', function (a, b, opts) {
...
+
+  if (!_.isArray(clientUrl)) {
+    return createUrlString(clientUrl);
+  }
+
+  let urlStrings = [];
+  for (const item of clientUrl) {
+    for (let [key, value] of _.toPairs(item)) {
+      key = key.toLowerCase();
+      const urlStr = ['ios', 'android'].includes(key)
```
mykola-mokhnach: should these magic strings be hardcoded? should we make it more flexible in case we'd like to have more options?

imurchie: I think that at the moment it is unnecessary, and it's not immediately obvious how it should best be done. So better solved when there is a need, rather than over-engineering it right now with a "solution" that might not do what we need.

commands-yml/commands/example.yml
```diff
@@ -31,8 +31,11 @@ example_usage:
       // CSharp Code here
 
 client_docs:
-  java: "https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/WebElement.html#click--"
-  python: "https://github.com/appium/python-client"
+  java:
+    - url: "https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/WebElement.html#click--"
+  python:
+    - android: "https://github.com/appium/python-client"
+      ios: "https://github.com/appium/python-client"
```
jlipps: does this line need a `-`? the recording screen one has a `-` for each platform

jlipps: also, both lines are the same...

docs/toc.js
```diff
@@ -101,6 +101,10 @@ module.exports = {
           ["Get Performance Data", "get-performance-data.md"],
           ["Performance Data Types", "performance-data-types.md"],
         ]],
+        ["Recording Screen", ["recording-screen",
```
jlipps: I would expect this to be phrased as either `Screen Recording` or `Recording the Screen`

commands-yml/commands/example.yml
imurchie: It works either way. If it is under a single dash, then the generating function is called with an array of one element with two properties, and if it is with two dashes it is called with an array of two elements each with one property. Either way it comes out the same.

As for this being the same, this is just the example. But I've changed them.

docs/toc.js
imurchie: Updated.

PR #11067, opened 2018-07-18
Refactor driver selection

imurchie:
## Proposed changes

On a boring flight with broken WiFi, I got to thinking about why the driver configuration is in the code at all. Couldn't it be in a configuration file like a sane system?

This is a first iteration on making it so. It is just what popped into my head, and I'm sure there are many improvements.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/drivers.json
+++ b/drivers.json
@@ -1,0 +1,84 @@
+{
+  "SelendroidDriver": {
+    "driverClass": "SelendroidDriver",
+    "packageName": "appium-selendroid-driver",
+    "automationName": "Selendroid",
+    "repository": "https://github.com/appium/appium-selendroid-driver"
+  },
+  "AndroidUiautomator2Driver": {
+    "driverClass": "AndroidUiautomator2Driver",
+    "packageName": "appium-uiautomator2-driver",
+    "automationName": "UiAutomator2",
+    "repository": "https://github.com/appium/appium-uiautomator2-driver"
+  },
+  "XCUITestDriver": {
+    "driverClass": "XCUITestDriver",
+    "packageName": "appium-xcuitest-driver",
+    "automationName": "XCUITest",
+    "repository": "https://github.com/appium/appium-xcuitest-driver"
+  },
+  "YouiEngineDriver": {
+    "driverClass": "YouiEngineDriver",
+    "packageName": "appium-youiengine-driver",
+    "automationName": "YouiEngine",
+    "repository": ""
+  },
+  "FakeDriver": {
+    "driverClass": "FakeDriver",
+    "packageName": "appium-fake-driver",
+    "automationName": "Fake",
+    "platformName": "Fake",
+    "repository": "https://github.com/appium/appium-fake-driver"
+  },
+  "AndroidDriver": {
+    "driverClass": "AndroidDriver",
+    "packageName": "appium-android-driver",
+    "platformName": "Android",
+    "repository": "https://github.com/appium/appium-android-driver",
+    "constraints": [{
+      "type": "capability",
+      "capability": "platformVersion",
+      "condition": ">=6.0.0",
+      "alternativeDriver": "AndroidDriver",
+      "message": "Consider setting 'automationName' capability to 'UiAutomator2' on Android >= 6, since UIAutomator framework is not maintained anymore by the OS vendor."
+    }]
+  },
+  "IosDriver": {
+    "driverClass": "IosDriver",
+    "packageName": "appium-ios-driver",
+    "platformName": "iOS",
+    "repository": "https://github.com/appium/appium-ios-driver",
+    "constraints": [{
+      "type": "capability",
+      "capability": "platformVersion",
+      "condition": ">=10.0.0",
+      "alternativeDriver": "XCUITestDriver",
+      "message": "Requested iOS support with version >= 10, using 'XCUITest' driver instead of UIAutomation-based driver, since the latter is unsupported on iOS 10 and up."
+    }]
+  },
+  "WindowsDriver": {
+    "driverClass": "WindowsDriver",
+    "packageName": "appium-windows-driver",
+    "platformName": "Windows",
+    "repository": ""
+  },
+  "MacDriver": {
+    "driverClass": "MacDriver",
+    "packageName": "appium-mac-driver",
+    "platformName": "Mac",
+    "repository": ""
+  },
+  "EspressoDriver": {
+    "driverClass": "EspressoDriver",
+    "packageName": "appium-espresso-driver",
+    "automationName": "Espresso",
+    "repository": "https://github.com/appium/appium-espresso-driver"
+  },
+  "TizenDriver": {
+    "driverClass": "TizenDriver",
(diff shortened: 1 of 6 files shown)


jlipps (commented):
I love this idea. I have not yet reviewed the code in detail, but what strikes me initially is that this a step on the way to Appium 2.0. I wonder if it would be easy enough to add a few more bits to this PR to basically get us what we need for "driver independence". IIRC, this is what else we'd need:

1. Ability to override location of drivers.json through a server flag (so somebody could start up appium with their own drivers.json, or so that Appium 2.0 could decide that this file belongs in `~/.appium-drivers.json` or similar
2. Ability to write new driver metadata into driver.json (so that in the future running `appium driver install foo` will add the driver foo
3. Ability to designate which drivers are installed vs just available to install. We could implement this a few different ways, for example having an 'installed' flag in drivers.json, or having essentially two files: `available-drivers.json` (which has metadata for the official appium drivers), and `drivers.json` which has metadata for only the installed drivers.

I understand these features would significantly increase the scope of this change, so I'm fine if we want to just review this as is and add the other stuff in a bit. But it struck me that this is already pretty close and at least wanted to make sure we have an eye to the future when reviewing the specifics of this proposal.

drivers.json
```diff
@@ -0,0 +1,84 @@
...
+  "AndroidDriver": {
+    "driverClass": "AndroidDriver",
+    "packageName": "appium-android-driver",
+    "platformName": "Android",
+    "repository": "https://github.com/appium/appium-android-driver",
+    "constraints": [{
+      "type": "capability",
+      "capability": "platformVersion",
+      "condition": ">=6.0.0",
+      "alternativeDriver": "AndroidDriver",
```
mykola-mokhnach: Perhaps `AndroidUiautomator2Driver` ?

mykola-mokhnach: Also, it should be possible to set multiple alernatives

drivers.json
```diff
@@ -0,0 +1,84 @@
...
+    "repository": "https://github.com/appium/appium-fake-driver"
+  },
+  "AndroidDriver": {
+    "driverClass": "AndroidDriver",
+    "packageName": "appium-android-driver",
+    "platformName": "Android",
+    "repository": "https://github.com/appium/appium-android-driver",
+    "constraints": [{
+      "type": "capability",
+      "capability": "platformVersion",
```
mykola-mokhnach: name?

drivers.json
```diff
@@ -0,0 +1,84 @@
...
+  },
+  "IosDriver": {
+    "driverClass": "IosDriver",
+    "packageName": "appium-ios-driver",
+    "platformName": "iOS",
+    "repository": "https://github.com/appium/appium-ios-driver",
+    "constraints": [{
+      "type": "capability",
+      "capability": "platformVersion",
+      "condition": ">=10.0.0",
```
mykola-mokhnach: should the switch to the alternative driver be enforced or just a warning message should be printed?

drivers.json
```diff
@@ -0,0 +1,84 @@
...
+    "driverClass": "XCUITestDriver",
+    "packageName": "appium-xcuitest-driver",
+    "automationName": "XCUITest",
+    "repository": "https://github.com/appium/appium-xcuitest-driver"
+  },
+  "YouiEngineDriver": {
+    "driverClass": "YouiEngineDriver",
+    "packageName": "appium-youiengine-driver",
+    "automationName": "YouiEngine",
+    "repository": ""
```
mykola-mokhnach: I'd rather just keep it unset

lib/driver-utils.js
```diff
@@ -0,0 +1,210 @@
...
+  } catch (err) {
+    log.errorAndThrow(`Unable to load driver '${driverClass}' from package '${packageName}': ${err.message}`);
+  }
+};
+
+function getDriverFromConstraints (caps, driverInfo) {
+  const getAlternativeDriver = function getAlternativeDriver (driverInfo) {
+    let alternativeDriver;
+    for (const constraint of (driverInfo.constraints || [])) {
+      if (constraint.type === 'capability') {
```
mykola-mokhnach: `switch` might be more readable

lib/driver-utils.js
```diff
@@ -0,0 +1,210 @@
...
+        if (constraint.capability === 'platformVersion') {
+          const platformVersion = semver.valid(semver.coerce(caps.platformVersion));
```
mykola-mokhnach: is it by design that `caps` variable is caught in the closure? Shall we use _.memoize instead?

lib/driver-utils.js
```diff
@@ -0,0 +1,210 @@
...
+              log.warn(constraint.message);
+            }
+            if (constraint.alternativeDriver) {
+              alternativeDriver = constraint.alternativeDriver;
+            }
+          }
+        }
+      }
+    }
+    return alternativeDriver;
```
mykola-mokhnach: I cannot understand this logic

lib/driver-utils.js
```diff
@@ -0,0 +1,210 @@
...
+    const modifiedPackageName = DRIVER_MAP[alternativeDriver].packageName;
+    if (modifiedDriverClass && modifiedPackageName) {
+      return [modifiedPackageName, modifiedDriverClass];
+    }
+  }
+  return [];
+}
+
+function getDriverFromAutomationName (caps) {
+  let packageName;
```
mykola-mokhnach: if (!_.isString(caps.automationName)) return [null, null];

lib/driver-utils.js
```diff
@@ -0,0 +1,210 @@
...
+      packageName = info.packageName;
+
+      // now check if there are constraints on this driver
+      const [modifiedPackageName, modifiedDriverClass] = getDriverFromConstraints(caps, info);
+      if (modifiedPackageName && modifiedDriverClass) {
+        packageName = modifiedPackageName;
+        driverClass = modifiedDriverClass;
+      }
+      // no need to go further
+      break;
```
mykola-mokhnach: we can return right here and get rid of intermediate variables

lib/driver-utils.js
```diff
@@ -0,0 +1,210 @@
...
+    packageName = info.packageName;
+
+    // now check if there are constraints on this driver
+    const [modifiedPackageName, modifiedDriverClass] = getDriverFromConstraints(caps, info);
+    if (modifiedPackageName && modifiedDriverClass) {
+      packageName = modifiedPackageName;
+      driverClass = modifiedDriverClass;
+    }
+    // no need to go further
+    break;
```
mykola-mokhnach: we could return right here

jlipps:
@imurchie let me know about my high-level comments and then I'm happy to do an in-depth review as well.

lib/driver-utils.js
farhanmannan: nice

mykola-mokhnach:
@imurchie Do you plan to finish this PR?

imurchie:
Yes. Just haven't had time.

PR #11201, opened 2018-08-17
Fix dependencies

imurchie:
## Proposed changes

Clean up dependencies a little bit:
1. Make `appium-base-driver` semver the same as the other drivers
2. Fix `appium-espresso-driver` range so we are getting the latest
3. Update `appium-tizen-driver`, whose version structure is such that it cannot be dealt with in the normal way

## Types of changes

## Checklist


package.json
```diff
@@ -55,7 +55,7 @@
     "continuation-local-storage": "3.x",
     "dateformat": "^3.0.3",
     "find-root": "^1.1.0",
-    "lodash": "=4.17.9",
+    "lodash": "^4.17.9",
```
mykola-mokhnach: what was the reason to fix this version?

imurchie: `wd` used to patch `lodash` with `underscore`'s string utilities, for reasons that made sense in 2014. If Appium had the same version of `lodash` then `_.truncate` would just return `...` no matter what the input. The "fix", before anyone could figure out what the actual problem was, was to pin `lodash` (when it was pinned, it seemed like probably a `lodash` issue, with 4.17.10). 

So, `wd` is fixed now, so this doesn't need to be pinned!

mykola-mokhnach (approved):

KazuCocoa (approved):
👍

Issue #11159, opened 2018-08-08
The revision field is always null in the /status output


mykola-mokhnach:
## The problem

When Selenium Grid requests /status for Appium node it always returns null as 'revision' field value 

## Environment

* Appium version (or git revision) that exhibits the issue: 1.8.2-beta
* Last Appium version that did not exhibit the issue (if applicable):
* Desktop OS/version used to run Appium:
* Node.js version (unless using Appium.app|exe): 8
* Mobile platform/version under test:
* Real device or emulator/simulator:
* Appium CLI or Appium.app|exe: CLI

## Details

The problem under the hood is that `getGitRev` method expects that git meta information is present for the package. Although it is not true for packages installed from npm, since such meta info is stripped out upon publishing. So the command `git rev-parse HEAD` always fails with non-zero code.

imurchie:
Yes. Release versions have never had a git revision.

mykola-mokhnach:
Closed as fixed

PR #11232, opened 2018-08-22
Return more appropriate info as server status and speed it up

mykola-mokhnach:
## Proposed changes

/status endpoint is the one which is called very often, so it must be as performant as possible. Although the problem described in https://github.com/appium/appium/issues/11159 is still relevant and should be addressed separately.

## Types of changes

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -5,7 +5,6 @@ import request from 'request-promise';
 import { exec } from 'teen_process';
 import { rootDir } from './utils';
 import logger from './logger';
-import B from 'bluebird';
 
 const APPIUM_VER = require(path.resolve(rootDir, 'package.json')).version;
 const GIT_META_ROOT = '.git';
@@ -22,34 +21,32 @@ function getNodeVersion () {
   return [Number(version[1]), Number(version[2])];
 }
 
-function updateBuildInfo () {
-  return new B((resolve) => {
-    getGitRev().then((sha) => { // eslint-disable-line promise/prefer-await-to-then
-      if (_.isEmpty(sha)) {
-        logger.warn(`Cannot retrieve git revision for Appium version ${APPIUM_VER}`);
-        return resolve();
-      }
+async function updateBuildInfo () {
+  let sha;
+  try {
+    sha = await getGitRev();
+    if (_.isEmpty(sha)) {
+      logger.warn(`Cannot retrieve git revision for Appium version ${APPIUM_VER}`);
+      return;
+    }
+  } catch (err) {
+    logger.warn(`Cannot retrieve git revision for Appium version ${APPIUM_VER}. ` +
+      `Original error: ${err.message}`);
+    return;
+  }
 
-      BUILD_INFO['git-sha'] = sha;
-      getGitTimestamp(sha)
-        .then((built) => { // eslint-disable-line promise/prefer-await-to-then
-          if (_.isEmpty(built)) {
-            logger.warn(`Cannot retrieve the timestamp for Appium git commit ${sha}`);
-          } else {
-            BUILD_INFO.built = built;
-          }
-          resolve();
-        }).catch((e) => {
-          logger.warn(`Cannot retrieve the timestamp for Appium git commit ${sha}. ` +
-            `Original error: ${e.message}`);
-          resolve();
-        });
-    }).catch((e) => {
-      logger.warn(`Cannot retrieve git revision for Appium version ${APPIUM_VER}. ` +
-        `Original error: ${e.message}`);
-      resolve();
-    });
-  });
+  BUILD_INFO['git-sha'] = sha;
+  try {
+    const built = await getGitTimestamp(sha);
+    if (_.isEmpty(built)) {
+      logger.warn(`Cannot retrieve the timestamp for Appium git commit ${sha}`);
+    } else {
+      BUILD_INFO.built = built;
+    }
+  } catch (err) {
+    logger.warn(`Cannot retrieve the timestamp for Appium git commit ${sha}. ` +
+      `Original error: ${err.message}`);
+  }
 }
 
 async function getGitRev () {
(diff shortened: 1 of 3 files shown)


lib/config.js
```diff
@@ -15,22 +15,37 @@ function getNodeVersion () {
 }
 
 async function getGitRev () {
-  let rev = null;
+  // FIXME: https://github.com/appium/appium/issues/11159
```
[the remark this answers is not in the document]
imurchie: Not really. Saving to a file would mean that the file would need to be checked in which would change the git SHA.

But it is a non-problem. For every version that is published on npm there is a record of exactly where in the code it was published from, in the form of a git tag. The version is known, and the tag can be inferred from the version.

The git SHA is only necessary when it is being run from source, in which case it is either running in a git context or someone is going out of their way to make their life as a debugger difficult.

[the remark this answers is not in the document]
mykola-mokhnach: > But it is a non-problem. For every version that is published on npm there is a record of exactly where in the code it was published from, in the form of a git tag. The version is known, and the tag can be inferred from the version.

Good idea. I can make it

mykola-mokhnach:
@imurchie how about now?

lib/config.js
```diff
@@ -14,24 +22,93 @@ function getNodeVersion () {
   return [Number(version[1]), Number(version[2])];
 }
 
+function updateBuildInfo () {
+  return new B((resolve) => {
+    getGitRev().then((sha) => { // eslint-disable-line promise/prefer-await-to-then
```
imurchie: Why all the old-style promise handling?
```js
try {
  const sha = await getGitRev();
  if (_.isEmpty(sha)) {
    logger.warn(`Cannot retrieve git revision for Appium version ${APPIUM_VER}`);
    return;
  }
  BUILD_INFO['git-sha'] = sha;

  try {
    const built = await getGitTimestamp(sha)
    if (_.isEmpty(built)) {
      logger.warn(`Cannot retrieve the timestamp for Appium git commit ${sha}`);
    } else {
      BUILD_INFO.built = built;
    }
  } catch (err) {
    logger.warn(`Cannot retrieve the timestamp for Appium git commit ${sha}. ` +
      `Original error: ${err.message}`);
  }
} catch (err) {
  logger.warn(`Cannot retrieve git revision for Appium version ${APPIUM_VER}. ` +
    `Original error: ${err.message}`);
}
```

lib/config.js
```diff
@@ -14,24 +22,93 @@ function getNodeVersion () {
...
-    rev = stdout.trim();
-  } catch (ign) {}
-  return rev;
+  if (await fs.exists(path.resolve(rootDir, GIT_META_ROOT))) {
+    const {stdout} = await exec(GIT_BINARY, ['rev-parse', 'HEAD'], {
+      cwd: rootDir
+    });
+    return stdout.trim();
+  } else {
+    const response = await request.get(`${GITHUB_API}/tags`, {
```
imurchie: Do we really want to start introducing the overhead of network calls just for the sake of getting the git SHA?  This means for each time the server starts it needs to make two network calls just to print out some information?

imurchie (commented):
I'm not convinced we want to start requiring network calls to github just to start the server.

lib/config.js
mykola-mokhnach: yep, I can change that

lib/config.js
mykola-mokhnach: Is it a problem? the calls are not blocking so we don't really care if and when they succeed. Although, this might give us (and clients) more useful information

imurchie: `await request.get` should be blocking.

For any released version of Appium, the exact SHA, as well as the tree at that point, can be retrieved through https://github.com/appium/appium/releases, or just by going to `https://github.com/appium/appium/tree/v<WHATEVER-VERSION>`.

mykola-mokhnach: Is there any reason why it should be blocking? The idea is to never block `getStatus` responses and only include the retrieved information there when it is available for us.

> For any released version of Appium, the exact SHA, as well as the tree at that point, can be retrieved through https://github.com/appium/appium/releases, or just by going to https://github.com/appium/appium/tree/v<WHATEVER-VERSION>.

I do use requests to official API endpoints, which return the information as nicely formatted json.

imurchie: I just mean it can easily be retrieved manually. If you know the SHA you have about as many steps to get to the tree for that commit as if you just know the npm version.

For the blocking part, `await` ensures that. You could do something like
```js
request.get(`${GITHUB_API}/tags`, {
  headers: {
    'User-Agent': `Appium ${APPIUM_VER}`
  }
}).then(function (res) {
  const resBodyObj = util.safeJsonParse(res);
  if (_.isArray(resBodyObj)) {
    for (const {name, commit} of resBodyObj) {
      if (name === `v${APPIUM_VER}` && commit && commit.sha) {
        BUILD_INFO['git-sha'] = commit.sha;
        break;
      }
    }
  }
});
```
and not care that when the first call is finished this will not have anything in it.

mykola-mokhnach: ok I got your point

lib/config.js
```diff
@@ -14,24 +21,93 @@ function getNodeVersion () {
...
+    return;
+  }
+
+  BUILD_INFO['git-sha'] = sha;
   try {
-    let {stdout} = await exec("git", ["rev-parse", "HEAD"], {rootDir});
-    rev = stdout.trim();
-  } catch (ign) {}
-  return rev;
+    const built = await getGitTimestamp(sha);
```
jlipps: this is still going to block on a network request, isn't it?

we probably also want to `catch` these requests and ignore failures so appium can run without internet

mykola-mokhnach: it should not. These methods are called from `updateBuildInfo`, which is not awaited

mykola-mokhnach:
@imurchie I have improved the logic there, so server startup should not be delayed now

imurchie (approved):

Issue #11300, opened 2018-09-02
Appium 'new session' response is not W3C compatible. This switches selenium-java client to old protocol and cause false-positive tests


kool79:
## The problem

Because of appium does not strictly conform w3c specification the Selenium WebDriver java client automatically  detects protocol as JsonWire. It cause wrong processing of w3c error responses which can lead to false-positive tests.

***This bug blocks appium usage with official selenium-java client*** 

## Environment

* Appium version (or git revision) that exhibits the issue: 1.9
* Selenium WebDriver client v3.14 (java)

## Details
  Appium sends response to `POST /session` command in the way that likes to old JsonWire protocol. (there is 'status' field) Selenium java client library use this response to detect protocol version used by client side  (w3c or json-wire).  Because selenium library switched to old protocol it incorrectly process w3c-compliant error responces from appium. Actually it does not terminate test execution in case of some errors which can cause false-positive tests.
  Details about protocol version detection you can find in [JsonWireProtocolResponse.java](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/JsonWireProtocolResponse.java) and [W3CHandshakeResponse.java](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/W3CHandshakeResponse.java). In short for w3c:
* `status` field must be absent, 
* `value` should be the json-object (ok with that)
* for `POST /session` the field `value` must include 'sessionId' field and 'capabilities' json object  

## Below are examples for current (appium v1.9) and expected responses:
Current responce to POST /session (truncated)
```
{
	"status":0,
	"value":{
		"webStorageEnabled":false,
		"locationContextEnabled":false,
		"newCommandTimeout":15,
		"resetOnSessionStartOnly":false,
		"deviceName":"iPhone 7",
		"platformName":"IOS",
		"automationName":"XCUITest",
		"udid":"A749AAB5-5BFE-4009-951F-BCF54DAC8754"
		.......
	},
	"sessionId":"7f631bee-3afa-421c-b083-75fa75bb07fa"
}
```
W3C responce (expected). Note: both `capabilities` and `sessionId` must belong to `value`
```
{
   "value": {
      "sessionId":"7f631bee-3afa-421c-b083-75fa75bb07fa",
      "capabilities":{
         "webStorageEnabled":false,
         "locationContextEnabled":false,
         "newCommandTimeout":15,
         "resetOnSessionStartOnly":false,
[...]

PR #11303, opened 2018-09-03
Write defaultCapabilities into firstMatch instead of alwaysMatch only if it is not overriden

mykola-mokhnach:
Partially addresses protocol setting issue in https://github.com/appium/appium/issues/11300#issuecomment-418061325

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -74,13 +74,15 @@ function parseCapsForInnerDriver (jsonwpCapabilities, w3cCapabilities, constrain
           }
         }
         isCapAlreadySet = isCapAlreadySet || _.has(removeW3CPrefixes(alwaysMatch), removeW3CPrefix(defaultCapKey));
-        if (!isCapAlreadySet) {
-          // Only add the default capability if it is not overridden
-          if (_.isEmpty(firstMatch)) {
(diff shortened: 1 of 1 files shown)


dpgraham (approved):

lib/utils.js
```diff
@@ -61,22 +61,38 @@ function parseCapsForInnerDriver (jsonwpCapabilities, w3cCapabilities, constrain
...
+          }
+        }
+        isCapAlreadySet = isCapAlreadySet || _.has(removeW3CPrefixes(alwaysMatch), removeW3CPrefix(defaultCapKey));
+        if (isCapAlreadySet) {
+          continue;
+        }
+
+        // Only add the default capability if it is not overridden
+        if (_.isEmpty(firstMatch)) {
+          w3cCapabilities.firstMatch = [{defaultCapKey: defaultCapValue}];
```
jlipps: doesn't this need to be `[{[defaultCapKey]: defaultCapValue}]`?

jlipps (commented):
i know this was merged, but may have found a bug

lib/utils.js
mykola-mokhnach: good catch, will fix

PR #11424, opened 2018-09-25
Fix typo in parallel testing docs

dpgraham:


mykola-mokhnach (approved):

docs/en/advanced-concepts/parallel-tests.md
```diff
@@ -6,7 +6,7 @@ The important capabilities:
 
 - `udid` the device id
 - `chromeDriverPort` the chromedriver port (if using webviews or chrome)
-- `systemPort` If you are using [appium-uiautomator2-driver](https://github.com/appium/appium-uiautomator2-driver), set a different system port for each Appium instanceset with `systemPort` capability since sometimes there can be a port conflict if different ports aren't used, such as in [this issue](https://github.com/appium/appium/issues/7745).
+- `systemPort` If you are using [appium-uiautomator2-driver](https://github.com/appium/appium-uiautomator2-driver), set a different system port for each Appium instance, set with `systemPort` capability since sometimes there can be a port conflict if different ports aren't used, such as in [this issue](https://github.com/appium/appium/issues/7745).
```
saikrishna321: @dpgraham instead of 'set a different system port for each Appium instance' can it be 'set a different system port for each android driver instance'

jlipps: maybe use the word "session" instead of "driver instance", since that is what users know it by? users just know they are creating appium sessions.

PR #11432, opened 2018-09-26
improve documentation

vikingden8:
driver.current_activity() and driver.currrent_package() should be driver.current_activity and driver.current_package because of using property feature in Python.


jlipps (approved):

jlipps:
right you are!

Issue #11504, opened 2018-10-09
add flag to avoid getting git rev


kazzmir:
## The problem

Appium sends a request to api.github.com to get the git rev, which ends up in the `config.BUILD_INFO` variable. These requests cause issues in my environment such as:
  * timeout errors from certain networks that cannot reach the github ip
  * api rate limiting enforced by github

As far as I can tell the relevant call graph is

`AppiumDriver::constructor -> config.updateBuildInfo() -> config.getGitRev()`

Could a flag be added to disable the github api check?

## Environment

* Appium version (or git revision) that exhibits the issue: 1.9.1
* Last Appium version that did not exhibit the issue (if applicable):
* Desktop OS/version used to run Appium: linux
* Node.js version (unless using Appium.app|exe):
* Mobile platform/version under test:
* Real device or emulator/simulator:
* Appium CLI or Appium.app|exe:

imurchie:
I personally would prefer all that code to be pulled out entirely. The revision adds no information on released versions of Appium, if the release was performed correctly. Appium run from source needs the revision sha in order to identify the code the server is running.

mykola-mokhnach:
The fix has been merged and will be published in appium@beta.3

PR #11506, opened 2018-10-09
Do not use Github API fallback by default

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/11504

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -21,19 +21,19 @@ function getNodeVersion () {
   return [Number(version[1]), Number(version[2])];
 }
 
-async function updateBuildInfo () {
-  const sha = await getGitRev(true);
+async function updateBuildInfo (useGithubApiFallback = false) {
+  const sha = await getGitRev(useGithubApiFallback);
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -126,7 +126,7 @@ function warnNodeDeprecations () {
 }
 
 async function showConfig () {
-  await updateBuildInfo();
+  await updateBuildInfo(true);
```
imurchie: This seems like the least important place to not try to check github, since this is only called when you run `appium --show-config`. The other place `updateBuildInfo` is called still will trigger checking github.

mykola-mokhnach: in appium.js `updateBuildInfo` is called without arguments, which sets `useGithubApiFallback` to the default value `false`

imurchie: Ignore this. I just realized the direction of the condition.

imurchie (approved):

jlipps (commented):
👍

PR #11640, opened 2018-11-06
Instruct animation disabling for Espresso

dpgraham:


KazuCocoa:
in espresso driver, below code exists. Is it good to address Espresso driver disables animation by default? (And it turns it in a delete session section.)

```js
    if (await this.adb.isAnimationOn()) {
      try {
        await this.adb.setAnimationState(false);
        this.wasAnimationEnabled = true;
      } catch (err) {
        logger.warn(`Unable to turn off animations: ${err.message}`);
      }
    }
```

dpgraham:
^ @saikrishna321 when you tested on the old Android version, did this message come up `Unable to turn off animations`.

KazuCocoa:
ah... I see..
I'm not sure where `Unable to turn off animations` comes since https://github.com/appium/io.appium.settings/blob/master/app/src/main/java/io/appium/settings/handlers/AnimationSettingHandler.java has no such message. But if it happens in particular environment, the message needs.

In my case, I've turned animation off using the same logic for Android API 17 ~ the latest one. They worked. Thus, I'm curious about which API level (or device) doesn't work the animation scale.

KazuCocoa:
ah... is the message in https://github.com/appium/appium-espresso-driver/blob/b11e90117cde4c709916abebe4712b29b8af99c0/lib/driver.js#L223 ?
I'm curious about the original message.

saikrishna321:
@dpgraham sorry i didnt notice that. 

@KazuCocoa 
> I've turned animation off using the same logic for Android API 17 ~ the latest one

How did u do this ?

KazuCocoa:
Mine is https://github.com/KazuCocoa/DroidTestHelper/blob/d5544158cbd01252f1dd99939d277f28a37eb7af/droidtesthelperlib/src/main/java/com/kazucocoa/droidtesthelperlib/HandleAnimations.kt . (Java https://github.com/KazuCocoa/DroidTestHelper/commit/6e9a674c2f227193c2b3a1997e01b10f6d4c29c1#diff-ba169d04c3574a74102c956df6421b23 )

The code is very similar to https://github.com/appium/io.appium.settings/blob/master/app/src/main/java/io/appium/settings/handlers/AnimationSettingHandler.java which has introduced in Appium

So, I'm curious about what error you can get as the original error you got.

KazuCocoa:
for after editing comment: @saikrishna321 
The disable animation logic is called in https://github.com/appium/appium-android-driver/blob/master/lib/driver.js#L259
Adb command sends a broadcast to io.appium.settings. It handles the animation scale.

KazuCocoa:
> So is the animation disabling not working for older emulators?
To be honest, I'm not sure since the handling animation logic doesn't change long time... But I can imagine some android emulators have such issue since Android API 27(8?)+ emulator has an issue which can't grant external storage permission in a particular situation, for example.

I'd second for adding this into _ troubleshooting_ 👍

KazuCocoa:
I investigated a bit in https://github.com/appium/appium-uiautomator2-server/issues/222

In my environment, Android 4.4, 5.0, 6.0 emulators raise an error, `E/AbstractSettingHandler: The permission android.permission.SET_ANIMATION_SCALE is not set. Cannot change state of animation.` by default.
After run `$ adb shell pm grant io.appium.settings android.permission.SET_ANIMATION_SCALE` by manual, disable animation worked well.

On Android 7.1.1 emulator, animation duration wasn't disabled.
image

On Android 8.1, everything worked.

So, I think we must call `adb shell pm grant io.appium.settings android.permission.SET_ANIMATION_SCALE` under Android 8 when Appium calls disableAnimation.

KazuCocoa:
^ I set `disableWindowAnimation: true,` in cap.

KazuCocoa:
In my comment https://github.com/appium/appium/pull/11640#issuecomment-436469370 , I called the grant permission every time in my own logic. So that worked.

KazuCocoa:
fyi @dpgraham @saikrishna321 
After https://github.com/appium/appium-android-driver/pull/465 , disabling animations via io.appium.settings app should work for API level 22- as well as API level 23+ devices

dpgraham:
I changed the wording to specify that it's _supposed_ to disable animation, but if it doesn't, then do it manually.

KazuCocoa (approved):
thanks! it looks reasonable

PR #11647, opened 2018-11-07
Recommend install FBSimulatorControl

dpgraham:


docs/en/drivers/ios-xcuitest.md
```diff
@@ -99,6 +99,13 @@ following desired capabilities:
 * `udid` - the specific id of the device to test on. This can also be set to
    `auto` if there is only a single device, in which case Appium will determine
    the device id and use it.
+   
+### Optional Setup
+
+Although not strictly required, we recommend installing FBSimulatorControl for better
+results.
+
+`brew install FBSimulatorControl`
```
mykola-mokhnach: not really:

```
# Get the Facebook Tap.
brew tap facebook/fb
# Install fbsimctl from master
brew install fbsimctl --HEAD
```

https://github.com/facebook/FBSimulatorControl/blob/master/fbsimctl/README.md

dpgraham: Yeah, I didn't read the documentation :p

mykola-mokhnach (approved):

docs/en/drivers/ios-xcuitest.md
```diff
@@ -99,6 +99,18 @@ following desired capabilities:
 * `udid` - the specific id of the device to test on. This can also be set to
    `auto` if there is only a single device, in which case Appium will determine
    the device id and use it.
+   
+### Optional Setup
+
+Although not strictly required, we recommend installing FBSimulatorControl for better
+results.
```
jlipps: I know this is merged, but I think it would have been nice to say something a little more specific than simply "better results". Something like "more reliable biometric verification, etc..." so that users know what's at stake in the choice, rather than simply that it's somehow "better"

PR #11731, opened 2018-11-23
Use CI to update documentation

dpgraham:
* Update generate-docs command to make a pull request with documentation changes
* Make a job that, when it's a tag, checks out appium.io, builds the docs and then makes a pull request


dpgraham:
@jlipps or @imurchie would one of you have a  chance to review this?

This will help keep appium.io up-to-date more frequently.

.travis.yml
```diff
@@ -24,11 +24,35 @@ jobs:
...
+        - npm install
+        - npm install -g uncommitted
+
+      script:
+        # Generate the docs
+        - npm run generate-docs
+
+        # If no docs were generated, don't do anything
+        - |
+          if [ !uncommitted ]; then
```
imurchie: Should just be able to check `git diff-index --quiet HEAD --` without needing the npm package.

.travis.yml
```diff
@@ -24,11 +24,35 @@ jobs:
       script: npm run test && npm run e2e-test
       node_js: "8"
     - stage:
```
imurchie: Can this stage be named?

.travis.yml
```diff
@@ -55,3 +79,24 @@ jobs:
         # Remove the dev depenedencies
         - npm prune
       script: npm run zip-and-upload
+    - stage: Update Documentation
+      if: tag
```
imurchie: It's unclear to me what this condition is doing.

imurchie (commented):
Sorry for missing this one. 

The approach seems great! Just a few clarifying questions. Travis config is never clear :(

.travis.yml
dpgraham: It evaluates if `TRAVIS_TAG=true` which means that it's a tag commit.

.travis.yml
dpgraham: When `stage` is empty, it means that it's in the same stage as the previously named `stage` and gets run concurrently with it.

imurchie: Oh yeah, I knew that.

.travis.yml
imurchie: So if we ever tag a commit for some other reason, it will cause this to be triggered? I'm fine with that, just curious.

dpgraham: @imurchie That's correct yes. If we're ever not ready to merge the docs yet, then we can just close the PR that's created by the job.

PR #11829, opened 2018-12-12
docs: Add more do not edit warnings

dpgraham:
We need to make it more obvious that these files shouldn't be edited because sometimes folks will edit the markdown. Added more invisible MD comments that instruct not to edit throughout the doc instead of at the header.


KazuCocoa (approved):
👍

SrinivasanTarget (approved):

PR #11882, opened 2018-12-20
Bug Clarification for TouchActions

rtmalone:
That this bug deals with **device** simulators needs to be clarified. I kept overlooking this section because I was running iOS **12** on an iPhone 8 simulator thinking I was the OS I needed to be concerned about. It's possible this may not be warranted, but I spent a lot of time overlooking this issue because of the original wording.

## Proposed changes

Clarify in the doc that the bug pertains to a device range and not the OS

## Types of changes

- [ X] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [x ] I have signed the [CLA](https://cla.js.foundation/appium/appium)

### DIFF
--- a/docs/en/writing-running-appium/touch-actions.md
+++ b/docs/en/writing-running-appium/touch-actions.md
@@ -85,7 +85,7 @@ MultiAction().add(action0).add(action1).perform()
 
 ### Bugs and Workarounds
 
-An unfortunate bug exists in the iOS 7.0 - 8.x Simulators where ScrollViews,
+An unfortunate bug exists in iOS Device 7.0 - 8.x Simulators where ScrollViews,
 CollectionViews, and TableViews don't recognize gestures initiated by
 UIAutomation (which Appium uses under the hood for iOS). To work around this,
 we have provided access to a different function, `scroll`, which in many cases


docs/en/writing-running-appium/touch-actions.md
```diff
@@ -85,7 +85,7 @@ MultiAction().add(action0).add(action1).perform()
 
 ### Bugs and Workarounds
 
-An unfortunate bug exists in the iOS 7.0 - 8.x Simulators where ScrollViews,
+An unfortunate bug exists in iOS Device 7.0 - 8.x Simulators where ScrollViews,
```
dpgraham: Maybe something like `bug exists in versions 7.0 to 8.x of iOS (Simulators only)`

rtmalone: I think because the wording didn't specifically point out that devices where the culprit is where I failed in the reading. I kept reading iOS and kept thinking it was a bug in the OS on simulators; not the *simulated device*. Seems silly now because that would be a really old OS version and the format of the number could infer it is a device we are talking about. But yet, I still didn't infer that in the moment which still supports my point. How about `bug exists in versions 7.0 to 8.x of iOS Devices (Simulator only)`?

rtmalone: > devices is a bit confusing too because device usually implies a real device

yeah....I guess I could agree with that. Probably something close to your original suggestion is the best that can be hoped for: `bug exists in versions 7.0 to 8.x of iOS Simulators`

dpgraham: `bug exists in iOS Simulators running on Xcode versions 7.0 to 8.x`

imurchie:
I agree that "device" usually connotes "_real_ device", so that could be confusing.

How about something like 
```md
An unfortunate bug exists in Simulators running iOS 7.0 - 8.x, where ScrollViews...
```

rtmalone:
This is getting further from clarity. The original wording is fine and I'm willing to admit it was just a personal misreading of the text and not something that can cause widespread misunderstanding.

PR #11947, opened 2019-01-05
Update system-time.yml

ki4070ma:
Removed "()" from device_time(python) since device_time has property decorator.
device_time() throws TypeError.

https://github.com/appium/python-client/blob/master/appium/webdriver/webdriver.py


KazuCocoa (approved):
Yes, thanks!

Issue #11967, opened 2019-01-09
1.11.0 release candidate, old versions in npm-shrinkwrap.json


john7doe:
## The problem

Looking at https://github.com/appium/appium/blob/1.11/npm-shrinkwrap.json it looks like it locks the version of the appium-* packages to some really old ones.

e.g. appium-adb is locked to version 6.19.0, when looking at the appium-adb repo I would have expected a ^7.0 version. The 1.10 release used 6.28.0 of the package.

Another example is appium-espresso-driver that is locked to 1.3 and the latest version released is 1.6.

imurchie:
Yes, something odd seems to have gone on here. On the bright side, the package is also not set up to be distributed with the `npm-shrinkwrap.json` file included, so you'll get the most recent dependencies anyway. :|

@dpgraham Can you have a look at this?

john7doe:
I was assuming that you were shipping with a `npm-shrinkwrap.json` file included?!?! 

If I understand npm correctly then, with the quite lax versions in `package.json`, there is nothing thats will  ensures that yesterdays `npm install appium@1.10` is the same as todays `npm install appium@1.10` ?

dpgraham:
I can look at that. I think I probably didn't clear the node modules and reinstall before ahronkwrapping.

john7doe:
1.9.0 did, it does not look like 1.10 or the latest rc did:
```
~/tmp $ curl -s https://registry.npmjs.org/appium/-/appium-1.11.0-rc.0.tgz | tar tv|grep npm-shrinkwrap.json
~/tmp $ curl -s https://registry.npmjs.org/appium/-/appium-1.10.0.tgz | tar tv |grep npm-shrinkwrap.json
~/tmp $ curl -s https://registry.npmjs.org/appium/-/appium-1.9.0.tgz | tar tv |grep npm-shrinkwrap.json
-rw-r--r--  0 0      0      453599 Oct 26  1985 package/npm-shrinkwrap.json
```

dpgraham:
@john7doe you're right, I never noticed.

@imurchie looks like `npm-shrinkwrap.json` is being ignored (https://github.com/appium/appium/pull/11461).

dpgraham:
I find it perpelixing that it's even _possible_ to ignore `npm-shrinkwrap.json`. It should be published and included by default.

dpgraham:
Also, I published a new rc (with shrinkwrap) and it has not outdated dependencies.

john7doe:
[...]

PR #11969, opened 2019-01-09
Shrinkwrap fixes

dpgraham:
1. Update `shrinkwrap-prod` script to prevent stale dependencies from getting into shrinkwrap 

How it works now:

* Delete `package-lock.json` if one exists (this is defensive, because if `package-lock.json` exists and `npm shrinkwrap` is called, it just renames `package-lock.json` to `npm-shrinkwrap.json`)
* Delete `node_modules`
* Install production modules (`npm instal --production`)
* Create shrinkwrap (`npm shrinkwrap`)
* Install dev dependencies, but don't update shrinkwrap (`npm install --no-shrinkwrap`)

2. Add `npm-shrinkwrap.json` to package.json files. NPM doesn't pack it by default, surprisingly
3. Add a script that does an `npm pack` dry-run and tests that `npm-shrinkwrap.json` was included

(related to https://github.com/appium/appium/issues/11967)

### DIFF
--- a/check-npm-pack-files.js
+++ b/check-npm-pack-files.js
@@ -25,4 +25,4 @@ if (!_.isEmpty(missingFiles)) {
   throw new Error(`Files '${missingFiles.join(', ')}' are not included in package.json "files". Please make sure these files are included before publishing.`);
 }
 
-process.exit(1);
+process.exit(0);
--- a/check-npm-pack-files.js
+++ b/check-npm-pack-files.js
@@ -3,13 +3,7 @@ const _ = require('lodash');
 
 const res = JSON.parse(childProcess.execSync('npm pack --dry-run --json --ignore-scripts', {encoding: 'utf8'}))[0];
 
-// Get list of the files being packed
-const fileNames = [];
-for (const file of res.files) {
-  fileNames.push(file.path);
-}
-
-// Get list of files we want to test for existence
+// List of files we are testing to make sure they are included in package
 const testFiles = [
   'npm-shrinkwrap.json', // Check that npm-shrinkwrap.json is being packed
   'LICENSE', // Check that license is included
@@ -17,7 +11,7 @@ const testFiles = [
 ];
 
 // Get list of files in `testFiles` that aren't in the list of packaged fileNames
-const missingFiles = _.without(testFiles, ...fileNames);
+const missingFiles = _.without(testFiles, ..._.map(res.files, 'path'));
 
 if (!_.isEmpty(missingFiles)) {
   throw new Error(`Files [${missingFiles.join(', ')}] are not included in package.json "files". ` +
--- a/check-npm-pack-files.js
+++ b/check-npm-pack-files.js
@@ -1,0 +1,28 @@
+const childProcess = require('child_process');
+const _ = require('lodash');
+
+const res = JSON.parse(childProcess.execSync('npm pack --dry-run --json --ignore-scripts', {encoding: 'utf8'}))[0];
+
+const fileNamesMap = {};
+for (const file of res.files) {
+  fileNamesMap[file.path] = true;
+}
+
+const testFiles = [
+  'npm-shrinkwrap.json', // Check that npm-shrinkwrap.json is being packed
+  'LICENSE', // Check that license is included
+  'build/lib/appium.js', // Sanity check that build files are being included by testing just one file
+];
+
+const missingFiles = [];
+for (const testFile of testFiles) {
+  if (!fileNamesMap[testFile]) {
+    missingFiles.push(testFile);
+  }
+}
+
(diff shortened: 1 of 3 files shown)


jlipps (approved):
yep, good catch.

jlipps:
oh yeah, weird that it can be ignored!

check-npm-pack-files.js
```diff
@@ -0,0 +1,28 @@
...
+  if (!fileNamesMap[testFile]) {
+    missingFiles.push(testFile);
+  }
+}
+
+if (!_.isEmpty(missingFiles)) {
+  throw new Error(`Files '${missingFiles.join(', ')}' are not included in package.json "files". Please make sure these files are included before publishing.`);
+}
+
+process.exit(1);
```
imurchie: Won't this indicate it is failing? And the npm command fail?

package.json
```diff
@@ -85,7 +86,7 @@
     "lint:fix": "gulp eslint --fix",
     "coverage": "gulp coveralls",
     "generate-docs": "node ./build/commands-yml/parse.js",
-    "shrinkwrap-prod": "rimraf package-lock.json && npm prune --production && npm shrinkwrap && npm install --no-shrinkwrap",
+    "shrinkwrap-prod": "rimraf package-lock.json && rimraf node_modules && npm install --production && npm shrinkwrap && node ./check-npm-pack-files && npm install --no-shrinkwrap",
```
imurchie: I'd personally prefer `shrinkwrap:prod`.

check-npm-pack-files.js
dpgraham: Yes. My bad. I always mix those up.

package.json
dpgraham: Yeah, I like that syntax too. Change was pushed.

check-npm-pack-files.js
```diff
@@ -0,0 +1,28 @@
+const childProcess = require('child_process');
+const _ = require('lodash');
+
+const res = JSON.parse(childProcess.execSync('npm pack --dry-run --json --ignore-scripts', {encoding: 'utf8'}))[0];
+
+const fileNamesMap = {};
```
mykola-mokhnach: I'd rather use res.files.map

check-npm-pack-files.js
```diff
@@ -0,0 +1,28 @@
...
+  fileNamesMap[file.path] = true;
+}
+
+const testFiles = [
+  'npm-shrinkwrap.json', // Check that npm-shrinkwrap.json is being packed
+  'LICENSE', // Check that license is included
+  'build/lib/appium.js', // Sanity check that build files are being included by testing just one file
+];
+
+const missingFiles = [];
```
mykola-mokhnach: filter

check-npm-pack-files.js
```diff
@@ -0,0 +1,28 @@
...
+for (const testFile of testFiles) {
+  if (!fileNamesMap[testFile]) {
+    missingFiles.push(testFile);
+  }
+}
+
+if (!_.isEmpty(missingFiles)) {
+  throw new Error(`Files '${missingFiles.join(', ')}' are not included in package.json "files". Please make sure these files are included before publishing.`);
```
mykola-mokhnach: This string seems to be quite long

dpgraham:
Changes pushed.

mykola-mokhnach (approved):

package.json
```diff
@@ -85,7 +86,7 @@
     "lint:fix": "gulp eslint --fix",
     "coverage": "gulp coveralls",
     "generate-docs": "node ./build/commands-yml/parse.js",
-    "shrinkwrap-prod": "rimraf package-lock.json && npm prune --production && npm shrinkwrap && npm install --no-shrinkwrap",
+    "shrinkwrap:prod": "rimraf package-lock.json && rimraf node_modules && npm install --production && npm shrinkwrap && node ./check-npm-pack-files && npm install --no-shrinkwrap",
```
john7doe: I understand the `npm install --no-shrinkwrap` part from a developer convenience standpoint. But the script does more that the name lead one to believe. 

Say someone wants to package up a production zip and think they can use `shrinkwrap:prod` then they will be surprised to by the last step.

imurchie (approved):

package.json
dpgraham: @john7doe if somebody was to make a build from the git repo, you wouldn't need to concern yourself with creating a shrinkwrap file. The shrinkwrap file tells NPM what to lock the dependencies to install when Appium is installed globally or as an externally dependency. When `npm install` is run from within the repo, the shrinkwrap is ignored. So the best practice for creating a bundle is to simply run `npm install --production` to get production-only dependencies.

PR #11979, opened 2019-01-10
Add deprecation notice when automationName not set

dpgraham:


jlipps (approved):
Looks good, I thought we had a whole deprecation engine for caps to get rid of boilerplate, but maybe that was axed with 1.0

lib/appium.js
```diff
@@ -173,6 +173,10 @@ class AppiumDriver extends BaseDriver {
     if (!_.isString(caps.platformName)) {
       throw new Error('You must include a platformName capability');
     }
+    if (_.isUndefined(caps.automationName)) {
```
mykola-mokhnach: I'd rather add the warning below in case `return` condition didn't match anything. We also would like to say something about incorrect values there

imurchie: I agree. This makes it seem like there is an alternative for some platforms, like if I wanted iOS 7 using Instruments I could supply _some_ `automationName` to get it, but that is not correct.

I would be happy to have an `automationName` for every platform, though, and then prefer it.

dpgraham: That's a good point. I'll give legacy IOS and Android drivers automation names....

What do you think about:
android = uiautomator1
ios=uiautomation

jlipps: 👍 i'm ambivalent between `UIAutomation` and `Instruments` for iOS though.

dpgraham:
Just pushed a commit.... I realized we probably shouldn't be deprecating empty `automationName` cap for mac, windows or tizen (anything other than Android or IOS). Made changes:

* Added Instruments and UiAutomator1 to automation names
* Moved deprecation warning into driver selector functions

lib/appium.js
dpgraham: `instruments` is probably better.... distinguishes it from Android

mykola-mokhnach (approved):

mykola-mokhnach:
Although there is still a nice patch https://github.com/appium/appium/pull/11067 from @imurchie waiting more than 6 months for its time

KazuCocoa (approved):
👍

imurchie (approved):

PR #12032, opened 2019-01-18
Update scroll.yml

juhlila:
## Proposed changes

there is a typo on ruby method example

## Types of changes

changed 'touch_action' to 'touch_actions'


KazuCocoa (approved):
thanks!

PR #12100, opened 2019-01-30
Update value to print out in for loop

biwkf:
Change value to print String contextName not contextNames within for loop


imurchie (approved):
Thanks for catching this!

PR #12110, opened 2019-01-31
Update auto-generated docs

triager:


KazuCocoa:
same as https://github.com/appium/appium/pull/12108/files

PR #12136, opened 2019-02-07
Update Execute Mobile Command docs

biwkf:
mobile:pressButton on iOS real device supports values: home, volumeup and volumedown

### DIFF
--- a/commands-yml/commands/mobile-command.yml
+++ b/commands-yml/commands/mobile-command.yml
@@ -46,7 +46,7 @@
       | mobile:installCertificate | refer to [IOS Pasteboard Guide](/docs/en/writing-running-appium/ios/ios-xctest-install-certificate.md)  | | |
       | mobile:getContexts | Retrieve available contexts, along with the url and title associated with each webview (see [get contexts](/docs/en/commands/context/get-contexts.md)) | | |
       | mobile:batteryInfo | Reads the battery information from the device under test | <none> | <none> |
-      | mobile:pressButton | Press a physical button. The supported button name is is _home_. _volumeup_ and _volumedown_ are available for real devices | `{name}` | `{name: "home"}` |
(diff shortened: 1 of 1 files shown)


commands-yml/commands/mobile-command.yml
```diff
@@ -46,7 +46,7 @@
       | mobile:installCertificate | refer to [IOS Pasteboard Guide](/docs/en/writing-running-appium/ios/ios-xctest-install-certificate.md)  | | |
       | mobile:getContexts | Retrieve available contexts, along with the url and title associated with each webview (see [get contexts](/docs/en/commands/context/get-contexts.md)) | | |
       | mobile:batteryInfo | Reads the battery information from the device under test | <none> | <none> |
-      | mobile:pressButton | Press a physical button. The supported button name is is _home_. _volumeup_ and _volumedown_ are available for real devices | `{name}` | `{name: "home"}` |
```
dpgraham: Why was `are available for real devices` taken out?

biwkf: Real device supports all 3 buttons. I was writing a test a few weeks ago that required pressing the home button. When I read the docs I didn't think that it was supported on real devices since it explicitly said "_volumeup_ and _volumedown_ are available for real devices". So I suggest removing available for real devices or add _home_ also

dpgraham: Okay, I understand now. I would change it to say that _home_ is supported on Simulators and real devices and _volumeup_ and _volumedown_ are available for real devices.

biwkf: Updated now after testing on simulator. I wrongly assumed that if those buttons worked on real devices they would work in simulator.

dpgraham (approved):

dpgraham:
Thank you @biwkf !

PR #12194, opened 2019-02-20
Add prepublish script to prune devDependencies from shrinkwrap

dpgraham:
Currently, we use a 'production-only' npm shrinkwrap on our release branches. The problem with this is that when you try running `npm ci` (clean install) it won't work because there's an incongruency between what's in `npm-shrinkwrap.json` and what's in `package.json` (it expects the dev deps to be there). 

I'd like to be able to use `npm ci` because it installs what's in the shrinkwrap (`npm install` doesn't respect shrinkwrap) and thus we can test Appium locally using the dependencies that will be shrinkwrapped and published.

The fix for this is a `prepublishOnly` script and a `publish` (postpublish) script

This `prepublishOnly` script runs the following commands:

```
npm ci --production --ignore-scripts // Clean install production dependencies found in npm-shrinkwrap.json
npm shrinkwrap // re-create the npm-shrinkwrap.json with only prod dependencies
npm install --only=dev --no-shrinkwrap // re-install dev dependencies; don't alter shrinkwrap
```

The `publish` script runs `git checkout HEAD -- npm-shrinkwrap.json` which restores the `npm-shrinkwrap.json` to what it was before the prepublish script.

When a publish happens, it will then only ever publish an `npm-shrinkwrap.json` that has prod dependencies (which is desirable). But, we can now start committing `npm-shrinkwrap.json` on release branches to include dev _and_ prod dependencies so that we don't have the `npm ci` problems.

(NOTE: the documentation for our new release flow is forthcoming)

### DIFF
--- a/docs/en/contributing-to-appium/developers-overview.md
+++ b/docs/en/contributing-to-appium/developers-overview.md
@@ -195,7 +195,7 @@ converted into the `npm-shrinkwrap.json` file.
 1. Determine whether we have a `patch` (bugfix), `minor` (feature), or `major` (breaking) release according to the principles of SemVer.
 1. Update `package.json` with the appropriate new version.
 1. Update the CHANGELOG/README with appropriate changes and submit for review as a PR, along with shrinkwrap and `package.json` changes. Wait for it to be merged, then pull it into the release branch.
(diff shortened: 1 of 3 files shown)


jlipps (approved):
yeah, makes sense to me.

imurchie (approved):

PR #12204, opened 2019-02-21
add some javascript clients to the clients list

jlipps:
wdio was missing, and one that i've been working on for the browser has now been open sourced.


dpgraham (approved):
Awesome! So we should be able to use the browser client in Appium Desktop?

KazuCocoa (approved):

PR #12253, opened 2019-03-04
fix: Skip prune shrinkwrap if no "npm-shrinkwrap.json"

dpgraham:
Can't do 'beta' releases because it tries to prune shrinkwrap when `npm-shrinkwrap.json` doesn't exist.

This checks that `npm-shrinkwrap.json` exists before pruning dependencies. If it doesn't exist, don't prune.


package.json
```diff
@@ -114,6 +114,7 @@
     "gulp": "^4.0.0",
     "handlebars": "^4.0.10",
     "mocha": "^6.0.0",
+    "path-exists-cli": "^1.0.0",
```
jlipps: i hate that we need another dev dep just to do a `ls` but, sigh, cross platform

jlipps (approved):

package.json
imurchie: Should be able to be `test -e npm-shrinkwrap.json && (...)`

dpgraham: Would that work in windows @imurchie ?

imurchie: No idea. But given the usage of this, is it necessary to be cross-platform?

dpgraham: @imurchie I'll change it to that since we only publish from Mac

PR #12308, opened 2019-03-14
Publish nightly Appium bundle

dpgraham:
Ported over the Appium bundle task from Travis to Azure Pipelines.

How it works:
* Every weeknight at 6pm PST (2am in Berlin) this pipeline is triggered and it creates an Appium bundle that gets uploaded to Github Releases (appium/appium-build-store) and uploaded to Sauce Storage so that it can be used in Appium staging tests
* When this build gets completed, the XCUITestDriver nightly build (and eventually Espresso and UiAutomator2) gets triggered and runs staging tests on Sauce Labs


KazuCocoa (approved):
This PR looks good to me.
> When this build gets completed, the XCUITestDriver nightly build (and eventually Espresso and UiAutomator2) gets triggered and runs staging tests on Sauce Labs

👍 looking forward to the combination

PR #12352, opened 2019-03-19
docs: fullReset happens before test not after done

dpgraham:


KazuCocoa (approved):

heiko-mueller:
Are you sure this is correct? I am using fullReset=true on Android real devices and the app is always being uninstalled *after* the test. This is the case with Appium server version 1.10.0 as well as 1.12.0. I am using the Appium Java client with a minimum set of capabilities (udid, deviceName, platformName, platformVersion, appActivity, automationName, autoGrantPermissions, app, appPackage, systemPort and fullReset).

KazuCocoa:
ah, both are correct if only `fullReset` is true. (with some conditions)

dpgraham:
Changes pushed

KazuCocoa (approved):

PR #12423, opened 2019-04-01
Remove invalid word from docs

WillBrock:
## Proposed changes

Removing the word `fuller` from the docs since it's not a correct english word.

## Types of changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Similarly, with Google's UiAutomator or Espresso you can only write tests in
 Java. Appium opens up the possibility of true cross-platform native app
 automation, for mobile and beyond. Finally!
 
-If you're new to Appium, or want a fuller description of what this is all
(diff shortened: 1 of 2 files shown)


imurchie:
> it's not a correct english word
This is untrue, though.

WillBrock:
Sorry, it's a correct word but is used incorrectly in this case.

imurchie:
Even that is not correct (there is a description on the page; there is a _fuller_ (as the comparative form of _full_) description on another page; there is not, for better or for worse, a _fullest_ description on any other page).

I'm ok with the change as it stands, though.

WillBrock:
You're right. I guess I've never seen it used this way.

jlipps (approved):

jlipps:
Yeah, `fuller` is OK, but I'm also fine with `more comprehensive`, which is a ... more comprehensive ... way of saying the same thing.

PR #12440, opened 2019-04-04
Add a possibility to parse initial values for Appium settings from capabilities

mykola-mokhnach:
## Proposed changes

It is sometimes necessary to set some particular appium setting per session. This feature will allow to pass default Appium settings values in capabilities, for example

```
{
        platformName: 'foo',
        browserName: 'bar',
        'settings[settingName]': 'baz',
        'settings[settingName2]': 'baz2',
}
```

so there is no need to separately call `updateSettings` API after the session is created.

## Types of changes

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -254,7 +254,13 @@ class AppiumDriver extends BaseDriver {
     jsonwpCaps = _.cloneDeep(jsonwpCaps);
     const jwpSettings = Object.assign({}, defaultSettings, pullSettings(jsonwpCaps));
     w3cCapabilities = _.cloneDeep(w3cCapabilities);
-    const w3cSettings = Object.assign({}, jwpSettings, pullSettings(w3cCapabilities));
+    const allW3cCaps = Object.assign({},
+      (w3cCapabilities || {}).alwaysMatch || {},
+      ((w3cCapabilities || {}).firstMatch || [])
+        .filter((x) => _.isPlainObject(x))
+        .reduce((acc, x) => Object.assign(acc, x), {})
+    );
+    const w3cSettings = Object.assign({}, jwpSettings, pullSettings(allW3cCaps));
 
     let protocol;
     let innerSessionId, dCaps;
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -248,12 +248,15 @@ class AppiumDriver extends BaseDriver {
    * @return {Array} Unique session ID and capabilities
    */
   async createSession (jsonwpCaps, reqCaps, w3cCapabilities) {
-    let {defaultCapabilities} = this.args;
-    defaultCapabilities = _.cloneDeep(defaultCapabilities);
+    const defaultCapabilities = _.cloneDeep(this.args.defaultCapabilities);
     const defaultSettings = pullSettings(defaultCapabilities);
     jsonwpCaps = _.cloneDeep(jsonwpCaps);
     const jwpSettings = Object.assign({}, defaultSettings, pullSettings(jsonwpCaps));
     w3cCapabilities = _.cloneDeep(w3cCapabilities);
+    // It is possible that the client only provides caps using JSONWP standard,
+    // although firstMatch/alwaysMatch properties are still present.
+    // In such case we assume the client understands W3C protocol and merge the given
+    // JSONWP caps to W3C caps
     const w3cSettings = Object.assign({}, jwpSettings);
     Object.assign(w3cSettings, pullSettings((w3cCapabilities || {}).alwaysMatch || {}));
     for (const firstMatchEntry of ((w3cCapabilities || {}).firstMatch || [])) {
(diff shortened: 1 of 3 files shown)


lib/appium.js
```diff
@@ -246,10 +248,16 @@ class AppiumDriver extends BaseDriver {
...
    */
   async createSession (jsonwpCaps, reqCaps, w3cCapabilities) {
-    const {defaultCapabilities} = this.args;
+    let {defaultCapabilities} = this.args;
+    defaultCapabilities = _.cloneDeep(defaultCapabilities);
+    const defaultSettings = pullSettings(defaultCapabilities);
+    jsonwpCaps = _.cloneDeep(jsonwpCaps);
+    const jwpSettings = Object.assign({}, defaultSettings, pullSettings(jsonwpCaps));
+    w3cCapabilities = _.cloneDeep(w3cCapabilities);
+    const w3cSettings = Object.assign({}, jwpSettings, pullSettings(w3cCapabilities));
```
jlipps: why are we merging jwp settings into w3c settings? below these seem like exclusive options

lib/utils.js
```diff
@@ -251,9 +251,43 @@ function getPackageVersion (pkgName) {
...
+  if (!_.isPlainObject(caps) || _.isEmpty(caps)) {
+    return {};
+  }
+
+  const result = {};
+  for (const [key, value] of _.toPairs(caps)) {
+    const match = /\bsettings\[(\S+)\]$/.exec(key);
+    if (!match) {
+      if (_.isPlainObject(value)) {
+        Object.assign(result, pullSettings(value));
```
jlipps: i'm having trouble imagining the need for a recursive case...?

test/utils-specs.js
```diff
@@ -178,4 +179,65 @@ describe('utils', function () {
       });
     });
   });
+
+  describe('pullSettings()', function () {
```
jlipps: what about a test for a setting that has an object as a value?

jlipps (commented):
nice idea! any reason you chose to put each setting as a string of the form `settings[setting]` rather than just adding a new cap called `settings` which has a value of type object? seems like that obviates the need for `pullSettings` entirely. might be a little harder to construct the cap client side, but every language i've seen has the possibility of complex capability values.

lib/utils.js
```diff
@@ -251,9 +251,43 @@ function getPackageVersion (pkgName) {
   return pkgInfo.version;
 }
 
+/**
+ * Pulls the initial values of Appium settings from the given capabilities argument.
```
KazuCocoa: Is it helpful to add https://github.com/appium/appium/blob/master/docs/en/advanced-concepts/settings.md around here?

lib/appium.js
mykola-mokhnach: I've seen the cases where the client sends jsonwp caps only, but the request contains firstMatch/alwaysMatch props. In such case we assume the client supports W3C and merge jwp caps to w3c ones

lib/utils.js
mykola-mokhnach: this might be the case for w3c caps, where they have firstMatch and alwaysMatch properties

test/utils-specs.js
mykola-mokhnach: sure, will add

lib/utils.js
mykola-mokhnach: will add

mykola-mokhnach:
> any reason you chose to put each setting as a string of the form settings[setting] rather than just adding a new cap called settings which has a value of type object? seems like that obviates the need for pullSettings entirely. might be a little harder to construct the cap client side, but every language i've seen has the possibility of complex capability values.

This is not very straightforward in java/C#, where one has to create a JSONObject first and put all the necessary values there. Also, I usually don't see a need to set multiple setting values at once. This syntax looks more natural to set a couple of settings. In the worst case we could also add a possibility to set `settings[*]`, for example, if this is really needed

lib/utils.js
mykola-mokhnach: I've changed the logic there. Now it's no more recursive

lib/utils.js
dpgraham: I'd like a link to http://appium.io/docs/en/advanced-concepts/settings/index.html so we know what "Settings" are.

lib/utils.js
```diff
@@ -251,9 +251,43 @@ function getPackageVersion (pkgName) {
...
+ * their values.
+ */
+function pullSettings (caps) {
+  if (!_.isPlainObject(caps) || _.isEmpty(caps)) {
+    return {};
+  }
+
+  const result = {};
+  for (const [key, value] of _.toPairs(caps)) {
+    const match = /\bsettings\[(\S+)\]$/.exec(key);
```
dpgraham: Maybe `const [, settingName] =  /\bsettings\[(\S+)\]$/.exec(key);`

dpgraham (changes requested):

lib/utils.js
mykola-mokhnach: this will throw if match is not an array

lib/utils.js
mykola-mokhnach: see below

lib/utils.js
dpgraham: Can we still have a variable called `settingName` to make it more clear what's being parsed.  

`const [, settingName] = /\bsettings\[(\S+)\]$/.exec(key) || []`;

dpgraham (approved):

PR #12401, opened 2019-03-28
add tvOS platform

KazuCocoa:
## Proposed changes

I've added `tvOS` platform.
`tvOS` also uses xcuites-driver so I've added `tvOS` platform with xcuitest automation name.

- related PR
    - https://github.com/appium/appium-xcuitest-driver/pull/911
- a test scenario
    - https://github.com/appium/ruby_lib_core/pull/204/files#diff-0d144d22af05542238b44094c9b1002eR30
- example

tvos

A note:
I'll add a doc for tvOS later as another PR

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla.js.foundation/appium/appium)
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #12480, opened 2019-04-11
add deviceType for tvOS support instead of adding a new platform name

KazuCocoa:
## Proposed changes

I've introduced tvOS in https://github.com/appium/appium/pull/12401/files
after a while, I considered adding `deviceType` was better than adding new platform since the tv was kind of new device/product in iOS(Apple) platform.

If we start considering such deviceType, we probably can think device variations such as watch, tv etc even if they are Android/iOS/Windows etc platform.

Will update xcuitest-driver and bump the version in package.json in this repository

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/docs/en/drivers/ios-xcuitest.md
+++ b/docs/en/drivers/ios-xcuitest.md
@@ -46,7 +46,7 @@ the value `XCUITest`. Of course, you must also include appropriate
 `platformName`, `platformVersion`, `deviceName`, and `app` capabilities, at
 a minimum.
 
-The `platformName` should be `iOS` for iPhone or iPad. tvOS devices are available if the `platformName` is `tvOS`.
+The `platformName` should be `iOS` for iPhone or iPad. `deviceType` should be `tv` for tvOS.
 
 - iOS
    ```json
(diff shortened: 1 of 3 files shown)


mykola-mokhnach (approved):

mykola-mokhnach:
@imurchie Are you ok with this change?

imurchie (commented):
I'm not sure I understand the reasoning behind this. `iOS` and `tvOS` are different platforms, even if both handled by XCUITest.

Why add another capability?

KazuCocoa:
When I managed android/android TV/fire TV  and used Appium before, I handled them like device type(similar to this deviceType caps) in my env. The name of xxxTv changed (Google TV => Android TV) once. The TV also needed different build like ios and tvOS applications. 
I was not sure tvOS name will change like the above case, but thinking as below was better, I thought. 


- Run tests on android platform on TV device
- Run tests on android platform on firetv device
- Run tests on ios(apple) platform on TV device

imurchie:
The handling at the AppiumDriver level will be gone once 2.0 is implemented, so it is not really a concern.

My main concern is that iOS and tvOS are entirely different platforms, not versions of each other. I strongly object to the concept of `deviceType` in this case (Android may be a different thing, since there the tv system _is_ derived from the phone system).

KazuCocoa:
I could understand 👌
Thanks for your opinion!

PR #12483, opened 2019-04-12
update maximize.yml code for ruby only

lrkaussie:
For Ruby file shows set_to, change to maximize


dpgraham (approved):
Approved, pending signing CLA.

PR #12484, opened 2019-04-12
update set window size.yml

lrkaussie:
change from maximize to resize_to in ruby code only


KazuCocoa (approved):
Thanks!
Could you sign the CLA?

imurchie (approved):

PR #12493, opened 2019-04-12
[wip] codecov

dpgraham:
Adds coverage reports to pull requests


.npmignore
```diff
@@ -0,0 +1,16 @@
+assets/
```
imurchie: This file is ignored. Having it will be confusing.

.travis.yml
```diff
@@ -13,7 +13,10 @@ jobs:
   include:
     - stage: Tests
       os: linux
-      script: npm run test && npm run e2e-test
+      script: |
+        npm run e2e-test
+        npx gulp coverage
+        npx codecov
       node_js: "11"
       after_script: npm run coverage
```
imurchie: Are we going to get rid of coveralls, then?

imurchie (commented):
I'd rather this be moved into `appium-gulp-plugins` so the coverage deps and whatnot are all kept in one place, and usable throughout.

dpgraham:
@imurchie yeah I just wanted to try it out. Maybe coveralls provides code coverage analysis within pull requests?

imurchie:
I feel like it used to, but I could be mistaken. If there is a better solution, I'm all for it, though.

Experiment away! Once it works well we can move it over (and put it under what we already map to `npm run coverage`, so we don't need to do anything in the packages that already do that.

dpgraham:
I just had to enable it on Coveralls (see example here: https://github.com/appium/appium/pull/12494#issuecomment-482669164)

Closing this in favour of keeping Coveralls.

PR #12574, opened 2019-05-02
update installation of fbsimctl

KazuCocoa:
## Proposed changes

[update]
Install idb for appium-idb instead of `fbsimctl`

----

https://github.com/facebook/idb has been renamed and changed the structure.
It requires to use Python 3.6. But we need Python 2.7 to build heapdump in nodejs.
I haven't ensured the new one also can call `fbsimctl`, too.

This PR is to guide users to install previous `fbsimctl` which is tagged as `1.0.0` in the repo instead of installing it via homebrew.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/docs/en/drivers/ios-xcuitest.md
+++ b/docs/en/drivers/ios-xcuitest.md
@@ -127,7 +127,7 @@ following desired capabilities:
 
 * Install idb for better handling of various iOS Simulator operations,
 such as: biometrics, geolocation setting and window focussing.
-    * Read https://github.com/appium/appium-idb#installation to install necessary libraries
+    * Read https://github.com/appium/appium-idb#installation to install necessary libraries (since Appium 1.14.0)
 
 * Install [AppleSimulatorUtils](https://github.com/wix/AppleSimulatorUtils)
 to use the [permissions capability](https://github.com/appium/appium-xcuitest-driver#desired-capabilities)
--- a/docs/en/drivers/ios-xcuitest.md
+++ b/docs/en/drivers/ios-xcuitest.md
@@ -129,10 +129,17 @@ following desired capabilities:
 such as: biometrics, geolocation setting and window focussing.
 
 ```
-# Get the Facebook Tap.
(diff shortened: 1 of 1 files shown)


docs/en/drivers/ios-xcuitest.md
```diff
@@ -129,10 +129,17 @@ following desired capabilities:
 such as: biometrics, geolocation setting and window focussing.
 
 ```
-# Get the Facebook Tap.
-brew tap facebook/fb
```
mykola-mokhnach: is the formula removed from brew?

KazuCocoa: No, but it only provides 0.4.0 and HEAD for now.
The 0.4.0 is old. HEAD fails to build, today.

I was able to checkout out v1.0.6 tag from the repo and got `/bin/fbsimctl`. But not sure the command will remain so far, sine commands built by `build.sh` decreased than `v1.0.0` tag.

Thus, so far, getting the repo, checkout `v1.0.0` and built binaries are safer way.

KazuCocoa:
note:
https://www.fbidb.io/docs/commands#set-a-simulators-location

docs/en/drivers/ios-xcuitest.md
```diff
@@ -125,15 +125,9 @@ following desired capabilities:
 
 ### Optional Setup
 
-* Install FBSimulatorControl for better handling of various iOS Simulator operations,
+* Install idb for better handling of various iOS Simulator operations,
```
mykola-mokhnach: It probably makes sense to mention that idb is only supported since 1.13.1

imurchie (approved):

docs/en/drivers/ios-xcuitest.md
KazuCocoa: 1.13.1 has not been released yet. 1.14.0-beta was released.
I would like to update either 1.13.1 or 1.14.0 here after the version has been fixed

PR #12630, opened 2019-05-10
Prepare 1.13.1 release that still supports Node 8

dpgraham:
This gets around the need for Node v10 support by reverting to v1.13.0 and then only updating XCUI and Espresso to versions that were published targeting Node v8


KazuCocoa (approved):

npm-shrinkwrap.json
```diff
@@ -2600,9 +2600,9 @@
       }
     },
     "appium-xcuitest-driver": {
-      "version": "2.113.2",
-      "resolved": "https://registry.npmjs.org/appium-xcuitest-driver/-/appium-xcuitest-driver-2.113.2.tgz",
-      "integrity": "sha512-dhid34cx99AG61O69gbIgvW51RUKtd9uQkKziQe0/XGApAozjwlxgMJo/pjshpuz2rXCMz3ilRcrLnscT36qmQ==",
+      "version": "2.114.0",
```
KazuCocoa: is it greater to bump https://github.com/appium/appium-xcuitest-driver/commits/master again and set here as 2.114.3? (They have fixes for WDA building/signing)

Issue #12579, opened 2019-05-02
--relaxed_security flag should whitelist features


jlipps:
## The problem

Currently, 2 or 3 features require the `--relaxed_security` flag. Let's imagine a scenario where we add a new feature down the road, which requires the flag, but which has drastically different security ramifications than the current set of features. Someone would upgrade Appium, and unwittingly open up a security hole they hadn't considered. 

## The solution

Should probably offer a whitelist of features, so it'd be like:

```
appium --relaxed_security=server_logs
appium --relaxed_security=all
appium --relaxed_security=mobile_shell,server_logs
```

imurchie:
I agree with this. Except the `'all'` option part, which I feel would open the same hole. These should be fully opt-in.

mykola-mokhnach:
Closed as fixed

PR #12778, opened 2019-06-15
add --allow-insecure and --deny-insecure server flags

jlipps:
## Proposed changes

The `--relaxed-security` flag is either on or off for all features. But an admin should be able to decide which insecure features to allow or not allow. Hence this PR adds the `--allow-insecure` flag which takes a comma separated list of strings pertaining to feature names. Driver authors can then read this flag in conjunction with features to determine whether or not to allow those features.

Accordingly `--relaxed-security` is deprecated so we should move to `--allow-insecure` only. Separate PRs to follow to update all existing instances of relaxedSecurityEnabled to use allowInsecure as well.

## Types of changes

## Checklist

- [X] I have signed the [CLA](https://cla.js.foundation/appium/appium)
- [X] Any dependent changes have been merged and published in downstream modules


lib/appium.js
```diff
@@ -321,10 +321,21 @@ class AppiumDriver extends BaseDriver {
...
+
+      // We want to assign security values directly on the driver. The driver
+      // should not read security values from `this.opts` because those values
+      // could have been set by a malicious user via capabilities, whereas we
+      // want a guarantee the values were set by the appium server admin
       if (this.args.relaxedSecurityEnabled) {
-        log.info(`Applying relaxed security to '${InnerDriver.name}' as per server command line argument`);
+        log.info(`Applying relaxed security to '${InnerDriver.name}' as per server command line argument. This flag is DEPRECATED. Use --allow-insecure instead.`);
         d.relaxedSecurityEnabled = true;
+      } else if (this.args.allowInsecure && this.args.allowInsecure.length > 0) {
```
mykola-mokhnach: !_.isEmpty(this.args.allowInsecure)

lib/parser.js
```diff
@@ -381,20 +381,32 @@ const args = [
           'session unless overridden by received capabilities.'
   }],
 
+  [['--allow-insecure'], {
+    dest: 'allowInsecure',
+    defaultValue: [],
+    type: parseAllowInsecureFeatures,
+    required: false,
+    example: 'execute-webdriver,adb-shell',
```
mykola-mokhnach: This value is the subject of discussion. I would put names of the particular mobile commands here.Also, the maximum length of command line parameters is limited. that is why it also makes sense to enumerate possible values in a file rather than as cmdline argument. The third problem I would like to mention is blacklisting vs whitelisting. I mean one could still provide `relaxedSecurityEnabled`, but disable particular commands with `denyInsecure` arg

jlipps: So @mykola-mokhnach to make sure I'm understanding you, you are saying:

1. We should keep `--relaxed-security` as a "turn on all insecure features" flag
2. We should have `--allow-insecure` as a whitelist (so can never be used with `--relaxed-security`
3. We should have `--deny-insecure` as a blacklist (so must be used with `--relaxed-security`)
4. We should optionally read feature names from a file and not just CLI. What is the right format for this kind of file, do you think?

I'm happy to implement this system if it's what we want. In this case I think it'd be a good idea to make a function in BaseDriver like `isFeatureAllowed(featureName)`, which does all the checking based on these three flags, so that individual drivers only ever need to call `isFeatureAllowed`, to keep their code clean and future-proof. Sound good?

jlipps: i went ahead and implemented that method in the most recent version of https://github.com/appium/appium-base-driver/pull/328

mykola-mokhnach: > We should have --allow-insecure as a whitelist (so can never be used with --relaxed-security

almost. These args should have priorities, where `relaxed-security` has the lowest one and `deny-insecure` has the highest one. So, for example, if one says `--relaxed-security --allow-insecure shell --deny-insecure shell` then no features are going to be enabled, because:

- `allow-insecure` overrides `relaxed-security` and only allows `shell`
- `deny-insecure` overrides everything else and denies `shell` completely

mykola-mokhnach: > make a function in BaseDriver like isFeatureAllowed(featureName)

mykola-mokhnach: > What is the right format for this kind of file, do you think

I would say each feature in the new line should be ok. Empty or unsupported features can be simply ignored

KazuCocoa:
link: https://github.com/appium/appium/issues/12579

jlipps:
ok @mykola-mokhnach i believe i implemented everything in the way we discussed, have another look.

docs/en/writing-running-appium/server-args.md
```diff
@@ -93,3 +93,5 @@ All flags are optional, but some are required in conjunction with certain others
 |`--localizable-strings-dir`|en.lproj|[DEPRECATED] - (IOS-only) the relative path of the dir where Localizable.strings file resides |`--localizable-strings-dir en.lproj`|
 |`--show-ios-log`|false|[DEPRECATED] - (IOS-only) if set, the iOS system log will be written to the console||
 |`--relaxed-security`|false|Disable additional security checks, so it is possible to use some advanced features, provided by drivers supporting this option. Only enable it if all the clients are in the trusted network and it is not the case if a client could potentially break out of the session sandbox.||
```
mykola-mokhnach: it makes sense to mention that both allow/deny-insecure have priority over that option

lib/appium.js
```diff
@@ -177,6 +177,11 @@ class AppiumDriver extends BaseDriver {
     // It is not recommended to access this property directly from the outside
     this.pendingDrivers = {};
 
+    // initialize security modes
```
mykola-mokhnach: these are not used

lib/parser.js
```diff
@@ -389,9 +390,37 @@ const args = [
...
+          'using the --deny-insecure flag',
     nargs: 0
   }],
+
+  [['--allow-insecure'], {
+    dest: 'allowInsecure',
+    defaultValue: [],
+    type: parseSecurityFeatures,
+    required: false,
+    example: 'execute-webdriver,adb-shell',
```
mykola-mokhnach: I'm still not sure how you are going to transform existing commands to these abbreviations. Or there will be another separate mapping?

lib/parser.js
```diff
@@ -789,6 +818,30 @@ function updateParseArgsForDefaultCapabilities (parser) {
   };
 }
 
+function parseSecurityFeatures (features) {
+  const splitter = (splitOn) => {
+    return (str) => {
```
mykola-mokhnach: returns are redundant

lib/parser.js
```diff
@@ -789,6 +818,30 @@ function updateParseArgsForDefaultCapabilities (parser) {
...
+      return str.toString().split(splitOn).map(s => s.trim()).filter(Boolean);
```
mykola-mokhnach:
```
`${str}` would be safer
```

lib/parser.js
```diff
@@ -789,6 +818,30 @@ function updateParseArgsForDefaultCapabilities (parser) {
...
+    };
+  };
+  let parsedFeatures;
+  try {
+    parsedFeatures = splitter(',')(features);
+  } catch (err) {
+    throw new Error('Could not parse value of --allow-insecure. Should be a list of strings separated by commas');
```
mykola-mokhnach: is it only applied to `allow-insecure` ?

mykola-mokhnach: > Should be a list of strings separated by commas

Can also be a file path

lib/parser.js
```diff
@@ -789,6 +818,30 @@ function updateParseArgsForDefaultCapabilities (parser) {
...
+  }
+
+  if (parsedFeatures.length === 1) {
```
mykola-mokhnach: && fs.exists (...)

lib/parser.js
```diff
@@ -789,6 +818,30 @@ function updateParseArgsForDefaultCapabilities (parser) {
...
+    // we might have a file which is a list of features
+    try {
+      const fileFeatures = fs.readFileSync(parsedFeatures[0], 'utf8');
+      parsedFeatures = splitter(os.EOL)(fileFeatures);
```
mykola-mokhnach: I'd rather split on '\n' and then trim. It might be possible that on Windows we still have a file in Unix format

lib/parser.js
KazuCocoa: Me, too.
Maybe, should we create a mapping file...?
execute-webdriver is probably easy since the command is a new command in routes.js. We can filter it with the command, _executeDriverScript_, for example.
The adb shell is defined in a part of mobile commands. How to map them... 🤔 https://developer.android.com/studio/command-line/adb#shellcommands

- adb-shell # all adb shell
-  _adb shell_ and one argument
    - e.g. adb-shell-pm (for [this](https://developer.android.com/studio/command-line/adb#pm)), adb-shell-pm, adb-shell-dpm, adb-shell-screencap, adb-shell-screenrecord...?

jlipps: I don't think we need any mapping. Different drivers can define their own feature names. All that matters is that the drivers (a) actually check whether a feature is enabled if that feature is insecure, and (b) document the feature names so people know what to use. I'll also create a new doc in the main docs area to list known feature names for the officially supported drivers.

lib/parser.js
jlipps: 👍

lib/appium.js
jlipps: Yeah, I need to put them into BaseDriver instead, oops.

docs/en/writing-running-appium/server-args.md
jlipps: But --allow-insecure does not have priority over --relaxed-security. If they are both used then --allow-insecure is doing nothing.

lib/parser.js
jlipps: is the call to `exists` more performant than just handling a failed call to `readFile`?

jlipps: it's ok, i like it better because we can give a better error message if the file can't be read.

mykola-mokhnach: `exists`, actually, verifies whether an item is available for read

jlipps:
full doc page for security features added.

lib/appium.js
```diff
@@ -321,10 +321,30 @@ class AppiumDriver extends BaseDriver {
...
-        log.info(`Applying relaxed security to '${InnerDriver.name}' as per server command line argument`);
+        log.info(`Applying relaxed security to '${InnerDriver.name}' as per ` +
+                 `server command line argument. All insecure features will be ` +
+                 `enabled unless explicitly disabled by --deny-insecure`);
         d.relaxedSecurityEnabled = true;
       }
+
+      if (!_.isEmpty(this.args.denyInsecure)) {
+        log.info(`Explicitly preventing use of insecure features ` +
+                 `${JSON.stringify(this.args.allowInsecure)} for this server`);
```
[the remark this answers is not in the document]
imurchie: This log line will be hard to read for anything but the simplest case. It would make more sense, I think, to have them listed on their own line, so it is abundantly clear what is and is not available.

jlipps:
OK, any more comments?

imurchie (approved):

lib/parser.js
KazuCocoa: I could imagine with https://github.com/appium/appium-base-driver/pull/328/files#diff-69c68e3397852a86397892936887db6aR258
What about addressing usage? of `isFeatureEnabled` for developers in docs/en/writing-running-appium/security.md later? It helps for other driver developers to get the security feature in their drivers

KazuCocoa (approved):
🎉

PR #12790, opened 2019-06-18
Update start-recording-screen.yml

edizzyy:
It can be confusing to pass the parameters without adding the given parameters under "options" - this is not stated anywhere in the documentation.

## Proposed changes

Add an explanation to use "options" parameter before using the json parameters on the documentation

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/commands-yml/commands/device/recording-screen/start-recording-screen.yml
+++ b/commands-yml/commands/device/recording-screen/start-recording-screen.yml
@@ -81,8 +81,10 @@ endpoint:
     - name: session_id
       description: ID of the session to route the command to
   json_parameters:
-    - short_description: The following parameters should be passed inside "options" key.
-    - name: remotePath
+    - name: options 
+      type: object
+      description: The following parameters of the action
+    - name: options.remotePath
       type: string
       description: The path to the remote location, where the resulting video should be uploaded.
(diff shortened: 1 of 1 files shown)


ediz12:
I am unsure how you would fix the documentation here - is there a specific documentation I have to follow based on appium's documentation style according to the build error logs?

imurchie:
There's no documentation on it, at the moment.

The problem you're having, though, is that there is no `short_description` attribute for that field, so the parser doesn't know what to do.

The way that this has been solved in the past can be seen in the JSON parameters for `touch-perform`: https://github.com/appium/appium/blob/master/commands-yml/commands/interactions/touch/touch-perform.yml#L102-L116

ediz12:
> There's no documentation on it, at the moment.
> 
> The problem you're having, though, is that there is no `short_description` attribute for that field, so the parser doesn't know what to do.
> 
> The way that this has been solved in the past can be seen in the JSON parameters for `touch-perform`: https://github.com/appium/appium/blob/master/commands-yml/commands/interactions/touch/touch-perform.yml#L102-L116

Fixed it accordingly - I believe it makes more sense now.

imurchie (approved):
Thanks for this! Definitely an improvement.

PR #12807, opened 2019-06-21
add docs for execute driver script command

jlipps:
Docs to support https://github.com/appium/appium-base-driver/pull/328


commands-yml/commands/session/execute-driver.yml
```diff
@@ -0,0 +1,128 @@
...
+  python:
+    |
+      # TODO fill out once client code is written
+  javascript_wd:
+    |
+      const script = `
+        const el = await driver.$('~foo');
+        await el.click();
+      `;
+      driver.executeScript(script, {timeout: 200});
```
KazuCocoa: Should here be blank? Or `executeDriver`?
If the wd client already has it, we should update `javascript_wd: false` in client support section, too.

jlipps: ah yes, it should be `executeDriver`, sorry. i was just making this up. it does not yet exist anywhere, but i wanted at least one example.

KazuCocoa (approved):

PR #12826, opened 2019-06-25
Specify minimum and maximum Node versions

dpgraham:
The "engines" variable in package.json should specify the minimum and the maximum NodeJS version that Appium supports. This will allow cloud providers to check what the maximum supported version is and bundle that as the binary when doing Appium releases.

Note that the engines variable doesn't actually stop users from using Appium for out-of-range Node versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "url": "https://github.com/appium/appium/issues"
   },
   "engines": {
-    "node": ">=8",
+    "node": ">=10 <=12",
     "npm": ">=6"
   },
   "main": "./build/lib/main.js",


umutuzgur (approved):

jlipps (approved):
we do a check in appium and fail though if the version isn't correct, right?

dpgraham:
@jlipps I don't know if we do. I think I remember installing Appium with an out-of-range NodeJS version and it still working.

I'm actually re-thinking this PR. We should probably just always be getting the latest stable Node version and using that.

imurchie:
We check the minimum, and fail if the node version is too low. https://github.com/appium/appium/blob/master/lib/config.js#L114-L119

dpgraham:
Okay that's good. I'm actually on the fence about this now if we should do this or not. I guess the only reason I would want to specify a max is in case a new, untested version of Node comes out and breaks Appium. But then again the downside of this approach is that we'd have to keep maintaining this.

dpgraham:
On further thought, I think this is unnecessary. I think it's fine to just assume LTS is the latest supported Node version.

Issue #12867, opened 2019-07-07
Cannot set altitude by setlocation


ki4070ma:
## The problem

Cannot set altitude by setlocation

## Environment

* Appium version (or git revision) that exhibits the issue: 1.14.0-beta.2
* Last Appium version that did not exhibit the issue (if applicable): N/A
* Desktop OS/version used to run Appium: macOS 10.14.5
* Node.js version (unless using Appium.app|exe): v12.4.0
* Npm or Yarn package manager: npm
* Mobile platform/version under test: Android 9
* Real device or emulator/simulator: Real device
* Appium CLI or Appium.app|exe: CLI

## Details

Just called below codes from python-client, but altitude isn't updated to args.

SampleAndroidTest.py
```
def test_01_test_name(self):
    self.driver.set_location(121.21, 11.56, 10)
    import time
    time.sleep(10)
    print(self.driver.location)
```

Terminal
```
➜  app_apidemo git:(master) ✗ python SampleAndroidTest.py
test_01_test_name (__main__.SimpleTest) ... {'latitude': 121.21, 'longitude': 11.56, 'altitude': 5e-324}
ok

----------------------------------------------------------------------
Ran 1 test in 25.996s

OK
```

I can set altitude by executing below adb commands.

```
➜  app_apidemo git:(master) ✗ adb -P 5037 -s 13f1fd2c shell am startservice -e longitude 11.56 -e latitude 121.21 -e altitude 0 io.appium.settings/.LocationService 
Starting service: Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] cmp=io.appium.settings/.LocationService (has extras) }
➜  app_apidemo git:(master) ✗ adb -P 5037 -s 13f1fd2c shell am broadcast -n io.appium.settings/.receivers.LocationInfoReceiver -a io.appium.settings.location      
Broadcasting: Intent { act=io.appium.settings.location flg=0x400000 cmp=io.appium.settings/.receivers.LocationInfoReceiver }
Broadcast completed: result=-1, data="121.21000 11.56000 0.00000"

➜  app_apidemo git:(master) ✗ adb -P 5037 -s 13f1fd2c shell am startservice -e longitude 11.56 -e latitude 121.21 -e altitude 10 io.appium.settings/.LocationService
Starting service: Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] cmp=io.appium.settings/.LocationService (has extras) }
[...]

PR #12906, opened 2019-07-13
docs: add limitation to Set Geolocation

ki4070ma:
## Proposed changes

altitude for setLocation as arg works on only Android real devices, so added its limitation to api doc for users. 

## Types of changes

## Checklist

## Further comments

None.
Link to https://github.com/appium/appium/issues/12867

### DIFF
--- a/commands-yml/commands/session/geolocation/set-geolocation.yml
+++ b/commands-yml/commands/session/geolocation/set-geolocation.yml
@@ -1,6 +1,6 @@
 ---
 name: Set Geolocation
-short_description: Set the current geo location. (altitude works on only Android real devices)
+short_description: Set the current geo location
 
 example_usage:
   java:
@@ -79,7 +79,7 @@ endpoint:
       description: The desired geolocation longitude
     - name: altitude
       type: number
-      description: The desired geolocation altitude (optional)
+      description: The desired geolocation altitude (optional. altitude works on real Android devices only)
 
 # Links to specifications. Should link to at least one specification
 specifications:
--- a/commands-yml/commands/session/geolocation/set-geolocation.yml
+++ b/commands-yml/commands/session/geolocation/set-geolocation.yml
@@ -1,6 +1,6 @@
 ---
 name: Set Geolocation
-short_description: Set the current geo location
+short_description: Set the current geo location. (altitude for Android real device only)
 
 example_usage:
   java:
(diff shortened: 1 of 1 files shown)


commands-yml/commands/session/geolocation/set-geolocation.yml
```diff
@@ -1,6 +1,6 @@
 ---
 name: Set Geolocation
-short_description: Set the current geo location
+short_description: Set the current geo location. (altitude works on only Android real devices)
```
KazuCocoa: What about mentioning the point in https://github.com/appium/appium/pull/12906/files#diff-bd878dad3adb5356b1a3c1aae26d6d85R82 rather than short description?

ki4070ma: I understand your points, but there is spec between clients and the server in my understanding, so users might miss its info there.

I'd mean it's same to here. (but above info is more details..)
https://github.com/appium/appium/blob/4785026e65cd5b96929895b7d1cd5f1e2cff0b32/commands-yml/commands/device/network/gsm-signal.yml#L3

@mykola-mokhnach 
Can I have your comments?

mykola-mokhnach: `altitude works on real Android devices only` is fine with me

mykola-mokhnach (approved):

commands-yml/commands/session/geolocation/set-geolocation.yml
KazuCocoa: I meant this case addressed only a parameter like https://github.com/appium/appium/blob/920197b7ba6b4fa1beda0c2f86bcc921667b2f98/commands-yml/commands/device/recording-screen/start-recording-screen.yml#L116 and https://github.com/appium/appium/blob/c596608bfde4ed0dc763952712e940000cfb2e67/commands-yml/commands/device/interactions/lock.yml#L78 rather than this command is available only for an android emulator like your link.

I'm fine we'll address parameters before describing parameters if it helps client users.
You can add `description` section to mention such details, too.

---

> spec between clients and the server

For Appium client/server developers, I would recommend referring https://github.com/appium/appium-base-driver/blob/master/lib/protocol/routes.js , btw.

ki4070ma: @KazuCocoa @mykola-mokhnach 
Thanks for your comments.

@KazuCocoa 
Thanks for your examples, I agree with you and will follow that way.

KazuCocoa (approved):
thanks!

PR #12920, opened 2019-07-16
docs: add cap docs for webview page detection on android

jlipps:
docs for https://github.com/appium/appium-android-driver/pull/553


KazuCocoa:
duplication of https://github.com/appium/appium/pull/13061

PR #13053, opened 2019-08-08
Check that pruned shrinkwrap has same dependencies as unpruned shrink…

dpgraham:
## Proposed changes

Add a check after pruning the shrinkwrap to verify that the pruned shrinkwrap has the same dependencies as the original shrinkwrap

## Approach

* Move `npm-shrinkwrap.json` to `npm-shrinkwrap-backup.json`
* Run `npm ci --production --ignore-scripts` to install production-only dependencies (same as before)
* Run `npm shrinkwrap` to create a new `npm-shrinkwrap.json` with production-only dependencies (same as before)
* Run `npm install --only=dev --no-shrinkwrap` to re-install dev dependencies without mucking with shrinkwrap (same as before)
* Run a script that checks `npm-shrinkwrap.json` is a subset of `npm-shrinkwrap-backup.json` to verify that the pinned production dependencies are unchanged
* Restore shrinkwrap by running `mv npm-shrinkwrap-backup.json npm-shrinkwrap.json` so that `npm-shrinkwrap.json` is restored to its original state


KazuCocoa (approved):

imurchie (approved):
Thanks for this!

PR #13461, opened 2019-10-21
WIP: CLI commands

jlipps:


jlipps:
The main question here is, how do we invoke NPM for the actual installation/update/uninstallation? Do we use it as a subprocess (which would look like assuming the command is on the system, and setting the cwd appropriately), or do we use it directly, as in `import npm from 'npm'`. I remember doing the latter with a project in the past, and there was no way to silence output. It didn't really seem designed to be used that way.

Right now I'm kind of leaning towards just creating a simple subprocess wrapper; what do you think @imurchie @mykola-mokhnach @KazuCocoa @umutuzgur?

lib/driver-cli.js
```diff
@@ -0,0 +1,91 @@
...
+        jsonResult = await install(args);
+        break;
+      case 'uninstall':
+        jsonResult = await uninstall(args);
+        break;
+      case 'update':
+        jsonResult = await update(args);
+        break;
+      default:
+        throw new Error(`Cannot handle driver command ${args.driverCommand}`);
```
mykola-mokhnach: I'd rather show the list of supported commands

jlipps: Actually this is just a programming error. I wanted to include a default block merely out of completeness; argparse will fail long before we get here if someone tries to use a different command.

```
~/C/a/appium ❯❯❯ node . driver foo                                                                                                                       
usage: /Users/jlipps/Code/appium/appium driver [-h]
                                               {list,install,uninstall,update}
                                               ...
/Users/jlipps/Code/appium/appium driver: error: argument "driverCommand": Invalid choice: foo (choose from [list, install, uninstall, update])
```

lib/driver-cli.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+async function list ({json, installed, showUpdates}) {
+  console.log(`Listing ${installed ? 'installed' : 'available'} drivers`);
+  const drivers = {};
+
+  for (const name of Object.keys(INSTALLED_DRIVERS)) {
+    drivers[name] = INSTALLED_DRIVERS[name];
+    drivers[name].installed = true;
```
mykola-mokhnach: why do we mutate things inside `list` method?

jlipps: you're right, there should be a defensive copy made

lib/driver-cli.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  if (showUpdates) {
+    // TODO
+  }
+  if (!json) {
+    for (const name of Object.keys(drivers)) {
+      const installTxt = drivers[name].installed ?
+        `@${drivers[name].version} [installed]` :
+        ' [not installed]';
```
mykola-mokhnach: do we also need a space at the start?

lib/driver-cli.js
```diff
@@ -0,0 +1,91 @@
...
+
+      console.log(`- ${name}${installTxt}`);
+    }
+  }
+  return drivers;
+}
+
+async function install ({json, driver, npm, github, git}) {
+  return {};
```
mykola-mokhnach: Missing TODOs

lib/driver-cli.js
jlipps: no, so that the output looks like `xcuitest@1.2.3 [installed]` or `xcuitest [not installed]`

lib/driver-parser.js
```diff
@@ -0,0 +1,119 @@
...
+    addHelp: true,
+    help: 'Access the driver management CLI commands'
+  });
+  const driverSubParsers = driverParser.addSubparsers({dest: 'driverCommand'});
+
+  const listParser = driverSubParsers.addParser('list', {
+    addHelp: true,
+    help: 'List available and installed drivers'
+  });
+  for (const arg of listArgs) {
```
mykola-mokhnach: we could deconstruct `arg` here

mykola-mokhnach: same below

lib/driver-parser.js
```diff
@@ -0,0 +1,119 @@
...
+
+  const uninstallParser = driverSubParsers.addParser('uninstall', {
+    addHelp: true,
+    help: 'Uninstall a driver'
+  });
+  for (const arg of uninstallArgs) {
+    uninstallParser.addArgument(arg[0], arg[1]);
+  }
+
+  const updateParser = driverSubParsers.addParser('update', {
```
mykola-mokhnach: Can we refactor this stuff? I observe a lot of duplication

package.json
```diff
@@ -44,6 +44,7 @@
     "@babel/runtime": "^7.6.0",
     "appium-base-driver": "^4.1.0",
     "appium-support": "2.x",
+    "appium-xcuitest-driver": "^3.2.0",
```
mykola-mokhnach: is this needed?

jlipps: no, it got saved by accident

lib/driver-parser.js
jlipps: 👍

lib/driver-parser.js
jlipps: 👍

jlipps:
The other difficulty so far with this approach is that argparse does not allow optional positional arguments. So to support `appium driver ...` means that we cannot have a gap in the positional arguments for the normal server running use case. i.e., `appium` by itself will throw a missing argument error. The way I have fixed that right now is to run the server via `appium server --arg1 --arg2 ...`. But this is a pretty big breaking change and I imagine people would get confused about having to run `appium server` instead of just `appium`. I thought I would peek at argv and use the following algorithm:

* if argv has only length 2 (node, appium), then add another item (node, appium, server)
* if argv has length > 2, and if argv[2] is not `driver` or `server` or `-h`, splice in `server` as argv[2]

I think this will solve the problem?

jlipps:
> I think this will solve the problem?

I implemented it and seems to work intuitively.

jlipps:
uh, the wrong branch got pushed :-(

PR #13463, opened 2019-10-21
Basic driver management for Appium 2.0

jlipps:
Initial work on Appium 2.0. Taking over from https://github.com/appium/appium/pull/13461. When merged, this will merely go into the appium/appium `2.0` branch (not master).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ sample-code/csharp/test/obj
 sample-code/csharp/test/bin
 sample-code/csharp/packages
 appium.zip
+.drivers.json
--- a/lib/cli-utils.js
+++ b/lib/cli-utils.js
@@ -1,0 +1,68 @@
+/* eslint-disable no-console */
+
+function errAndQuit (json, msg) {
+  if (json) {
+    console.log(JSON.stringify({error: msg}));
+  } else {
+    console.error(msg.red);
+  }
+  process.exit(1);
+}
+
+function log (json, msg) {
+  !json && console.log(msg);
+}
+
+function startSpinner (msg) {
+  process.stdout.write(msg);
+
+  let count = 1;
+  const numDots = 5;
+  const dotTime = 500;
+  const cursorStart = msg.length;
+
+  const interval = setInterval(() => {
+    if (count % (numDots + 1) === 0) {
+      process.stdout.cursorTo(cursorStart);
+      for (let i = 0; i < numDots; i++) {
+        process.stdout.write(' ');
+      }
+      process.stdout.cursorTo(cursorStart);
+    } else {
+      process.stdout.write('.');
+    }
+    count++;
+  }, dotTime);
+
+  return () => {
+    clearInterval(interval);
+    process.stdout.write('\n');
+  };
+}
+
+function logSpin (json, msg) {
+  if (json) {
+    return () => {};
+  }
+
+  return startSpinner(msg);
+}
+
+async function spinWith (json, msg, fn) {
+  const stopSpinner = logSpin(json, msg);
+  let res;
+  try {
+    res = await fn();
+  } finally {
+    stopSpinner();
+  }
+  return res;
+}
+
+export {
+  errAndQuit,
+  log,
+  startSpinner,
+  spinWith,
+  logSpin
+};
--- a/lib/cli-utils.js
+++ b/lib/cli-utils.js
@@ -2,9 +2,9 @@
 
 function errAndQuit (json, msg) {
   if (json) {
-    console.log(JSON.stringify({error: msg}));
+    console.log(JSON.stringify({error: msg.toString()}));
   } else {
-    console.error(msg.red);
+    console.error(msg.toString().red);
   }
   process.exit(1);
 }
--- a/lib/driver-cli.js
+++ b/lib/driver-cli.js
@@ -123,9 +123,9 @@ async function install ({appiumHome, json, driver, installType}) {
 
   if (installType === INSTALL_TYPE_LOCAL) {
     const msg = 'Linking driver from local path';
-    const pkgJsonData = await spinWith(json, msg, async () => {
-      return await linkPackage(appiumHome, installSpec);
-    });
+    const pkgJsonData = await spinWith(json, msg, async () => (
+      await linkPackage(appiumHome, installSpec))
+    );
     driverData = getDriverFields(pkgJsonData);
   } else if (installType === INSTALL_TYPE_GITHUB) {
     if (installSpec.split('/').length !== 2) {
@@ -198,9 +198,9 @@ async function installViaNpm ({appiumHome, json, driver, pkgName, pkgVer}) {
   const specMsg = npmSpec === driver ? '' : ` using NPM install spec '${npmSpec}'`;
   const msg = `Installing '${driver}'${specMsg}`;
   try {
-    const pkgJsonData = await spinWith(json, msg, async () => {
-      return await installPackage(appiumHome, pkgName, pkgVer);
-    });
+    const pkgJsonData = await spinWith(json, msg, async () => (
+      await installPackage(appiumHome, pkgName, pkgVer)
+    ));
     return getDriverFields(pkgJsonData);
   } catch (err) {
     throw new Error(`Encountered an error when installing package: ${err.message}`);
@@ -217,12 +217,9 @@ function getDriverFields (pkgJsonData) {
   const {driverName, automationName, platformNames, mainClass} = appium;
 
   const reqFields = ['driverName', 'automationName', 'platformNames', 'mainClass'];
-  const missingFields = [];
-  for (const reqField of reqFields) {
-    if (!appium[reqField]) {
-      missingFields.push(reqField);
-    }
-  }
+  const missingFields = reqFields.reduce((acc, field) => (
+    appium[field] ? acc : acc.concat(field)
+  ), []);
 
   if (!_.isEmpty(missingFields)) {
     throw new Error(`Installed driver did not expose correct fields for compability ` +
--- a/lib/driver-cli.js
+++ b/lib/driver-cli.js
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import path from 'path';
 import { KNOWN_DRIVERS, validateDriversYaml, writeInstalledDriver,
          INSTALL_TYPE_LOCAL, INSTALL_TYPE_NPM, INSTALL_TYPE_GIT,
          INSTALL_TYPE_GITHUB } from './drivers';
@@ -127,6 +128,7 @@ async function install ({appiumHome, json, driver, installType}) {
       await linkPackage(appiumHome, installSpec))
     );
     driverData = getDriverFields(pkgJsonData);
+    driverData.installPath = driverData.pkgName;
   } else if (installType === INSTALL_TYPE_GITHUB) {
     if (installSpec.split('/').length !== 2) {
       throw new Error(`Github driver spec ${installSpec} appeared to be invalid; ` +
@@ -191,17 +193,21 @@ async function install ({appiumHome, json, driver, installType}) {
 }
 
 async function installViaNpm ({appiumHome, json, driver, pkgName, pkgVer}) {
+  let installPath = pkgName;
   if (!pkgName) {
     pkgName = driver;
+    installPath = path.posix.basename(pkgName);
   }
   const npmSpec = `${pkgName}${pkgVer ? '@' + pkgVer : ''}`;
   const specMsg = npmSpec === driver ? '' : ` using NPM install spec '${npmSpec}'`;
   const msg = `Installing '${driver}'${specMsg}`;
   try {
     const pkgJsonData = await spinWith(json, msg, async () => (
-      await installPackage(appiumHome, pkgName, pkgVer)
+      await installPackage(path.resolve(appiumHome, installPath), pkgName, pkgVer)
     ));
-    return getDriverFields(pkgJsonData);
+    const driverData = getDriverFields(pkgJsonData);
+    driverData.installPath = installPath;
+    return driverData;
   } catch (err) {
     throw new Error(`Encountered an error when installing package: ${err.message}`);
   }
--- a/lib/driver-cli.js
+++ b/lib/driver-cli.js
@@ -1,0 +1,221 @@
+import _ from 'lodash';
+import { KNOWN_DRIVERS, validateDriversJson, writeInstalledDriver,
+         INSTALL_TYPE_LOCAL, INSTALL_TYPE_NPM, INSTALL_TYPE_GIT,
+         INSTALL_TYPE_GITHUB } from './drivers';
+import { getLatestVersion, installPackage, linkPackage } from './npm';
+import { errAndQuit, log, spinWith } from './cli-utils';
+import { util } from 'appium-support';
+
+let INSTALLED_DRIVERS;
+
+async function runDriverCommand (args) {
+  // TODO driver.json file should be locked while any of these commands are
+  // running to prevent weird situations
+  let jsonResult = null;
+  const logFn = (msg) => log(args.json, msg);
+  try {
+    INSTALLED_DRIVERS = validateDriversJson(logFn);
+    switch (args.driverCommand) {
+      case 'list':
+        jsonResult = await list(args);
+        break;
+      case 'install':
+        jsonResult = await install(args);
+        break;
+      case 'uninstall':
+        jsonResult = await uninstall(args);
+        break;
+      case 'update':
+        jsonResult = await update(args);
+        break;
+      default:
+        throw new Error(`Cannot handle driver command ${args.driverCommand}`);
+    }
+  } catch (err) {
+    errAndQuit(args.json, err.message);
+  }
+
+  if (args.json) {
+    console.log(JSON.stringify(jsonResult)); // eslint-disable-line no-console
+  }
+}
+
+async function list ({json, showInstalled, showUpdates}) {
+  const lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} drivers`;
+  const drivers = {};
+
+  for (const name of Object.keys(INSTALLED_DRIVERS)) {
+    drivers[name] = INSTALLED_DRIVERS[name];
+    drivers[name].installed = true;
+  }
+
+  if (!showInstalled) {
+    for (const name of Object.keys(KNOWN_DRIVERS)) {
+      if (!drivers[name]) {
+        drivers[name] = {
+          pkgName: KNOWN_DRIVERS[name],
+          installed: false,
+        };
+      }
+    }
+  }
+
+  await spinWith(json, lsMsg, async () => {
+    if (showUpdates) {
+      for (const name of Object.keys(drivers)) {
+        const {version, pkgName, installType} = drivers[name];
+        drivers[name].updateVersion = null;
+        if (installType === INSTALL_TYPE_NPM) {
+          const latestVersion = await getLatestVersion(pkgName);
+          if (!version || util.compareVersions(latestVersion, '>', version)) {
+            drivers[name].updateVersion = latestVersion;
+          }
+        }
+      }
+    }
+  });
+
+  if (!json) {
+    for (const name of Object.keys(drivers)) {
+      const {installType, installSpec, installed, updateVersion, version} = drivers[name];
+      let typeTxt;
+      switch (installType) {
+        case INSTALL_TYPE_GIT:
+        case INSTALL_TYPE_GITHUB:
+          typeTxt = `(cloned from ${installSpec})`.yellow;
+          break;
+        case INSTALL_TYPE_LOCAL:
+          typeTxt = `(linked from ${installSpec})`.magenta;
+          break;
+        default:
+          typeTxt = '(NPM)';
+      }
+      const installTxt = installed ?
+        `@${version.yellow} ${('[installed ' + typeTxt + ']').green}` :
+        ' [not installed]'.grey;
+      const updateTxt = showUpdates && updateVersion ?
+        ` [${updateVersion} available]`.magenta :
+        '';
+
+      console.log(`- ${name.yellow}${installTxt}${updateTxt}`); // eslint-disable-line no-console
+    }
+  }
+  return drivers;
+}
+
+async function install ({json, driver, local, npm, github, git}) {
+  log(json, `Attempting to find and install driver '${driver}'`);
+
+  let driverData, installType, installSpec = driver;
+
+  if (local) {
+    installType = INSTALL_TYPE_LOCAL;
+    const msg = 'Linking driver from local path';
+    const pkgJsonData = await spinWith(json, msg, async () => {
+      return await linkPackage(installSpec);
+    });
+    driverData = getDriverFields(pkgJsonData);
+  } else if (github) {
+    installType = INSTALL_TYPE_GITHUB;
+    if (installSpec.split('/').length !== 2) {
+      throw new Error(`Github driver spec ${installSpec} appeared to be invalid; ` +
+                      'it should be of the form <org>/<repo>');
+    }
+    driverData = await installViaNpm(json, installSpec);
+  } else if (git) {
+    installType = INSTALL_TYPE_GIT;
+    driverData = await installViaNpm(json, installSpec);
+  } else {
+    installType = INSTALL_TYPE_NPM;
+
+    // drivers installed via npm can include versions or tags after the '@'
+    // sign, so check for that
+    const [name, pkgVer] = installSpec.split('@');
+    let pkgName;
+
+    if (npm) {
+      // if we're installing a named package from npm, we don't need to check
+      // against the appium driver list; just use the installSpec as is
+      pkgName = name;
+    } else {
+      // if we're installing a named appium driver (like 'xcuitest') we need to
+      // dereference the actual npm package ('appiupm-xcuitest-driver'), so
+      // check it exists and get the correct package
+      const knownNames = Object.keys(KNOWN_DRIVERS);
+      if (!_.includes(knownNames, name)) {
+        const msg = `Could not resolve driver; are you sure it's in the list ` +
+                    `of supported drivers? ${JSON.stringify(knownNames)}`;
+        throw new Error(msg);
+      }
+      pkgName = KNOWN_DRIVERS[name];
+    }
+
+    driverData = await installViaNpm(json, driver, pkgName, pkgVer);
+  }
+
+  if (!driverData.driverName || !driverData.version || !driverData.automationName ||
+      !driverData.platformNames || !driverData.mainClass) {
+    throw new Error(`Installed driver did not expose correct fields for ` +
+                    `compability with Appium.`);
+  }
+
+  const {driverName} = driverData;
+  delete driverData.driverName;
+
+  if (_.includes(Object.keys(INSTALLED_DRIVERS), driverName)) {
+    throw new Error(`A driver named '${driverName}' is already installed. ` +
+                    `Did you mean to update? 'appium driver update'. See ` +
+                    `installed drivers with 'appium driver list --installed'.`);
+  }
+
+  driverData.installType = installType;
+  driverData.installSpec = installSpec;
+  await writeInstalledDriver(driverName, driverData);
+
+  // refresh the installed driver manifest
+  INSTALLED_DRIVERS = validateDriversJson((msg) => log(json, msg));
+
+  // log info for the user
+  log(json, `Driver ${driverName}@${driverData.version} successfully installed`.green);
+  log(json, `- automationName: ${driverData.automationName.green}`);
+  log(json, `- platformNames: ${JSON.stringify(driverData.platformNames).green}`);
+
+  return INSTALLED_DRIVERS;
+}
+
+async function installViaNpm (json, driver, pkgName, pkgVer) {
+  if (!pkgName) {
+    pkgName = driver;
+  }
+  const npmSpec = `${pkgName}${pkgVer ? '@' + pkgVer : ''}`;
+  const specMsg = npmSpec === driver ? '' : ` using NPM install spec '${npmSpec}'`;
+  const msg = `Installing '${driver}'${specMsg}`;
+  try {
+    const pkgJsonData = await spinWith(json, msg, async () => {
+      return await installPackage(pkgName, pkgVer);
+    });
+    return getDriverFields(pkgJsonData);
+  } catch (err) {
+    throw new Error(`Encountered an error when installing package: ${err.message}`);
+  }
+}
+
+function getDriverFields (pkgJsonData) {
+  const {appium, name, version} = pkgJsonData;
+  const {driverName, automationName, platformNames, mainClass} = appium;
+  return {version, pkgName: name, driverName, automationName, platformNames, mainClass};
+}
+
+function uninstall (/*{json, driver}*/) {
+  // TODO implementation
+  return {};
+}
+
+function update (/*{json, driver}*/) {
+  // TODO implementation
+  return {};
+}
+
+export {
+  runDriverCommand
+};
--- a/lib/driver-cli.js
+++ b/lib/driver-cli.js
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import { KNOWN_DRIVERS, validateDriversJson, writeInstalledDriver,
+import { KNOWN_DRIVERS, validateDriversYaml, writeInstalledDriver,
          INSTALL_TYPE_LOCAL, INSTALL_TYPE_NPM, INSTALL_TYPE_GIT,
          INSTALL_TYPE_GITHUB } from './drivers';
 import { getLatestVersion, installPackage, linkPackage } from './npm';
@@ -12,9 +12,10 @@ async function runDriverCommand (args) {
   // TODO driver.json file should be locked while any of these commands are
   // running to prevent weird situations
   let jsonResult = null;
-  const logFn = (msg) => log(args.json, msg);
+  const {json, appiumHome} = args;
+  const logFn = (msg) => log(json, msg);
   try {
-    INSTALLED_DRIVERS = validateDriversJson(logFn);
+    INSTALLED_DRIVERS = await validateDriversYaml(appiumHome, logFn);
     switch (args.driverCommand) {
       case 'list':
         jsonResult = await list(args);
@@ -28,19 +29,26 @@ async function runDriverCommand (args) {
       case 'update':
         jsonResult = await update(args);
         break;
+      case 'verify':
+        // TODO create a 'verify' command that just checks driver.json and then
+        // that everything in driver.json can be required OK.
+        // verify could also have a --reinstall flag which installs any drivers
+        // which are listed in the drivers.json but which cannot be required
+        // for whatever reason
+        break;
       default:
         throw new Error(`Cannot handle driver command ${args.driverCommand}`);
     }
   } catch (err) {
-    errAndQuit(args.json, err.message);
+    errAndQuit(json, err);
   }
 
-  if (args.json) {
+  if (json) {
     console.log(JSON.stringify(jsonResult)); // eslint-disable-line no-console
   }
 }
 
-async function list ({json, showInstalled, showUpdates}) {
+async function list ({appiumHome, json, showInstalled, showUpdates}) {
   const lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} drivers`;
   const drivers = {};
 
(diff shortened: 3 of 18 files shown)


jlipps:
@mykola-mokhnach wanted to give you a chance to review the initial code as well.

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
...
+  tizen: 'appium-tizen-driver',
+};
+
+function validateDriversJson () {
+  try {
+    INSTALLED_DRIVERS = require(DRIVERS_JSON);
+  } catch (err) {
+    throw new Error(`Appium had trouble loading the driver installation ` +
+                    `cache file (${DRIVERS_JSON}). Ensure it exists and is ` +
+                    `readable. Specific error: ${err}`);
```
mykola-mokhnach: err.message

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
...
+  }
+  const automationNames = [];
+  let msg = '';
+  for (const driverName of Object.keys(INSTALLED_DRIVERS)) {
+    const platformNames = INSTALLED_DRIVERS[driverName].platformNames;
+    if (!_.isArray(platformNames)) {
+      msg += `Drivers.json entry for '${driverName}' has missing or ` +
```
mykola-mokhnach: file name can be a constant

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
...
+              `${msg}Please ensure you have installed drivers correctly ` +
+              `using the Appium CLI.`);
+  }
+
+  return INSTALLED_DRIVERS;
+}
+
+function printAvailableDrivers () {
+  const driverNames = Object.keys(INSTALLED_DRIVERS);
+  if (!driverNames.length) {
```
mykola-mokhnach: _.isEmpty

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
...
+    const installed = INSTALLED_DRIVERS[driverName];
+    const aNameMatches = installed.automationName.toLowerCase() ===
+                         automationName.toLowerCase();
+    const lower = (x) => x.toLowerCase();
+    const pNameMatches = _.includes(installed.platformNames.map(lower),
+                                    platformName.toLowerCase());
+
+    if (aNameMatches && pNameMatches) {
+      return {driverName, ...installed};
+    } else if (aNameMatches) {
```
mykola-mokhnach: else is redundant

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
...
+    log.info(`  - ${driverName}@${ver} (automationName '${aName}')`);
+  }
+}
+
+function getDriverBySupport (automationName, platformName) {
+  for (const driverName of Object.keys(INSTALLED_DRIVERS)) {
+    const installed = INSTALLED_DRIVERS[driverName];
+    const aNameMatches = installed.automationName.toLowerCase() ===
+                         automationName.toLowerCase();
+    const lower = (x) => x.toLowerCase();
```
mykola-mokhnach: use _.toLower()

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
...
+    const ver = INSTALLED_DRIVERS[driverName].version;
+    const aName = INSTALLED_DRIVERS[driverName].automationName;
+    log.info(`  - ${driverName}@${ver} (automationName '${aName}')`);
+  }
+}
+
+function getDriverBySupport (automationName, platformName) {
+  for (const driverName of Object.keys(INSTALLED_DRIVERS)) {
+    const installed = INSTALLED_DRIVERS[driverName];
+    const aNameMatches = installed.automationName.toLowerCase() ===
```
mykola-mokhnach: is automationName always set?

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
...
+
+  if (!_.isString(automationName)) {
+    throw new Error('You must include an automationName capability');
+  }
+
+  log.info(`Attempting to find matching driver for automationName ` +
+           `'${automationName}' and platformName '${platformName}'`);
+
+  try {
+    const {driverName, mainClass, pkg, version} = getDriverBySupport(automationName, platformName);
```
mykola-mokhnach: what the function name means (BySupport) ?

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
...
+    log.info(`The '${driverName}' driver was installed and matched caps.`);
+    const driver = require(pkg)[mainClass];
+    return {driver, version};
+  } catch (err) {
+    const msg = `Could not find a driver for automationName ` +
+                `'${automationName}' and platformName ${platformName}'. ` +
+                `Have you installed a driver that supports those ` +
+                `capabilities? Run 'appium driver list --installed' to see. ` +
+                `(Lower-level error: ${err})`;
```
mykola-mokhnach: err.message

lib/main.js
```diff
@@ -90,6 +93,19 @@ function logServerPort (address, port) {
...
+function fixArgvForParser () {
+  // this function mutates argv to ensure that argparse gets the 'server'
+  // subcommand in the case where the user provides no subcommand
+  if (process.argv.length < 2) {
+    process.argv.push('server');
+    return;
+  }
+
+  if (!_.includes(['driver', 'server', '-h'], process.argv[2])) {
+    process.argv.splice(2, 0, 'server');
```
mykola-mokhnach: is it necessary to mutate process.argv ?

jlipps: yes, argparse reads global process.argv. We could re-set it after running the parser to be on the safe side though.

lib/drivers.js
```diff
@@ -0,0 +1,157 @@
+import path from 'path';
+import _ from 'lodash';
+import log from './logger';
+
+// TODO the location of the driver data file should be configurable
+const DRIVERS_JSON = path.resolve(__dirname, '..', '..', '.drivers.json');
```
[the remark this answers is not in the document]
jlipps: I was thinking about this. If someone reinstalls appium, then all the drivers need to also be reinstalled. But if .drivers.json lives in the home folder, it will be out of sync. Thoughts on this design?

lib/drivers.js
jlipps: get the driver based on its supported automation name and platform names. kinda long-winded :-)

lib/drivers.js
jlipps: Yes, we error out earlier in the flow if it's not set.

lib/drivers.js
jlipps: 👍

jlipps:
Another design question: currently, there are a lot of appium server flags that really only pertain to specific drivers. For example `chromedriverExecutablePath` is not relevant to the other drivers.

For some of these, it's ok to just remove them, because users can set them via desired capabilities instead.

But some of them would be a security risk as a desired cap (like being able to execute any binary on the system with the cap above), so it should really be set when the server is started by the system admin. Here are some options I've thought of for how to deal with this:

1. add a new server arg `--driver-args` that allows a JSON object or file to be defined, something like `{"uiautomator2": {"chromedriverExecutablePath": "/usr/bin/chromedriver"}}`. Then Appium will pass these args along if that driver is instantiated.
2. Each driver could export its own custom argparse args, and when Appium starts up, it first checks installed drivers, and retrieves their custom server args, and merges them in with its own args. So `--chromedriverExecutablePath` would still be available as a server arg for anyone who has installed the uiautomator2 driver.

Other ideas?

lib/drivers.js
jlipps: actually, it's not redundant. it's handling the case where `aNameMatches && !pNameMatches`

lib/drivers.js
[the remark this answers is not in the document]
jlipps: When the Appium server starts, should it verify that what's in .drivers.json matches what's installed in node_modules? We could then have two new driver commands which are mentioned in an error message:

```
appium driver verify  # prints out errors if anything in .drivers.json is not actually installed
appium driver reinstall  # reinstalls any that don't exist
```

On this kind of model it would make sense to have .drivers.json in $HOME, maybe should be called `.appium-drivers.json`, or we have a dir `.appium/drivers.json`, in case we need more stuff there in the future.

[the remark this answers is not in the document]
mykola-mokhnach: in case with Appium Desktop we usually cannot write into its installation folder

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+  } catch (err) {
+    errAndQuit(args.json, err.message);
+  }
+
+  if (args.json) {
+    console.log(JSON.stringify(jsonResult));
+  }
+}
+
+function startSpinner (msg) {
```
mykola-mokhnach: I'd rather put these helpers into separate module

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+      if (installType === INSTALL_TYPE_NPM) {
+        const latestVersion = await getLatestVersion(pkgName);
+        if (!version || semver.gt(latestVersion, version)) {
+          drivers[name].updateVersion = latestVersion;
+        }
+      }
+    }
+  }
+
+  stopSpinner();
```
mykola-mokhnach: should this be called in finally block?

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+    }
+  }
+
+  if (showUpdates) {
+    for (const name of Object.keys(drivers)) {
+      const {version, pkgName, installType} = drivers[name];
+      drivers[name].updateVersion = null;
+      if (installType === INSTALL_TYPE_NPM) {
+        const latestVersion = await getLatestVersion(pkgName);
+        if (!version || semver.gt(latestVersion, version)) {
```
mykola-mokhnach: use appium-support.util.compareVersions

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+  return startSpinner(msg);
+}
+
+async function list ({json, showInstalled, showUpdates}) {
+  const lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} drivers`;
+  const stopSpinner = logSpin(json, lsMsg);
+
+  const drivers = {};
+
+  for (const name of Object.keys({...INSTALLED_DRIVERS})) {
```
mykola-mokhnach: what is the purpose of `{...INSTALLED_DRIVERS}` ?

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+
+function logSpin (json, msg) {
+  if (json) {
+    return () => {};
+  }
+
+  return startSpinner(msg);
+}
+
+async function list ({json, showInstalled, showUpdates}) {
```
mykola-mokhnach: I would say we need file-based locking mechanism to prevent parallel driver installation, like npm or apt does

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+async function install ({json, driver, local, npm, github, git}) {
+  log(json, `Attempting to find and install driver '${driver}'`);
+
+  let driverData, installType, installSpec = driver;
+
+  if (local) {
+    installType = INSTALL_TYPE_LOCAL;
+    const stopSpinner = logSpin(json, 'Linking driver from local path');
+    const pkgJsonData = await linkPackage(installSpec);
+    stopSpinner();
```
mykola-mokhnach: finally

lib/npm.js
```diff
@@ -0,0 +1,102 @@
...
+
+const APPIUM_ROOT = path.resolve(__dirname, '..', '..');
+
+async function npmExec (cmd, args, {json} = {json: false}, execOpts = {}) {
+  args.unshift(cmd);
+  if (json) {
+    args.push('-json');
+  }
+  execOpts.cwd = APPIUM_ROOT;
+  const {stdout, stderr, code} = await exec('npm', args, execOpts);
```
mykola-mokhnach: npm is a batch script on windows, so this won't work there

lib/drivers.js
mykola-mokhnach: you can keep `if (aNameMatches)`. I was only talking about `else`

lib/drivers.js
```diff
@@ -0,0 +1,198 @@
...
+              `${msg}Please ensure you have installed drivers correctly ` +
+              `using the Appium CLI.`);
+  }
+
+  return INSTALLED_DRIVERS;
+}
+
+async function writeInstalledDriver (name, data) {
+  INSTALLED_DRIVERS[name] = data;
+  await fs.writeFile(DRIVERS_JSON, JSON.stringify(INSTALLED_DRIVERS));
```
mykola-mokhnach: should we also set encoding explicitly?

lib/drivers.js
```diff
@@ -0,0 +1,198 @@
...
+function printAvailableDrivers () {
+  const driverNames = Object.keys(INSTALLED_DRIVERS);
+  if (_.isEmpty(driverNames)) {
+    log.info('No drivers have been installed. Use the "appium driver" ' +
+             'command to install the one(s) you want to use.');
+    return;
+  }
+
+  log.info('Available drivers:');
+  for (const driverName of Object.keys(INSTALLED_DRIVERS)) {
```
mykola-mokhnach: why not deconstruct directly here?

lib/drivers.js
```diff
@@ -0,0 +1,198 @@
...
+    const ver = INSTALLED_DRIVERS[driverName].version;
+    const aName = INSTALLED_DRIVERS[driverName].automationName;
+    log.info(`  - ${driverName}@${ver} (automationName '${aName}')`);
+  }
+}
+
+function getDriverBySupport (automationName, platformName) {
+  for (const driverName of Object.keys(INSTALLED_DRIVERS)) {
```
mykola-mokhnach: why not deconstruct here?

lib/drivers.js
```diff
@@ -0,0 +1,198 @@
...
+    // if the file just doesn't exist, call it empty
+    return {};
+  }
+
+  const automationNames = [];
+  let msg = '';
+  for (const driverName of Object.keys(INSTALLED_DRIVERS)) {
+    const platformNames = INSTALLED_DRIVERS[driverName].platformNames;
+    if (!_.isArray(platformNames)) {
+      msg += `Entry for '${driverName}' has missing or ` +
```
mykola-mokhnach: I would also print the actual values in the error message

lib/drivers.js
```diff
@@ -0,0 +1,198 @@
...
+  INSTALL_TYPE_GITHUB,
+  INSTALL_TYPE_LOCAL,
+  INSTALL_TYPE_NPM
+];
+
+function validateDriversJson () {
+  try {
+    INSTALLED_DRIVERS = require(DRIVERS_JSON);
+  } catch (err) {
+    if (existsSync(DRIVERS_JSON)) {
```
mykola-mokhnach: what is the reason to not use the async implementation?

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+    console.error(msg.red);
+  }
+  process.exit(1);
+}
+
+function log (json, msg) {
+  !json && console.log(msg);
+}
+
+async function runDriverCommand (args) {
```
mykola-mokhnach: I would be happy if functions could have docstrings with parameters and returns types descriptions

lib/driver-cli.js
mykola-mokhnach: same below

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+
+  // log info for the user
+  log(json, `Driver ${driverName}@${driverData.version} successfully installed`.green);
+  log(json, `- automationName: ${driverData.automationName.green}`);
+  log(json, `- platformNames: ${JSON.stringify(driverData.platformNames).green}`);
+
+  return INSTALLED_DRIVERS;
+}
+
+async function installViaNpm (json, driver, pkgName, pkgVer) {
```
mykola-mokhnach: pkgVer can be undefined though

mykola-mokhnach: also, this function has too many args. Consider packing them into an object

lib/npm.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  return ret;
+}
+
+async function getLatestVersion (pkg) {
+  return (await npmExec('view', [pkg, 'dist-tags'], {json: true})).json.latest;
+}
+
+async function installPackage (pkg, ver) {
```
mykola-mokhnach: ver can be null/undefined

lib/npm.js
```diff
@@ -0,0 +1,102 @@
...
+    // we parsed a valid json response, so if we got an error here, return that
+    // message straightaway
+    if (res.json.error) {
+      throw new Error(res.json.error);
+    }
+
+    // if we tried to install a package via git/github, we might not have been
+    // certain of the package name, so if we can retrieve it unambiguously from
+    // the json report of the install, do so
+    const names = _.uniq(res.json.added.concat(res.json.updated).map(x => x.name));
```
mykola-mokhnach: use `[...arr1, ...arr2]` format for concatenation

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+  }, dotTime);
+
+  return () => {
+    clearInterval(interval);
+    process.stdout.write('\n');
+  };
+}
+
+function logSpin (json, msg) {
+  if (json) {
```
mykola-mokhnach: ternary operator would be shorter

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+        const latestVersion = await getLatestVersion(pkgName);
+        if (!version || semver.gt(latestVersion, version)) {
+          drivers[name].updateVersion = latestVersion;
+        }
+      }
+    }
+  }
+
+  stopSpinner();
+  if (!json) {
```
mykola-mokhnach: if json return drivers

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+      }
+      pkgName = KNOWN_DRIVERS[name];
+    }
+
+    driverData = await installViaNpm(json, driver, pkgName, pkgVer);
+  }
+
+  if (!driverData.driverName || !driverData.version || !driverData.automationName ||
+      !driverData.platformNames || !driverData.mainClass) {
+    throw new Error(`Installed driver did not expose correct fields for ` +
```
mykola-mokhnach: I'd rather explicitly say which field(s) are missing

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+  log(json, `- platformNames: ${JSON.stringify(driverData.platformNames).green}`);
+
+  return INSTALLED_DRIVERS;
+}
+
+async function installViaNpm (json, driver, pkgName, pkgVer) {
+  if (!pkgName) {
+    pkgName = driver;
+  }
+  const npmSpec = `${pkgName}${pkgVer ? '@' + pkgVer : ''}`;
```
mykola-mokhnach: npm also supports special identifiers, like `latest` or `beta`. Are they properly handled?

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+  const specMsg = npmSpec === driver ? '' : ` using NPM install spec '${npmSpec}'`;
+  const msg = `Installing '${driver}'${specMsg}`;
+  const stopSpinner = logSpin(json, msg);
+  try {
+    const pkgJsonData = await installPackage(pkgName, pkgVer);
+    stopSpinner();
```
mykola-mokhnach: should be in finally

lib/driver-cli.js
```diff
@@ -0,0 +1,265 @@
...
+}
+
+function log (json, msg) {
+  !json && console.log(msg);
+}
+
+async function runDriverCommand (args) {
+  let jsonResult = null;
+  try {
+    INSTALLED_DRIVERS = validateDriversJson();
```
mykola-mokhnach: we probably would need some command, like reset, which is going to recreate/resync drivers json if it was damaged/edited manually. With the current approach there is no way to launch Appium if this config is damaged

lib/main.js
```diff
@@ -27,6 +29,7 @@ async function preflightChecks (parser, args, throwInsteadOfExit = false) {
     }
     warnNodeDeprecations();
     validateServerArgs(parser, args);
+    validateDriversJson();
```
mykola-mokhnach: I think this verification must be forgivable

lib/drivers.js
jlipps: oh i see 👍

lib/main.js
jlipps: I found a different solution that involves overriding the parseArgs method. Yay for no mutation.

lib/drivers.js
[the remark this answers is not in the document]
jlipps: Yeah, good point. We should have a `.appium` dir which can host `.appium/drivers.json` and `.appium/plugins.json` and maybe the driver npm modules themselves. (see forthcoming comment)

lib/driver-cli.js
jlipps: ah yes, good point.

jlipps: in practice, every error thrown ends up halting the process, so i think it's fine. but finally is much safer/cleaner.

lib/driver-cli.js
jlipps: 👍

lib/driver-cli.js
jlipps: 👍

lib/driver-cli.js
jlipps: I thought I was protecting against mutation of INSTALLED_DRIVERS. but i wasn't.

lib/driver-cli.js
jlipps: you mean lock the drivers.json file during operations?

lib/npm.js
jlipps: i'm able to run commands on my windows machine just fine using `npm`

lib/drivers.js
jlipps: 👍

lib/drivers.js
jlipps: `Object.keys(INSTALLED_DRIVERS)` is an array, not an object.

lib/drivers.js
jlipps: going to refactor this whole bit so it's cleaner and can allow validation to proceed when some drivers are valid and some aren't.

lib/driver-cli.js
mykola-mokhnach: not only that. There might be a mess when two parallel appium instances are trying to preform changes to the npm registry

lib/drivers.js
mykola-mokhnach: I mean you can iterate `_.toPairs()` and then deconstruct it to `[driverName, {version, automationName}]`

lib/driver-cli.js
jlipps: I promise to add them once the actual methods and parameters are approved.

lib/driver-cli.js
jlipps: Yes, any npm identifier will work here. We just pass it straight from the command line.

lib/drivers.js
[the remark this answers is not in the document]
jlipps: @mykola-mokhnach I had the thought that we could just keep all the installed modules in a location separate from appium, and then simply `require` that full path rather than the bare module name. This is also nice because when people reinstall Appium, their drivers don't get blown away.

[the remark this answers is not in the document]
jlipps: This seems to work well in practice!

lib/driver-cli.js
jlipps: i've added a todo for this, my thought was as follows:

```
# verify driver.json and that every driver referred to there can be required OK
# just print out the status
appium driver verify

# perform verification but then reinstall anything which couldn't be required
appium driver verify --reinstall
```

mykola-mokhnach:
I'm not quite sure how much we need that, but usually one adds the actual schema version number into configs, so we can easily alert the schema and still handle it properly in never appium versions and/or perform config migration. This touches both drivers config and the config we add into package.json for each driver

lib/driver-cli.js
mykola-mokhnach: would it fix a problem with damaged config?

jlipps: what kind of config problems could be fixable automatically? here are the kinds of problems i can think of:

1. driver is mentioned in config but is not npm installed. this case would be fixed with `--reinstall`
2. driver is npm installed but not mentioned in config. the only way to fix this automatically would be to require the package.json of every module in the appium home dir, and check for appium driver metadata. this seems ... slow. also, what if it was npm linked in instead of installed? or simply copied in? there's no way to guarantee that a package in node_modules can be re-installed via NPM.
3. config for driver has a missing key. how could we automatically generate the value for a missing key?
4. config for driver has an invalid value for a key. how could we decide how to fix it automatically?
5. config for driver has an additional key beyond the valid ones. this could be removed automatically, but what's the point?

Can you think of any other "damaged" cases and how we might fix them? I can't, which is why it seemed a simple reinstall of missing modules is the best we can do.

jlipps:
> I'm not quite sure how much we need that, but usually one adds the actual schema version number into configs, so we can easily alter the schema and still handle it properly in newer appium versions and/or perform config migration.

A schema number is a good idea 👍

lib/driver-cli.js
jlipps: @mykola-mokhnach how would you go about adding the lock?

lib/npm.js
jlipps: that's ok, it's checked in callee.

mykola-mokhnach:
I personally prefer YAML as main schema format, since it supports comments and is easier for manual edit

lib/driver-cli.js
mykola-mokhnach: I just wanted to make it simple at the beginning. I mean it probably would be enough to go without config for the MVP and simply use npm for packages control.

jlipps:
> I personally prefer YAML as main schema format, since it supports comments and is easier for manual edit

@imurchie @KazuCocoa and anyone else, what do you think? In my mind, staying with JSON was a nice and easy thing to do. I do agree YAML's a bit easier to read, though just pretty-printing the JSON would go a long way here.

KazuCocoa:
I'm for JSON basically.

But as Mykola commented as _comments_, we can leave comments in the YAML file.
If the file will be managed by multiple persons, such comments will be helpful.
(Nested YAML will be complicated to manage, but I think this file won't be so.)

Is this about 'driver.json', btw?
https://github.com/appium/appium/pull/13463/files#diff-726dd6ffdf869f938cc3750167ab1521R12

imurchie:
yaml is definitely better. Comments, partials, etc.

jlipps:
> Is this about 'driver.json', btw?
Yes.

jlipps:
> yaml is definitely better. Comments, partials, etc.

@imurchie do we use a yaml library in our codebase somewhere already?

jlipps:
this one looks pretty great: https://www.npmjs.com/package/yaml

jlipps:
OK, I switched to YAML. I like it better too.

.gitignore
```diff
@@ -53,3 +53,4 @@ sample-code/csharp/test/obj
 sample-code/csharp/test/bin
 sample-code/csharp/packages
 appium.zip
+.drivers.json
```
mykola-mokhnach: is this still needed?

lib/cli-utils.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  return startSpinner(msg);
+}
+
+async function spinWith (json, msg, fn) {
+  const stopSpinner = logSpin(json, msg);
+  let res;
+  try {
+    res = await fn();
```
mykola-mokhnach: `return await fn()`

finally block will be properly executed after return

lib/driver-cli.js
```diff
@@ -0,0 +1,247 @@
...
+        break;
+      case 'uninstall':
+        jsonResult = await uninstall(args);
+        break;
+      case 'update':
+        jsonResult = await update(args);
+        break;
+      case 'verify':
+        // TODO create a 'verify' command that just checks driver.json and then
+        // that everything in driver.json can be required OK.
```
mykola-mokhnach: is it still .json?

lib/driver-cli.js
```diff
@@ -0,0 +1,247 @@
...
+          pkgName: KNOWN_DRIVERS[name],
+          installed: false,
+        };
+      }
+    }
+  }
+
+  await spinWith(json, lsMsg, async () => {
+    if (showUpdates) {
+      for (const [, data] of _.toPairs(drivers)) {
```
mykola-mokhnach: you can use _.values(drivers) if keys are not needed

mykola-mokhnach: `for (const {version, pkgName, installType} of _.values(drivers))`

lib/driver-cli.js
```diff
@@ -0,0 +1,247 @@
...
+
+async function installViaNpm ({appiumHome, json, driver, pkgName, pkgVer}) {
+  if (!pkgName) {
+    pkgName = driver;
+  }
+  const npmSpec = `${pkgName}${pkgVer ? '@' + pkgVer : ''}`;
+  const specMsg = npmSpec === driver ? '' : ` using NPM install spec '${npmSpec}'`;
+  const msg = `Installing '${driver}'${specMsg}`;
+  try {
+    const pkgJsonData = await spinWith(json, msg, async () => {
```
mykola-mokhnach: `await spinWith(json, msg, () => installPackage(appiumHome, pkgName, pkgVer))`

lib/driver-cli.js
```diff
@@ -0,0 +1,247 @@
...
+    throw new Error(`Installed driver did not have an 'appium' section in its ` +
+                    `package.json file as expected`);
+  }
+
+  const {appium, name, version} = pkgJsonData;
+  const {driverName, automationName, platformNames, mainClass} = appium;
+
+  const reqFields = ['driverName', 'automationName', 'platformNames', 'mainClass'];
+  const missingFields = [];
+  for (const reqField of reqFields) {
```
mykola-mokhnach: I'd use reduce here

lib/drivers.js
```diff
@@ -0,0 +1,232 @@
...
+    if (aNameMatches) {
+      throw new Error(`Driver '${driverName}' supports automationName ` +
+                      `'${automationName}', but Appium could not find ` +
+                      `support for platformName '${matchPlatformName}'. Supported ` +
+                      `platformNames are: ` +
+                      JSON.stringify(platformNames));
+    }
+  }
+
+  throw new Error(`Could not find installed driver to support given caps`);
```
mykola-mokhnach: I'd rather explicitly state the caps

lib/npm.js
```diff
@@ -0,0 +1,108 @@
+import _ from 'lodash';
+import path from 'path';
+import { exec } from 'teen_process';
+
+async function npmExec (cmd, args, opts, execOpts = {}) {
+  const json = opts.json;
+  if (!opts.appiumHome) {
+    throw new Error('appiumHome was not provided to npmExec');
+  }
+  execOpts.cwd = opts.appiumHome;
```
mykola-mokhnach: it is not a good practice to mutate object arguments

mykola-mokhnach:
I was also thinking about the following scenario:
- I install Appium uia2 driver
- Appium uia1 driver is installed as a dependency: do we show it in the installed drivers list?
- I see the list of driver and I want to uninstall/downgrade uia1 driver, because I don't use it: what do we do in such case? do we also uninstall the depending uia2 driver?
or
- I installed uia1 driver of particular version
- Now I want to install uia2 driver, but that one is not compatible with uia1 driver having this version: what would be the behaviour in such case?

lib/driver-cli.js
jlipps: I’m actually mutating the value of drivers though, so I need access to it.

lib/driver-cli.js
jlipps: Was just too long for one line.

jlipps:
> I was also thinking about the following scenario...

Would it work to `npm install` each driver into its own subdirectory of the driver registry? That way UIA1 and UIA2 could be installed side-by-side.

lib/driver-cli.js
jlipps: I used https://github.com/appium/appium-support/pull/179 to address this.

jlipps:
OK @mykola-mokhnach I think I've addressed all the concerns. Can you review the code again?

lib/cli-utils.js
```diff
@@ -0,0 +1,68 @@
+/* eslint-disable no-console */
+
+function errAndQuit (json, msg) {
+  if (json) {
+    console.log(JSON.stringify({error: msg.toString()}));
```
mykola-mokhnach, line 5:
```
`${msg}` is more secure than msg.toString()
```

lib/cli-utils.js
```diff
@@ -0,0 +1,68 @@
...
+  } else {
+    console.error(msg.toString().red);
+  }
+  process.exit(1);
+}
+
+function log (json, msg) {
```
mykola-mokhnach, line 12: it would be nice to have docstrings on these methods

lib/driver-cli.js
```diff
@@ -0,0 +1,250 @@
...
+        break;
+      default:
+        throw new Error(`Cannot handle driver command ${args.driverCommand}`);
+    }
+  } catch (err) {
+    errAndQuit(json, err);
+  }
+
+  if (json) {
+    console.log(JSON.stringify(jsonResult)); // eslint-disable-line no-console
```
mykola-mokhnach, line 48: should the output be prettified?

lib/driver-cli.js
```diff
@@ -0,0 +1,250 @@
...
+    throw new Error(`Installed driver did not have an 'appium' section in its ` +
+                    `package.json file as expected`);
+  }
+
+  const {appium, name, version} = pkgJsonData;
+  const {driverName, automationName, platformNames, mainClass} = appium;
+
+  const reqFields = ['driverName', 'automationName', 'platformNames', 'mainClass'];
+  const missingFields = reqFields.reduce((acc, field) => (
+    appium[field] ? acc : acc.concat(field)
```
mykola-mokhnach, line 227: [...acc, field]

lib/driver-cli.js
```diff
@@ -0,0 +1,250 @@
...
+  if (!pkgJsonData.appium) {
+    throw new Error(`Installed driver did not have an 'appium' section in its ` +
+                    `package.json file as expected`);
+  }
+
+  const {appium, name, version} = pkgJsonData;
+  const {driverName, automationName, platformNames, mainClass} = appium;
+
+  const reqFields = ['driverName', 'automationName', 'platformNames', 'mainClass'];
+  const missingFields = reqFields.reduce((acc, field) => (
```
mykola-mokhnach, line 226: could we use lodash _.difference helper instead?

lib/drivers.js
```diff
@@ -0,0 +1,242 @@
...
+// type 'appium driver install 'name'', rather than having to specify the full
+// npm package. I.e., these are the officially recognized drivers.
+const KNOWN_DRIVERS = {
+  uiautomator2: 'appium-uiautomator2-driver',
+  xcuitest: 'appium-xcuitest-driver',
+  youiengine: 'appium-youiengine-driver',
+  windows: 'appium-windows-driver',
+  mac: 'appium-mac-driver',
+  espresso: 'appium-espresso-driver',
+  tizen: 'appium-tizen-driver',
```
mykola-mokhnach, line 24: flutter?

lib/drivers.js
```diff
@@ -0,0 +1,242 @@
...
+};
+
+const INSTALL_TYPE_NPM = 'npm';
```
mykola-mokhnach, line 27: this could be a map

lib/drivers.js
```diff
@@ -0,0 +1,242 @@
...
+const INSTALL_TYPE_GITHUB = 'github';
+const INSTALL_TYPE_GIT = 'git';
+const INSTALL_TYPES = [
+  INSTALL_TYPE_GIT,
+  INSTALL_TYPE_GITHUB,
+  INSTALL_TYPE_LOCAL,
+  INSTALL_TYPE_NPM
+];
+
+async function validateDriversYaml (appiumHome, logFn = log.error) {
```
mykola-mokhnach, line 38: I'd rather extract all the schema-related stuff to a separate class

lib/npm.js
```diff
@@ -0,0 +1,125 @@
+import _ from 'lodash';
+import path from 'path';
+import { exec } from 'teen_process';
+import { mkdirp, util } from 'appium-support';
+
+const INSTALL_LOCKFILE = '.appium.install.lock';
+const LINK_LOCKFILE = '.appium.link.lock';
+
+async function npmExec (cmd, args, opts, execOpts = {}) {
```
mykola-mokhnach, line 9: how about putting these npm-related functions into a class?

lib/server-parser.js
```diff
@@ -0,0 +1,377 @@
+import fs from 'fs';
+import _ from 'lodash';
+import { DEFAULT_BASE_PATH } from 'appium-base-driver';
+
+// serverArgs will be added to the `server` (default) subcommand
+const serverArgs = [
```
mykola-mokhnach, line 6: I think it makes sense to put server args into a separate json file

lib/server-parser.js
```diff
@@ -0,0 +1,377 @@
...
+    help: 'Set which insecure features are not allowed to run in this server\'s sessions. ' +
+          'Features are defined on a driver level; see documentation for more details. ' +
+          'This should be either a comma-separated list of feature names, or a path to ' +
+          'a file where each feature name is on a line. Features listed here will not be ' +
+          'enabled even if also listed in --allow-insecure, and even if --relaxed-security ' +
+          'is turned on.',
+  }],
+];
+
+function parseSecurityFeatures (features) {
```
mykola-mokhnach, line 304: There is a potential issue that some third party drivers could mimic the same feature names. Perhaps, it would make sense to prefix each feature name with the corresponding automation name it affects (e.g. UiAutomator2:adb_shell)

lib/driver-cli.js
```diff
@@ -0,0 +1,250 @@
...
+  } catch (err) {
+    errAndQuit(json, err);
+  }
+
+  if (json) {
+    console.log(JSON.stringify(jsonResult)); // eslint-disable-line no-console
+  }
+}
+
+async function list ({appiumHome, json, showInstalled, showUpdates}) {
```
mykola-mokhnach, line 52: it makes sense to have the commands as members of the same class, since they share common args

.gitignore
mykola-mokhnach, line 56: isn't this in yaml format now?

lib/driver-cli.js
```diff
@@ -0,0 +1,250 @@
...
+let INSTALLED_DRIVERS;
+
+async function runDriverCommand (args) {
+  // TODO driver.json file should be locked while any of these commands are
+  // running to prevent weird situations
+  let jsonResult = null;
+  const {json, appiumHome} = args;
+  const logFn = (msg) => log(json, msg);
+  try {
+    INSTALLED_DRIVERS = await validateDriversYaml(appiumHome, logFn);
```
mykola-mokhnach, line 19: I wouldn't focus on the format here, but rather on the business logic. how about changing the func name to  `validateDriversConfig` ?

jlipps:
this was merged on accident, but i force-pushed to the target branch, so it is now unmerged. i'm going to open a new PR for this once i've resolved the comments here.

lib/driver-cli.js
jlipps: i don't think so, we're not trying to find the difference between two arrays. we're trying to find which keys from a short list don't exist on an object. there's probably a way to recast this problem in terms of _.difference but i don't think it would be clearer than reduce.

lib/drivers.js
jlipps: i am not sure how this would improve the situation? the reason for making it a constant is to not have to worry about the particular string value that we use in the cli to refer to 'npm' type installs.

lib/server-parser.js
jlipps: this is a good point, however given that we can't prevent the third party drivers from just `rm -rf /` anyway, there doesn't seem to be much use in trying to create a restricted system. to even use a 3rd party driver you have to trust it totally. authors of 3rd party drivers can opt into appium's security model, just the way we do as 1st party authors. if they want to do something nefarious, they don't need to worry about breaking our security model, they can just do it.

lib/server-parser.js
jlipps: they can't totally be in a json file since they reference js functions, but i moved them into their own file.

PR #13635, opened 2019-11-25
feat: Bump base driver to version 5

mykola-mokhnach:
## Proposed changes

These changes are needed for https://github.com/appium/appium-base-driver/pull/370

## Types of changes

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -371,9 +371,6 @@ class AppiumDriver extends BaseDriver {
         });
       }
 
-      // this is an async function but we don't await it because it handles
-      // an out-of-band promise which is fulfilled if the inner driver
-      // unexpectedly shuts down
       this.attachUnexpectedShutdownHandler(d, innerSessionId);
 
       log.info(`New ${InnerDriver.name} session created successfully, session ` +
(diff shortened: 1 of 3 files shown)


lib/appium.js
```diff
@@ -405,25 +402,33 @@ class AppiumDriver extends BaseDriver {
...
-        return;
-      }
-      log.warn(`Closing session, cause was '${e.message}'`);
-      log.info(`Removing session ${innerSessionId} from our master session list`);
-      await sessionsListGuard.acquire(AppiumDriver.name, () => {
-        delete this.sessions[innerSessionId];
-      });
+  attachUnexpectedShutdownHandler (driver, innerSessionId) {
+    const removeSessionFromMasterList = (cause) => {
+      log.warn(`Closing session, cause was '${(cause || {}).message}'`);
```
imurchie: Since this is the only use of `cause`, the default can be in the signature
```js
const removeSessionFromMasterList = (cause = {}) => {
```

imurchie (approved):

lib/appium.js
mykola-mokhnach: changed

PR #14133, opened 2020-04-01
update uninstallOtherPackages documentation

titusfortner:
## Proposed changes

I'm assuming this is the actual correct behavior, since the Appium driver [expects this to be a String](https://github.com/appium/appium-android-driver/blob/e2a7e4b5407b845ce7f21b8e4b0be0c2cc0fd01a/lib/desired-caps.js#L193)

## Types of changes

## Checklist


titusfortner:
Looks like this would also apply to: `otherApps`?

imurchie:
It is expected to be a string that is
1. '*'
1. the name of the package
1. parsable as JSON, into an array

imurchie:
So a comma-separated list will _not_ work. multiple packages are sent as 
```
'["some.app","some.other.app"]'
```

imurchie:
See https://github.com/appium/appium-android-driver/blob/e2a7e4b5407b845ce7f21b8e4b0be0c2cc0fd01a/test/unit/driver-specs.js#L312-L345

titusfortner:
Ugh, ok, so supporting that in Java with a static typed method will be fun. :)

Thanks.

imurchie:
Why is it a problem? It's still a string.

titusfortner:
You've got me confused again. If it is a String with brackets, not a serialized JSON Array, then how is this the correct syntax? `["io.appium.example1", "io.appium.example2"]`
Are you saying that (in Java) I need: `"[\"io.appium.example1\", \"io.appium.example2\"]"`

imurchie:
Now you've got me confused. It is a string, which can be an array of strings serialized as JSON. In Java, you need, as you say, `"[\"io.appium.example1\", \"io.appium.example2\"]"`. My example just had single quotes on the outside. It should all probably be more like
```
> packages = ["io.appium.example1", "io.appium.example2"]
[ 'io.appium.example1', 'io.appium.example2' ]
> uninstallOtherPackages = JSON.stringify(packages)
'["io.appium.example1","io.appium.example2"]'
>
```

titusfortner:
So if it is just a stringified list, why is this specified as an array?
https://github.com/appium/appium-android-driver/blob/e2a7e4b5407b845ce7f21b8e4b0be0c2cc0fd01a/lib/desired-caps.js#L91

titusfortner:
In neither Java nor Ruby do I think directly in JSON, I think in the primitives that a user interacts with directly and I let a library serialize as needed.
So I need to know what primitives I'm putting into the values of a`MutableCapabilities` instance so that when the Selenium bindings does its magic on behalf of the Appium bindings it is sending the right syntax to Appium server.

imurchie:
All of them really ought to accept an Array, or a String that represents and array. There was a move to the String for some reason but I cannot remember the details and I cannot find the conversation.

PR #14170, opened 2020-04-10
Fix the problematic code snippet for Java.

gluxhappy:
The given code won't in the real world and the user will get an error when running the given code snippet with uiautomation=uiautomator2.  The setting ```waitForIdleTimeout``` only works with Integer/Long value. 
Ref: http://appium.io/docs/en/advanced-concepts/settings/index.html


mykola-mokhnach (approved):
Thanks for the update

PR #14301, opened 2020-05-07
Initial Appium 2.0 changes

jlipps:
The new version of https://github.com/appium/appium/pull/13463


dpgraham:
Nice. Looks good so far. We're going to want to make some changes to the drivers before we can push this out.

* Automate releases and semantic changelog generation so we don't have to do a big manual release for every driver
* Shrinkwrap all of the drivers so that the installations are deterministic

jlipps:
@mykola-mokhnach all your suggestions have been implemented in this PR now.

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+
+    for (const [
+      name,
+      {installType, installSpec, installed, updateVersion, version}
+    ] of _.toPairs(drivers)) {
+      let typeTxt;
+      switch (installType) {
+        case INSTALL_TYPE_GIT:
+        case INSTALL_TYPE_GITHUB:
+          typeTxt = `(cloned from ${installSpec})`.yellow;
```
KazuCocoa, line 92: just a question.
Can like `--log-no-colors` option work for CLI such as `yellow` etc?

KazuCocoa (approved):
looks good to me for this change. Promising 🎉

.travis.yml
```diff
@@ -17,12 +17,16 @@ jobs:
   include:
     - stage: Tests
       name: node 10
-      before_script: npm outdated || true
+      before_script:
+        - npm outdated || true
+        - node . driver install --source=npm appium-fake-driver@0.8.0
```
mykola-mokhnach, line 22: do we need to pin the version here?

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+import { errAndQuit, log, spinWith, JSON_SPACES } from './cli-utils';
+import { util, fs } from 'appium-support';
+
+async function runDriverCommand (args) {
+  // TODO driver config file should be locked while any of these commands are
+  // running to prevent weird situations
+  let jsonResult = null;
+  const {json, appiumHome} = args;
+  const logFn = (msg) => log(json, msg);
+  const config = new DriverConfig(appiumHome, logFn);
```
mykola-mokhnach, line 16: 😍

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
+import _ from 'lodash';
+import path from 'path';
+import { KNOWN_DRIVERS } from './drivers';
+import DriverConfig, { INSTALL_TYPE_NPM, INSTALL_TYPE_GIT, INSTALL_TYPE_GITHUB,
+                       INSTALL_TYPE_LOCAL } from './driver-config';
+import NPM from './npm';
+import { errAndQuit, log, spinWith, JSON_SPACES } from './cli-utils';
+import { util, fs } from 'appium-support';
+
+async function runDriverCommand (args) {
```
mykola-mokhnach, line 10: it would be nice to have some docstring on that. It is not very obvious what `args`is

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+class DriverCommand {
+
+  constructor ({config, json}) {
+    this.config = config;
+    this.json = json;
+    this.npm = new NPM(this.config.appiumHome);
+  }
+
+  async execute (args) {
+    if (!this[args.driverCommand]) {
```
mykola-mokhnach, line 39: I'd rather use _.isFunction
Also, it makes sense to check directly the instance prototype instead of `this`. We don't want any standard javascript Object methods to match

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+      throw new Error(`Cannot handle driver command ${args.driverCommand}`);
+    }
+    return await this[args.driverCommand](args);
+  }
+
+  async list ({showInstalled, showUpdates}) {
+    const lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} drivers`;
+    const drivers = {};
+
+    for (const name of Object.keys(this.config.installedDrivers)) {
```
mykola-mokhnach, line 49: This could also use `reduce`

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+  if (json) {
+    console.log(JSON.stringify(jsonResult, null, JSON_SPACES)); // eslint-disable-line no-console
+  }
+}
+
+class DriverCommand {
+
+  constructor ({config, json}) {
+    this.config = config;
+    this.json = json;
```
mykola-mokhnach, line 34: what would you say if we rename this var to `isJsonLog` ?

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+          drivers[name] = {
+            pkgName: KNOWN_DRIVERS[name],
+            installed: false,
+          };
+        }
+      }
+    }
+
+    await spinWith(this.json, lsMsg, async () => {
+      if (showUpdates) {
```
mykola-mokhnach, line 66: `if (!showUpdates) return`

it is easier to read the code if the nesting level is low

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+          typeTxt = '(NPM)';
+      }
+      const installTxt = installed ?
+        `@${version.yellow} ${('[installed ' + typeTxt + ']').green}` :
+        ' [not installed]'.grey;
+      const updateTxt = showUpdates && updateVersion ?
+        ` [${updateVersion} available]`.magenta :
+        '';
+
+      console.log(`- ${name.yellow}${installTxt}${updateTxt}`); // eslint-disable-line no-console
```
mykola-mokhnach, line 107: should we put` eslint-disable-line no-console` to the top?

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    return drivers;
+  }
+
+  async install ({driver, installType}) {
+    log(this.json, `Attempting to find and install driver '${driver}'`);
+
+    let driverData, installSpec = driver;
```
mykola-mokhnach, line 116: it is recommended to define each var in a separate line

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+      }
+
+      driverData = await this.installViaNpm({driver, pkgName, pkgVer});
+    }
+
+    const {driverName} = driverData;
+    delete driverData.driverName;
+
+    if (this.config.isInstalled(driverName)) {
+      throw new Error(`A driver named '${driverName}' is already installed. ` +
```
mykola-mokhnach, line 168: should this really be an error?

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+  getDriverFields (pkgJsonData) {
+    if (!pkgJsonData.appium) {
+      throw new Error(`Installed driver did not have an 'appium' section in its ` +
+                      `package.json file as expected`);
+    }
+
+    const {appium, name, version} = pkgJsonData;
+    const {driverName, automationName, platformNames, mainClass} = appium;
+
+    const reqFields = ['driverName', 'automationName', 'platformNames', 'mainClass'];
```
mykola-mokhnach, line 215: this could be moved to the top of the module

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+      ));
+      const driverData = this.getDriverFields(pkgJsonData);
+      driverData.installPath = installPath;
+      return driverData;
+    } catch (err) {
+      throw new Error(`Encountered an error when installing package: ${err.message}`);
+    }
+  }
+
+  getDriverFields (pkgJsonData) {
```
mykola-mokhnach, line 206: like I said, some documentation on what the method accepts and returns would be super nice here

lib/driver-config.js
```diff
@@ -0,0 +1,193 @@
...
+      }
+
+      if (!_.isString(mainClass)) {
+        p.push({err: 'Missing or incorrect driver class name', val: mainClass});
+      }
+    }
+
+    const problemSummaries = [];
+    for (const [driverName, problems] of _.toPairs(foundProblems)) {
+      if (!_.isEmpty(problems)) {
```
mykola-mokhnach, line 118: if (_.isEmpty(problems)) continue

lib/driver-config.js
```diff
@@ -0,0 +1,193 @@
...
+        this.log(summary);
+      }
+    }
+
+    return drivers;
+  }
+
+  async write () {
+    const yamlData = {schemaRev: CONFIG_SCHEMA_REV, drivers: this.installedDrivers};
+    await fs.writeFile(this.configFile, YAML.stringify(yamlData), 'utf-8');
```
mykola-mokhnach, line 143: utf-8 or utf8?

lib/main.js
```diff
@@ -6,19 +6,22 @@ import logger from './logger'; // logger needs to remain first of imports
...
 } from './config';
+import DriverConfig from './driver-config';
+import { runDriverCommand } from './driver-cli';
 import { AppiumDriver } from './appium';
 import registerNode from './grid-register';
 import { inspectObject } from './utils';
 
 
-async function preflightChecks (parser, args, throwInsteadOfExit = false) {
+async function preflightChecks (parser, args, driverConfig, throwInsteadOfExit = false) {
```
mykola-mokhnach, line 24: it looks suspicious if a method needs more than 3 arguments, it, probably, needs to be refactored or the arguments should be combined

lib/npm.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  async exec (cmd, args, opts, execOpts = {}) {
+    let { cwd, json, lockFile } = opts;
+    if (!cwd) {
+      cwd = this.appiumHome;
+    }
+    // ensure the directory we want to install inside of exists
+    await mkdirp(cwd);
+    execOpts.cwd = cwd;
```
mykola-mokhnach, line 22: I'd rather avoid mutating the original arguments

lib/npm.js
```diff
@@ -0,0 +1,132 @@
...
+
+    args.unshift(cmd);
+    if (json) {
+      args.push('-json');
+    }
+    let runner = async () => await exec('npm', args, execOpts);
+    if (lockFile) {
+      const acquireLock = util.getLockFileGuard(path.resolve(cwd, lockFile));
+      const _runner = runner;
```
mykola-mokhnach, line 31: why this is needed?

lib/npm.js
```diff
@@ -0,0 +1,132 @@
...
+    await mkdirp(cwd);
+    execOpts.cwd = cwd;
+
+    args.unshift(cmd);
+    if (json) {
+      args.push('-json');
+    }
+    let runner = async () => await exec('npm', args, execOpts);
+    if (lockFile) {
+      const acquireLock = util.getLockFileGuard(path.resolve(cwd, lockFile));
```
mykola-mokhnach, line 30: is cwd always writeable? I'd rather use tempdir

lib/server-args.js
```diff
@@ -0,0 +1,348 @@
...
+                      `from file ${parsedFeatures[0]} but got error: ${err.message}`);
+    }
+  }
+
+  return parsedFeatures;
+}
+
+function parseDefaultCaps (caps) {
+  try {
+    // use synchronous file access, as `argparse` provides no way of either
```
mykola-mokhnach, line 33: it is synchronous only in scope of the current process, but not cross-process

lib/server-args.js
```diff
@@ -0,0 +1,348 @@
...
+  }
+  caps = JSON.parse(caps);
+  if (!_.isPlainObject(caps)) {
+    throw 'Invalid format for default capabilities';
+  }
+  return caps;
+}
+
+export default [
+  [['--shell'], {
```
mykola-mokhnach, line 52: so how about moving these into a separate JSON config file? such json files could be loaded with simple `import` operator like ES modules.

package.json
```diff
@@ -42,19 +42,8 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.6.0",
-    "appium-android-driver": "^4.20.0",
     "appium-base-driver": "^6.0.1",
-    "appium-espresso-driver": "^1.0.0",
```
mykola-mokhnach, line 47: 😻

lib/driver-cli.js
```diff
@@ -0,0 +1,249 @@
...
+  } catch (err) {
+    errAndQuit(json, err);
+  }
+
+  if (json) {
+    console.log(JSON.stringify(jsonResult, null, JSON_SPACES)); // eslint-disable-line no-console
+  }
+}
+
+class DriverCommand {
```
mykola-mokhnach, line 30: would it be possible to have e2e tests for all these commands?

lib/cli-utils.js
```diff
@@ -0,0 +1,96 @@
...
+ */
+function log (json, msg) {
+  !json && console.log(msg);
+}
+
+/**
+ * Start a CLI spinner
+ * @param {string} msg - what to label the spinner with
+ */
+function startSpinner (msg) {
```
christian-bromann, line 32: Would it make sense to use a package like [ora](https://www.npmjs.com/package/ora) here?

.travis.yml
jlipps: i suppose it's safe to not pin it unless we run into a problem

lib/driver-cli.js
jlipps: ok makes sense

lib/driver-cli.js
jlipps: yep, will explain

lib/driver-cli.js
jlipps: :+1

lib/driver-cli.js
jlipps: i tend to think so, especially since we might be in a json context where there's no other log output. if something doesn't happen the way the user expects i would take it to be an error.

lib/npm.js
jlipps: cwd can't be random, since we have to run the npm commands in specific directories inside of APPIUM_HOME. the user needs to ensure that it's writeable.

lib/server-args.js
jlipps: this code was just moved over, it is not new. i'd prefer not to think about whether this needs re-architecting as part of this PR.

lib/npm.js
mykola-mokhnach: Does it make sense to have a separate check for the folder writability?

lib/npm.js
jlipps: because if we don't alias this, then we end up with a recursive function which is not what we want.

lib/npm.js
jlipps: Yes, I'll put a check in the config read section, which is always executed at the beginning of any driver subcommand.

lib/server-args.js
jlipps: a couple reasons i don't like this:

1) json is like js just with less flexibility and more quotation marks
2) some of these cannot be represented easily in JSON, for example:

```js
  [['-pa', '--base-path'], {
    required: false,
    defaultValue: DEFAULT_BASE_PATH,
    dest: 'basePath',
    example: '/path/prefix',
    help: 'Base path to use as the prefix for all webdriver routes running' +
          `on this server (default: ${DEFAULT_BASE_PATH})`
  }],
```

or

```js
  [['-dc', '--default-capabilities'], {
    dest: 'defaultCapabilities',
    defaultValue: {},
    type: parseDefaultCaps,
    required: false,
    example: '[ \'{"app": "myapp.app", "deviceName": "iPhone Simulator"}\' ' +
             '| /path/to/caps.json ]',
    help: 'Set the default desired capabilities, which will be set on each ' +
          'session unless overridden by received capabilities.'
  }],
```

(Note the use of variables here corresponding to strings and functions)

3) this file is already loaded with a simple `import` operator...

but maybe I'm missing something about your suggestion?

lib/driver-cli.js
jlipps: i'll see what i can do...

lib/appium.js
```diff
@@ -266,6 +90,14 @@ class AppiumDriver extends BaseDriver {
     log.info(introString);
   }
 
+  /**
+   * This is just an alias for driver.js's method, which is necessary for
```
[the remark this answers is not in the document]
jlipps: we can consider a refactor of the test system, but I would not want to attempt that in this PR

lib/npm.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+    // ensure the directory we want to install inside of exists
+    await mkdirp(cwd);
+    execOpts.cwd = cwd;
+
+    args.unshift(cmd);
+    if (json) {
+      args.push('-json');
+    }
+    let runner = async () => await exec('npm', args, execOpts);
```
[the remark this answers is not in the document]
jlipps: The npm commands run by appium will never interface with any other project or user. They'll be run under the hood in a secret directory that is totally managed by Appium. It's therefore up to us whether to use npm or yarn. I picked npm because the user is guaranteed to have it on their system if they're already using Appium.

lib/cli-utils.js
jlipps: ora looks great! i didn't know about it. i'll see if i can switch to use it.

lib/server-args.js
mykola-mokhnach: no, this makes sense. Let's keep it as is, but isolate the args list itself from the rest of the code

lib/driver-cli.js
jlipps: --log-no-colors is not available as part of the driver cli, only part of the server cli, so this would be a moot question.

lib/driver-cli.js
```diff
@@ -0,0 +1,271 @@
...
+      throw new Error(`Cannot handle driver command ${args.driverCommand}`);
+    }
+    return await this[args.driverCommand](args);
+  }
+
+  async list ({showInstalled, showUpdates}) {
+    const lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} drivers`;
+    const installedNames = Object.keys(this.config.installedDrivers);
+    const knownNames = Object.keys(KNOWN_DRIVERS);
+    const drivers = [...installedNames, ...knownNames].reduce((drivers, name) => {
```
mykola-mokhnach, line 60: I'd rather rename the first lambda argument to `acc` to avoid confusion

lib/driver-config.js
```diff
@@ -0,0 +1,204 @@
...
+const INSTALL_TYPES = [
+  INSTALL_TYPE_GIT,
+  INSTALL_TYPE_GITHUB,
+  INSTALL_TYPE_LOCAL,
+  INSTALL_TYPE_NPM
+];
+
+
+export default class DriverConfig {
+  constructor (appiumHome, logFn = log.error) {
```
mykola-mokhnach, line 25: should `bind` be used here?

lib/driver-config.js
```diff
@@ -0,0 +1,204 @@
...
+      if (await fs.exists(this.configFile)) {
+        // if the file exists and we couldn't parse it, that's a problem
+        throw new Error(`Appium had trouble loading the driver installation ` +
+                        `cache file (${this.configFile}). Ensure it exists and is ` +
+                        `readable. Specific error: ${err.message}`);
+      }
+
+      // if appium home exists but the dir is not readable/writable, fs.access will error
+      try {
+        await fs.access(this.appiumHome);
```
mykola-mokhnach, line 52: this might not work on Windows in case of missing NTFS permissions

mykola-mokhnach: https://groups.google.com/forum/#!topic/nodejs/qmZtIwDRSYo

lib/drivers.js
```diff
@@ -0,0 +1,81 @@
...
+  mac: 'appium-mac-driver',
+  espresso: 'appium-espresso-driver',
+  tizen: 'appium-tizen-driver',
+  flutter: 'appium-flutter-driver',
+};
+
+function getDriverBySupport (drivers, matchAutomationName, matchPlatformName) {
+  for (const [driverName, driverData] of _.toPairs(drivers)) {
+    const {automationName, platformNames} = driverData;
+    const aNameMatches = automationName.toLowerCase() === matchAutomationName.toLowerCase();
```
mykola-mokhnach, line 23: is automationName always set?

mykola-mokhnach (approved):

test/cli-e2e-specs.js
```diff
@@ -0,0 +1,98 @@
+import path from 'path';
```
mykola-mokhnach, line 1: 👍 

More tests are always appreciated

lib/drivers.js
jlipps: yes, it is validated in config load

PR #14454, opened 2020-06-15
feat: add driver update cli command

jlipps:


lib/cli/args.js
```diff
@@ -380,7 +380,14 @@ const driverUpdateArgs = [
...
     example: 'xcuitest',
-    help: 'Name of the driver to update, or the word "all" to update all. To see available updates, run "appium driver list --installed --updates"',
+    help: 'Name of the driver to update, or the word "installed" to update all installed drivers. To see available updates, run "appium driver list --installed --updates"',
+  }],
+  [['--unsafe'], {
+    required: false,
+    defaultValue: false,
+    action: 'storeTrue',
+    nargs: 0,
+    help: 'Update beyond a major revision or to a beta version, which might have breaking changes or bugs',
```
mykola-mokhnach, line 390: major and beta a different things. I would rather put them as separate options

lib/cli/driver.js
```diff
@@ -75,15 +76,20 @@ class DriverCommand {
...
-          if (!version || util.compareVersions(latestVersion, '>', version)) {
-            data.updateVersion = latestVersion;
-          }
+      for (const [driver, data] of _.toPairs(drivers)) {
+        const {installed, installType} = data;
+        if (!installed) {
+          // don't need to check for updates on drivers that aren't installed
+          continue;
+        }
+        if (installType !== INSTALL_TYPE_NPM) {
```
mykola-mokhnach, line 85: we could combine this condition with the previous one using ||

lib/cli/driver.js
```diff
@@ -262,9 +274,88 @@ class DriverCommand {
     return this.config.installedDrivers;
   }
 
-  update (/*{driver}*/) {
-    // TODO implementation
-    return {};
+  async update ({driver, unsafe}) {
+    const all = driver === UPDATE_ALL;
```
mykola-mokhnach, line 278: how about giving a more descriptive name to this var (`shouldUpdateAll`)

lib/cli/driver.js
```diff
@@ -262,9 +274,88 @@ class DriverCommand {
...
+    const updates = {};
+    for (const d of driversToUpdate) {
+      try {
+        await spinWith(this.isJsonOutput, `Checking if driver '${d}' is updatable`, () => {
+          if (this.config.installedDrivers[d].installType !== INSTALL_TYPE_NPM) {
+            throw new Error(`Driver '${d}' was not installed via npm, so we can't check for updates`);
+          }
+        });
+        const update = await spinWith(this.isJsonOutput, `Checking if driver '${d}' needs an update`, async () => {
+          const update = await this.checkForDriverUpdate(d);
```
mykola-mokhnach, line 294: const {safeUpdate, unsafeUpdate} = ...

lib/cli/driver.js
```diff
@@ -262,9 +274,88 @@ class DriverCommand {
...
+          }
+          return update;
+        });
+        if (!unsafe && !update.safeUpdate) {
+          throw new Error(`Driver '${d}' has a major revision update ` +
+                          `(${update.current} => ${update.unsafeUpdate}), which could include ` +
+                          `breaking changes. If you want to apply this update, re-run with --unsafe`);
+        }
+        const updateVer = unsafe && update.unsafeUpdate ? update.unsafeUpdate : update.safeUpdate;
+        await spinWith(this.isJsonOutput, `Updating driver '${d}' from ${update.current} to ${updateVer}`, async () => {
```
mykola-mokhnach, line 306: do we need to enclose the lambda body into curly brackets here?

lib/cli/driver.js
```diff
@@ -262,9 +274,88 @@ class DriverCommand {
...
-    // TODO implementation
-    return {};
+  async update ({driver, unsafe}) {
+    const all = driver === UPDATE_ALL;
+    // if we're specifically requesting an update for a driver, make sure it's installed
+    if (!all && !this.config.isInstalled(driver)) {
+      throw new Error(`Driver '${driver}' was not installed, so can't be updated`);
+    }
+    const driversToUpdate = all ? Object.keys(this.config.installedDrivers) : [driver];
+    const errors = {};
```
mykola-mokhnach, line 284: it makes sense to either add a comment about what will be keys ands values of these maps

lib/cli/driver.js
```diff
@@ -262,9 +274,88 @@ class DriverCommand {
...
+    }
+    if (this.isJsonOutput) {
+      return updates;
+    }
+    log(this.isJsonOutput, 'Update report:');
+    for (const [d, update] of _.toPairs(updates)) {
+      log(this.isJsonOutput, `- Driver ${d} updated: ${update.from} => ${update.to}`.green);
+    }
+    for (const [d, err] of _.toPairs(errors)) {
+      if (/no updates available/.test(err) || /not installed via npm/.test(err)) {
```
mykola-mokhnach, line 322: please move magic strings to constants. This might break easily if one changes the original message

lib/cli/driver.js
```diff
@@ -262,9 +274,88 @@ class DriverCommand {
...
+  }
+
+  async checkForDriverUpdate (driver) {
+    // TODO decide how we want to handle beta versions?
+    // this is a helper method, 'driver' is assumed to already be installed here, and of the npm
+    // install type
+    const {version, pkgName} = this.config.installedDrivers[driver];
+    let unsafeUpdate = await this.npm.getLatestVersion(pkgName);
+    let safeUpdate = await this.npm.getLatestSafeUpgradeVersion(pkgName, version);
+    if (!util.compareVersions(unsafeUpdate, '>', version)) {
```
mykola-mokhnach, line 338: it would be nice to have some unit tests for this logic

lib/cli/driver.js
```diff
@@ -262,9 +274,88 @@ class DriverCommand {
...
+      unsafeUpdate = null;
+    }
+    if (safeUpdate && !util.compareVersions(safeUpdate, '>', version)) {
+      // even the safe update is not later than the current, so it is not actually an update
+      safeUpdate = null;
+    }
+    return {current: version, safeUpdate, unsafeUpdate};
+  }
+
+  async updateDriver (driver, version) {
```
mykola-mokhnach, line 354: docstrings would be nice

lib/cli/driver.js
```diff
@@ -262,12 +273,130 @@ class DriverCommand {
     return this.config.installedDrivers;
   }
 
-  update (/*{driver}*/) {
-    // TODO implementation
-    return {};
+  /**
+   * Attempt to update one or more drivers using NPM
+   *
+   * @param [object] updateSpec - object containing a 'driver' key whose value is the name of
```
mykola-mokhnach, line 279: We usually use the [typedef](https://github.com/appium/appium-adb/blob/master/lib/tools/adb-commands.js#L965) format to describe such things

lib/cli/driver.js
```diff
@@ -262,12 +273,130 @@ class DriverCommand {
...
+   * a driver, or the string specified in UPDATE_ALL to designate the updating of all drivers, and
+   * also an 'unsafe' key, which when true will allow Appium to upgrade the driver past a major
+   * revision boundary.
+   *
+   * @returns [object] - object with 'errors' and 'updates' keys. Errors is a map of driver names
```
mykola-mokhnach, line 284: `{Object}`

lib/cli/driver.js
```diff
@@ -262,12 +273,130 @@ class DriverCommand {
...
+
+    // 'updates' will have driver names as keys and update objects as values, where an update
+    // object is of the form {from: versionString, to: versionString}
+    const updates = {};
+
+    for (const d of driversToUpdate) {
+      try {
+        await spinWith(this.isJsonOutput, `Checking if driver '${d}' is updatable`, () => {
+          if (this.config.installedDrivers[d].installType !== INSTALL_TYPE_NPM) {
+            throw new Error(`Driver '${d}' ${NOT_NPM_MSG}`);
```
mykola-mokhnach, line 308: does it make sense to have different error classes instead?

lib/cli/driver.js
```diff
@@ -262,12 +273,130 @@ class DriverCommand {
...
+        errors[d] = err;
+      }
+    }
+
+    log(this.isJsonOutput, 'Update report:');
+    for (const [d, update] of _.toPairs(updates)) {
+      log(this.isJsonOutput, `- Driver ${d} updated: ${update.from} => ${update.to}`.green);
+    }
+    for (const [d, err] of _.toPairs(errors)) {
+      if (new RegExp(NO_UPDATES_MSG).test(err) || new RegExp(NOT_NPM_MSG).test(err)) {
```
mykola-mokhnach, line 340: if ([NO_UPDATES_MSG, NOT_NPM_MSG].map((x) -> new RegExp(_.escapeRegExp(x)).some((p) -> p.test(err)))

lib/cli/driver.js
```diff
@@ -262,12 +273,130 @@ class DriverCommand {
...
+    }
+
+    return {updates, errors};
+  }
+
+  /**
+   * Given a driver name, figure out what its highest possible version upgrade is, and also the
+   * highest possible safe upgrade.
+   *
+   * @param [string] driver - name of driver
```
mykola-mokhnach, line 356: Please use curly brackets for types to align with the common documentation style

lib/cli/driver.js
```diff
@@ -262,12 +273,130 @@ class DriverCommand {
...
+   *
+   * @returns [object] - object with keys 'current', 'safeUpdate', and 'unsafeUpdate'. The values
```
mykola-mokhnach, line 358: ⬆️

test/cli-specs.js
```diff
@@ -0,0 +1,67 @@
+// transpile:mocha
```
mykola-mokhnach, line 1: 😍

lib/cli/driver.js
```diff
@@ -9,6 +9,9 @@ import NPM from './npm';
 import { errAndQuit, log, spinWith, JSON_SPACES } from './utils';
 import { util, fs } from 'appium-support';
 
+const UPDATE_ALL = 'installed';
```
mykola-mokhnach, line 12: should we have documented somewhere the set of rules for drivers naming and which of these names are reserved?

lib/cli/npm.js
```diff
@@ -54,6 +55,56 @@ export default class NPM {
...
+    return this.getLatestSafeUpgradeFromVersions(curVersion, allVersions);
+  }
+
+  /**
+   * Given a current version and a list of all versions for a package, return the version which is
+   * the highest safely-upgradable version (meaning not crossing any major revision boundaries, and
+   * not including any alpha/beta/rc versions)
+   *
+   * @param [string] curVersion - the current version of a package
+   * @param [array] allVersion - a list of version strings
```
mykola-mokhnach, line 71: Please add `return` description

mykola-mokhnach, line 71: `@param {Array<string>} allVersions`

mykola-mokhnach (approved):
Had some minor comments, but in general I like it

KazuCocoa (approved):

lib/cli/driver.js
jlipps: yeah, this should be in the documentation. i don't see this as dangerous though, if someone creates a driver called 'installed' it will just never be updatable using this command. would be a dumb move though.

also i don' tmind if people create drivers with the same name, just whichever one is installed first will be the one that's used, and others won't be able to be installed.

jlipps:
ok, all feedback addressed. i think this is good to go? once it's merged, i'll put together a short driver cli doc, and then we can release a 2.0 beta to start seeing where the issues might be with this approach.

jlipps:
any more thoughts @mykola-mokhnach or is this good to go?

Issue #14490, opened 2020-06-28
ms:waitForAppLaunch is not recognized by appium


krevathi1912:
## The problem

I am automating UWP app using appium+winappdriver. The app launch time is more, so i am using "ms:waitForAppLaunch" capability. Appium is not recognizing this capability and takes the capability as "waitForAppLaunch". Below is the code and log. I am using python language

## Environment

* Appium version (or git revision) that exhibits the issue: 1.16.0
* Last Appium version that did not exhibit the issue (if applicable):
* Desktop OS/version used to run Appium: Windows OS
* Node.js version (unless using Appium.app|exe): v10.16.1
* Npm or Yarn package manager: 6.9.0
* Mobile platform/version under test: Windows UWP app
* Real device or emulator/simulator: Windows 10 Laptop
* Appium CLI or Appium.app|exe: Appium CLI

## Link to Appium logs

https://gist.github.com/krevathi1912/4f62f9042052e4f7cbd260b14c80ccc5

## Code To Reproduce Issue [ Good To Have ]

        desired_caps['platformName'] = "Windows"
        desired_caps['deviceName'] = "WindowsPC"
        desired_caps['app'] = "..................!App"
        desired_caps["ms:ms:waitForAppLaunch"] = 50
        self.driver = webdriver.Remote('http://localhost:4723/wd/hub', desired_caps)

mykola-mokhnach:
This issue is that WAD only works in JWP mode, while prefixes are defined as a part of W3C spec. Everything is mixed around. You could add `appium:` prefix to the capability name as a workaround: `appium:ms:waitForAppLaunch`

mykola-mokhnach:
I've also added a PR to avoid stripping third party prefixes from cap names: https://github.com/appium/appium/pull/14492, but it will only be present in the next Appium release

krevathi1912:
@mykola-mokhnach  Thanks for the response. I have tried using the workaround appium:ms:waitForAppLaunch, still appium says ms:waitForAppLaunch as not recognized. Also the WinAppDriver did not wait for 50 secs after an app launch is initiated prior to attaching to the application session. Attached the log

https://gist.github.com/krevathi1912/4c56d992e774fa44acf028f2af870df5

mykola-mokhnach:
Appium's job is to only send session capabilities. Everything else is up to WAD, which is maintained by Microsoft. Unfortunately, we cannot help anything more there

krevathi1912:
ok, thank you very much for the response

[...]

PR #14492, opened 2020-06-28
fix: Only strip known prefixes from capability names

mykola-mokhnach:
## Proposed changes

We strip all prefixes from cap names, which is sometimes not something we'd like to achieve. It only makes sense to deal with the prefixes that we know about. Addresses https://github.com/appium/appium/issues/14490

## Types of changes

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -73,14 +73,14 @@ function parseCapsForInnerDriver (jsonwpCapabilities, w3cCapabilities, constrain
         // Check if the key is already present in firstMatch entries
         for (const firstMatchEntry of (w3cCapabilities.firstMatch || [])) {
           if (_.isPlainObject(firstMatchEntry)
-              && _.has(removeW3CPrefixes(firstMatchEntry), removeW3CPrefix(defaultCapKey))) {
+              && _.has(removeAppiumPrefixes(firstMatchEntry), removeAppiumPrefix(defaultCapKey))) {
             isCapAlreadySet = true;
             break;
           }
         }
         // Check if the key is already present in alwaysMatch entries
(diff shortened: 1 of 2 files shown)


test/utils-specs.js
```diff
@@ -172,6 +174,20 @@ describe('utils', function () {
     });
   });
 
+  describe('removeAppiumPrefixes()', function () {
```
KazuCocoa, line 177: 👍

KazuCocoa (approved):

lib/utils.js
```diff
@@ -95,15 +95,15 @@ function parseCapsForInnerDriver (jsonwpCapabilities, w3cCapabilities, constrain
...
-      jsonwpCapabilities = Object.assign({}, removeW3CPrefixes(defaultCapabilities), jsonwpCapabilities);
+      jsonwpCapabilities = Object.assign({}, removeAppiumPrefixes(defaultCapabilities), jsonwpCapabilities);
     }
   }
 
   // Get MJSONWP caps
   if (hasJSONWPCaps) {
     protocol = MJSONWP;
     desiredCaps = jsonwpCapabilities;
-    processedJsonwpCapabilities = removeW3CPrefixes({...desiredCaps});
```
KazuCocoa, line 106: ah, we removed w3c prefix for jsonwp

jlipps:
seems OK to me, the only downside appears to be that in the appium 2.0 world, each driver will be responsible for reimplementing logic to strip out its own prefixes--or it must direct users to use the 'appium:' prefix.

mykola-mokhnach:
> must direct users to use the 'appium:' prefix

I assume it already does

KazuCocoa (approved):

PR #14527, opened 2020-07-07
docs: add appium 2.0 driver cli docs

jlipps:
Add docs for the driver CLI functionality

### DIFF
--- a/docs/en/drivers/driver-cli.md
+++ b/docs/en/drivers/driver-cli.md
@@ -1,0 +1,77 @@
+## The Appium 2.0 Driver Command Line Interface (CLI)
+
+Appium calls the bits of software that enable automation of a certain platform using a certain automation technology a "driver". There are drivers for iOS, Android, and other platforms. As of Appium 2.0, Appium does not download and install any drivers by default when you install Appium. There are too many drivers for this to be practical. Instead, you use a special command line interface to direct Appium to install and register drivers for you, to make them available for your use. This doc highlights the various commands available within this interface.
+
+### Overview
+
+All of these commands are available after typing `appium driver`. For example, in this doc, the command defined as `list` would be used in the following manner:
+
+```
+appium driver list
+```
+
+*Global Flags*
+
+There are a number of flags available on all commands:
+
+|Flag|Description|Default|
+|----|-----------|-------|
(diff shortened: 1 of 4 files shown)


docs/en/drivers/driver-cli.md
```diff
@@ -0,0 +1,74 @@
...
+
+Note: this will update the driver(s) to the highest available minor version, but will *not* update to a new major version, since it might contain breaking changes. To update to a new major version, use the `--unsafe` flag.
+
+|Flag|Description|Default|
+|----|-----------|-------|
+|`--unsafe`|Upgrade to a new major version if available|None|
+
+|Example|Description|
+|-------|-----------|
+|`appium driver update xcuitest`|Update the XCUITest drivers to the latest safe version|
```
mykola-mokhnach, line 64: I think it makes sense to add a possibility to be able to set the particular version number to install or to update to.

docs/en/drivers/driver-cli.md
```diff
@@ -0,0 +1,74 @@
+## The Appium Driver Command Line Interface (CLI)
+
+Appium calls the bits of software that enable automation of a certain platform using a certain automation technology a "driver". There are drivers for iOS, Android, and other platforms. As of Appium 2.0, Appium does not download and install any drivers by default when you install Appium. There are too many drivers for this to be practical. Instead, you use a special command line interface to direct Appium to install and register drivers for you, to make them available for your use. This doc highlights the various commands available within this interface.
```
mykola-mokhnach, line 3: It makes sense to mention that in the title, that we are talking about Appium 2.0+

mykola-mokhnach (approved):

docs/en/drivers/driver-cli.md
jlipps: this won't be published though until 2.0 is published, right, so at that point won't it be moot?

mykola-mokhnach: publishing Appium2.0 does not mean that people would stop using 1.0 immediately

jlipps: ok, fair enough!

KazuCocoa:
👍

PR #14556, opened 2020-07-14
feat: add support for plugins

jlipps:
This PR adds the following:

* a new `appium plugin` CLI which has the same shape and behavior as the driver CLI (so lots of refactoring to share code)
* the ability for plugins to modify the Appium express app and http server
* the ability for plugins to add new (or override old) routes/commands to the method map
* the ability for plugins to handle specific server commands (either to wrap the original behavior or to replace it)

As of now, every installed plugin is always active. A next step is to make it so that you have to opt into installed plugins, via server args or caps.

This depends on some BaseDriver changes (https://github.com/appium/appium-base-driver/pull/422)

The plugin used for testing is the new [@appium/fake-plugin](https://github.com/appium/appium-fake-plugin), which is built on a base class ([@appium/base-plugin](https://github.com/appium/appium-base-plugin)) that everyone can use to build their own plugins.


lib/appium.js
```diff
@@ -344,32 +346,136 @@ class AppiumDriver extends BaseDriver {
     }
   }
 
+  pluginsToWrapCmd (cmd) {
+    return this.plugins.filter((plugin) => plugin.wrapCommands.includes(cmd));
```
mykola-mokhnach, line 350: we need to check if `wrapCommands` property exists first. an invalid plugin should not crash the whole server

lib/appium.js
```diff
@@ -344,32 +346,136 @@ class AppiumDriver extends BaseDriver {
...
+  }
+
+  pluginsToHandleCmd (cmd) {
+    return this.plugins.filter((plugin) => {
+      if (plugin.commands === true) {
+        return true;
+      }
+      if (_.isArray(plugin.commands) && plugin.commands.includes(cmd)) {
```
mykola-mokhnach, line 358: `_.includes(plugin.commands, cmd)` would do the job

mykola-mokhnach: the whole method could be rewritten to

`return this.plugins.filter((p) -> plugin.commands === true || _.includes(plugin.commands, cmd))`

lib/appium.js
```diff
@@ -344,32 +346,136 @@ class AppiumDriver extends BaseDriver {
...
+        return true;
+      }
+      return false;
+    });
+  }
+
   async executeCommand (cmd, ...args) {
```
mykola-mokhnach, line 365: this method is becoming too big. How about extracting parts of it into separate methods?

lib/cli/args.js
```diff
@@ -329,73 +333,97 @@ const globalDriverArgs = [
...
+  ];
+}
+
+function makeUpdateArgs (type) {
+  return [
+    ...globalExtensionArgs,
+    [[type], {
+      type: 'string',
+      example: type === DRIVER_TYPE ? DRIVER_EXAMPLE : PLUGIN_EXAMPLE,
+      help: `Name of the ${type} to update, or the word "installed" to update all installed ${type}s. To see available updates, run "appium ${type} list --installed --updates"`,
```
mykola-mokhnach, line 406: some strings are too long

mykola-mokhnach:
The PR is complicated and hard to read and understand without a spec and any documentation. Thus, I would not be able to review it properly

lib/appium.js
jlipps: actually i just need to delete this, it's obsolete

jlipps:
> The PR is complicated and hard to read and understand without a spec and any documentation. Thus, I would not be able to review it properly

@mykola-mokhnach i understand. I will work on updating the appium 2.0 plugin spec to reflect the new reality. In the meantime, does the extensive commenting on the base plugin help? https://github.com/appium/appium-base-plugin/blob/master/lib/plugin.js

jlipps:
@mykola-mokhnach I updated the [Appium 2.0 draft spec (plugins section)](https://gist.github.com/jlipps/651b62316603400cabc95ff0f9faf70f#appium-plugins) so you can read my intentions here in prose.

lib/appium.js
jlipps: I'll see what I can do about a refactor here.

jlipps:
OK @mykola-mokhnach I cleaned up a few things per your request, did some refactoring, and finished implementing the spec as described in that gist. What do you think?

lib/appium.js
```diff
@@ -344,35 +346,142 @@ class AppiumDriver extends BaseDriver {
     }
   }
 
+  pluginsToHandleCmd (cmd) {
+    return this.plugins.filter((p) => p.commands === true || _.includes(p.commands, cmd));
```
mykola-mokhnach, line 350: ` || _.isArray(p.commands) && _.includes(p.commands, cmd)`

lib/appium.js
```diff
@@ -344,35 +346,142 @@ class AppiumDriver extends BaseDriver {
...
-    if (isAppiumDriverCommand(cmd)) {
-      return await super.executeCommand(cmd, ...args);
-    }
+    // now we define a 'handled' object which will keep track of which plugins have handled this
+    // command. we care about this because (a) multiple plugins can handle the same command, and
+    // (b) there's no guarantee that a plugin will actually call the next() method which runs the
+    // original command execution. This results in a situation where the command might be handled
+    // by some but not all plugins, or by plugin(s) but not by the default behavior. So start out
+    // this object declaring that the default handler has not been executed.
+    const handled = {default: false};
```
mykola-mokhnach, line 388: can we have a better name for this var, that makes the fact it is used by plugins more obvious?

lib/appium.js
```diff
@@ -344,35 +346,142 @@ class AppiumDriver extends BaseDriver {
...
+
+    // now we define an async function which will be passed to plugins, and successively wrapped
+    // if there is more than one plugin that can handle the command. To start off with, the async
+    // function is defined as calling the default behavior, i.e., whichever of the 3 cases above is
+    // the appropriate one
+    const defaultBehavior = async () => {
+      log.info(`Executing default handling behavior for command '${cmd}'`);
```
mykola-mokhnach, line 395: this line will be logged for each command. I don't think it is needed

lib/appium.js
```diff
@@ -344,35 +346,142 @@ class AppiumDriver extends BaseDriver {
...
+    this.logPluginHandlerReport({cmd, handled});
+
+    return res;
+  }
+
+  wrapCommandWithPlugins ({cmd, args, next, handled}) {
+    // get any plugins which are registered as handling this command
+    const plugins = this.pluginsToHandleCmd(cmd);
+
+    plugins.length && log.info(`Plugins which can handle cmd '${cmd}': ${plugins.map((p) => p.name)}`);
```
mykola-mokhnach, line 425: we should do a short return here if there are no plugins

lib/appium.js
```diff
@@ -344,35 +346,142 @@ class AppiumDriver extends BaseDriver {
...
+    // now we can go through each plugin and wrap `next` around its own handler, passing the *old*
+    // next in so that it can call it if it wants to
+    for (const plugin of plugins) {
+      // need an IIFE here because we want the value of next that's passed to plugin.handle to be
+      // exactly the value of next here before reassignment; we don't want it to be lazily
+      // evaluated, otherwise we end up with infinite recursion of the last `next` to be defined.
+      handled[plugin.name] = false; // we see a new plugin, so add it to the 'handled' object
+      next = ((_next) => async () => {
+        log.info(`Plugin ${plugin.name} is now handling cmd '${cmd}'`);
+        handled[plugin.name] = true; // if we make it here, this plugin has attempted to handle cmd
```
mykola-mokhnach, line 436: how do we guarantee that plugin names are unique? I think it would be more logical to generate uuid for each plugin and use it as a key

lib/appium.js
```diff
@@ -344,35 +346,142 @@ class AppiumDriver extends BaseDriver {
...
+      }
 
-    let res = {
-      protocol: dstSession.protocol
+      // here we know that we are executing a session command, and have a valid session driver
+      return await dstSession.executeCommand(cmd, ...args);
     };
 
+    // now take our default behavior, wrap it with any number of plugin behaviors, and run it
+    const wrappedCmd = this.wrapCommandWithPlugins({cmd, args, handled, next: defaultBehavior});
```
mykola-mokhnach, line 413: we should avoid wrapping a command if no plugins are defined for it

lib/appium.js
```diff
@@ -344,35 +346,142 @@ class AppiumDriver extends BaseDriver {
...
+    const res = await this.executeWrappedCommand({wrappedCmd, protocol});
+
+    this.logPluginHandlerReport({cmd, handled});
```
mykola-mokhnach, line 416: this should only be executed if the command was wrapped

lib/cli/extension-command.js
```diff
@@ -0,0 +1,400 @@
...
+    this.npm = new NPM(this.config.appiumHome);
+    this.knownExtensions = {}; // this needs to be overridden in final class
+  }
+
+  async execute (args) {
+    const cmd = args[`${this.type}Command`];
+    if (!_.isFunction(ExtensionCommand.prototype[cmd])) {
+      throw new Error(`Cannot handle ${this.type} command ${cmd}`);
+    }
+    return await this[cmd].bind(this)(args);
```
mykola-mokhnach, line 31: this is hard to read %)

lib/cli/extension-command.js
```diff
@@ -0,0 +1,400 @@
...
+import { log, spinWith } from './utils';
+import { INSTALL_TYPE_NPM, INSTALL_TYPE_GIT, INSTALL_TYPE_GITHUB,
+         INSTALL_TYPE_LOCAL } from '../extension-config';
+
+const UPDATE_ALL = 'installed';
+
+class NotUpdatableError extends Error {}
+class NoUpdatesAvailableError extends Error {}
+
+export default class ExtensionCommand {
```
mykola-mokhnach, line 16: it would be much easier to read this class if it had docstrings on its methods

lib/cli/parser-helpers.js
```diff
@@ -28,6 +28,17 @@ function parseSecurityFeatures (features) {
   return parsedFeatures;
 }
 
+function parsePluginNames (names) {
+  const splitter = (splitOn, str) => `${str}`.split(splitOn).map((s) => s.trim()).filter(Boolean);
```
mykola-mokhnach, line 32: I don't see a need in this method. Also, an exception here might only be thrown if `names` is not a string or it is empty/blank. I would prefer an explicit check though

lib/extension-config.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    const problemSummaries = [];
+    for (const [extName, problems] of _.toPairs(foundProblems)) {
+      if (_.isEmpty(problems)) {
+        continue;
+      }
+      // remove this extension from the list since it's not valid
+      delete exts[extName];
+      problemSummaries.push(`${this.extensionType} ${extName} had errors and will not ` +
```
mykola-mokhnach, line 56: 👍

lib/main.js
```diff
@@ -128,16 +132,33 @@ async function main (args = null) {
...
+
+  // find any plugin name which has been installed, and which has been requested for activation by
+  // using the --plugins flag, and turn each one into an instantiated plugin object, so we can send
+  // them as objects to the server init. we also want to send/assign them to the umbrella driver so
+  // it can use them to wrap command execution
+  const plugins = Object.keys(pluginConfig.installedExtensions).filter((pluginName) =>
+    _.includes(args.plugins, pluginName) ||
+    (args.plugins.length === 1 && args.plugins[0] === USE_ALL_PLUGINS)
+  ).map((pluginName) => {
+    const PluginClass = pluginConfig.require(pluginName);
```
mykola-mokhnach, line 150: I'd rather put this under try/catch

mykola-mokhnach:
it looks like tests are not very happy

lib/appium.js
jlipps: So maybe only writing it if we have plugins that might run?

lib/appium.js
jlipps: yeah, good thought

lib/appium.js
jlipps: Plugin names are guaranteed to be unique because the`appium plugin install` command won't let you install one if another one with the same name is installed. Plugin names are also keys in the yaml, so even if you edit it by hand no duplicates should show up. It's up to the ecosystem to not step on each other's toes (or as a user, you need to choose which plugin of two that register identical names you want to use).

jlipps:
> it looks like tests are not very happy

This depends on the basedriver code which has not been published, so tests will fail until that PR is merged, published, and this one updated.

lib/appium.js
mykola-mokhnach: we anyway write logs if a command is handled by a plugin

jlipps: But as we are reading the logs, we also want to see at what point a plugin allows the original command to be handled. I think keeping this line is good as long as we are running in a context with plugins.

lib/appium.js
jlipps: actually, looking at this, what is the advantage of a short return? if the list is empty, the loop won't be executed anyway. is it just that it is more clear to you?

lib/appium.js
jlipps: if no plugins are defined by it, `next` is returned as-is, so nothing is wrapped. Really this variable could be called `possiblyWrappedCmd` but that seems obtuse.

lib/cli/parser-helpers.js
jlipps: empty/blank string would be safe here. but i'll add a check for non-string.

jlipps:
ok, basedriver PR has been merged and brought in here, so let's see how CI does.

PR #14866, opened 2020-11-05
feat: Add Safari driver to the list of available automation names

mykola-mokhnach:
## Proposed changes

https://github.com/appium/appium-safari-driver

This driver could be useful if one wants to automate web pages in Safari either on Desktop or Mobile platform. It will also be necessary to update some documents, but I'm not quite sure where to...

## Types of changes


KazuCocoa (approved):
🎉

jlipps (commented):
awesome! can you make the same PR to the 2.0 branch? should just be one line

PR #14895, opened 2020-11-18
merge master into 2.0

jlipps:
This was not a straightforward merge mostly due to @mykola-mokhnach 's argparse updates. I think I've applied them all correctly to the new situation; we'll see what the tests say.


lib/cli/parser.js
```diff
@@ -5,15 +5,27 @@ import { sharedArgs, serverArgs, extensionArgs } from './args';
 import { DRIVER_TYPE, PLUGIN_TYPE } from '../extension-config';
 import { rootDir } from '../utils';
 
+
+function makeDebugParser (parser) {
```
mykola-mokhnach, line 9: is this necessary?

lib/cli/parser.js
```diff
@@ -24,31 +36,34 @@ function getParser (debug = false) {
   // add the 'driver' and 'plugin' subcommands
   addExtensionsToParser(sharedArgs, subParsers, debug);
 
-  // modify the parseArgs function to insert the 'server' subcommand if the
+  // modify the parse_args function to insert the 'server' subcommand if the
   // user hasn't specified a subcommand or the global help command
-  parser._parseArgs = parser.parseArgs;
-  parser.parseArgs = function (args, namespace) {
+  parser._parse_args = parser.parse_args;
```
mykola-mokhnach, line 41: shouldn't we use bind?

lib/cli/parser.js
jlipps: Yes, otherwise when we run tests of the parser, the entire mocha process will exit instead of just throw an error. In the past, we could include a `debug` option to parser creation, but that option was removed in argparse 2.0, so now we have to do this (or use a subclass of the parser class which overrides `exit` but we're in JS so we can just do this)

lib/cli/parser.js
jlipps: yeah, probably a good idea, even though it seemed to work fine without for forever. will update.

jlipps:
@mykola-mokhnach your questions have been addressed, anything else?

mykola-mokhnach (approved):
LGTM

PR #14936, opened 2020-12-08
feat: add --keep-alive-timeout CLI arg to adjust server connection timeout

jlipps:
This is the Appium server side of https://github.com/appium/appium-base-driver/pull/443

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -66,6 +66,15 @@ const args = [
           'on this server'
   }],
 
+  [['-ka', '--keep-alive-timeout'], {
+    required: false,
+    default: null,
+    dest: 'keepAliveTimeout',
+    type: 'int',
+    help: 'Number of seconds the Appium server should apply as both the keep-alive timeout ' +
+          'and the connection timeout for all requests. Defaults to 600 (10 minutes).'
+  }],
+
   [['-ca', '--callback-address'], {
     required: false,
     dest: 'callbackAddress',
--- a/lib/main.js
+++ b/lib/main.js
@@ -130,13 +130,17 @@ async function main (args = null) {
   await logStartupInfo(parser, args);
   let appiumDriver = new AppiumDriver(args);
   let routeConfiguringFunction = makeRouter(appiumDriver);
-  let server = await baseServer({
+  const serverOpts = {
     routeConfiguringFunction,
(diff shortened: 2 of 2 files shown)


lib/parser.js
```diff
@@ -66,6 +66,15 @@ const args = [
           'on this server'
   }],
 
+  [['-ct', '--connection-timeout'], {
```
mykola-mokhnach, line 69: this is actually --keep-alive-timeout. Lets not mix different entities

lib/parser.js
```diff
@@ -66,6 +66,15 @@ const args = [
...
+    required: false,
+    default: null,
+    dest: 'keepAliveTimeout',
+    type: 'int',
+    help: 'Number of seconds the Appium server should apply as the connection timeout for all ' +
```
mykola-mokhnach, line 74: We should mention the default value

mykola-mokhnach: connection timeout is something different. Keepalive works on the lower level than HTTP does (TCP one)

jlipps: we use the same value for both, in the code, though...?

jlipps:
Updated to `--keep-alive-timeout`

KazuCocoa (approved):
The title should also be --keep-alive-timeout :)

lib/parser.js
```diff
@@ -66,6 +66,15 @@ const args = [
           'on this server'
   }],
 
+  [['-ka', '--keep-alive-timeout'], {
+    required: false,
+    default: null,
+    dest: 'keepAliveTimeout',
+    type: 'int',
+    help: 'Number of seconds the Appium server should apply as the connection timeout for all ' +
```
mykola-mokhnach, line 74: https://stackoverflow.com/questions/49704708/what-is-a-connection-timeout-during-a-http-request

Although, keep alive is https://nodejs.org/api/http.html#http_server_keepalivetimeout

jlipps: @mykola-mokhnach I see. the main point of my change has to do with the connection timeout aspect. What would you want to see changed here to make things clear, if anything?

mykola-mokhnach: The code is fine. I just thought about tuning the help string to avoid possible misunderstandings

PR #14938, opened 2020-12-08
feat: add --keep-alive-timeout CLI arg to adjust server connection timeout

jlipps:
This is the 2.0 branch version of https://github.com/appium/appium/pull/14936/files


KazuCocoa (approved):
The title should be --kee-alive-timeout :)

PR #14985, opened 2020-12-31
docs: changelog for Appium v1.20

dpgraham:


CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
...
+
+Appium 1.20.0 is a minor release
+
+### General
+* Does not destory sockets explicitly against a client [appium-base-driver#437](https://github.com/appium/appium-base-driver/pull/437)
+* Connection timeout to each driver respects `--keep-alive-timeout` configuration [appium-base-driver#443](https://github.com/appium/appium-base-driver/pull/443). Default to 10 minutes
+
+### Android General
+* `InvalidContextError` error is thrown if running instrumentation process was dead
+* Fix Android paths on Windows [appium-adb#558](https://github.com/appium/appium-adb/pull/558)
```
mykola-mokhnach, line 12: these a paths to the emulator config files

CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
...
+
+### Android(UIAutomator2)
+* Appium adds `io.appium.settings`, `io.appium.uiautomator2.server` and `io.appium.uiautomator2.server.test` as the device's Doze whitelist to keep working [appium-uiautomator2-driver#420](https://github.com/appium/appium-uiautomator2-driver/pull/420)
+* Add capabilities:
```
mykola-mokhnach, line 16: should this list be empty?

CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
...
+
+### Android(Espresso)
+* Appium adds `io.appium.settings` and `io.appium.espressoserver.test` as the device's Doze whitelist to keep working [appium-espresso-driver#627](https://github.com/appium/appium-espresso-driver/pull/627)
+* Add `mobile:` functions:
```
mykola-mokhnach, line 20: extensions

CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
...
+* Breaking changes
+    * Support over Xcode 10.2, iOS 12.2 (Drop supporting Xcode 10.0 and 10.1)
+    * `accessibility id`, `name` and `id` lookup strategies now find elements by `name`(`wdName`) attributes in page source: [WebDriverAgent#414](https://github.com/appium/WebDriverAgent/pull/414)
+        * Previously, they found elements by `name`(`wdName`) and `value`(`wdValue`)
+        * Please use `predicate` strategy to find `value`(`wdValue`) attribute like [this change](https://github.com/appium/ruby_lib_core/pull/282)
+* Add capabilities:
+    * `resultBundlePath` and `resultBundleVersion` to allow to specify the path to the result bundle of WebDriverAgent xcodebuild [WebDriverAgent#410](https://github.com/appium/WebDriverAgent/pull/410)
+    * `safariIgnoreWebHostnames` to provide a list of hostnames that the Safari automation tools should ignore [appium-xcuitest-driver#1258](https://github.com/appium/appium-xcuitest-driver/pull/1258)
+* Add capabilities:
+    * `customSnapshotTimeout` which was renamed from `snapshotTimeout`
```
mykola-mokhnach, line 35: Actually, the whole idle mechanism handling for WDA has been rewritten.
For a user this means:
- If a custom snapshot fails (e.g. `customSnapshotTimeout` expires or if it is set to zero) then WDA falls back to the default snapshotting strategy. "Custom" is a snapshot, that includes `visible` and `enabled` attributes whose values are retrieved directly from the Accessibility manager. The default snapshotting mechanism is trying to guess these values based on an internal algorithm, which is not always precise. In previous WDA versions an error was thrown in such case and no snapshot was taken.
- A new setting/capability `waitForIdleTimeout` has been added. It is equal to 10 float seconds by default. If the application under test is blocking for longer than its value due to quiescence checks then WDA would forcefully skip them and continue with the test execution. It makes sense to change `waitForIdleTimeout` value as a workaround to run your tests if the application under test has issues with idling and it is not possible to fix the source code of it. It is not recommended to set the value to zero or too low numbers (less than 1-2 seconds) though. Important: this is still a workaround, so there is no guarantee it is going to always work. Consider rather fixing your application source code, because XCTest uses idle intervals to send commands to the accessibility manager. You may get unexpected testing results or application crashes if such intervals don't exist or are too tiny.
- Added `animationCoolOffTimeout` setting. By default it equals 2 float seconds. The value tells WDA for how long to block after a single touch action was performed, for example click, if there is an animation is following it (for example a list moving animation after swipe gesture). Zero value tells WDA to not block on animations and to immediately continue with the next test action.
- Setting `shouldWaitForQuiescence` capability to `true` effectively sets `waitForIdleTimeout` to zero and is not recommended for use

CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
...
+* Snapshot/quiescence improvements e.g. [WebDriverAgent#404](https://github.com/appium/WebDriverAgent/pull/404), [WebDriverAgent#407](https://github.com/appium/WebDriverAgent/pull/407)
+* Add a possibility to select elements by indexes [WebDriverAgent#417](https://github.com/appium/WebDriverAgent/pull/417)
+* Fix parsing SSL output from OpenSSL output [appium-xcuitest-driver#1256](https://github.com/appium/appium-xcuitest-driver/pull/1256)
+
+### iOS(Safari)
+
+https://github.com/appium/appium-safari-driver
```
mykola-mokhnach, line 42: I would rather add a basic description for each driver here, like what it is supposed to automate, which technology it is based on. Could be just a first paragraph of driver's README. And a list to the actual readme at the end

CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
...
+* Add capabilities:
+
+### Android(Espresso)
+* Appium adds `io.appium.settings` and `io.appium.espressoserver.test` as the device's Doze whitelist to keep working [appium-espresso-driver#627](https://github.com/appium/appium-espresso-driver/pull/627)
+* Add `mobile:` functions:
+    * `uiautomatorPageSource` returns the page source dump by UIAutomator [appium-espresso-driver#628](https://github.com/appium/appium-espresso-driver/pull/628)
+* Improves XPath lookup performance [appium-espresso-driver#637](https://github.com/appium/appium-espresso-driver/pull/637)
+
+### iOS(XCUITest)
+* Support M1 chip based Mac, Xcode 12.3
```
mykola-mokhnach, line 25: Snapshots caching logic has been rewritten. This should grant better performance while performing WDA API calls

docs/en/about-appium/platform-support.md
```diff
@@ -13,7 +13,7 @@ iOS automation is supported with two drivers:
 
 Please refer to these driver docs for setup instructions.
 
-* Versions: 9.0 and up (as a rule, Appium supports the latest two iOS versions)
+* Versions: 10.0 and up (as a rule, Appium supports the latest two iOS versions)
```
mykola-mokhnach, line 16: iOS 12.2 (the one corresponding to Xcode 10.2)

CHANGELOG.md
```diff
@@ -1,3 +1,66 @@
...
+* Add settings:
+    * `useResourcesForOrientationDetection` to use application resource propertires to determine the current device orientation [appium-uiautomator2-server#389](https://github.com/appium/appium-uiautomator2-server/pull/389)
+
+### Android(Espresso)
+* Appium adds `io.appium.settings` and `io.appium.espressoserver.test` as the device's Doze whitelist to keep working [appium-espresso-driver#627](https://github.com/appium/appium-espresso-driver/pull/627)
+* Add `mobile:` extensions:
+    * `uiautomatorPageSource` returns the page source dump by UIAutomator [appium-espresso-driver#628](https://github.com/appium/appium-espresso-driver/pull/628)
+* Improves XPath lookup performance [appium-espresso-driver#637](https://github.com/appium/appium-espresso-driver/pull/637)
+
+### iOS(XCUITest)
```
mykola-mokhnach, line 25: I think it also makes sense to mention https://github.com/appium/appium-ios-simulator/pull/298

CHANGELOG.md
```diff
@@ -1,3 +1,66 @@
...
+    1258)
+    * `waitForIdleTimeout` (Please read the below settings section)
+* Add settings: (Please read [Settings API](http://appium.io/docs/en/advanced-concepts/settings/index.html) for more details)
+    * `customSnapshotTimeout` which was renamed from `snapshotTimeout` sets how much time is allowed to resolve a single accessibility snapshot with custom attributes
+    * `waitForIdleTimeout` to customize the time for waiting until the application under test is idling
+        * The value `zero` (not recommended) is equal to `waitForQuiescence` to `false`
+        * **Important**: this is still a workaround, so there is no guarantee it is going to always work. Please consider rather fixing your application source code, because XCTest uses idle intervals to send commands to the accessibility manager. You may get unexpected testing results or application crashes if such intervals don't exist or are too tiny.
+    * `animationCoolOffTimeout` tells WDA for how long to block after a single touch action was performed.
```
mykola-mokhnach, line 43: not really. It would only block if there is an active animation after an action. If no animation has been detected then no blocking would occur.

CHANGELOG.md
KazuCocoa: in iOS General ✅

CHANGELOG.md
```diff
@@ -56,15 +56,13 @@ Read https://github.com/appium/appium-safari-driver for more details.
 
 ### Mac2
```
mykola-mokhnach, line 57: for Safari driver it makes sense to mention that it does not support webviews automation, while geckodriver does

mykola-mokhnach (approved):

PR #15018, opened 2021-01-14
fix: issue with driver/plugin install and plugin driver access

jlipps:
Extension install was failing mysteriously if someone happened to have a package.json or node_modules dir higher up in the directory tree than the appium home dir.

Also, plugins were getting references to the Appium umbrella driver, when they should have gotten access to the inner driver, so fix that.

### DIFF
--- a/lib/cli/npm.js
+++ b/lib/cli/npm.js
@@ -1,7 +1,7 @@
 import path from 'path';
 import semver from 'semver';
 import { exec } from 'teen_process';
-import { mkdirp, util, system } from 'appium-support';
+import { fs, mkdirp, util, system } from 'appium-support';
 
 const INSTALL_LOCKFILE = '.appium.install.lock';
 const LINK_LOCKFILE = '.appium.link.lock';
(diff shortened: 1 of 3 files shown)


mykola-mokhnach (approved):

lib/cli/npm.js
```diff
@@ -20,6 +20,15 @@ export default class NPM {
     // ensure the directory we want to install inside of exists
     await mkdirp(cwd);
 
+    // not only this, this directory needs a 'package.json' inside of it, otherwise, if any
+    // directory in the filesystem tree ABOVE cwd happens to have a package.json or a node_modules
+    // dir in it, NPM will install the module up there instead (silly NPM)
+    const dummyPkgJson = path.resolve(cwd, 'package.json');
+    const dummyPkg = {};
+    if (!await fs.exists(dummyPkgJson)) {
+      await fs.writeFile(dummyPkgJson, JSON.stringify(dummyPkg));
```
mykola-mokhnach, line 29: `await fs.writeFile(dummyPkgJson, '{}');` should be fine

package.json
```diff
@@ -11,7 +11,7 @@
     "firefoxos",
     "testing"
   ],
-  "version": "2.0.0-beta.6",
+  "version": "2.0.0-beta.7",
```
mykola-mokhnach, line 14: should the version be bumped as a part of this PR?

jlipps: i already published this version having forgotten to push the rev change, so it does not need to be a part of this PR but it is a moot point, so i am just getting it in now.

PR #15024, opened 2021-01-15
misc appium 2.0 updates

jlipps:
* Since this stuff will be the responsibility of a plugin, migrate image docs to the plugin
* Also, now that the plugin is real, include it in the list of official plugins.
* Add a postinstall script that looks for --drivers or --plugins flag and auto-installs latest versions of any official drivers or plugins in a list
* Require Node >= 12
* Bring in new basedriver and update tests/caps parsing logic due to breaking changes
* Remove wd from testsuite in favor of webdriverio


KazuCocoa (approved):
🎉

KazuCocoa:
ah, https://github.com/appium/appium/blob/6014b96ba30f500df1bbe5495ef3b5c2953825a9/test/driver-e2e-specs.js#L110 changed in fakedriver??

jlipps:
bringing in the basedriver updates broke tests, will figure it out

KazuCocoa:
`npm run test` worked. e2e is failing now

jlipps:
OK, I *think* I've fixed the e2e tests. This was a fair amount of work bringing in those w3c changes. It also necessitated the removal of wd from our testsuite in favor of webdriverio, since wd's w3c support is not good.

mykola-mokhnach:
> It also necessitated the removal of wd from our testsuite in favor of webdriverio, since wd's w3c support is not good.

I'm afraid we'd have to do the same for ALL the other drivers tests

postinstall.js
```diff
@@ -0,0 +1,57 @@
...
+  if (!driverEnv && !pluginEnv) {
+    console.log('Not auto-installing any drivers or plugins');
+    return;
+  }
+
+  let extension;
+  try {
+    extension = require('./build/lib/cli/extension');
+  } catch (e) {
+    throw new Error(`Could not load extension CLI file; has the project been transpiled? (${e})`);
```
mykola-mokhnach, line 17: e.message?

postinstall.js
```diff
@@ -0,0 +1,57 @@
...
+
+  const {DEFAULT_APPIUM_HOME, DRIVER_TYPE, PLUGIN_TYPE} = require('./build/lib/extension-config');
+  const {runExtensionCommand} = extension;
+  const appiumHome = process.env.npm_config_appium_home || DEFAULT_APPIUM_HOME;
+  const specs = [[DRIVER_TYPE, driverEnv], [PLUGIN_TYPE, pluginEnv]];
+
+  for (const [type, extEnv] of specs) {
+    if (extEnv) {
+      for (const ext of extEnv.split(',')) {
+        const extList = await runExtensionCommand({
```
mykola-mokhnach, line 28: I assume we should be more flexible here and skip possible failing exensions + log errors rather than terminating the whole postinstall process

jlipps: hmmm, yeah good point

jlipps:
> I'm afraid we'd have to do the same for ALL the other drivers tests

We'll either need to do this eventually or maintain WD better

PR #15035, opened 2021-01-20
move to different style of plugin command execution

jlipps:
This is part 2/3 of @christian-bromann's proposed change.

Plugins PR here: https://github.com/appium/appium-plugins/pull/3. Tests will mutually fail till these are merged.

### DIFF
--- a/lib/appium.js
+++ b/lib/appium.js
@@ -356,10 +356,9 @@ class AppiumDriver extends BaseDriver {
   }
 
   pluginsToHandleCmd (cmd) {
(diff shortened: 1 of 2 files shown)


jlipps:
FWIW I verified this works locally by running all unit and e2e tests with the plugins and basedriver updates applied. So will plan on merging those then re-running CI here to make it work.

jlipps:
hooray, it all seems to work

KazuCocoa (approved):

PR #15078, opened 2021-02-02
feat: allow plugins to avoid proxying certain routes

jlipps:
enables behavior like in https://github.com/appium/appium-plugins/pull/4


KazuCocoa (approved):
seems a good idea. (I also read https://github.com/appium/appium-plugins/pull/4)

jlipps:
published in beta 10

PR #15106, opened 2021-02-11
docs: Document doubleClick uia2 gesture

mykola-mokhnach:
## Proposed changes

Based on https://github.com/appium/appium-uiautomator2-driver/pull/441

## Types of changes


Prasanth-chockalingam:
Appium tooks more time to load the elements of our iOS application.

1. Generally when page as many elements the 'App source ' is not fully loaded, sometimes loading,sometimes not.
2. At times I am getting this error in appium terminal 'ECONNREFUSED 127.0.0.1:8100' .
3. On refresh it takes 10-15 mins to load the screens
For Android no issues, iOS alone i could not work with Appium inspector to inspect elements.

Can I get help on this.
Not sure this is the right place to put my queries, navigated via several link here

Issue #14441, opened 2020-06-11
Appium Monorepo Proposal


christian-bromann:
I propose to move Appiums core packages into a single repo to streamline the development of all non driver packages. I reviewed the packages in the Appium GitHub org and identified the following packages to be included:

* appium-remote-debugger
    * Dependencies:
        * appium-base-driver
        * appium-ios-device
        * appium-support
    * DevDependencies
        * appium-ios-simulator
* appium-support
* node-simctl
    * Dependencies:
        * appium-support
        * appium-xcode
* appium-xcode
    * Dependencies:
        * appium-support
* appium-adb
    * Dependencies:
        * appium-support
    * DevDependencies
* appium-ios-simulator
    * Dependencies:
        * appium-support
        * appium-xcode
* appium-base-driver
    * Dependencies:
        * appium-support
* appium-doctor
    * Dependencies:
        * appium-support
        * appium-adb
* appium-idb
    * Dependencies:
        * appium-support
* appium-ios-device
    * Dependencies:
        * appium-support
* eslint-config-appium

## Setup

The setup of the mono repo should be aligned to Lernas best practices. To avoid unnecessary complexity I recommend to not include Appiums gulp plugins and rather use NPM scripts. For example to watch all packages and compile them on file change we could use a [custom script](https://github.com/webdriverio/webdriverio/blob/master/scripts/watch.js) that watches all packages at the same time.

I also recommend to switch from the current test stack (Mocha / sinon) to Jest. It comes with an assertion and mocking library built in and can remove a lot dependencies and helper packages such as `appium-test-support`. It also simplifies tracking the coverage across the whole repository.

## Package Naming Convention

The ability to pin versions of inter dependencies turns out to be very valuable when developing in a mono repo setup. Therefor I recommend to move all packages into an `@appium` namespace, e.g.:

- `appium-remote-debugger` -> `@appium/remote-debugger`
- `node-simctl` -> `@appium/node-simctl`

This allows us to start at v2.0 across all packages and pin them to each other. At the same time it simplifies the transition to Appium 2.0.

## Testing Strategies

I am not much familiar with what is being tested in the packages mentioned above but ideally they should not need a different environment other than Node.js. For example the `appium-adb` package should not need ADB to test its functionality. It is up to the driver to ensure that their commands work on the actual environment they are suppose to control. Given that there will be a lot of tests to be run they need to run fast and should not require any special dependencies to keep it simple to execute for collaborators.

## Transition

[...]

PR #15335, opened 2021-05-06
WIP: example of retaining history for monorepo

boneskull:
## Proposed changes

I'm opening this PR so I can show an example of what retaining the history of other projects would look like when collapsing into a monorepo.  If we just copy files, of course, we lose all history.  If it is not important to retain the history, then we can skip this step.

This PR pulls `appium-doctor` into `packages/appium-doctor`.  No changes have been made to the code other than moving the files.  The files were moved via a commit in my `appium-doctor` working copy, then this changeset was merged into the `appium` repo.

Note that it's not possible to _rebase_ and retain the history, as this would require each changeset in `appium-doctor` to have _always_ been operating on files in `packages/appium-doctor` to avoid many, many conflicts. 

If doing this "for reals," we'd want the changeset added to `appium-doctor` to actually get pushed to that repo in some branch.

Please let me know what you think of this strategy.

Ref: #14441

cc @christian-bromann 

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the [CLA](https://cla.js.foundation/appium/appium)


boneskull:
(do not bother running checks on this code)

boneskull:
the main drawback I'm aware of: it adds a merge commit, so maybe less nice if you're trying to keep the history merge-free.

it's possible that some dumb git tools will fail to understand "file history" and conflate `package.json` with `packages/appium-doctor/package.json`, due to the unrelated histories.  I'll try to confirm if it's a problem.

w/r/t rebasing: when you rebase _n_ commits from one branch to another, rebasing starts with the oldest "new" changeset, and moves forward up to the latest changeset.  so the first thing it'd do is try to merge the initial commit of `appium-doctor` onto the current state of the `2.0` branch.  the initial commit of `appium-doctor` has its own `package.json` in the root (and probably many other conflicting filenames).  _every changeset_ in `appium-doctor` builds on this, so _every changeset_ will have conflicts which will need to be manually resolved when rebasing.  there _might_ be a git resolution strategy that avoids this, but I have no idea if such a thing exists or how to use it.

merging _would_ have this same problem if I didn't add a changeset in `appium-doctor` which _moved_ the entire project into a `packages/appium-doctor` subdirectory.  when merging, git doesn't care about the old changesets, e.g., when `appium-doctor`'s `package.json` was in its root--it only cares about the state of the repo in the merge's two parent changesets.  because there is no `packages/appium-doctor` in `2.0`, there are no conflicts.

jlipps:
ah, i get it now. yeah in general we prefer rebasing but for something like this a merge might be OK. i'm also honestly not too opposed to just bringing it in one commit and saying go look at this old repo for history... what do you think @mykola-mokhnach ?

mykola-mokhnach:
> ah, i get it now. yeah in general we prefer rebasing but for something like this a merge might be OK. i'm also honestly not too opposed to just bringing it in one commit and saying go look at this old repo for history... what do you think @mykola-mokhnach ?

I am fine with either approach

boneskull:
it's easier to just copy the files, of course.

of the repos that we intend to live in the monorepo, are there any that you'd specifically like to retain history for?

it's certainly possible to say "go look over there" if you want to see the history, but such a direction is not well-supported by IDEs 😄

PR #14562, opened 2020-09-16
Set up appium/appium mono repo

christian-bromann:
Set up a new monorepo for our Appium base packages, including:

* appium-support (`@appium/support` on NPM)
* appium-base-driver (`@appium/base-driver` on NPM)
* appium-doctor (`@appium/doctor` on NPM)
* eslint-config-appium (`@eslint-config-appium on NPM)

refs #14441

PR #15343, opened 2021-05-07
chore: create appium monorepo; closes #14441

### LINKED ISSUE #14441: Appium Monorepo Proposal

boneskull:
## UPDATE - See https://github.com/appium/appium/pull/15343#issuecomment-843529470

_This PR supersedes PR #14562, but ought to be very similar._

## Proposed changes

This is my first attempt at making this repo a monorepo and moving appium-proper into `packages/appium`.

I had tried to make this work with npm's "workspaces", but it seems lifecycle scripts don't run as you'd expect them to upon install.  Instead, I'm using Lerna, which provides more features for monorepos.

Changes include:

- The root `package.json` is now a package named `appium-monorepo`, which is `private`.  The monorepo should not be published.
- All dev dependencies remain in the root `package.json`.
- `gulpfile.js` _moved_ into `packages/appium/gulpfile.js`, but this should be **temporary**--`appium-gulp-plugins` makes certain assumptions which a monorepo violates, and I was unable to get it working without patching `appium-gulp-plugins`.
- Most of the `package.json` `scripts` moved into `packages/appium/package.json` due to the above.  This too should be considered **temporary**.
- Adds a root `postinstall` lifecycle script which calls `lerna bootstrap` to install/link packages.  I don't see a `package-lock.json` in `packages/appium/`; this needs investigating.
- Adds a root `test` lifecycle script which executes `npm run test` in all packages.  The tests do not pass, but I don't know if this is my fault.
- The `clean` script will purge all `node_modules` folders in each package, as well.
- I removed `scripts/release*.sh`, since it seems they would fail anyway--`lerna` should be used for releases instead
- The shrinkwrap-related stuff is untested.
- Renamed `tags` prop in `packages/appium/package.json` to `keywords`.  I don't know if `tags` is supported.
- Removed an unused `projectRoot` prop in the gulpfile.

## Types of changes

## Checklist


boneskull:
Looks like I need to sort out the `package-lock.json` issue w/ Lerna.

boneskull:
Next steps:

1. [ ] Fix CI / Lerna
2. [ ] Move `appium-gulp-plugins` into the monorepo.
3. [ ] Fix `appium-gulp-plugins` to work properly with this setup
4. [ ] Move scripts and `gulpfile.js` back into root.
5. [ ] Get tests passing

jlipps:
Let me know once you get to a place where you're reasonably sure your changes aren't causing test failures and I'll help investigate any. There's some pretty wild stuff that happens in test setup in some places that may be making various assumptions and need rethinking.

jlipps (commented):
things generally look great here. one thought is that the docs can be moved to the root outside packages (potentially with some of their own tooling). it's not a package and not related to anything else. more like a set of dev scripts. or if we ultimately make the docs a sort of template app that allows driver/plugin authors to publish their own docs, it would be its own package. either way just a note to flag a future improvement of moving docs generation etc out of the main appium package.

lerna.json
```diff
@@ -0,0 +1,26 @@
+{
+  "packages": [
+    "packages/*"
+  ],
+  "version": "2.0.0-beta.1",
```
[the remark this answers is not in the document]
boneskull: this only applies if we are not going to version the subpackages separately.  I basically just copied it from what Christian did.  unless there's a real need (shrinkwrap, perhaps?), keeping the versions in sync will probably be more trouble than it's worth (lerna advises against this).

the publishing flow is still TBD anyway--I am hoping to defer the decision for now

scripts/release-branch.sh
```diff
@@ -1,13 +0,0 @@
-git fetch
```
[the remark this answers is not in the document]
boneskull: okay, I can revert them, but the scripts seem to point to directories that don't exist, so I figured it was unused

boneskull:
hey, build's green!

jlipps:
Just thinking of some other tasks for this project in case you haven't got them listed:

- [ ] deprecate any old NPM packages that have been moved to the `@appium` scope
- [ ] add a 'Repo moved' notice to the top of the readme of all the ported repos
- [ ] move the old repos to the appium-boneyard org to help keep our org clean

boneskull:
I think we'll want to pull `fake-driver` in here too, then after that I'll merge the history for the base driver.

boneskull:
wait, guess not. it's somewhere else

boneskull:
# UPDATE

## packages now in monorepo

- `appium`
- `@appium/base-driver`
- `@appium/doctor`
- `@appium/eslint-config-appium`
- `@appium/fake-driver`
- `@appium/gulp-plugins`
- `@appium/support`
- `@appium/test-support`

All package names have changed _except_ `appium`.

## Notes

- The entire Git history from `appium-base-driver` is part of this PR.  Before merging, I [moved _all_ files in this project into a subdir, `packages/base-driver`](https://github.com/appium/appium/pull/15343/commits/a55d3136341a05b602939f08d90d6671ea540324). 

- Some devDeps are duplicated between the root monorepo (where _all_ dev deps should live) and `@appium/gulp-plugins`, which wants to provide its own as _dependencies_.  Moving these dev deps into the root enables IDE integration (debugging, eslint, mocha, etc), whereas an IDE will have a difficult time finding the dev deps in `packages/gulp-plugins/node_modules`.

- Lerna's "hoist" mode does not seem to improve installation time.

- The `prepare` scripts (transpilation) to be run for each package (except the eslint config) mean npm workspaces cannot be used at this time (or likely ever).  npm workspaces work differently than Lerna's `bootstrap`--in Lerna, the root packages are installed, _then_ the leaf packages are installed.  npm installs leaves _first_ (like a depth-first search), which means the dev deps wouldn't be in place anyway to run `prepare`.  On the one hand, forcing a build upon install isn't really ideal, but neither is npm's behavior. `¯\_(ツ)_/¯` 

## Future Work (Not part of this PR)

- Figure out what to do with the `README`s
- Design & document new publishing workflow
- Refactor doc gen
- Consolidate or remove Gulp entirely; dedupe dev deps
- Future of `@appium/test-support`?
- Build & test optimizations

jlipps:
Another task for later:
* update all packages which depend on the moved packages to use the new name (and push a deprecation notice to the old names?). This should be the case for all the moved packages except for base driver (since the basedriver for Appium 1.x has not been moved, and can still be developed/published as appium-base-driver, while we continue appium 2.0 basedriver work with `@appium/base-driver`

.eslintrc
```diff
@@ -1,3 +1,14 @@
...
+    {
+      "files": "packages/fake-driver/**/*",
+      "rules": { "require-await": "off" }
+    },
+    { "files": "packages/support/**/*", "globals": { "BigInt": "readonly" } },
+    {
+      "files": "packages/*/test/**/*",
+      "rules": { "func-names": "off" }
+    }
+  ]
```
boneskull, line 13: @jlipps 

this file consolidates all `.eslintrc` files across the packages.

however, the `lint` gulp task also does whatever it does.  I'm a little concerned `eslint --fix` won't have the same result as `gulp lint --fix`.  it's _probably_ not a problem, but we should keep an eye open for linting churn.

babel.config.json
```diff
@@ -0,0 +1,39 @@
+{
```
boneskull, line 1: @jlipps

`babel.config.json` (or `babel.config.js`) is the "root" babel config as of Babel v7.  a `.babelrc` file in any package will merge or override (not clear...) the two config files.  

this is roughly the same as the `.babelrc` file in `gulp-plugins` -- but trying to output some source maps here for debugging & stack traces.  I'm _pretty sure_ this file is not yet used anywhere _except_ when debugging in an IDE.

babel.config.json
```diff
@@ -0,0 +1,39 @@
...
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "10"
```
mykola-mokhnach, line 7: it makes sense to bump it to 12 now. Probably, in another PR

docs/en/contributing-to-appium/appium-from-source.md
```diff
@@ -11,10 +11,11 @@ request; for more information on how to run tests, keep reading!
 Appium is written in JavaScript, and run with the [Node.js](https://nodejs.org/) engine. Currently
 version 6+ is supported. While Node.js can be installed globally on the system,
 a version manager is _highly_ recommended.
+
 * NVM - [https://github.com/creationix/nvm](https://github.com/creationix/nvm)
 * N - [https://github.com/tj/n](https://github.com/tj/n)
 
-Your Node.js installation will include the [NPM](https://www.npmjs.com/) package manager, which Appium
+Your Node.js installation will include the [npm](https://www.npmjs.com/) package manager, which Appium
```
mykola-mokhnach, line 18: version 6+

docs/en/contributing-to-appium/appium-from-source.md
```diff
@@ -174,4 +156,4 @@ the commit from occurring.
 
 Once code is committed and a [pull request](https://help.github.com/articles/about-pull-requests/)
 is made to the correct Appium respository on [GitHub](https://github.com/), Appium build system
-will run all of the functional tests.
+will run all of the functional tests.
```
mykola-mokhnach, line 159: nit: trailing spaces

package.json
```diff
@@ -1,112 +1,98 @@
...
+    "lerna": "^4.0.0",
+    "lint-staged": "^11.0.0",
+    "mjpeg-server": "^0.3.0",
     "mocha": "^8.0.1",
     "pre-commit": "1.x",
+    "querystring": "^0.2.0",
+    "serve-static": "^1.13.1",
     "sinon": "^9.0.0",
     "validate.js": "^0.13.0",
+    "wd": "^1.14.0",
```
mykola-mokhnach, line 90: note: this is something we don't want to depend anymore :)

package.json
```diff
@@ -1,112 +1,98 @@
 {
```
mykola-mokhnach, line 1: does dependabot handle monorepo dependencies updates properly?

packages/appium/gulpfile.js
```diff
@@ -47,8 +46,7 @@ boilerplate({
   test: {
     files: ['${testDir}/**/*-specs.js']
   },
-  testTimeout: 160000,
-  preCommitTasks: ['eslint', 'once'],
```
mykola-mokhnach, line 51: so we got rid of all precommit tasks or we still keep the eslint one?

packages/appium/lib/config.js
```diff
@@ -44,11 +44,23 @@ async function updateBuildInfo (useGithubApiFallback = false) {
   }
 }
 
+/**
+ * Finds the Git metadata dir (see `GIT_META_ROOT`)
+ *
+ * This is needed because Appium cannot assume `package.json` and `.git` are in the same
+ * directory.  Monorepos, see?
```
mykola-mokhnach, line 51: 🙈

packages/appium/lib/config.js
```diff
@@ -44,11 +44,23 @@ async function updateBuildInfo (useGithubApiFallback = false) {
...
+ * @returns {string|void} Path to dir or `undefined` if not found
```
mykola-mokhnach, line 52: we usually put it as `?string`

packages/appium/lib/config.js
```diff
@@ -44,11 +44,23 @@ async function updateBuildInfo (useGithubApiFallback = false) {
...
+ */
+async function findGitRoot () {
+  return await findUp(GIT_META_ROOT, {cwd: rootDir, type: 'directory'});
```
mykola-mokhnach, line 55: cannot we use the existing glob package from appium-support for this purpose instead of adding a new dependency?

packages/appium/test/cli-e2e-specs.js
```diff
@@ -4,10 +4,14 @@ import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import { tempDir, fs, mkdirp, util } from 'appium-support';
 import { KNOWN_DRIVERS } from '../lib/drivers';
+import { PROJECT_ROOT as cwd } from './helpers';
 
 chai.should();
 chai.use(chaiAsPromised);
-const cwd = path.resolve(__dirname, '..', '..');
+
+// cannot use `require.resolve()` here (w/o acrobatics) due to the ESM context.
```
mykola-mokhnach, line 12: acrobatics 😄

packages/appium/test/helpers.js
```diff
@@ -0,0 +1,35 @@
+import path from 'path';
+import {insertAppiumPrefixes} from '../lib/utils';
+import findUp from 'find-up';
+import getPort from 'get-port';
+
+const TEST_HOST = 'localhost';
```
mykola-mokhnach, line 6: I would rather prefer 127.0.0.1
Some hosts may resolve to ipv6 address by default, which we don't want

packages/appium/test/helpers.js
```diff
@@ -0,0 +1,35 @@
...
+const W3C_PREFIXED_CAPS = {...insertAppiumPrefixes(BASE_CAPS)};
+const W3C_CAPS = {
+  alwaysMatch: {...W3C_PREFIXED_CAPS},
+  firstMatch: [{}],
+};
+
+let TEST_PORT;
+/**
+ * Returns a free port; one per process
+ * @returns {Promise<number>} a free port
```
mykola-mokhnach, line 29: we skip `Promise` and only put the actual value type, since an IDE knows anyway that the result of an `async` function is always a promise

packages/appium/test/plugin-e2e-specs.js
```diff
@@ -8,26 +8,31 @@ import { remote as wdio } from 'webdriverio';
...
 
 chai.should();
 chai.use(chaiAsPromised);
-const FAKE_PLUGIN_DIR = path.resolve(__dirname, '..', '..', 'node_modules', '@appium', 'fake-plugin');
-const FAKE_DRIVER_DIR = path.resolve(__dirname, '..', '..', 'node_modules', 'appium-fake-driver');
-const TEST_SERVER = `http://${TEST_HOST}:${TEST_PORT}`;
+
+const FAKE_PLUGIN_DIR = path.join(PROJECT_ROOT, 'node_modules', '@appium', 'fake-plugin');
+const FAKE_DRIVER_DIR = path.join(PROJECT_ROOT, 'packages', 'fake-driver');
+let TEST_SERVER;
```
mykola-mokhnach, line 19: I would prefer to define variables closer to the scope where they are used (`describe` in such case)

docs/en/contributing-to-appium/appium-from-source.md
boneskull: we should adopt prettier, or something 😄

package.json
boneskull: ¯\_(ツ)_/¯ I'll research this

docs/en/contributing-to-appium/appium-from-source.md
mykola-mokhnach: I  have simply enabled trimming in my VSCode

package.json
boneskull: yeah, need to update fake-driver.  that will be a separate PR

packages/appium/gulpfile.js
boneskull: the only precommit task is now eslint, and it's defined in `package.json` only

packages/appium/lib/config.js
boneskull: we should lint the docstrings 😄

packages/appium/lib/config.js
boneskull: I mean, we could, but we'd end up writing our own `find-up`.

packages/appium/test/cli-e2e-specs.js
boneskull: not sure if it's even possible with babel.  babel lets you do things you couldn't otherwise, but I don't think it supports all of Node.js' native ESM handling

packages/appium/test/helpers.js
boneskull: that's true that the IDE knows, but if we're being explicit about `return await` it would be consistent to be explicit about this, yes?

packages/appium/test/plugin-e2e-specs.js
boneskull: ok

boneskull:
The CLA check is "pending" because apparently some _other_ commit authors have not yet signed it.  Or at least, that's my best guess.  The bot is dumb.

packages/appium/test/helpers.js
boneskull: actually, I think I'm going to leave it, since `localhost` is hardcoded throughout the tests and it will be a pretty sizable diff.

boneskull: see https://github.com/appium/appium/issues/15399

packages/appium/lib/config.js
boneskull: afaict there's no wildcard that tells glob/minimatch to navigate to an arbitrary parent directory from cwd.  you'd need to start from the fs root and work forwards, which would be...slow.  so you'd need to search for `.git` using glob, and if it's not there, navigate to `..`, then search again (but skip the directory we just came from), and do so in a loop.  I imagine this is basically what `find-up` is doing

packages/appium/lib/config.js
boneskull: `?string` means "`string` or `null`", and this function never returns `null`.

package.json
boneskull: it looks like it doesn't handle this very well at present time.  

renovate does, however, and I think that provides something similar. I'll investigate in a subsequent PR.

packages/appium/lib/config.js
mykola-mokhnach: this function might return undefined though

package.json
boneskull: see https://github.com/appium/appium/issues/15400

docs/en/contributing-to-appium/appium-from-source.md
boneskull: trailing spaces are significant in github-flavored markdown, so that might be bad.  even with an `.editorconfig` file -- which most editors/IDEs support -- you'd have to skip trimming markdown.  prettier is smart enough to know how to handle this.

packages/appium/test/helpers.js
mykola-mokhnach: I never use Promise notation anywhere for async returns. Yolo :)

babel.config.json
boneskull: see https://github.com/appium/appium/issues/15401

packages/appium/lib/config.js
mykola-mokhnach: np, I'm just usually being a bit paranoidal about adding new dependecies because some of them might exponentially increase the size of the linked modules tree %)

packages/appium/lib/config.js
jlipps: isn't that what `void` means?

packages/appium/test/helpers.js
jlipps: we should just port everything to typescript and be done with it :trollface:

jlipps (approved):

mykola-mokhnach (approved):

christian-bromann (commented):
Some comments:

- `packages/base-driver/lib/basedriver/...` vs. `packages/basedriver/lib/basedriver/desired-caps.js `?
- is it worth maybe moving appium-express server package into its own sub package or will it be completely absorbed by the basedriver? Maybe it makes sense anyway, my feeling is that if we can keep sub packages reasonable small it simplifies contributing to it - it also helps in cases e.g. where we want to migrate technologies like move to TypeScript where it allows us to migrate package by package

packages/appium/test/helpers.js
boneskull: @jlipps imo it's not needed.  properly-written docstrings are easily understood by typescript, and you can just use it as a type linter (and you get that info in the IDE!).   I wrote [report-toolkit](https://github.com/ibm/report-toolkit) this way.

PR #15356, opened 2021-05-11
feat: add "run" command to the client interface inorder to allow the …

khanayan123:
…running of scripts against drivers

## Proposed changes

To create a generic script running CLI command that allows commands of the form appium driver run <driverName> <scriptName>.

## Types of changes

## Checklist

- [] I have added necessary documentation (if appropriate)

## Further comments
I have not added necessary documentation yet. Wanted to get some feedback on the PR first. The E2E tests in this PR require the merging of this branch: https://github.com/appium/appium-fake-driver/pull/76 
in order to work.


jlipps (changes requested):
looking really good! i made some larger comments and will look at tests once we have those other things hammered out.

lib/cli/args.js
```diff
@@ -415,11 +415,27 @@ function makeUpdateArgs (type) {
   ];
 }
 
+function makeRunArgs (type) {
+  return [
+    ...globalExtensionArgs,
+    [[type], {
+      type: 'str',
+      help: `Name of the ${type} to run a script from, for example: ` + type === DRIVER_TYPE ? DRIVER_EXAMPLE : PLUGIN_EXAMPLE,
```
jlipps, line 423: Break this into different lines for readability as above, I think our max length is around 100 chars

lib/cli/extension-command.js
```diff
@@ -486,4 +487,49 @@ export default class ExtensionCommand {
...
+  /**
+   * Runs a script cached inside the "scripts" field under "appium"
+   * inside of the driver/plugins "package.json" file
+   *
+   * @param {string} ext - name of the extension to run a script from
+   * @param {string} scriptName - name of the script to run
+   */
+  async run ({ext, scriptName}) {
+    if (!_.has(this.config.installedExtensions, ext)) {
+      const msg = `please install the ${this.type} first`;
```
jlipps, line 500: no need to define the string and throw in a separate line. just put the string literal in the Error constructor

lib/cli/extension-command.js
```diff
@@ -486,4 +487,49 @@ export default class ExtensionCommand {
...
+    const extConfig = this.config.installedExtensions[ext];
+
+    if (!_.has(extConfig, 'scripts')) {
+      throw new Error(`The ${this.type} named '${ext}' does not contain the ` +
+                      `"scripts" field underneath the "appium" field in its package.json`);
+    }
+
+    const extScripts = extConfig.scripts;
+
+    if (!(scriptName in extScripts)) {
```
jlipps, line 513: use `_.has` as in the other places

lib/cli/extension-command.js
```diff
@@ -486,4 +487,49 @@ export default class ExtensionCommand {
...
+      throw new Error(`The ${this.type} named '${ext}' does not support the script: '${scriptName}'`);
+    }
+
+    const runner = new SubProcess(`node`, [extScripts[scriptName]],
+                                  {cwd: this.config.getExtensionRequirePath(ext)});
+
+    runner.on('stream-line', (line) => {
+      log(this.isJsonOutput, line);
```
jlipps, line 521: here we can also keep an array of output lines that we can use in the JSON return

lib/cli/extension-command.js
```diff
@@ -486,4 +487,49 @@ export default class ExtensionCommand {
...
+    });
+
+    await runner.start();
+
+    try {
+      await runner.join();
+      log(this.isJsonOutput, `${scriptName} successfuly ran`.green);
+      return {scriptName};
```
jlipps, line 529: it would be more useful to return the output than just the script name, as an array of output lines that we capture above

lib/cli/extension-command.js
```diff
@@ -486,4 +487,49 @@ export default class ExtensionCommand {
...
+    } catch (err) {
+      log(this.isJsonOutput, `Encountered an error when running the script: ${err.message}`.red);
+      return {error: err.message};
```
jlipps, line 532: in addition to the error message, it would be nice to include any output in the JSON response, since the error message might be something generic like 'exited with code 1'.

jlipps (changes requested):

test/cli-e2e-specs.js
```diff
@@ -142,4 +143,87 @@ describe('Driver CLI', function () {
...
+    it('should run a valid plugin, valid script, and result in success', async function () {
+      await clear();
+      const pluginName = 'fake';
+      const scriptName = 'fake-success';
+      await run('install', [fakePluginDir, '--source', 'local', '--json']);
+      const out = JSON.parse(await run('run', [pluginName, scriptName, '--json']));
+      out.should.not.include.key('error');
+    });
+    it('should run a valid plugin, valid error prone script, and return error in json', async function () {
+      await clear();
```
jlipps, line 212: in this and the next test, we theoretically don't need to clear and install, right? what about just clearing and installing once in a `before` for this whole describe?

jlipps (approved):

jlipps:
also, looks like CI is failing. any ideas?

ayvnkhan:
> also, looks like CI is failing. any ideas?
I'd assume it is because this branch https://github.com/appium/appium-plugins/pull/7 hasn't been merged yet.

jlipps:
now these changes need to be applied to the new appium directory location in the 2.0 branch

packages/appium/lib/cli/extension-command.js
```diff
@@ -486,4 +487,51 @@ export default class ExtensionCommand {
...
+    const extConfig = this.config.installedExtensions[ext];
+
+    if (!_.has(extConfig, 'scripts')) {
+      throw new Error(`The ${this.type} named '${ext}' does not contain the ` +
+                      `"scripts" field underneath the "appium" field in its package.json`);
+    }
+
+    const extScripts = extConfig.scripts;
+
+    if (!_.has(extScripts, scriptName)) {
```
mykola-mokhnach, line 512: I'd also check if it is a plain object here

packages/appium/lib/cli/extension-command.js
```diff
@@ -486,4 +487,51 @@ export default class ExtensionCommand {
...
     await this.config.write();
   }
-}
+
+  /**
+   * Runs a script cached inside the "scripts" field under "appium"
+   * inside of the driver/plugins "package.json" file
+   *
+   * @param {string} ext - name of the extension to run a script from
+   * @param {string} scriptName - name of the script to run
```
mykola-mokhnach, line 496: missing the description of the returned result and of possible thrown exceptions

packages/appium/lib/cli/extension-command.js
```diff
@@ -486,4 +487,51 @@ export default class ExtensionCommand {
...
+    const extScripts = extConfig.scripts;
+
+    if (!_.has(extScripts, scriptName)) {
+      throw new Error(`The ${this.type} named '${ext}' does not support the script: '${scriptName}'`);
+    }
+
+    const runner = new SubProcess(`node`, [extScripts[scriptName]],
+                                  {cwd: this.config.getExtensionRequirePath(ext)});
+
+    const output = [];
```
mykola-mokhnach, line 519: I'm not quite sure we'd like to keep the whole output. It might create OOM in the worst case. I'd prefer to limit it's size and organize a ring buffer

packages/appium/lib/cli/extension-command.js
```diff
@@ -486,4 +487,51 @@ export default class ExtensionCommand {
...
+
+    runner.on('stream-line', (line) => {
+      output.push(line);
+      log(this.isJsonOutput, line);
+    });
+
+    await runner.start();
```
mykola-mokhnach, line 526: this might miserably fail if there was no output for more than 5 seconds

packages/appium/lib/cli/extension-command.js
```diff
@@ -509,29 +519,39 @@ export default class ExtensionCommand {
 
     const extScripts = extConfig.scripts;
 
+    if (!_.isPlainObject(extScripts)) {
+      throw new Error(`"scripts" field must be a plain object`);
```
mykola-mokhnach, line 523: I would also add the extension name here, because there might be multiple

packages/appium/lib/cli/extension-command.js
```diff
@@ -509,29 +519,39 @@ export default class ExtensionCommand {
...
       output.push(line);
       log(this.isJsonOutput, line);
     });
 
-    await runner.start();
+    try {
+      // start and wait 5 seconds before throwing an error
+      await runner.start(null, 5 * 1000);
+    } catch (err) {
+      log(this.isJsonOutput, `Encountered an error when running the script: ${err.message}`.red);
```
mykola-mokhnach, line 544: I would also add the script name here

packages/appium/lib/cli/utils.js
```diff
@@ -52,9 +52,24 @@ async function spinWith (json, msg, fn) {
   }
 }
 
+let createRingBuffer = function (length) {
```
mykola-mokhnach, line 55: is there any particular reason why this must be an anonymous function?

packages/appium/lib/cli/utils.js
```diff
@@ -52,9 +52,24 @@ async function spinWith (json, msg, fn) {
...
+
+  let pointer = 0, buffer = [];
+
+  return {
+    getBuf () {return buffer;},
+    get (key) {return buffer[key];},
```
mykola-mokhnach, line 61: I would call it index rather than key

packages/appium/lib/cli/utils.js
```diff
@@ -52,9 +52,24 @@ async function spinWith (json, msg, fn) {
   }
 }
 
+let createRingBuffer = function (length) {
+
+  let pointer = 0, buffer = [];
+
+  return {
+    getBuf () {return buffer;},
```
mykola-mokhnach, line 60: is linter happy about it?

packages/appium/lib/cli/extension-command.js
```diff
@@ -488,12 +488,22 @@ export default class ExtensionCommand {
     await this.config.write();
   }
 
+  /**
+   * @typedef RunOutput
+   * @property {string|undefined} error - error message if script ran unsuccessfuly, otherwise undefined
+   * @property {array} output - script output
```
mykola-mokhnach, line 494: `{string[]} or {Array<string>}

packages/appium/lib/cli/extension-command.js
```diff
@@ -488,12 +488,22 @@ export default class ExtensionCommand {
     await this.config.write();
   }
 
+  /**
+   * @typedef RunOutput
+   * @property {string|undefined} error - error message if script ran unsuccessfuly, otherwise undefined
```
mykola-mokhnach, line 493: I usually use `{?string}`, but the current one is also ok

packages/appium/lib/cli/utils.js
```diff
@@ -52,9 +52,24 @@ async function spinWith (json, msg, fn) {
   }
 }
 
+let createRingBuffer = function (length) {
+
+  let pointer = 0, buffer = [];
```
mykola-mokhnach, line 57: I prefer to define each variable in a separate row to improve readability

packages/appium/lib/cli/utils.js
```diff
@@ -52,9 +52,24 @@ async function spinWith (json, msg, fn) {
...
+
+  return {
+    getBuf () {return buffer;},
+    get (key) {return buffer[key];},
+    push (item) {
+      buffer[pointer] = item;
+      pointer = (length + pointer + 1) % length;
```
mykola-mokhnach, line 64: I think it might be confusing if we start recording recent logs to the beginning of the buffer. I would rather shift the buffer to the left and delete the very first record in order to add the last one, so that the original order is preserved

packages/appium/lib/cli/extension-command.js
```diff
@@ -486,4 +487,71 @@ export default class ExtensionCommand {
...
+    const output = createRingBuffer(50);
+
+    runner.on('stream-line', (line) => {
+      output.push(line);
+      log(this.isJsonOutput, line);
+    });
+
+    try {
+      // start and wait 5 seconds before throwing an error
+      await runner.start(null, 5 * 1000);
```
mykola-mokhnach, line 542: I usually avoid creating magic number and put them to constants instead

packages/appium/lib/cli/extension-command.js
```diff
@@ -486,4 +487,71 @@ export default class ExtensionCommand {
...
+
+    const output = createRingBuffer(50);
+
+    runner.on('stream-line', (line) => {
+      output.push(line);
+      log(this.isJsonOutput, line);
+    });
+
+    try {
+      // start and wait 5 seconds before throwing an error
```
mykola-mokhnach, line 541: I'm actually not sure whether we need to wait for any output here at all. A script is not obliged to print anything. I would simply use `start(0)` and let it throw when `join` is called

packages/appium/lib/cli/extension-command.js
```diff
@@ -491,7 +491,7 @@ export default class ExtensionCommand {
   /**
    * @typedef RunOutput
    * @property {?string|undefined} error - error message if script ran unsuccessfuly, otherwise undefined
-   * @property {array} output - script output
+   * @property {string[]|Array<string>} output - script output
```
mykola-mokhnach, line 494: I meant one of these types annotations is ok, not both of them at once :)

packages/appium/lib/cli/utils.js
```diff
@@ -52,9 +52,30 @@ async function spinWith (json, msg, fn) {
...
+    this.buffer = [];
+  }
+  getBuff () {
+    return this.buffer;
+  }
+  dequeue () {
+    this.buffer.shift();
+  }
+  enqueue (item) {
+    if (this.buffer.length === this.size) {
```
mykola-mokhnach, line 68: `>=` condition looks safer to me

packages/appium/lib/cli/extension-command.js
```diff
@@ -486,4 +487,65 @@ export default class ExtensionCommand {
...
+
+    if (!_.isPlainObject(extScripts)) {
+      throw new Error(`The ${this.type} named '${ext}' "scripts" field must be a plain object`);
+    }
+
+    if (!_.has(extScripts, scriptName)) {
+      throw new Error(`The ${this.type} named '${ext}' does not support the script: '${scriptName}'`);
+    }
+
+    const runner = new SubProcess(`node`, [extScripts[scriptName]],
```
mykola-mokhnach, line 530: nit: we usually format such arguments like: 
```
const runner = new SubProcess(`node`, [extScripts[scriptName]], {
  cwd: this.config.getExtensionRequirePath(ext)
});
```

mykola-mokhnach (approved):
had some minor comments

packages/appium/lib/cli/utils.js
```diff
@@ -52,9 +52,30 @@ async function spinWith (json, msg, fn) {
   }
 }
 
+class RingBuffer {
+  constructor (size = 50) {
+    this.size = size;
+    this.pointer = 0;
```
jlipps, line 58: `this.pointer` doesn't seem to be used anywhere. should we delete it?

packages/appium/test/cli-e2e-specs.js
```diff
@@ -15,6 +15,7 @@ const executable = path.join(cwd, 'packages', 'appium', 'build', 'lib', 'main.js
 
 describe('Driver CLI', function () {
   let appiumHome;
+  const localFakeDriverPath = path.resolve(cwd, 'packages', 'fake-driver');
```
jlipps, line 18: this resolution strategy will work only if running tests from the main appium directory. what if we're running them inside packages/appium? It might be better to do something like `path.resolve(__dirname, '..', '..', '..', 'fake-driver')` (with an extra '..' in the list because the transpiled code will run from `packages/appium/build`, not `packages/appium`)

jlipps: i'm ok if we leave this as it is right now, i understand you were just following the existing strategy.

packages/appium/test/cli-e2e-specs.js
```diff
@@ -146,4 +146,81 @@ describe('Driver CLI', function () {
...
+      const driverName = 'foo';
+      await chai.expect(run('run', [driverName, 'bar', '--json'])).to.eventually.be.rejectedWith(Error);
+    });
+  });
+});
+
+describe('Plugin CLI', function () {
+  let appiumHome;
+  const fakePluginDir = path.resolve(cwd, 'node_modules', '@appium', 'fake-plugin');
+  before(async function () {
```
jlipps, line 180: we duplicate all this before, after, run, and clear logic. seems like we could easily make them shared helper functions? (and there's probably no downside to sharing `appiumHome` between the driver and plugin tests here.)

jlipps (changes requested):
everything looks great! just a couple thoughts as well as a request to deduplicate some of the test code.

jlipps (approved):

jlipps:
looks like you're out of CI probation now at least :-)

ayvnkhan:
> looks like you're out of CI probation now at least :-)

haha thank god

packages/appium/lib/cli/extension-command.js
```diff
@@ -486,4 +487,51 @@ export default class ExtensionCommand {
...
+                      `"scripts" field underneath the "appium" field in its package.json`);
+    }
+
+    const extScripts = extConfig.scripts;
+
+    if (!_.has(extScripts, scriptName)) {
+      throw new Error(`The ${this.type} named '${ext}' does not support the script: '${scriptName}'`);
+    }
+
+    const runner = new SubProcess(`node`, [extScripts[scriptName]],
```
mykola-mokhnach, line 516: Consider using `process.execPath` instead of `node` as the latter won't work on Windows. See https://stackoverflow.com/questions/49934315/how-to-get-path-of-nodejs-executable-on-runtime

packages/appium/lib/cli/extension-command.js
mykola-mokhnach: still not resolved. Please pass a single `0` argument to the `start` call to avoid any checks

packages/appium/lib/cli/extension-command.js
mykola-mokhnach: still not resolved. Is the method supposed to return a string a an array of strings?

packages/appium/lib/cli/parser.js
```diff
@@ -98,8 +98,16 @@ function addExtensionsToParser (sharedArgs, subParsers, debug = false) {
...
       {command: 'update', args: extensionArgs[type].update,
        help: `Update installed ${type}s to the latest version`},
+      {command: 'run', args: extensionArgs[type].run,
+       help: `Run a script (cached inside the ${type}s package.json under the ` +
+             `“scripts” field inside the “appium” field) from an installed ${type}`}
     ];
 
+    if (type === DRIVER_TYPE) {
+      parserSpecs.push({command: 'run', args: extensionArgs[type].run,
+                        help: `Run a script (cached inside the ${type}s package.json under the “scripts” field inside the “appium” field) against an installed ${type}`});
```
mykola-mokhnach, line 108: this line is too long

packages/appium/lib/cli/extension-command.js
ayvnkhan: Seems like I must've messed up during git rebase, let me try again these comments should have been resolved

packages/appium/lib/cli/parser.js
```diff
@@ -98,6 +98,9 @@ function addExtensionsToParser (sharedArgs, subParsers, debug = false) {
        help: `Uninstall a ${type}`},
       {command: 'update', args: extensionArgs[type].update,
        help: `Update installed ${type}s to the latest version`},
+      {command: 'run', args: extensionArgs[type].run,
+       help: `Run a script (cached inside the ${type}s package.json under the ` +
```
jlipps, line 102: I would say `defined` rather than `cached`. I'm not sure what cached would mean here.

jlipps (approved):

packages/appium/lib/cli/parser.js
jlipps: also, `${type}'s`

jlipps (approved):

jlipps:
\o/ nice work!

PR #15386, opened 2021-05-18
feat: define appium home for plugins to handle multiple drivers

KazuCocoa:
## Proposed changes

This issue is still WIP.

When I managed different versions' drivers and plugins on a host, I got a use case I wanted to use a particular driver version with plugins. To reduce installation time, we can have multiple `appium-home` on a host for drivers and can switch them when an appium server starts. 

For now, plugins could have multiple modules in a session, but drivers could be used only one module in a session.
So, like the below case could do.

```
appium --appium-home=/path/to/driver/home --appium-plugin-home=/path/to/plugin/home
```

```
appium --appium-home=/path/to/driver/which/has/another/version --appium-plugin-home=/path/to/plugin/home
```

This case may help to reduce installation time to manage multiple driver versions on a host and use the same plugin versions across them. This case may happen frequently than managing a plugin as various versions.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/lib/cli/args.js
+++ b/lib/cli/args.js
@@ -24,6 +24,13 @@ const sharedArgs = [
     dest: 'appiumHome',
   }],
 
+  [['-aph', '--appium-plugin-home'], {
+    required: false,
+    default: null,
+    help: 'The path to the directory where Appium will keep installed plugins. "--appium-home" is more general configuration. This argument can be used when you want to manage plugins separately.',
+    dest: 'appiumPluginHome',
+  }],
+
   [['--log-filters'], {
     dest: 'logFilters',
     default: null,
--- a/lib/cli/extension.js
+++ b/lib/cli/extension.js
@@ -22,14 +22,13 @@ async function runExtensionCommand (args, type) {
   if (!extCmd) {
     throw new TypeError(`Cannot call ${type} command without a subcommand like 'install'`);
   }
-  let {json, suppressOutput, appiumHome} = args;
+  let {json, suppressOutput, appiumHome, appiumPluginHome} = args;
   if (suppressOutput) {
     json = true;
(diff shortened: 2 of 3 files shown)


jlipps (commented):
this looks interesting, is there a scenario where we would also want `--appium-driver-home`? also does using this new flag do the right thing in terms of creating an appium manifest file in the plugin home directory, etc?

jlipps:
also we probably want to rebase this into the monorepo setup after that is merged

KazuCocoa:
> --appium-driver-home

I considered the case, but probably only for plugins is enough for now.
As appium, drivers are mondatory to enable `automationName`s. So `--appium-home` can think this home is mainly for drivers. if we want to manage driver versions, e.g. for development and for production in our development case, we can define multiple `--appium-home` for each usage. Plugins are kindly an optional, so we can extract plugin version manageemnt outside drivers. 

> does using this new flag do the right thing in terms of creating an appium manifest file in the plugin home directory, etc?

Still considering.
Probably its better to split `drivers` and `plugins` from `extensions.yaml` for each directory so far.

jlipps:
@KazuCocoa do you want to rebase this on the monorepo?

lib/cli/args.js
```diff
@@ -24,6 +24,13 @@ const sharedArgs = [
     dest: 'appiumHome',
   }],
 
+  [['-aph', '--appium-plugin-home'], {
```
Dharin-shah, line 27: `--plugin-home` as an option also required i believe

KazuCocoa:
will rebase & update this PR this weekend or around :)

KazuCocoa:
=> https://github.com/appium/appium/pull/15500

PR #15388, opened 2021-05-18
Add driver plugin server arg injection ability

khanayan123:
## Proposed changes

Include ability for appium to inject user cli arguments to the driver/plugin

## Types of changes

## Checklist

## Further comments

I have not added necessary documentation yet, waiting for feedback. This feature depends on these branches to be merged: 
https://github.com/appium/appium-fake-driver/pull/77
https://github.com/appium/appium-base-driver/pull/486
https://github.com/appium/appium-plugins/pull/9


test/driver-e2e-specs.js
```diff
@@ -77,6 +85,53 @@ describe('FakeDriver - via HTTP', function () {
     });
   });
 
+  describe('cli args handling', function () {
+    it('should recieve user cli args for driver if passed in', async function () {
+      let driver = await wdio({...wdOpts, capabilities: caps});
+      const {sessionId} = driver;
+      try {
+        const {data} = await axios.get(`${baseUrl}/${sessionId}/fakecliargs`);
+        JSON.stringify(data.value).should.eql('{}');
```
jlipps, line 94: might be more simpler to say `data.value.should.eql({})`?

jlipps: (we don't need to test `JSON.stringify`)

test/driver-e2e-specs.js
```diff
@@ -77,6 +85,53 @@ describe('FakeDriver - via HTTP', function () {
...
+    it('should recieve empty user cli args from a driver if no arguments were passed in', async function () {
+      await serverClose();
+      let args = {port: TEST_PORT, host: TEST_HOST, appiumHome, driverArgs: FAKE_DRIVER_ARGS};
+      await serverStart(args);
+
+      let driver = await wdio({...wdOpts, capabilities: caps});
+      const {sessionId} = driver;
+      try {
+        const {data} = await axios.get(`${baseUrl}/${sessionId}/fakecliargs`);
+        JSON.stringify(data.value).should.eql(FAKE_ARGS);
```
jlipps, line 111: if we treated fake args as an object before we don't need to stringify here for comparison

test/parser-specs.js
```diff
@@ -83,6 +83,27 @@ describe('Server Parser', function () {
     parsed.allowInsecure.should.eql(['feature1', 'feature2', 'feature3']);
     parsed.denyInsecure.should.eql(['nofeature1', 'nofeature2', 'nofeature3']);
   });
+  it('should parse default driver args correctly from a string', function () {
+    let fakeArgs = `{"sillyWebServerPort":1234,"host":"hey"}`;
+    let fakeDriverName = `fake`;
+    let fakeDriverArgs = `{"${fakeDriverName}": ${fakeArgs}}`;
```
jlipps, line 89: i would probably compose all these as objects and then stringify in the parse invocation

test/parser-specs.js
```diff
@@ -83,6 +83,27 @@ describe('Server Parser', function () {
...
+    let args = p.parse_args(['--driver-args', fakeDriverArgs]);
+    args.driverArgs.should.eql(JSON.parse(fakeDriverArgs));
+  });
+  it('should parse default driver args correctly from a file', function () {
+    let fakeArgs = `{"sillyWebServerPort":1234,"host":"hey"}`;
+    let fakeDriverName = `fake`;
+    let fakeDriverArgs = `{"${fakeDriverName}": ${fakeArgs}}`;
+    let args = p.parse_args(['--driver-args', 'test/fixtures/driverArgs.json']);
```
jlipps, line 97: probably want to `path.resolve(__dirname, 'fixtures', 'driverArgs.json')` so it works on windows and isn't dependent on the cwd

jlipps: (may also need to go up to the root and back down because .json fixtures might not be copied into the transpiled code)

test/plugin-e2e-specs.js
```diff
@@ -146,4 +147,80 @@ describe('FakePlugin', function () {
...
+        await server.close();
+      }
+    });
+
+    it('should recieve user cli args for plugin if passed in', async function () {
+      const driver = await wdio(wdOpts);
+      const {sessionId} = driver;
+
+      try {
+        const {data} = await axios.get(`${baseUrl}/${sessionId}/fakecliargs`);
```
jlipps, line 169: now that I think about it I'm a little concerned that both the driver and the plugin establish the same new GET path. would be nice if they were a bit different so that we were sure in a scenario like this we are getting the plugin's CLI args and not fakedriver's

jlipps (changes requested):

docs/en/writing-running-appium/server-args.md
```diff
@@ -40,20 +40,20 @@ All flags are optional, but some are required in conjunction with certain others
 |`--nodeconfig`|null|Configuration JSON file to register appium with selenium grid|`--nodeconfig /abs/path/to/nodeconfig.json`|
 |`-ra`, `--robot-address`|0.0.0.0|IP Address of robot|`--robot-address 0.0.0.0`|
 |`-rp`, `--robot-port`|-1|port for robot|`--robot-port 4242`|
-|`--chromedriver-port`|9515|Port upon which ChromeDriver will run|`--chromedriver-port 9515`|
-|`--chromedriver-executable`|null|ChromeDriver executable full path||
+|`--driver-args '{"uiautomator2": {"chromedriver-port": 9515}}`|9515|Port upon which ChromeDriver will run|`--driver-args '{"uiautomator2": {"chromedriver-port": 9515}}`|
```
jlipps, line 43: the appium 2.0 documentation will not include driver specific info like this; we should move these values to a table in the uiautomator2 driver readme!

docs/en/writing-running-appium/server-args.md
```diff
@@ -40,20 +40,20 @@ All flags are optional, but some are required in conjunction with certain others
...
+|`--driver-args '{"uiautomator2": {"chromedriver-executable": '/path/to/file'}}`|null|ChromeDriver executable full path||
```
jlipps, line 44: since we're doing JSON, we should use camelCase for all arg names, so `chromedriverExecutable`

docs/en/writing-running-appium/server-args.md
```diff
@@ -40,20 +40,20 @@ All flags are optional, but some are required in conjunction with certain others
...
 |`--isolate-sim-device`|false|Xcode 6 has a bug on some platforms where a certain simulator can only be launched without error if all other simulator devices are first deleted. This option causes Appium to delete all devices other than the one being used by Appium. Note that this is a permanent deletion, and you are responsible for using simctl or xcode to manage the categories of devices used with Appium.||
 |`--tmp`|null|Absolute path to directory Appium can use to manage temporary files, like built-in iOS apps it needs to move around. On *nix/Mac defaults to /tmp, on Windows defaults to C:\Windows\Temp||
 |`--trace-dir`|null|Absolute path to directory Appium use to save ios instruments traces, defaults to <tmp dir>/appium-instruments||
 |`--debug-log-spacing`|false|Add exaggerated spacing in logs to help with visual inspection||
-|`--suppress-adb-kill-server`|false|(Android-only) If set, prevents Appium from killing the adb server instance||
+|`--driver-args '{"uiautomator2": {"suppress-adb-kill-server": false}}'`|false|(Android-only) If set, prevents Appium from killing the adb server instance||
```
jlipps, line 52: `suppressAdbKillServer`

lib/appium.js
```diff
@@ -155,7 +155,8 @@ class AppiumDriver extends BaseDriver {
 
       const {
         driver: InnerDriver,
-        version: driverVersion
+        version: driverVersion,
+        driverName: driverName
```
jlipps, line 159: when we have `key: key` we can replace it with just `key`, so this can be:

```js
const {
  driver: InnerDriver,
  version: driverVersion,
  driverName
} = ...
```

lib/cli/parser-helpers.js
```diff
@@ -44,7 +44,7 @@ function parsePluginNames (names) {
   }
 }
 
-function parseDefaultCaps (capsOrPath) {
+function parseJsonStringOrFile (capsOrPath) {
```
jlipps, line 47: 👍

lib/utils.js
```diff
@@ -33,6 +34,15 @@ function inspectObject (args) {
   }
 }
 
+function parseExtensionArgs (extensionArgs, extensionName) {
+  const parsedExtensionArgs = _.isString(extensionArgs) ? parseJsonStringOrFile(extensionArgs) : {};
+  const extensionSpecificArgs = _.has(parsedExtensionArgs, extensionName) ? parsedExtensionArgs[extensionName] : {};
+  if (!_.isPlainObject(extensionSpecificArgs)) {
+    throw new Error(`The passed in extension argument is not a plain object.`);
```
jlipps, line 41: `Driver or plugin arguments must be plain objects`

test/driver-e2e-specs.js
```diff
@@ -20,6 +19,8 @@ chai.use(chaiAsPromised);
 
 const TEST_SERVER = `http://${TEST_HOST}:${TEST_PORT}`;
 const FAKE_DRIVER_DIR = path.resolve(__dirname, '..', '..', 'node_modules', 'appium-fake-driver');
+const FAKE_ARGS = {'sillyWebServerPort': 1234, 'host': 'hey'};
```
jlipps, line 22: no need for quotes around the keys in this line and the next

jlipps (changes requested):
couple small things, this is looking really good!

jlipps:
oh, and we need to apply all this work on top of the new state of 2.0, which will involve putting it all in the `packages/appium` directory, and hopefully not much else!

docs/en/writing-running-appium/server-args.md
mykola-mokhnach: if these capabilities are the same as driver capabilities then it makes sense to just provide default capabilities json instead to simplify their validation. Also, I would prefer to have a possibility to load that stuff either from JSON or YAML file, which might be more convenient taking into account the amount of formatting that should be applied to the command line

lib/utils.js
```diff
@@ -33,6 +34,15 @@ function inspectObject (args) {
   }
 }
 
+function parseExtensionArgs (extensionArgs, extensionName) {
+  const parsedExtensionArgs = _.isString(extensionArgs) ? parseJsonStringOrFile(extensionArgs) : {};
```
mykola-mokhnach, line 38: there is no point to continue here if args are empty

lib/utils.js
```diff
@@ -33,6 +34,15 @@ function inspectObject (args) {
...
+  const extensionSpecificArgs = _.has(parsedExtensionArgs, extensionName) ? parsedExtensionArgs[extensionName] : {};
```
mykola-mokhnach, line 39: same here

mykola-mokhnach: it is ok to use `parsedExtensionArgs[extensionName]` here since it will be undefined if the item does not exist

docs/en/writing-running-appium/server-args.md
mykola-mokhnach: it makes sense to either set default capabilities in args rather than set each value separately. It will also help with input validation

packages/appium/lib/utils.js
```diff
@@ -33,6 +34,18 @@ function inspectObject (args) {
   }
 }
 
+function parseExtensionArgs (extensionArgs, extensionName) {
+  if (!_.isString(extensionArgs)) {
+    return {};
+  }
+  const parsedExtensionArgs = parseJsonStringOrFile(extensionArgs);
+  const extensionSpecificArgs = parsedExtensionArgs[extensionName] ? parsedExtensionArgs[extensionName] : {};
```
mykola-mokhnach, line 42: the ternary operator usage is redundant here. `const extensionSpecificArgs = parsedExtensionArgs[extensionName]` is enough

mykola-mokhnach: otherwise the following condition looks suspicious

packages/base-driver/lib/basedriver/commands/find.js
```diff
@@ -60,115 +47,6 @@ commands.findElementsFromElement = async function findElementsFromElement (strat
   return await this.findElOrElsWithProcessing(strategy, selector, true, elementId);
 };
 
-/**
```
mykola-mokhnach, line 63: should these change be in that PR?

docs/en/writing-running-appium/server-args.md
```diff
@@ -40,20 +40,18 @@ All flags are optional, but some are required in conjunction with certain others
 |`--nodeconfig`|null|Configuration JSON file to register appium with selenium grid|`--nodeconfig /abs/path/to/nodeconfig.json`|
 |`-ra`, `--robot-address`|0.0.0.0|IP Address of robot|`--robot-address 0.0.0.0`|
 |`-rp`, `--robot-port`|-1|port for robot|`--robot-port 4242`|
-|`--chromedriver-port`|9515|Port upon which ChromeDriver will run|`--chromedriver-port 9515`|
-|`--chromedriver-executable`|null|ChromeDriver executable full path||
+
 |`--show-config`|false|Show info about the appium server configuration and exit||
 |`--no-perms-check`|false|Bypass Appium's checks to ensure we can read/write necessary files||
 |`--strict-caps`|false|Cause sessions to fail if desired caps are sent in that Appium does not recognize as valid for the selected device||
 |`--isolate-sim-device`|false|Xcode 6 has a bug on some platforms where a certain simulator can only be launched without error if all other simulator devices are first deleted. This option causes Appium to delete all devices other than the one being used by Appium. Note that this is a permanent deletion, and you are responsible for using simctl or xcode to manage the categories of devices used with Appium.||
```
mykola-mokhnach, line 47: we should think about cleaning up some outdated/unused args as well as driver-specific ones

mykola-mokhnach: cc @jlipps

jlipps: yep, definitely don't want people using `--isolate-sim-device` anymore!

packages/appium/test/driver-e2e-specs.js
```diff
@@ -81,6 +88,53 @@ describe('FakeDriver - via HTTP', function () {
...
+      } finally {
+        await driver.deleteSession();
+      }
+    });
+  });
+
+  describe('cli args handling for different driver', function () {
+    it('should recieve empty user cli args from a driver if cli arguments created for a different driver', async function () {
+      await serverClose();
+      let args = {port: TEST_PORT, host: TEST_HOST, appiumHome, driverArgs: `{"xcuitest": ${JSON.stringify(FAKE_ARGS)}}`};
```
jlipps, line 124: this would make more sense as:

```js
let args = {..., driverArgs: JSON.stringify({xcuitest: FAKE_ARGS})}
```

jlipps: also, in general i see you using `let` a lot when you're not actually changing the value of the variable later on. our rule is to basically use `const` in every case. if you find you need to change the value of a variable, then you can go back up and change it to `let`.

jlipps (changes requested):
looking really good! just a few more thoughts

packages/appium/lib/appium.js
```diff
@@ -405,7 +407,9 @@ class AppiumDriver extends BaseDriver {
     sessionId = _.truncate(sessionId, {length: 11});
     return this.pluginClasses.map((PluginClass) => {
       const name = `${PluginClass.pluginName} (${sessionId})`;
-      return new PluginClass(name);
+      const opts = {};
```
mykola-mokhnach, line 410: could be 

```
const opts = {
  pluginArgs: parseExtensionArgs...
};
```

packages/appium/lib/cli/args.js
```diff
@@ -1,14 +1,14 @@
...
-  parseSecurityFeatures, parseDefaultCaps,
+  parseSecurityFeatures, parseJsonStringOrFile,
   parsePluginNames, parseInstallTypes, parseDriverNames
 } from './parser-helpers';
 import {
   INSTALL_TYPES, DEFAULT_APPIUM_HOME,
   DRIVER_TYPE, PLUGIN_TYPE
 } from '../extension-config';
 import {
-  DEFAULT_CAPS_ARG, StoreDeprecatedDefaultCapabilityAction
```
mykola-mokhnach, line 11: is this action still used anywhere?

packages/appium/lib/cli/parser-helpers.js
```diff
@@ -58,7 +58,7 @@ function parsePluginNames (names) {
   }
 }
 
-function parseDefaultCaps (capsOrPath) {
+function parseJsonStringOrFile (capsOrPath) {
```
mykola-mokhnach, line 61: The error message here is confusing for driverArgs, since it says about capabilities. We should either have a possibility to customize it via arguments or make it neutral

docs/en/writing-running-appium/server-args.md
```diff
@@ -15,83 +15,28 @@ All flags are optional, but some are required in conjunction with certain others
...
-|`-r`, `--backend-retries`|3|(iOS-only) How many times to retry launching Instruments before saying it crashed or timed out|`--backend-retries 3`|
 |`--session-override`|false|Enables session override (clobbering)||
-|`-l`, `--pre-launch`|false|Pre-launch the application before allowing the first session (Requires --app and, for Android, --app-pkg and --app-activity)||
 |`-g`, `--log`|null|Also send log output to this file|`--log /path/to/appium.log`|
 |`--log-level`|debug|Set the server log level for console and logfile (specified as `console-level:logfile-level`, with both being the same if only one value is supplied). Possible values are `debug`, `info`, `warn`, `error`, which are progressively less verbose.|`--log-level error:debug`|
 |`--log-timestamp`|false|Show timestamps in console output||
 |`--local-timezone`|false|Use local timezone for timestamps||
 |`--log-no-colors`|false|Do not use colors in console output||
 |`-G`, `--webhook`|null|Also send log output to this HTTP listener|`--webhook localhost:9876`|
-|`--safari`|false|(IOS-Only) Use the safari app||
```
mykola-mokhnach, line 34: finally 🎉

packages/appium/lib/cli/args.js
ayvnkhan: in lib/config.js

packages/appium/lib/cli/parser-helpers.js
```diff
@@ -82,8 +82,8 @@ function parseJsonStringOrFile (capsOrPath) {
     return result;
   } catch (e) {
     const msg = loadedFromFile
-      ? `Default capabilities in '${capsOrPath}' must be a valid JSON`
-      : `Default capabilities must be a valid JSON`;
+      ? `Passed in argument '${capsOrPath}' must be a valid JSON`
```
mykola-mokhnach, line 85: How about `The provided value of 'capsOrPath' ...` ?

mykola-mokhnach:
The CI seem to be still unhappy

ayvnkhan:
this branch depends on https://github.com/appium/appium-base-driver/pull/495

jlipps:
Most recent CI failure:

```
1) FakeDriver - via HTTP
       cli args handling for empty args
         should not recieve user cli args if none passed in:
     Error: Request failed with status code 404
      at createError (node_modules/axios/lib/core/createError.js:16:15)
      at settle (node_modules/axios/lib/core/settle.js:17:12)
      at IncomingMessage.handleStreamEnd (node_modules/axios/lib/adapters/http.js:260:11)
      at IncomingMessage.emit (domain.js:470:12)
      at endReadableNT (internal/streams/readable.js:1336:12)
      at processTicksAndRejections (internal/process/task_queues.js:82:21)
```

Seems different than before; in this case it looks like the updated fake driver is not being used?

packages/fake-driver/lib/driver.js
```diff
@@ -6,15 +6,16 @@ import commands from './commands';
...
+  constructor (opts = {}, shouldValidateCaps = true, driverArgs = {}) {
+    super(opts, shouldValidateCaps, driverArgs);
     this.appModel = null;
     this.curContext = 'NATIVE_APP';
     this.elMap = {};
     this.focusedElId = null;
     this.maxElId = 0;
     this.caps = {};
     this.fakeThing = null;
+    this.driverArgs = driverArgs;
```
jlipps, line 18: maybe we should add a TODO that we won't need this line once we're relying completely on basedriver from appium 2.0

jlipps (approved):

packages/fake-driver/lib/driver.js
boneskull: wait... I thought we _were_ completely relying on basedriver from appium 2.0?

jlipps: in general, drivers themselves are not yet restricted to use in appium 2.0 only. unless we want to fork each driver so that there is the 2.0 version of the driver and the 1.x version of the driver.

PR #15420, opened 2021-05-24
configure lerna for publishing; closes #15403

### LINKED ISSUE #15403: publishing workflow for monorepo
This needs implementation & documentation thereof.

#15402 depends upon this task.

boneskull:
1. Uses `sync-monorepo-packages` to keep `README.md` and `packages/appium/README.md` in sync.  Since `README` gets published to npm but _also_ shows in the GitHub repo, both places need a `README`. In lieu of a custom `README`, we can just sync it for now.  This is called via a `preversion` script.
1. Various fields are synced in `package.json`.  This does not include dependencies—but those _in common_ should probably be kept in sync as well.
1. Lerna uses conventional commits and Git tags to figure out what to do.  It should automatically determine what needs releasing and what version it should be released as.
1. Tags have been pushed to the repo for Lerna to use.  In the case where the changesets are functionally identical to the latest code in the old repositories, the version number has been retained, and the tag is of the format `[@scope/]<package-name>@<version>`.  In the case where the old repos had _unreleased_ changesets, I’ve tagged with a new patch release and adopted the `-rc.0` suffix.  So if we were at v5.5.0 in `appium-gulp-plugins`, `@appium/gulp-plugins` is tagged as `@appium/gulp-plugins@5.5.1-rc.0`.  I have not modified the version in `package.json`.
 1. As with adopting any new release/publishing strategy, we should be meticulous about what we’re doing until we’re comfortable with how it works.  Lerna supports dry runs; we should use them.
1. Documentation on usage will arrive in next changes to this PR.
1. We should try to use `lerna-changelog`; I was unsuccessful at getting it to do anything on my first try, so that can be a separate PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,17 @@
   "version": "0.0.0",
   "private": true,
   "description": "Automation for Apps.",
+  "keywords": [
+    "automation",
+    "javascript",
+    "selenium",
+    "webdriver",
+    "ios",
+    "android",
+    "firefoxos",
+    "testing"
+  ],
+  "homepage": "https://appium.io",
   "bugs": {
     "url": "https://github.com/appium/appium/issues"
   },
@@ -27,6 +38,10 @@
     "precommit-msg": "echo 'Pre-commit checks...' && exit 0",
     "reinstall": "npm run clean && npm install",
     "start": "node ./packages/appium",
+    "sync-pkgs": "npm run sync-pkgs:appium-readme && npm run sync-pkgs:common-fields && npm run sync-pkgs:keywords",
+    "sync-pkgs:appium-readme": "sync-monorepo-packages --force --no-package-json --packages=packages/appium README.md",
+    "sync-pkgs:common-fields": "sync-monorepo-packages --fields=author,license,repository,bugs,homepage,engines",
+    "sync-pkgs:keywords": "sync-monorepo-packages --field=keywords --packages='packages/*' --packages='!packages/eslint-config-appium'",
     "test": "lerna exec gulp once",
     "test:appium": "lerna exec --scope=appium gulp once",
     "test:base-driver": "lerna exec --scope=@appium/base-driver gulp once",
@@ -40,6 +55,7 @@
     "test:gulp-plugins": "lerna exec --scope=@appium/gulp-plugins gulp once",
     "test:support": "lerna exec --scope=@appium/support gulp once",
     "test:test-support": "lerna exec --scope=@appium/test-support gulp once",
+    "preversion": "npm run sync-pkgs && git add -A packages/appium/README.md packages/*/package.json",
     "watch": "lerna exec watch"
   },
   "pre-commit": [
@@ -86,6 +102,7 @@
     "querystring": "^0.2.0",
     "serve-static": "^1.13.1",
     "sinon": "^9.0.0",
+    "sync-monorepo-packages": "^0.3.2",
     "validate.js": "^0.13.0",
     "wd": "^1.14.0",
     "webdriverio": "^6.12.0",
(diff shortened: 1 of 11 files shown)


packages/base-driver/package.json
```diff
@@ -53,9 +53,12 @@
...
     "webdriverio": "^6.0.2",
     "ws": "^7.0.0"
-  },  
+  },
   "devDependencies": {
     "@appium/gulp-plugins": "^5.4.0",
     "@appium/eslint-config-appium": "^4.7.0"
+  },
+  "publishConfig": {
+    "access": "public"
```
jlipps, line 62: the 2.0 branch of basedriver (which this is based on) should probably only be published as "beta" until we explicitly opt out of that mode.

boneskull: ok

boneskull: wait--do you mean `next`?

boneskull:
OK, so, it's going to be difficult to write up the docs without doing an initial publish.

Once we merge this, running:

```bash
npx lerna publish from-package
```

will publish the scoped packages. e.g.

```bash
lerna notice cli v4.0.0
lerna info versioning independent
lerna WARN Unable to determine published version, assuming "@appium/doctor" unpublished.
lerna WARN Unable to determine published version, assuming "@appium/base-driver" unpublished.
lerna WARN Unable to determine published version, assuming "@appium/eslint-config-appium" unpublished.
lerna WARN Unable to determine published version, assuming "@appium/fake-driver" unpublished.
lerna WARN Unable to determine published version, assuming "@appium/gulp-plugins" unpublished.
lerna WARN Unable to determine published version, assuming "@appium/test-support" unpublished.
lerna WARN Unable to determine published version, assuming "@appium/support" unpublished.

Found 7 packages to publish:
 - @appium/base-driver => 8.0.0-beta.6
 - @appium/doctor => 1.16.0
 - @appium/eslint-config-appium => 4.7.0
 - @appium/fake-driver => 2.2.0
 - @appium/gulp-plugins => 5.5.0
 - @appium/support => 2.53.0
 - @appium/test-support => 1.3.3

? Are you sure you want to publish these packages? (ynH)
```

Thereafter, I think we'll need to feel it out.  If we want to _only release one package_, for example, the workflow would need to be different than just publishing the changes to all packages at once.  It _shouldn't really matter_ if we publish a bunch of new versions at once, but idk.

package.json
```diff
@@ -40,6 +55,7 @@
     "test:gulp-plugins": "lerna exec --scope=@appium/gulp-plugins gulp once",
     "test:support": "lerna exec --scope=@appium/support gulp once",
     "test:test-support": "lerna exec --scope=@appium/test-support gulp once",
+    "preversion": "npm run sync-pkgs && git add -A packages/appium/README.md packages/*/package.json",
```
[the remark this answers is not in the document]
boneskull: this doesn't sync dependencies; just a bunch of non-critical metadata fields.  keeping dependencies in sync is another tool 😄

boneskull:
OK, this is ready for now.  As mentioned [above](https://github.com/appium/appium/pull/15420#issuecomment-847334952), I'll be able to document the publishing workflow after we get those scoped packages published.

package.json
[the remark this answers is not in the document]
jlipps: 👍

boneskull:
One thing I'm not sure of is if we configure `lerna version` _not_ to immediately push the tag, will `lerna publish` _also_ not push the tag to GH, but _only_ publish to npm?  I don't really love automatic tag-pushing when versioning, but with publishing I want it.

jlipps (approved):

jlipps:
i guess we'll need to try it and see. probably not a difficult PR to Lerna if we want a different option

jlipps:
Alright, initial publish ran successfully

PR #15426, opened 2021-05-25
startRecordingScreen() not available on Javascript

joshuapinter:
When trying to call this you get `startRecordingScreen is not a function`.

The docs are misleading and they point to the erroneous line in the commands file. I'm not sure why it's not supported on Javascript but it should at least be better documented that it isn't.

## Proposed changes

Docs change.

## Types of changes

## Checklist

- [ ] I have signed the [CLA](https://cla.js.foundation/appium/appium)

## Further comments


jlipps:
Seems to be supported in wdio still: https://webdriver.io/docs/api/appium#startrecordingscreen

jlipps:
If you get an error when using that client library, I would submit an issue to WDIO instead.

joshuapinter:
Yeah, it's confusing to know where the core issue lies. I found some corroboration with Bitrise on their article for recording your UI tests as well: 

image

https://devcenter.bitrise.io/tips-and-tricks/screen-recording-your-ui-tests-with-appium/

I'll post an issue there and link it back to here but we're on the latest versions of everything and it just doesn't appear to be supported. And these docs don't point to anywhere meaningful like they do for something like `saveScreenshot`, etc.

joshuapinter:
I dug a little more into this. First off, we are using `wd` and not `wdio`. Looking at `wdio`'s docs, I think you're right, they definitely seem to support this. But `wd` definitely does not.

So, I'll update this commit to keep the `wdio` references and show that `wd` does not support it. 

I really wish they did because the migration from `wd` to `wdio` is not insignificant but I would love to be able to use `startRecordingScreen` without rolling our own solution.

jlipps:
Unfortunately WD is no longer maintained so sooner or later you'll probably want to move to WDIO anyway.

joshuapinter:
Thanks for the heads up, @jlipps. That's what I was gleaning as well. I'll put it on The List. :)

mykola-mokhnach (approved):

PR #15433, opened 2021-05-27
update releasing documentation

boneskull:
This is...about as good as I can figure out without actually doing it myself.


jlipps (approved):
LGTM. sensible and thorough. i can give it an actual try soon.

KazuCocoa (approved):
🎉

PR #15436, opened 2021-05-28
chore: update @appium/fake-driver to use @appium/base-driver

### LINKED ISSUE #15425: @appium/fake-driver: update to use @appium/base-driver & wdio
Currently `@appium/fake-driver` is on the pre-appium-2.x-compatible branch of `appium-base-driver`. It needs to use `@appium/base-driver`, which is appium-2.x-compatible.  

I don't know how involved this is going to be, but I'm going to look at it anyway.  This may be split into two PRs.

boneskull:
BREAKING CHANGE: `fake-driver` now depends upon `@appium/base-driver@8.x`

## `@appium/fake-driver`

- need to use w3c capabilities only
- switch from `wd` to `wdio` in tests
- many, many fixes to support these

## `@appium/base-driver`

- fixed a dead URL in a comment
- updated the "logging" tests to manually supply w3c capabilities.  `createSession()` does it for you, but `executeCommand('createSession')` does not.
- display the name of the driver under test when executing base driver's test suite with other drivers
- add missing `releaseActions` endpoint
 
Ref: #15436

### DIFF
--- a/packages/base-driver/test/basedriver/driver-tests.js
+++ b/packages/base-driver/test/basedriver/driver-tests.js
@@ -12,7 +12,10 @@ chai.use(chaiAsPromised);
 // wrap these tests in a function so we can export the tests and re-use them
 // for actual driver implementations
 function baseDriverUnitTests (DriverClass, defaultCaps = {}) {
-  describe('BaseDriver', function () {
+  // to display the driver under test in report
+  const className = DriverClass.name || '(unknown driver)';
+
+  describe(`BaseDriver (as ${className})`, function () {
     let d, w3cCaps;
 
     beforeEach(function () {
@@ -75,12 +78,12 @@ function baseDriverUnitTests (DriverClass, defaultCaps = {}) {
       let sessions = await d.getSessions();
 
       sessions.length.should.equal(1);
-      sessions[0].should.eql({
-        id: d.sessionId,
-        capabilities: {
-          'deviceName': 'Commodore 64',
-          'platformName': 'Fake'
-        }
+      sessions[0].should.include({
+        id: d.sessionId
+      });
+      sessions[0].capabilities.should.include({
+        'deviceName': 'Commodore 64',
+        'platformName': 'Fake'
       });
     });
 
@@ -393,7 +396,7 @@ function baseDriverUnitTests (DriverClass, defaultCaps = {}) {
       beforeEach(async function () {
         beforeStartTime = Date.now();
         d.shouldValidateCaps = false;
-        await d.executeCommand('createSession', null, null, defaultCaps);
+        await d.executeCommand('createSession', null, null, {alwaysMatch: {...defaultCaps}, firstMatch: [{}]});
       });
       describe('#eventHistory', function () {
         it('should have an eventHistory property', function () {
--- a/packages/base-driver/test/basedriver/driver-tests.js
+++ b/packages/base-driver/test/basedriver/driver-tests.js
@@ -82,8 +82,8 @@ function baseDriverUnitTests (DriverClass, defaultCaps = {}) {
         id: d.sessionId
       });
       sessions[0].capabilities.should.include({
-        'deviceName': 'Commodore 64',
-        'platformName': 'Fake'
+        deviceName: 'Commodore 64',
+        platformName: 'Fake'
       });
     });
 
--- a/packages/fake-driver/lib/fake-app.js
+++ b/packages/fake-driver/lib/fake-app.js
@@ -1,5 +1,5 @@
-import {fs} from '@appium/support';
-import {readFileSync, existsSync} from 'fs';
+import { fs } from '@appium/support';
+import { readFileSync, existsSync } from 'fs';
 import path from 'path';
 import XMLDom from 'xmldom';
 import xpath from 'xpath';
--- a/packages/fake-driver/lib/fake-app.js
+++ b/packages/fake-driver/lib/fake-app.js
@@ -1,12 +1,10 @@
-import B from 'bluebird';
-import fs from 'fs';
+import {readFile} from 'fs/promises';
+import {readFileSync} from 'fs';
 import path from 'path';
 import XMLDom from 'xmldom';
 import xpath from 'xpath';
 import log from './logger';
 import { FakeElement } from './fake-element';
-
-const readFile = B.promisify(fs.readFile);
 
 const SCREENSHOT = path.resolve(__dirname, '..', '..', 'screen.png');
 
@@ -71,7 +69,7 @@ class FakeApp {
   }
 
   async loadApp (appPath) {
-    log.info('Loading Mock app model');
+    log.info(`Loading Mock app model at ${appPath}`);
     let data = await readFile(appPath);
     log.info('Parsing Mock app XML');
     this.rawXml = data.toString();
@@ -151,7 +149,7 @@ class FakeApp {
   }
 
   getScreenshot () {
-    return fs.readFileSync(SCREENSHOT, 'base64');
+    return readFileSync(SCREENSHOT, 'base64');
   }
 
 }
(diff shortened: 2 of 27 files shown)


packages/fake-driver/lib/logger.js
```diff
@@ -1,4 +1,4 @@
-import { logger } from 'appium-support';
+import { logger } from '@appium/support';
```
boneskull, line 1: I am not entirely sure why this wasn't caught before.

packages/fake-driver/test/helpers.js
```diff
@@ -1,17 +1,21 @@
...
-
-const TEST_HOST = 'localhost';
+const TEST_HOST = '127.0.0.1';
 const TEST_PORT = 4774;
-const TEST_APP = path.resolve(__dirname, '..', '..', 'test', 'fixtures', 'app.xml');
+
+// XXX: what dir are we in? it depends on how we're running this test. the .xml file doesn't move, but we do!
+// we may be running with @babel/register; otherwise, we're probably in `build/test/` which _does not_ contain the xml file.
+const appFixturePath = path.join(__dirname, 'fixtures', 'app.xml');
+const TEST_APP = existsSync(appFixturePath) ? appFixturePath : path.join(__dirname, '..', '..', 'test', 'fixtures', 'app.xml');
```
boneskull, line 11: two potential fixes here

1. stop using babel 😄
2. copy the xml file into `build` during transpilation

packages/fake-driver/package.json
```diff
@@ -30,9 +30,9 @@
     "screen.png"
   ],
   "dependencies": {
+    "@appium/base-driver": "^8.0.0-beta.6",
```
KazuCocoa, line 33: 👍

KazuCocoa (approved):
👍 
lint? :)

packages/base-driver/test/basedriver/driver-tests.js
```diff
@@ -75,12 +78,12 @@ function baseDriverUnitTests (DriverClass, defaultCaps = {}) {
...
-        id: d.sessionId,
-        capabilities: {
-          'deviceName': 'Commodore 64',
-          'platformName': 'Fake'
-        }
+      sessions[0].should.include({
+        id: d.sessionId
+      });
+      sessions[0].capabilities.should.include({
+        'deviceName': 'Commodore 64',
```
mykola-mokhnach, line 85: I assume there is no need to quote key names

packages/fake-driver/lib/fake-app.js
```diff
@@ -1,13 +1,11 @@
-import B from 'bluebird';
-import fs from 'fs';
+import {readFile} from 'fs/promises';
```
mykola-mokhnach, line 1: readFile with promises could be imported from fs submodule of appium-support

packages/fake-driver/test/helpers.js
mykola-mokhnach: option #2 looks reasonable to be

mykola-mokhnach (approved):

mykola-mokhnach:
Tests are still not quite happy though

packages/base-driver/test/basedriver/driver-tests.js
boneskull: no, but we don't auto-format the code and eslint doesn't seem to have a rule enabled about it, so ... we should address that

jlipps: i thought we had an eslint rule. i guess not. we should add a rule

packages/fake-driver/lib/fake-app.js
jlipps, line 1: i don't know if we can have an eslint rule about this, but we differentiate named imports from object destructuring in our whitespace:

```
import { foo } from 'bar';
export { thing1, thing2 };
```
vs
```
const {foo} = bar;
const thing = {x: 1, y: 2};
```

packages/fake-driver/lib/logger.js
jlipps: probably some other package still had appium-support as a dep?

packages/fake-driver/test/helpers.js
jlipps: yeah this bites us in a few different places, where we have to unreasonably path out to the package root and then down into the build dir. if there's an easy way to copy fixtures during transpilation, great!

packages/fake-driver/lib/fake-app.js
boneskull: eslint is obviously not configured to care about this...  I do pretty well with an eslint + prettier combo and would be gung-ho about adding prettier here, if you are interested.

boneskull:
test failure is expected since I added a route... unfortunately that seems to prevent the rest of the suite from running.

packages/fake-driver/lib/fake-app.js
jlipps: I'm up for that. Are you saying it could help us make this distinction?

jlipps:
The solution is to update that failing test with the (now) correct hash: `3351a966`

packages/fake-driver/lib/fake-app.js
boneskull: I'm curious: why promisify the `fs` api with bluebird?  performance reasons?

jlipps: Before promise versions of the node apis were shipped with Node, we put promisified versions in appium-support since we used them everywhere. I'm not sure why we would keep using them in appium-support though now, unless there are functional differences (@mykola-mokhnach ?)

boneskull: (I've pushed this change)

mykola-mokhnach: > Before promise versions of the node apis were shipped with Node, we put promisified versions in appium-support since we used them everywhere. I'm not sure why we would keep using them in appium-support though now, unless there are functional differences (@mykola-mokhnach ?)

Exactly. As for appium-support we could now replace bluebird usage with async file operations natively provided by node. I don't see any point in changing method signatures because it would require too many changes in depending modules

boneskull: I tried it, actually, and prettier makes no distinction.  you may have a tough time making this work.

packages/base-driver/test/basedriver/driver-tests.js
jlipps: even if there's not a rule, let's unquote these key names as part of this review

packages/fake-driver/lib/fake-app.js
```diff
@@ -1,14 +1,13 @@
-import B from 'bluebird';
-import fs from 'fs';
+import {fs} from '@appium/support';
```
jlipps, line 1: even though there's not a rule, let's stick with the `{ <name> }` spacing style if we notice it.

packages/fake-driver/test/general-tests.js
```diff
@@ -1,7 +1,7 @@
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
-import { initSession, deleteSession, DEFAULT_CAPS } from './helpers';
-import { W3CActions } from 'wd';
+import chaiWebdriverIOAsync from 'chai-webdriverio-async';
```
jlipps, line 3: o_O there's a package for this??

packages/fake-driver/test/helpers.js
jlipps: should we create an issue to track this task?

jlipps (changes requested):
couple style nits, otherwise this is ready to go!

packages/fake-driver/test/general-tests.js
boneskull: yeah.  async webdriver isn't the most fun imo... this makes it a little easier. sometimes.

boneskull:
@jlipps Created #15471

jlipps (approved):

jlipps:
thanks so much @boneskull! this was incredible, and no doubt incredibly tedious, work.

PR #15438, opened 2021-05-29
docs: Appium 1 EOL plan

mykola-mokhnach:
## Proposed changes

I see a need for such document for both developers and users of Appium. It adds more transparency and visibility to what is being done and to what is expected to be done. I'm not quite sure where to put this document, it could even stay as a PR, although it is important to keep it publicly visible for the lifetime of this transition period.

## Types of changes

### DIFF
--- a/docs/en/about-appium/eol.md
+++ b/docs/en/about-appium/eol.md
@@ -1,0 +1,51 @@
+# Appium version 1 End Of Life (EOL) Plan
+
+Appium 1 has been around for quite a while and now is the time to move
+project development forward and to provide even better automation
+experience with Appium version 2.
+
+## Major Changes
+
+The main differences of Appium 1 in comparison to version 2 are the following:
+
+- Platform drivers are now managed separately from the Appium foundation. This means
+  each driver could now be updated/installed/removed separately from the umbrella one.
+- The foundation modules have been moved to the `@appium` monorepo managed by
+  [Lerna](https://github.com/lerna/lerna). These modules are: appium (now `@appium/appium`),
+  appium-base-driver (now `@appium/base-driver`), appium-fake-driver (now `@appium/fake-driver`),
+  appium-support (now `@appium/support`), appium-doctor (now `@appium/doctor`),
+  eslint-config-appium (now `@appium/eslint-config-appium`), appium-test-support (now `@appium/test-support`),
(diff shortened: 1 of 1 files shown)


docs/en/about-appium/eol.md
```diff
@@ -0,0 +1,51 @@
...
+  appium-gulp-plugins (now `@appium/gulp-plugins`).
+- Added support of plugins
+- Added a CLI to manage drivers and plugins
+- Dropped support of JSONWP capabilities. Only W3C-compatible clients are now supported.
+- The list of supported command line arguments has been modified to exclude driver-specific ones.
+- Drivers compatible to Appium 2 must include a special manifest in their package.json
+
+## Transition Plan
+
+The preliminary date of Appium 2 release is *1st of January 2022*. After this date we won't
```
KazuCocoa, line 27: @jlipps for this date?

jlipps: It should be much sooner, like in the next few months

jlipps: however i'm ok if we want a maintenance plan for 1.x that extends beyond 2.0 release a bit

jlipps (commented):
this is awesome thanks @mykola-mokhnach! i like your suggestions, let's keep talking as a team about these things for a bit before merging/publicizing

docs/en/about-appium/eol.md
KazuCocoa: We can release 1.22.0 as the final version as 1.x, or 1.23.0?
(At least, we should release 1.22.0 with this 2.0 announcement on the release page to notify this 2.0 update widely I think.)

For now, it seems we can mark this line as tentative...? (and merge this PR to modify this good documentation more?)

KazuCocoa (approved):

docs/en/about-appium/eol.md
jlipps: I think we should say preliminary release date is like Aug 1 2021. and after this date we won't add features to 1.x. but will continue to provide urgent bugfixes to 1.x through Dec 31 2021.

boneskull: let's not commit to any dates until we can commit to dates 😄

KazuCocoa: seems good for me👌

jlipps:
closed in favor of #15711

PR #15472, opened 2021-06-07
fix(fake-driver): copy non-JS files into build dir at transpile time; closes #15471

### LINKED ISSUE #15471: fix relative paths to non-JS files in test code
A few places in test code reference files (e.g., `app.xml`) relative to the current module.  This would be fine, except that non-JS don't get transpiled, and thus do not make it into `build/`, which bamboozles relative path calculation--`./build/..` is not the same path as `./..`.

To fix this (for now), _copy_ non-JS files into `build/`.

boneskull:
## Proposed changes

In `fake-driver`, configured Gulp to copy `app.xml` and `screen.png` into appropriate directories as a post-transpilation step.  

Closes #15471

## Types of changes

## Checklist

### DIFF
--- a/packages/fake-driver/gulpfile.js
+++ b/packages/fake-driver/gulpfile.js
@@ -3,8 +3,13 @@
 const gulp = require('gulp');
 const boilerplate = require('@appium/gulp-plugins').boilerplate.use(gulp);
 
+gulp.task('copy-screen.png', () => gulp.src('./screen.png').pipe(gulp.dest('./build/')));
+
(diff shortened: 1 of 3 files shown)


jlipps (commented):
I think this works for now; it might be a bit more extensible if we made the change in appium-gulp-plugins so that when we call the `boilerplate` function, we can simply pass in an array of files to some `staticFiles` or `copyFiles`as an option.

boneskull:
> I think this works for now; it might be a bit more extensible if we made the change in appium-gulp-plugins so that when we call the `boilerplate` function, we can simply pass in an array of files to some `staticFiles` or `copyFiles`as an option.

I think YAGNI for now -- the only pkg that does anything like this is fake-driver

PR #15502, opened 2021-06-14
chore(gulp-plugins): enable consumers to use mocha's "exit" flag

boneskull:
unit tests _should_ not need `--exit`, but this is real life, so things aren't perfect.

boilerplate configs can pass `{test: {exit: true}}` to cause Mocha to force-exit.

note that this is enabled by default for e2e tests (which _also_ shouldn't be necessary, but one thing at a time...)

### DIFF
--- a/packages/gulp-plugins/lib/tasks/unit-test.js
+++ b/packages/gulp-plugins/lib/tasks/unit-test.js
@@ -14,7 +14,8 @@ const configure = function configure (gulp, opts, env) {
       timeout: opts.testTimeout,
       traceWarnings: opts.test.traceWarnings,
       traceDeprecation: opts.test.traceWarnings,
-      color: true
+      color: true,
+      exit: Boolean(opts.test.exit)
     };
     // set env so our code knows when it's being run in a test env
     process.env._TESTING = 1;


packages/gulp-plugins/lib/tasks/unit-test.js
```diff
@@ -14,7 +14,8 @@ const configure = function configure (gulp, opts, env) {
       timeout: opts.testTimeout,
       traceWarnings: opts.test.traceWarnings,
       traceDeprecation: opts.test.traceWarnings,
-      color: true
+      color: true,
+      exit: Boolean(opts.test.exit)
```
mykola-mokhnach, line 18: we usually use `!!opts.test.exit`

mykola-mokhnach (approved):

packages/gulp-plugins/lib/tasks/unit-test.js
boneskull: for stuff like this we need ESLint rules... really, we shouldn't be talking about style in PRs.  it should all be automated.

when I was a new JS developer, I'd see `!!` and wonder what sort of clever trickery was going on.  I'd never seen `!!` before in any language.  so, that's my rationale for using `Boolean()` -- it should be more clear to newcomers to the language.

jlipps: I'm fine to standardize on either, I agree `Boolean` is more approachable. Without a lint rule, it will be hard to make any standardization changes however. Is there an eslint rule for this?

jlipps (approved):

KazuCocoa (approved):

packages/gulp-plugins/lib/tasks/unit-test.js
boneskull: if there isn't, it's pretty straightforward to make one

PR #15506, opened 2021-06-15
chore: upgrade sync-monorepo-packages

boneskull:
this fixes some problems with the command-line options, and maybe some security fixes too


boneskull:
note: this tool is used upon publish to sync various `package.json` fields across the packages

jlipps (approved):

PR #15580, opened 2021-06-25
experimental(base-driver): publish @appium/base-driver as hybrid cjs/esm module

### LINKED ISSUE #15578: Appium 2.0: distribute dual CJS/ESM packages
The situation:

- Currently Appium only distributes CJS modules.
- Appium 2.0's minimum required Node.js version is v12.x, which has ESM support.
- Consumers writing ESM would be better off consuming ESM instead of CJS.
- _Appium_ would be better off consuming ESM directly than CJS; it will simplify testing and the importing of local packages. If we are in `@appium/foo` and `import '@appium/bar'`, we will be importing a transpiled CJS module, which can be painful in a development environment.  We should try to get as close to the source as possible.
- We have a hard dependency on building before testing, which _may_ be mitigated if we publish ESM.  The only blocker is using JS syntax unsupported in Node.js v12.x.  Depending on the syntax, it may be worth just _not using it_ in order to skip the build step entirely.  Since we already _need_ a build step, currently, there's no extra cost associated with using bleeding-edge syntax.  But if we didn't _need_ the build step, using bleeding-edge syntax would not be free.
- The goal is to improve both the _user experience_ and the _developer experience_.

Proposed solution:

- Publish each package in this monorepo (and `appium-ios` and forthcoming `appium-android`) as a dual CJS/ESM package.
- Create a reusable configuration/workflow/whatever for this to happen automatically.
- Evaluate the use of bleeding-edge syntax and decide whether it is best to just _wait_ until Node.js v12.x goes EOL.

Gotchas/warnings/potential-sources-of-peril:

- Publishing a dual package is not necessarily _tricky_, but it must be done correctly or consumers will break.  Since we're talking about a major release, this is less of a concern, but should be completely avoidable if we're careful.
- We will may need to _explicitly declare_ what modules are considered "public" and which are "private" (or implementation details).  For example, if in CJS you wanted to `require('@appium/base-driver/lib/boobah')` and wanted to move to ESM and `import` this module, we must _explicitly declare_ that `lib/boobah` is exposed as such.  [More info](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#packages_package_entry_points)
- This typically creates some "stub" modules, which can be annoying.  Further, some modules may require the `.cjs` extension.

boneskull:
I was curious how difficult this would be, so I tried it out.  There are some undocumented rules around what the `exports` field in `package.json` can contain, which tripped me up--but otherwise it wasn't so bad.

While `@appium/base-driver` is still built as before, the tests run against the original ESM sources.  I have not tested this with anything except Node.js v16, so we'll see how that goes.

In older Node.js versions, there's likely going to be a warning emitted by Node.js that conditional exports are experimental (they are no longer experimental).

Because `@appium/fake-driver` (which depends upon `@appium/base-driver`), is _not_ a hybrid (dual) module, it runs its tests against its build artifacts, and thusly against `@appium/base-driver`'s CJS build artifacts.  So this should show that **both CJS and ESM versions of `@appium/base-driver` work as expected.**

Changes of note:

- All _relative_ imports _must_ contain a file extension _unless_ they are explicitly declared in `exports` field of `package.json`
- `@appium/base-driver` defaults to ESM, but uses conditional exports to redirect the entry point into `build/index.js` when used via `require('@appium/base-driver')`
- `.js` files living in `build/` are now explicitly treated as CJS modules via the stub `package.json` copied there (which just contains `{"type": "commonjs"}` and applies to this directory only)
- any `.js` file outside of `build/` which is a CJS file **must** have a `.cjs` extension (e.g., `gulpfile.cjs`)
- it's too bad gulp does not recognize `gulpfile.cjs`, so we have to give it an explicit path to `gulpfile.cjs`
- replaced usages of `__dirname`, `__filename` with `import.meta` and `URL`
- added package `babel-plugin-transform-import-meta` to translate finding filepaths from `import.meta.url` into `__filename`, etc.
- completely disabled `import/no-unresolved` since it does not understand Node.js' `exports` field
- gulp changes:
    - `@appium/gulp-plugins`: copy `.babelrc` into `build` since this package is tightly-bound to that dir 😜
    - the `hybridModule` `boilerplate()` option will now cause a hybrid (dual module) build; used only in `@appium/base-driver`
    - needed to copy the entirety of `@appium/base-driver`'s `static` dir into `build` for tests. it still doesn't work too well. a better solution would be to actually put `static` in `lib` so we don't need to look at parent paths
    - XXX: `@appium/base-driver` attempts to retrieve its version from `package.json`, however, `build/package.json` is a stub, so there's a conditional here. a better solution would be to actually copy the `version` field into the stub `build/package.json`
- drivers (including `@appium/fake-driver`) should not `import` or `require()` `@appium/base-driver/build/test/basedriver`; just `import '@appium/base-driver/test` instead (this is a custom export).  this will **BREAK** if a package attempts to load the old path
- increased `babel-env`'s target to Node.js v12

Closes #15578

**UPDATE**: 

- Reverted the `gulpfile.cjs` nonsense and just rewrote `@appium/base-driver`'s gulpfile as ESM.
- Fixed some quoting issues in the `sync-pkgs:keywords` script


boneskull:
(Most of the 66 files changed are simply adding the `.js` extension to imported relative paths)

boneskull:
OK, so the [current failure](https://github.com/appium/appium/pull/15580/checks?check_run_id=2935131867) is related to the phony `package.json` in `build/`.  In addition to the `version` field, it would need the `engines` field as well, since `appium/lib/config.js` is looking for that.

Further, I needed to "downgrade" some syntax Node.js v12 does not support (optional chaining, nullish coalescing).

I am not going to continue with this PR until we decide we want to go in this direction.  This is how I see our options:

1. Status quo w/ sources using **latest** syntax -- publish CJS only, uses a build step.
2. Hybrid module w/ sources using **Node.js-v12-supported** syntax - publish both CJS and ESM; uses a build step for CJS.
3. Hybrid module w/ sources using **latest** syntax - publish both CJS and ESM; uses a build step _for each_
4. ESM only w/ sources using **Node.js-v12-supported** syntax - publish ESM; _no build step whatsoever_.
5. ESM only w/ sources using latest syntax - publish ESM; uses a build step

Here's a breakdown of the major advantages/disadvantages, as I see it:

1. Status quo:
    1. Advantage: nothing to change.  
    2. Advantage: we can use whatever syntax we want. 
    3. Disadvantage: consumers (including Appium!) cannot consume the ESM we write.   
    4. Disadvantage: source maps always needed when debugging.  
    5. Disadvantage: value of publishing CJS is dubious.
2. Hybrid w/ Node.js v12 syntax:
    1. Advantage: no source maps & better debugging if using ESM (including any other Appium packages using ESM).  
    2. Disadvantage: manual management of package exports
    3. Disadvantage: value of publishing CJS is dubious.
    4. Neutral: can upgrade syntax (which?) when Node.js v12 goes EOL
3. Hybrid with latest syntax:
    1. Advantage: sources can use latest syntax.
    2. Disadvantage: manual management of package exports. 
    3. Disadvantage: builds needed for both ESM and CJS. 
    4. Disadvantage: value of publishing CJS is dubious.  
4. ESM-only w/ Node.js v12 syntax:
    1. Advantage - no source maps for consumers, period.
    2. Advantage: no build step whatsoever.  
    3. Disadvantage: cannot use latest syntax
    4.  Neutral: can upgrade syntax (which?) when Node.js v12 goes EOL
5. ESM-only w/ latest syntax:
    1. Advantage - can use latest syntax.
    2. Disadvantage: source maps always needed when debugging.
    3. Disadvantage: build step still necessary.

boneskull:
(if we choose a Node.js-v12-syntax-only path, we can restrict the syntax via ESLint rules)

jlipps:
Thanks, this is a real helpful overview. Let's mull it over for a bit. Right now I'm kind of inclined to say let's keep the status quo on the strength of its 2 advantages.

PR #15638, opened 2021-07-16
feat: add driver/plugin arg parsing to appium

khanayan123:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

Add parseSeverArg method to base-driver. This is done to minimize code duplication by adding a method to base-driver which drivers can use to parse server args passed in by users. This is apart of appiums goal of having drivers handle cli server args. 

## Types of changes

## Checklist

- [] Any dependent changes have been merged and published in downstream modules

## Further comments

view this pr to see example usage:
https://github.com/appium/appium-xcuitest-driver/pull/1307


packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -420,6 +420,44 @@ function parseCapsArray (cap) {
...
+ *
+ * @param {object} opts - driver opts
+ * @param {object} serverArgs - serverArgs
+ * @param {Array<String>} knownArgNames - argNames to parse for
+ * @param {object} argsConstraints - Constraints for argNames
+ * @return {object}
+ */
+
+function parseServerArgs (opts, serverArgs, knownArgNames, argsConstraints) {
+  let acc = parseKnownArgs(serverArgs, knownArgNames, argsConstraints);
```
jlipps, line 449: why is this called `acc` and not something like `args`?

jlipps (approved):

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -420,6 +420,44 @@ function parseCapsArray (cap) {
...
+  acc = validateCaps(acc, argsConstraints);
+  for (let arg in argsConstraints) {
+    if (!_.has(acc, arg)) {
+      if (_.has(opts, arg)) {
```
mykola-mokhnach, line 453: there is no need to nest these ifs

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -420,6 +420,44 @@ function parseCapsArray (cap) {
...
+ * arg constraints for the arg names, and returns a combined object containing opts
+ * and parsed server args
+ *
+ * @param {object} opts - driver opts
+ * @param {object} serverArgs - serverArgs
+ * @param {Array<String>} knownArgNames - argNames to parse for
+ * @param {object} argsConstraints - Constraints for argNames
+ * @return {object}
+ */
+
```
mykola-mokhnach, line 447: extra line break

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -420,6 +420,44 @@ function parseCapsArray (cap) {
   throw new Error(`must provide a string or JSON Array; received ${cap}`);
 }
 
+function parseKnownArgs (serverArgs, knownArgNames, argsConstraints) {
```
mykola-mokhnach, line 423: I would rather put these methods to a separate .js file called `arguments.js`

jlipps (approved):

packages/base-driver/index.js
```diff
@@ -60,3 +60,8 @@ export { processCapabilities, isStandardCap, validateCaps };
 import * as ws from './lib/express/websocket';
 const { DEFAULT_WS_PATHNAME_PREFIX } = ws;
 export { DEFAULT_WS_PATHNAME_PREFIX };
+
+// Arguments
+import * as args from './lib/basedriver/arguments';
```
mykola-mokhnach, line 65: we could import parseServerArgs directly: `import { parseServerArgs } from './lib/basedriver/arguments';`

packages/base-driver/lib/basedriver/arguments.js
```diff
@@ -0,0 +1,41 @@
+import _ from 'lodash';
+import { validateCaps } from './capabilities';
+
+function parseKnownArgs (serverArgs, knownArgNames, argsConstraints) {
```
mykola-mokhnach, line 4: it looks like our knownArgNames are always equal to keys of argsConstraints, which makes this function simpler

packages/base-driver/lib/basedriver/arguments.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param {Array<String>} knownArgNames - argNames to parse for
+ * @param {object} argsConstraints - Constraints for argNames
+ * @return {object}
+*/
+function parseServerArgs (opts, serverArgs, knownArgNames, argsConstraints) {
+  let args = parseKnownArgs(serverArgs, knownArgNames, argsConstraints);
+  args = validateCaps(args, argsConstraints);
+  for (let arg in argsConstraints) {
+    if (!_.has(args, arg) && _.has(opts, arg)) {
+      args[arg] = opts[arg];
```
mykola-mokhnach, line 33: this loop is probably redundant. `assign` call only overrides values in the first object argument that are defined in the following arguments, so it is ok to only have the actual values to override in `args` object.

mykola-mokhnach:
See https://github.com/appium/appium-xcuitest-driver/pull/1307/files#r672827891

jlipps (approved):

jlipps:
what's up with CI?

ayvnkhan:
Check out fake-driver/lib/driver.js for example implementation. Once/if approved I will incorporate these new changes to plugins and drivers

packages/appium/lib/appium.js
```diff
@@ -5,8 +5,8 @@ import { findMatchingDriver } from './drivers';
 import { BaseDriver, errors, isSessionCommand } from '@appium/base-driver';
 import B from 'bluebird';
 import AsyncLock from 'async-lock';
-import { parseExtensionArgs, parseCapsForInnerDriver, pullSettings } from './utils';
-import { util } from '@appium/support';
+import { parseExtensionArgs, parseCapsForInnerDriver, pullSettings, parseDriverPluginArgs } from './utils';
+import { util } from 'appium-support';
```
mykola-mokhnach, line 9: Please use import from the monorep module (`'@appium/support'`)

packages/appium/lib/appium.js
```diff
@@ -167,7 +167,13 @@ class AppiumDriver extends BaseDriver {
       let runningDriversData, otherPendingDriversData;
       const parsedInnerDriverArgs = parseExtensionArgs(this.args.driverArgs, driverName);
 
-      const d = new InnerDriver(this.args, true, parsedInnerDriverArgs);
+      if (_.has(InnerDriver, 'argsConstraints')) {
+        const driverArgsConstraints = InnerDriver.argsConstraints;
```
mykola-mokhnach, line 171: we probably don't need this intermediate variable

packages/appium/lib/appium.js
```diff
@@ -167,7 +167,13 @@ class AppiumDriver extends BaseDriver {
...
+        // parse/verify CLI args and merge them to this.args if OK or throw an exception
+        parseDriverPluginArgs(this.args, parsedInnerDriverArgs, driverArgsConstraints);
```
mykola-mokhnach, line 173: I would prefer this method to return parsed args rather than mutate them

packages/appium/lib/appium.js
```diff
@@ -167,7 +167,13 @@ class AppiumDriver extends BaseDriver {
       let runningDriversData, otherPendingDriversData;
       const parsedInnerDriverArgs = parseExtensionArgs(this.args.driverArgs, driverName);
 
-      const d = new InnerDriver(this.args, true, parsedInnerDriverArgs);
+      if (_.has(InnerDriver, 'argsConstraints')) {
```
mykola-mokhnach, line 170: is makes no sense to call this branch if parsedInnerDriverArgs is empty

packages/appium/lib/utils.js
```diff
@@ -46,6 +46,43 @@ function parseExtensionArgs (extensionArgs, extensionName) {
...
+ *
+ * @param {object} args - Args
+ * @param {object} driverPluginArgs - Driver or Plugin args
+ * @param {object} argsConstraints - Constraints for arguments
+ * @return {object}
+*/
+function parseDriverPluginArgs (args, driverPluginArgs, argsConstraints) {
+  if (_.isEmpty(driverPluginArgs) || _.isEmpty(argsConstraints)) {
+    return args;
+  } else {
```
mykola-mokhnach, line 78: else is redundant here. The previous `if` is a classic guard clause

packages/appium/lib/utils.js
```diff
@@ -46,6 +46,43 @@ function parseExtensionArgs (extensionArgs, extensionName) {
...
+    let parsedArgs = parseKnownArgs(driverPluginArgs, argsConstraints);
```
mykola-mokhnach, line 79: const

packages/appium/lib/utils.js
```diff
@@ -46,6 +46,43 @@ function parseExtensionArgs (extensionArgs, extensionName) {
...
 }
 
+
+function parseKnownArgs (driverPluginArgs, argsConstraints) {
+  const knownArgNames = Object.keys(argsConstraints);
+  return _.toPairs(driverPluginArgs).reduce((args, [argName, argValue]) => {
+    if (knownArgNames.includes(argName)) {
+      args[argName] = argValue;
+    } else {
+      const knownArgs = Object.keys(argsConstraints);
```
mykola-mokhnach, line 56: don't we already have knownArgNames above ?

packages/appium/lib/appium.js
mykola-mokhnach: it might also be useful to add a debug log here about which args/values were parsed

packages/appium/lib/utils.js
```diff
@@ -46,6 +46,41 @@ function parseExtensionArgs (extensionArgs, extensionName) {
...
+ * and parsed driver/plugin args. If driverPluginArgs or argsConstraints is empty, args is returned
+ * back
+ *
+ * @param {object} args - Args
+ * @param {object} driverPluginArgs - Driver or Plugin args
+ * @param {object} argsConstraints - Constraints for arguments
+ * @return {object}
+*/
+function parseDriverPluginArgs (args, driverPluginArgs, argsConstraints) {
+  if (!_.isEmpty(driverPluginArgs) && !_.isEmpty(argsConstraints)) {
```
mykola-mokhnach, line 75: I usually prefer guard style:

```
if (_.isEmpty(driverPluginArgs) || _.isEmpty(argsConstraints)) {
  return args;
}

...

```

packages/appium/lib/utils.js
```diff
@@ -46,6 +46,41 @@ function parseExtensionArgs (extensionArgs, extensionName) {
...
+    } else {
+      throw new Error(`"${argName}" is not a recognized key. Are you sure it's in the list ` +
+                      `of supported keys? ${JSON.stringify(knownArgNames)}`);
+    }
+    return args;
+  }, {});
+}
+
+/**
+ * Takes in a set of args, driver/plugin args passed in by user, and arg constraints
```
mykola-mokhnach, line 64: The documentation must mention that args object is mutated there

mykola-mokhnach: I'm actually in general not quite sure if it is a good idea to mutate the `args` property of the umbrella driver and whether this is not going to create unexpected conflicts for parallel sessions. @jlipps Could you please confirm it is safe?

jlipps: I think in general we want to move away from mutation unless there's a strong reason for it. We could avoid it pretty easily here by simply adding `args = _.cloneDeep(args);` as the first line of this method. then the caller can decide if it wants to mutate or not. but the place to mutate would be there and probably not in an incidental parser helper function.

jlipps (commented):
I agree with @mykola-mokhnach's comments. But the bigger question is whether we want to follow this approach or try to merge this work with the args schema stuff from @boneskull. Right now I'm kind of leaning towards the args schema approach just to standardize all our arg docs/parsing/logic.

jlipps:
closing in favor of https://github.com/appium/appium/pull/15806

PR #15664, opened 2021-07-26
feat(appium): config file support

boneskull:
This is the first pass at support for config files for Appium.

There's quite a bit happening here:

- A JSON Schema v7 defines the shape of the config file _and_ is reused to define the command-line options (`lib/appium.schema.json`). Providing a JSON schema allows IDEs to automatically validate JSON config files (and YAML files too, but I haven't looked into exactly how this works) and provide contextual help and suggestions.  This is how something like an `.eslintrc.json` is validated by your IDE.
- A new module, `lib/config-file` exports some functions for loading the config file and schema.
- _Defaults_ (e.g., `--port=4723`) are defined _in the schema file_.
- Because we want the precedence CLI ARGS > CONFIG FILE > DEFAULTS, we cannot rely on `argparse` to handle defaults, since anything defined here will take precedence over a config file.  This behavior is changed so that `argparse` is no longer responsible for defaults.
- Currently, this supports "shared" options and options for `appium server` only (thus it's a WIP until I figure out plugin/driver options)
- Much of the hard-coded options in `lib/cli/args` were ripped out, since these are now defined in the schema
- Some options remain hardcoded in `lib/config.js` which are _not_ explicitly mentioned in the server args (e.g., some iOS-specific ones).  I am not sure what should happen with these, but they _look_ like they are leftovers from 1.x.
- Valid config files can be named `.appiumrc`, `.appiumrc.{cjs|js|json|yaml|yml}`, `appium.config.{js,cjs}`. Also the `appium` prop of the closest `package.json` can be used
- An invalid config file currently _throws_
- Human-readable and helpful error messages for config problems
- Pulled in `sinon-chai`
- Schema type declaration (`lib/appium-config.d.ts`) generated by `scripts/generate-schema-declaration.js` & [json-schema-to-typescript](https://npm.im/json-schema-to-typescript). This file is under VCS, but is automatically regenerated via pre-commit hook when the schema file has changed.  This file will be useful for consumers, but also type-checking the JS in the `config-file` module will fail w/o it.
- Could use more E2E tests _and_ a snapshot library, since we may be comparing rather large strings or objects.
- There are some samples.
- `driver-args` and `plugin-args` for now broken

* * * 

I wanted to get us discussing this, since I think supporting config files would be a huge win.  I kind of need it for my own sanity, too. 😄 

## Driver/Plugin-specific Config

Now, I have not touched anything to do with driver- or plugin-specific configuration, but this is how I envision it would work:

1. Drivers/plugins (hereby known as "_extensions_") will export one of the following, known hereby as an "_options declaration_":
    - A JSON schema file describing their own options, _OR_
    - An [Ajv](https://npm.im/ajv)-compatible object describing their own options.  If you've built a custom ESLint rule that accepts options, then you are familiar with the syntax--there's precedent for this sort of thing, anyway.
2. Based on the options declaration, Appium-the-CLI would accept and validate extension-specific options, and display them in `--help` as needed.
3. Appium-the-project would define any non-standard metadata fields for extensions to use in the options declaration.  We already have a couple (`appiumDest` means "the name of the object property on which to store this option"; `appiumAliases` define aliases for command-line options).  One may be `appiumConfigOnly` which would hide or disable the option from the CLI, but allow it in a config file/object.  This is mainly a matter of documentation and implementing whatever else we decide extensions need.
4. Extensions will do basic validation via the options declaration (e.g., this option must be an integer / email address / pass this regex), and this happens automatically for both config files and CLI flags.  But, like as seen in this PR, custom validation functions can be provided.
5. This would piggyback onto whatever was added in #15388.

Further (and this is not yet fully implemented, but stubbed out), extension-specific configuration could live in a user's config file:

```yaml
server:
  port: 31337
  # more server options
driver:
  # driver-specific options
  myDriver:
    foo: bar
plugin:
  # plugin-specific options
  myPlugin:
    baz: quux
```

Currently, we have users provide paths to external JSON config files or paste in raw JSON on the CLI (ugh).  Instead, they could use a single config file if they wish.  The current `driver-args` and `plugin-args` are **not implemented** in this PR, but we could use them for the old-style "external file", drop support entirely, soft-deprecate, etc.  Or just leave as-is in the name of backwards compat (which would necessitate adding them to the schema).

## Questions

- What do we think of the general idea of making the JSON schema the single source of truth (for most options)?
- What should the CLI do if it encounters an invalid config file?
- Would it be helpful to add a subcommand to validate a config file (I'm thinking _yes_)? Or better to live in `appium-doctor`?
- Does the proposal about handling extension (driver/plugins) options sound reasonable?
- I realize there is "TypeScript" creeping in here.  But _I will take a blood oath and promise I will not try to get us writing TypeScript_.  I am, however, a proponent of careful use of JS docstrings, which provides many benefits for the contributor, extension author, or anyone wishing to use Appium as a platform. The in-IDE TypeScript language server provides much more introspection ("intellisense") than traditional JSDoc.  We are also able to leverage third-party types (DefinitelyTyped) and any types published by our dependencies.  _Are you cool with this_? 

## Note

- We should also add a schema for the base server capabilities. Not sure if that should be its own schema.
- I don't think this is a good contender for a backport to 1.x, but happy to hear your thoughts.

This is a **draft** since it's incomplete & broken in the case of driver/plugin options.  In my testing, it seems to work OK otherwise.  You don't need to pick at the implementation (yet) but I _would_ like some general feedback otherwise.

cc @avynkhan

# **UPDATE** Sep 28 2021

> Description of latest changes from a236575dc4ec7638279044bbcfe102ae6dccad45

### Development
- Added many `@types/*` packages (and `type-fest`) for Intellisense and type-checking
- Added `sinon-chai` for making assertions on Sinon functions using Chai
- Added `rewiremock` for module-level mocking (improves unit test isolation)
- Added `through2` for a custom gulp task to convert `appium/lib/appium-config-schema.js` to a JSON file
- Added `json-schema-to-typescript` for converting aforementioned JSON file to a `.d.ts` file (which gets published)
### Production (applies to `appium` only)
- Added `libconfig` for multi-format config file loading, which is a smaller drop-in replacement for `cosmiconfig`
- Added `ajv` for schema support and validation
- Added `ajv-formats` to understand optional value formats (e.g., `ipv4`)
- Added `@sidvind/better-ajv-errors` which provides nice error output on the CLI if the config file is invalid

## New Modules of Note

- `lib/schema.js` consumes `ajv` to create and modify a schema
-  lib/config-file.js` consumes `lilconfig` to read config files, validate them against the schema, and provide the result as CLI args
- `lib/cli/schema-args.js` which serves as an adapter between the schema and `argparse`
- `lib/appium-config-schema.js` which is the root schema. Extensions add to this before it is registered & compiled by `ajv`.

## Modifications

- The single source of truth for the CLI _and_ config file options is the schema, with the caveats that:
    - Command-line arguments (for `appium server` are effectively a flat list, but the schema/config file is not.
    - Extension-specific configuration lives in an extension-type-specific property (`driver` or `plugin`) in the config file
        - Under each (`driver` or `plugin`), each extension may have a configuration section corresponding to the `automationName` ??
        - The shape of the extension-specific options _can_ be defined by a schema (and referenced in the `appium.schema` prop of the extension's `package.json`)
        - If an extension does not provide a schema, no validation will occur on any options found here -- TODO needs tests
    - This does _not_ affect _CLI commands_ added by extensions; only options for `server`.
- "Shared" options `appiumHome` and `log-filters` are removed
- TODO: Fully remove `pluginArgs` and `driverArgs` and pass in the proper values instead
- `lib/main.js`
    - Now exports a few public APIs
    - split into functions `init` and `main`.  `init` is intended to be run via a consumer, which does what the old `main` did up to starting the server.  This bisection is not perfect, as `init` still does some CLI-related things; it needs to be further broken up.
    - `driverConfig.read()` and `pluginConfig.read()` are now called in `lib/cli/parser.js`.
    - Merges the CLI args, config file args, and defaults, in that order of precedence.
- `lib/grid-register.js`
    - Supports embedded configuration instead of just a path to a JSON file.
    - Not clear to me if `nodeconfig` is a thing we need to keep?
- `lib/extension-config.js`
    - Given `APPIUM_HOME` is where we keep the extensions, this file now defines `APPIUM_HOME`. It is exported from the `appium` package.
- `lib/driver-config.js`
    - **Fix** the case where two drivers attempt to use the same name.  It appears this was intended, but did not work correctly.
    - Read, validate, and register the schema, if it exists.  TODO: We should probably not register the schema here, but validating it is reasonable.
- `lib/plugin-config.js`
    - **TODO**: Ensure this works with schemas
- `lib/config.js`
    - Remove unused validations (now occur via schema validation)
    - Rename `getNonDefaultArgs` to `getNonDefaultServerArgs` to be more specific
    - Rewrote `getNonDefaultServerArgs`, which became much more complicated due to adding a config file and schema defaults to the mix. Flexed my lodash-fu
- `lib/cli/parser.js`
    - Add another bit in the `parse_args()` monkeypatch which _unflattens_ the resulting `args` object for extensions.  `--driver-foo-bar=baz` would be stored in prop `driverFooBar`, but this moves it into `driver: {foo: {bar: 'baz'}}`, which will be provided to each appropriate extension (TODO)
    - `addExtensionsToParser` renamed to `addExtensionCommandsToParser` which is more specific.
    - Removed `getDefaultServerArgs` since those now come from the schema.
    - Removed shared args -- TODO: need to remove these from other functions as well; stubbed with empty array
    - `getParser` is now async.
    - TODO: `call to `utils.getExtensionArgs` (recently added) needs to be removed since `driverArgs` and `pluginArgs` no longer exist.
- `lib/cli/parser-helpers.js`
    - Allow security features to just be an array (as allowed in schema)
    - TODO: needs test
- `lib/cli/extension.js`
    - Use `APPIUM_HOME` exported from `lib/extension-config` instead of `args.appiumHome`
- `lib/cli/args.js`
    - Remove all server argument definitions in lieu of schema; leave a few that do not make sense in a config file (like `--config`, which determines the location of the config file!)
    - Server args are now pulled from the schema via a call to `toParserArgs()` with overrides for various validation functions (which are confusingly "types" in `argparse` parlance)
    - Instantiates the `DriverConfig` and `PluginConfig` instances, which are exported.  TODO: maybe move these into `extension-config`?? idk
- Tests
    - Added unit tests for `extension-config`, `grid-register` which were missing
    - Added unit tests for new `schema` and `schema-args` and `config-file` modules
    - Added e2e test for config file handling
    - Other necessary modifications
    - Fix path to `fake-driver` in `driver-e2e-specs`
    - Removed stuff from `parser-specs` which no longer applies

## Other

- Fixed fake-driver schema port definition (`min/max` to `minimum/maximum`)
- Added `scripts/generate-schema-declarations.js` which uses `json-schema-to-typescript` to create `appium-config.d.ts`
- Added config file samples in `.json` and `.yaml` formats
- Added custom gulp task to `appium/gulpfile.js` to generate schema in JSON format, since its "single source of truth" is actually just a big JS object
- Moved `appium` CLI-related tests into `test/cli`
- Added some declarations into `packages/appium/types`; the schema-related one is generated by the script mentioend above, but `types.d.ts` is just some stuff that was not expressible in a docstring


packages/appium/.appiumrc.json
```diff
@@ -0,0 +1,31 @@
+{
+  "server": {
+    "address": "0.0.0.0",
+    "port": 31337,
```
mykola-mokhnach, line 4: is it the new default port?

packages/appium/.appiumrc.json
```diff
@@ -0,0 +1,31 @@
...
+    "log": "/tmp/appium.log",
+    "log-no-colors": false,
+    "long-stacktrace": false,
+    "no-perms-check": false,
+    "nodeconfig": "",
+    "plugins": "all",
+    "relaxed-security": false,
+    "session-override": false,
+    "strict-caps":1,
+    "tmp": "/tmp",
```
mykola-mokhnach, line 27: would it work on Windows?

packages/appium/.appiumrc.json
```diff
@@ -0,0 +1,31 @@
+{
```
mykola-mokhnach, line 1: I prefer Yaml files more, since they support comments

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,293 @@
...
+        err.message = `Config file not found at user-provided path: ${filepath}`;
+      } else if (err instanceof SyntaxError) {
+        err.message = `Config file at user-provided path ${filepath} is invalid:\n${err.message}`;
+      }
+      throw err;
+    }
+  }
+  log.debug('No config file specified; searching...');
+  const result = await lc.search();
+  if (!result || !result.filepath) {
```
mykola-mokhnach, line 49: could be `!result?.filepath`

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,293 @@
...
+/**
+ * Inits Ajv, compiles the schema, and returns the validation function.
+ * Whenever this function is called, the validation function's `errors` property will be reset.
+ * If there are errors during validation, it will be set again.
+ */
+const validator = _.once(
+  /**
+   * @returns {import('ajv').ValidateFunction<import('./appium-config').AppiumConfigurationSchema>}
+   */
+  () => {
```
mykola-mokhnach, line 64: I prefer to use named functions in such context, since their names are visible in the debugger stack

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,293 @@
...
+    const {normalizeKeys = true, pretty = true} = opts;
+    /** @type {ReadConfigFileResult} */
+    let retval;
+    const errors = validateConfigFile(result.config);
+    if (errors.length) {
+      /** @type {string} */
+      let json;
+      try {
+        json = await fs.readFile(result.filepath, 'utf8');
+      } catch {}
```
mykola-mokhnach, line 111: is such syntax allowed?

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,293 @@
...
+ * Converts flag aliases to actual flags.
+ * Any flag shorter than 3 characters is prefixed with a single dash, otherwise two.
+ * @param {string} alias - A flag alias (e.g. `verbose`) to convert to a flag (e.g., `--verbose`)
+ * @throws {TypeError} If `alias` is falsy
+ */
+function aliasToFlag (alias) {
+  if (!alias) {
+    throw new TypeError('falsy alias value');
+  }
+  return alias.length < 3 ? `-${alias}` : `--${alias}`;
```
mykola-mokhnach, line 154: magic number here?

packages/appium/lib/main.js
```diff
@@ -21,7 +21,8 @@ import { runExtensionCommand } from './cli/extension';
 import { AppiumDriver } from './appium';
 import registerNode from './grid-register';
 import { inspectObject } from './utils';
-
+import { readConfigFile, getDefaultsFromSchema } from './config-file';
```
mykola-mokhnach, line 24: in the imports above spaces are missing before and after parentheses

packages/appium/lib/main.js
```diff
@@ -172,13 +169,34 @@ async function main (args = null) {
...
+  // the order of precendece is:
+  // 1. command line args
+  // 2. config file
+  // 3. defaults from config file
+  args = _.defaults(args, configResult.config?.server, getDefaultsFromSchema({prop: 'server'}));
+
   await logsinkInit(args);
 
+  if (configResult.config) {
+    const configFilepath = path.relative(process.cwd(), configResult.filepath);
```
mykola-mokhnach, line 189: is `filepath` always relative?

packages/appium/lib/main.js
```diff
@@ -172,13 +169,34 @@ async function main (args = null) {
...
+  }
+
   // if the user has requested the 'driver' CLI, don't run the normal server,
   // but instead pass control to the driver CLI
   if (args.subcommand === DRIVER_TYPE || args.subcommand === PLUGIN_TYPE) {
+    // TODO: extract driver/plugin-specific config from config file here, and merge
+    // it with the args.
     await runExtensionCommand(args, args.subcommand);
-    process.exit();
+    return;
```
mykola-mokhnach, line 199: why not to return?
does this change have anything to do with the original topic?

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,293 @@
...
+      const reason = betterAjvErrors(schema, result.config, errors, {
+        json,
+        format: pretty ? 'cli' : 'js',
+      });
+      retval = reason ? {...result, errors, reason} : {...result, errors};
+    } else {
+      retval = {...result, errors};
+    }
+    if (normalizeKeys) {
+      // XXX blah
```
mykola-mokhnach, line 121: ?

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,293 @@
...
+      retval.config = objectKeysToCamelCase(retval.config);
+      ['sever', 'plugin', 'driver'].forEach((key) => {
```
mykola-mokhnach, line 123: magic array?

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,293 @@
...
+    case 'integer':
+      argOpts.type = 'int';
+    // fallthrough
+    default:
+      if (assignDefaults) {
+        argOpts.default = subSchema.default ?? null;
+      }
+      break;
+  }
+  if (Array.isArray(subSchema.enum) && !_.isEmpty(subSchema.enum)) {
```
mykola-mokhnach, line 203: lodash also has _.isArray

packages/appium/lib/config.js
```diff
@@ -157,8 +158,9 @@ async function showConfig () {
 }
 
 function getNonDefaultArgs (parser, args) {
-  return parser.rawArgs.reduce((acc, [, {dest, default: defaultValue}]) => {
-    if (args[dest] && args[dest] !== defaultValue) {
+  const defaultsFromSchema = getDefaultsFromSchema({prop: 'server', assignDefaults: true});
+  return parser.rawArgs.reduce((acc, [, {dest}]) => {
+    if (args[dest] && args[dest] !== defaultsFromSchema[dest]) {
```
mykola-mokhnach, line 163: should we also check if `defaultsFromSchema[dest]` is defined?

packages/appium/lib/config.js
```diff
@@ -213,8 +215,6 @@ function validateServerArgs (parser, args) {
   }
 
   const validations = {
-    port: checkValidPort,
```
mykola-mokhnach, line 216: why these validators are removed?

packages/appium/test/config-schema.specs.cjs
```diff
@@ -0,0 +1,16 @@
+const Ajv = require('ajv');
```
mykola-mokhnach, line 1: is this file needed?

packages/appium/test/parser-specs.js
```diff
@@ -60,14 +60,14 @@ describe('Server Parser', function () {
     (() => {p.parse_args(['-dc', 'does/not/exist.json']);}).should.throw();
   });
   it('should parse --allow-insecure correctly', function () {
-    p.parse_args([]).allowInsecure.should.eql([]);
+    p.parse_args([]).should.have.property('allowInsecure', undefined);
```
mykola-mokhnach, line 63: does the parser in base driver code still work as expected if undefined is passed instead of an empty array?

packages/appium/.appiumrc.json
boneskull: lol, sorry, that was just for debugging

packages/appium/lib/config-file.js
boneskull: yes

packages/appium/lib/config-file.js
boneskull: indeed it is.

packages/appium/lib/config-file.js
boneskull: I mean... yes, sort of.  all of our aliases less than length 3 are usable with `-`, and anything longer is `--`.

packages/appium/lib/main.js
boneskull: I'm not sure I understand this question.  any filepath is always relative to the current working directory.  I can change this to display as absolute, but as far as I recall this was just for output.  config files _tend_ to be close to the cwd, practically speaking, so I thought it appropriate to display this way.

packages/appium/lib/main.js
boneskull: it doesn't.  I can revert it for this PR, but you shouldn't `process.exit()` while relying so heavily on async code.

packages/appium/lib/config.js
boneskull: mm, probably

packages/appium/lib/config.js
boneskull: they are handled by the schema itself, and validation of the arguments thru `ajv`

packages/appium/test/config-schema.specs.cjs
boneskull: nope, again a debugging thing

packages/appium/test/parser-specs.js
boneskull: how would you pass in `undefined`?  you mean just calling `parse_args()`?

boneskull:
@mykola-mokhnach I appreciate the attention, but I'm not really ready for the implementation to be reviewed...

packages/appium/lib/config.js
boneskull: those other ones that I _didn't_ remove... I have no idea what to do with those, because they aren't in `--help`

packages/appium/lib/config-file.js
boneskull: I think I was unhappy about the magic array, as you noted...

KazuCocoa:
Seems great. Appium 2.0 will be more configurable with drivers/plugins than 1.x, so this idea seems great for me.

> we want the precedence CLI ARGS > CONFIG FILE > DEFAULTS

👍 

> What should the CLI do if it encounters an invalid config file?

For now, I'd like to raise an error and stop the CLI. Perhaps current behavior also stops the process if cli was invalid... (my thought might be wrong tho..)
Then, ideally, information about which part is invalid helps to understand the reason.

> Would it be helpful to add a subcommand to validate a config file (I'm thinking yes)? Or better to live in appium-doctor?

As subcommand seems good to me +1

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,389 @@
...
+ * @returns {import('ajv').ErrorObject[]} Array of errors, if any.
+ */
+export function validateConfig (value) {
+  const validator = getValidator(APPIUM_CONFIG_SCHEMA_ID);
+  if (!validator) {
+    throw new Error(
+      `Could not find schema with ID: ${APPIUM_CONFIG_SCHEMA_ID}`,
+    );
+  }
+  return !validator(value) &&
```
jlipps, line 118: NIT: can we put `!validator(value) && _.isArray(validator.errors)` on the same line so the semantics of this ternary operator are more transparent?

packages/appium/lib/config-file.js
```diff
@@ -0,0 +1,389 @@
...
+ * @param {AppiumConfiguration} config - Configuration object
+ * @returns {NormalizedAppiumConfiguration} New object with camel-cased keys.
+ */
+function normalizeConfig (config) {
+  const jsonSchema =
+    /** @type {import('./schema').AppiumConfigJsonSchemaType} */ (
+      getValidator(APPIUM_CONFIG_SCHEMA_ID)?.schema
+    );
+  if (jsonSchema) {
+    return _.mapKeys(
```
jlipps, line 182: since it's a larger block of code, can we separate out rather than nest the mapping functions?

jlipps: there's already enough logical duplication in this set of maps, should we just go full recursive at this point?

packages/appium/lib/grid-register.js
```diff
@@ -9,20 +9,31 @@ const hubUri = (config) => {
...
 
-async function registerNode (configFile, addr, port, basePath) {
-  let data;
-  try {
-    data = await fs.readFile(configFile, 'utf-8');
-  } catch (err) {
-    logger.error(`Unable to load node configuration file to register with grid: ${err.message}`);
-    return;
+/**
+ * Registers a new node with a selenium gbrid
```
jlipps, line 13: grid

packages/appium/lib/types.d.ts
```diff
@@ -0,0 +1,27 @@
+import {AppiumConfiguration} from './appium-config';
+
+/*
+ * Utility types
+ */
+
+export type CamelCase<S extends string> =
```
jlipps, line 7: whoa did not know you could define restricted types of string based on templates!

jlipps (approved):

packages/appium/lib/driver-config.js
```diff
@@ -9,17 +13,24 @@ export default class DriverConfig extends ExtensionConfig {
   getConfigProblems (driver) {
     const problems = [];
     const automationNames = [];
-    const {platformNames, automationName} = driver;
+    const {platformNames, automationName, schema, installSpec, pkgName} = driver;
```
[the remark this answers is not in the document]
jlipps: Yeah. there is a schema insofar as I wrote those functions to validate it. But it'd probably make more sense to use Ajv etc now that we have all this machinery in place.

packages/appium/lib/extension-config.js
```diff
@@ -7,6 +7,7 @@ import YAML from 'yaml';
 
 const DRIVER_TYPE = 'driver';
 const PLUGIN_TYPE = 'plugin';
+const SERVER_TYPE = 'server';
```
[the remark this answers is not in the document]
boneskull: yes, it felt weird to me as well.  I am not sure where it'd make sense.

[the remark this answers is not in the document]
jlipps: maybe something like this in the arg parsing file or wherever we most use `SERVER_TYPE`:

```
import { DRIVER_TYPE, PLUGIN_TYPE } from './extension-config';
const DRIVER_SUBCOMMAND = DRIVER_TYPE;
const PLUGIN_SUBCOMMAND = PLUGIN_TYPE;
const SERVER_SUBCOMMAND = 'server';
```

packages/appium/lib/config-file.js
boneskull: yeah, I don't like this function.  working on something better

sakibguy:
very good!

packages/appium/lib/extension-config.js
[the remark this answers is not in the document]
boneskull: no longer applicable

packages/appium/lib/types.d.ts
boneskull: yeah, it's pretty new.  I don't really understand it too well; I asked a colleague for some assistance.

boneskull:
closing in lieu of #15938, #15937, #15936 and #15935

PR #15695, opened 2021-08-05
[WIP] refactor: Move args constraints to a class property

mykola-mokhnach:
## Proposed changes

This is actually the example implementation of my proposal from https://github.com/appium/appium-xcuitest-driver/pull/1307/files#r672825746

Such implementation would have the following pros:
- We don't keep unnecessary/unused data in the driver instance (`driverArgs`)
- We only have to add minimal changes to the base driver and inherited ones (basically to define args constraints only). There is no need to change constructors
- This scheme would work perfectly for both v1 and v2 drivers
- All args handling logic is located in a single place - the umbrella driver

I'm happy to collaborate on this proposal further.

## Types of changes

cc @jlipps @ayvnkhan @KazuCocoa

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -166,8 +166,8 @@ class AppiumDriver extends BaseDriver {
 
       let runningDriversData, otherPendingDriversData;
       const parsedInnerDriverArgs = parseExtensionArgs(this.args.driverArgs, driverName);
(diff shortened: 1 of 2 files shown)


jlipps (approved):
i like it! does fake-driver in this repo have any changes that go in tandem with this?

KazuCocoa (approved):
Looks like a great idea. Is CI not happy..?

PR #15702, opened 2021-08-06
chore: housecleaning

boneskull:
Just a bunch of random (mostly path-related) fixes.  These problems were surfaced by trying to run tests outside of the Gulp context (in the IDE, via Wallaby, `@babel/register`, etc.) and/or when working on #15664.

It undoes a couple of the changes I made earlier regarding path resolution--in particular, avoiding stuff that native Node.js ESM does not understand--since it's unlikely we will go that direction in the foreseeable future.

Not sure why the `appium-support` references were still hanging around; must have missed those earlier.

## Monorepo

- Make `npm run clean` work cross-platform
- Sort root `package.json`
- Update Babel configs to suppress warning about `@babel/plugin-proposal-private-property-in-object`
- Add `.nyc_output` and `*.lerna_backup` to `.gitignore`
- Replace references to `appium-support`
- Sync `package.json` fields

## appium

- `lib/config`: `package.json` resolution
- split some tests out of `config-specs.js` and into `config-e2e-specs.js` since they are very clearly not unit tests
- Improve `PROJECT_ROOT` resolution, which is still a hack
- `./fixtures/driverArgs.json` resolution

## @appium/base-driver

- `lib/basedriver/driver`: `package.json` resolution

## @appium/doctor

- `lib/doctor`: `package.json` resolution
- `lib/utils`: `package.json` resolution & exposed `resetLog()` which undoes what `configureBinaryLog()` does (for testing purposes)
- `utils-specs.js`: clean up after `onLogMessage` test
- copy fixtures into `build` upon transpile
- Upgrade `@appium/gulp-plugins`

## @appium/gulp-plugins

- Ensure `postTranspile` is run when `transpile` is run before tests
- Fix `.babelrc` like `babel.config.json`
-
## @appium/support

- Fixture resolution
- Copy fixtures when transpiling

### DIFF
--- a/packages/appium/test/driver-e2e-specs.js
+++ b/packages/appium/test/driver-e2e-specs.js
@@ -329,7 +329,9 @@ describe('FakeDriver - via HTTP', function () {
   });
 });
 
-describe('Logsink', function () {
+// TODO this test only works if the log has not previously been initialized in the same process.
+// there seems to be some global state that is not cleaned up between tests.
+describe.skip('Logsink', function () {
   let server = null;
   let logs = [];
   let logHandler = function (level, message) {
--- a/packages/support/lib/fs.js
+++ b/packages/support/lib/fs.js
@@ -1,4 +1,6 @@
 // jshint ignore: start
+import _ from 'lodash';
+import path from 'path';
 import _fs from 'fs';
 import rimraf from 'rimraf';
 import ncp from 'ncp';
@@ -9,12 +11,14 @@ import glob from 'glob';
 import crypto from 'crypto';
 import klaw from 'klaw';
 import sanitize from 'sanitize-filename';
+import findRoot from 'find-root';
 import { pluralize } from './util';
 import log from './logger';
 import Timer from './timing';
 
 const mkdirAsync = B.promisify(_fs.mkdir);
 const ncpAsync = B.promisify(ncp);
+const findRootCached = _.memoize(findRoot);
 
 const fs = {
   async hasAccess (path) {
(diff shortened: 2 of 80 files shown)


KazuCocoa (approved):
👍

packages/appium/test/driver-e2e-specs.js
```diff
@@ -329,7 +329,7 @@ describe('FakeDriver - via HTTP', function () {
   });
 });
 
-describe('Logsink', function () {
+describe.skip('Logsink', function () {
```
mykola-mokhnach, line 332: is there a way to mention why the particular test suite is skipped?

mykola-mokhnach (approved):

boneskull:
Note that we can solve issues with the paths to test fixtures easily by using a "copy files" task in Gulp.  _However_, we often want to get info (typically `version`) out of a `package.json`, which we cannot easily copy into `build` due to `package.json` being _special_.  For example, if we copy a `package.json` with a `main` field of `build/foo.js` into `build/package.json`, we will get a warning from Node.js about an invalid main module due to `build/build/foo.js` not existing.  This would also complicate publishing.  

A better thing to do, perhaps, is actually just _inline_ whatever we need from `package.json` at compile time.  Don't know if Babel can do this, but something like Rollup certainly can--or perhaps there's a Gulp task out there to do it.

boneskull:
Further, I think it'd make sense in the meantime to put a helper function in `@appium/support` to handle this.

boneskull:
Fixed a couple more issues and then added commit 6629e63 to make finding `package.json` less awful.

boneskull:
Did you know that `chai` is a singleton?  This means if you

```js
import chai from 'chai';
import cap from 'chai-as-promised';

chai.use(cap);
```

This affects all other test files which _also_ `import chai`.  Other test files could then assume `chai-as-promised` is present, and "forget" to `use()` it.  However, if those other test files are run _standalone_, then they will fail.  

Anyway, I fixed a couple of these problems, but I'm unsure if they are fixed everywhere.  

My recommendation to avoid this in the future (and you may hate this) is to actually just throw the `chai` initialization in some `test/setup.js` and run `mocha --require test/setup.js test/my-test.spec.js`.  Then you can either use `import chai` by itself in every file (without needing to `use()`, because Mocha will be guaranteed to have done this already).  You could also just create `global.should = chai.should()` and avoid `import chai` entirely 🙂

jlipps:
I like that idea!

boneskull:
will merge this after someone eyeballs the new functions in `fs`

packages/support/lib/fs.js
```diff
@@ -141,6 +145,39 @@ const fs = {
...
     });
+  },
+  /**
+   * Reads the closest `package.json` file from absolute path `dir`.
+   * @param {string} dir - Directory to search from
+   * @throws {TypeError} If `dir` is not a nonempty string or relative path
+   * @throws {Error} If there were problems finding or reading a `package.json` file
+   * @returns {object} A parsed `package.json`
+   */
+  readPackageJsonFrom (dir) {
```
mykola-mokhnach, line 156: It would be nice to have an e2e test for the newly added methods

packages/support/lib/fs.js
```diff
@@ -1,4 +1,6 @@
 // jshint ignore: start
```
mykola-mokhnach, line 1: How changes in these modules get versioned? Does CI bump their versions automatically?

packages/support/lib/fs.js
boneskull: I totally wrote tests for these but they must have gotten stashed or otherwise lost.  I'll dig them up and push

packages/support/lib/fs.js
boneskull: The version is determined at publish time by Lerna, based on conventional commit messages

boneskull:
@jlipps 

> I like that idea!

Which?  Using a `test/setup.js` or that _and_ setting `global.should`?

jlipps:
@boneskull both of them

PR #15719, opened 2021-08-12
chore: configure FOSSA

boneskull:
## Proposed changes

FOSSA was giving us a bunch of warnings about copyleft licenses, but these licenses are only present in our examples.  This configuration file ~will~ should tell FOSSA to ignore that stuff.

## Types of changes

Adds a `.fossa.yml`


boneskull:
I think I'm about satisfied with this.  FOSSA now seems to give results I'd expect.

I also reconfigured FOSSA so that it will only "fail" if there are license issues; FOSSA's security checks will still happen, but will not affect the build status.

boneskull:
still a couple things to fiddle with on the FOSSA side

jlipps (approved):

boneskull:
This is not working properly, as the builds aren't getting correctly associated with the project.  I've reached out to FOSSA support... we'll see what they say.

boneskull:
Ref: https://support.fossa.com/hc/en-us/requests/3086 (unsure if others can view)

jlipps:
Nope, can't view.

jlipps (approved):
is this ready to go? are you sure no more links to sample-code in the docs?

boneskull:
@jlipps It doesn't work properly, so no.  There _are_ references to `sample-code`, but those are things specifically pointing to the `master` branch, so I didn't think I should change them.

boneskull:
@jlipps This works now, so merging it.

PR #15720, opened 2021-08-12
chore: global chai

boneskull:
This PR makes `chai` and `should()` _globals_ when using Mocha:

- `chai` is configured with `chai-as-promised`
- A `.mocharc.js` requires `./test/setup.js` _from the monorepo root_.
- Updated the ESLint config to recognize the globals.
- Updated Wallaby config

All instances of `import chai...` and the like have been removed from all test files.

* * *

I'm wondering if this configuration should live in `@appium/test-support`?  It only seems to be consumed by `@appium/doctor`, so wasn't sure.

Consumers of `@appium/gulp-plugins` will not be affected by this change.


jlipps (approved):
*chefs kiss* beautiful

PR #15733, opened 2021-08-16
chore: get rid of shrinkwrap and related functionality

jlipps:
The main reason we had shrinkwrap with Appium 1.x was to provide the functionality that package-lock.json does for production releases, but without the annoying developer experience. We also wanted the ability to publish beta versions of Appium with no shrinkwrap (with the rationale that beta users could then simply get the latest drivers with a reinstall).

The latter motivation is moot in Appium 2.0 since you can get any drivers you want now independent of Appium version. And if we still want 100% reproducible Appium server production installs, moving forward we can use package-lock (or if we don't like the developer experience, simply pin deps to specific versions and let renovate update as necessary).


jlipps:
@mykola-mokhnach do you have any issues with this given the explanation above?

mykola-mokhnach:
I still prefer the good old existing shrinkwrapping strategy for releases and no strict versions pinning (e.g. leave it as is) and want to stick to my guns.

jlipps:
Can you please explain your rationale again? I don't think I fully understand it.

mykola-mokhnach:
So, basically I don't really like the idea about strict dependencies pinning, because it is convenient if beta always the latest and the greatest.
And releases must have strict dependencies to avoid unexpected issues related to submodules updates, which could only be achieved by shrinkwrapping/locking them at the time this particular release is done.

jlipps:
> it is convenient if beta always the latest and the greatest

With Appium 1.x, 99% of the motivation for this was because the drivers were bundled with Appium. So for people to get a fix in WDA, they had to install the Appium beta. With Appium 2.x, this will no longer be necessary. So the only convenience we get from not pinning dependencies with Appium 2.x is if we make a change to basedriver or one of those core packages.

And now with our monorepo setup, it's easy to run 1 command to publish a new beta of Appium with updated subpackages if that's what we want to do.

And with Renovate auto-merging green PRs, having dependencies pinned will actually be less work than it was for us previously, with the added benefit that each new dependency will get a build check (before, with version ranges, we would only find out if some dependency broke our stuff randomly).

> And releases must have strict dependencies to avoid unexpected issues related to submodules updates, which could only be achieved by shrinkwrapping/locking them at the time this particular release is done.

Pinning deps also does this. It also ensures all developers are running the same versions of subpackages, which is not the case currently.

mykola-mokhnach:
> And now with our monorepo setup, it's easy to run 1 command to publish a new beta of Appium with updated subpackages if that's what we want to do.

It is painful to publish a new beta for each fix. I would like to reduce the amount of manual actions.

jlipps:
> It is painful to publish a new beta for each fix

It's really not all that painful: `npx lerna publish`. And I would argue that every time we make a fix to the Appium core code (remember this is not driver code), we should just publish a new version anyway. I don't think we need the lengthy beta periods anymore, because most of that was due to wanting to ensure driver stability. That can be handled by simply publishing beta versions of the drivers now if we want.

In the meantime, I would like to publish new versions of Appium 2.0, but I cannot, since the shrinkwrap stuff appears to be broken and I'm not sure exactly how it's supposed to work anymore. Are you able to fix it?

mykola-mokhnach:
> In the meantime, I would like to publish new versions of Appium 2.0, but I cannot, since the shrinkwrap stuff appears to be broken and I'm not sure exactly how it's supposed to work anymore. Are you able to fix it?

No idea, main module publishing process was always a black box to me and I have never touched it. 
I was thinking about automating the stuff and we could do it by attaching CI to a new tag event. E.g. if one pushes a new tag to master then GitHub CI picks it up, verify its valid and publishes everything automatically without any manual intractions

jlipps:
> I was thinking about automating the stuff and we could do it by attaching CI to a new tag event.

I like this idea. But it seems like even slightly more work:

```
git tag -a "publish2" -m "trigger publish"
git push origin 2.0
```

vs

```
npx lerna publish
```

(Lerna keeps track of all package tags and versions for us already).

jlipps:
But if we want to do things without any kind of command line interaction whatsoever, we could wrap up `npx lerna publish` in a github action and just trigger it manually via github's UI. the action would clone the repo, run all tests, and then run `npx lerna publish`

mykola-mokhnach:
> But if we want to do things without any kind of command line interaction whatsoever, we could wrap up npx lerna publish in a github action and just trigger it manually via github's UI. the action would clone the repo, run all tests, and then run npx lerna publish

I would be happy to have something like that, - one magic button and the job is done

jlipps:
OK, I'm happy to work on the magic publish button thing! But right now I am blocked since the shrinkwrap situation keeps `npx lerna publish` from working correctly. That is one reason I thought now is a good time to get rid of the shrinkwrap. I guess you and I just disagree on that, so I'd love to hear from some other maintainers. @KazuCocoa @boneskull can you weigh in on the shrinkwrap vs no-shrinkwrap conversation?

boneskull:
I am not really familiar with using shrinkwrap.  I don't think it's necessary for Appium 2.x until it's obvious that it is.  We can always change it back.

Regarding 100%-automated publishing: I tend not to use it because I feel the one-step manual release is easy enough.  That said it'd be particularly useful for beta/prereleases.

KazuCocoa:
I guess the current main point is appium dependencies in `appium` package, so when we get changes in base-driver, base-plugin and support. https://github.com/appium/appium/blob/2.0/packages/appium/package.json#L54-L56

Currently `shrinkwrap` affects only when we push `appium` package. So, in `beta` channel, we do not need to re-push `appium` package to bump other relevant packages like base-driver's minor/patch release in beta. (latest channel has pinned all versions with shrinkwrap)

After 2.0 (without shrinkwrap), `npx lerna publish` to push `appium` package triggers other package releases in https://github.com/appium/appium/tree/2.0/packages as well. Other packages like base-driver in packages are also triggered in the appium package release. They have pinned dependencies. (So, users install the same dependencies versions)
Is it correct?

So, for example, when we want to release a new path base-driver version, then we run `npx lerna publish`. Then, packages in  https://github.com/appium/appium/tree/2.0/packages are also published with the latest dependencies. (with increasing the version automatically..?)

In beta release case also the same...?
So, I mean when we want to push appium 2.1 beta with base-driver 9.0 (for example), we run `npx lerna publish`. The 2.1 beta refers to the base-driver 9.0 in beta channel. When we add a small change in base-driver and want to make the version 9.0.1. Then, we should run `npx lerna publish`.

In this case, the appium 2.1 beta refers to base-driver 9.0.1 after the push command, or we also should bump appium 2.1 beta to appium 2.1.1 beta to refer to the base-driver 9.0.1? So, if installing `appium@beta` refers to appium 2.1 and it refers to base-driver 9.0, we should bump appium version to 2.1.1 to refer to the 9.0.1...?

Personally, if this version lock (with/without shrinkwrap) affects only this repository packages (base-driver, base-plugin, support) and we should call `npx lerna publish` more or less to push a new version in this repository's packages, it is enough to make release easier.

jlipps:
> After 2.0 (without shrinkwrap), npx lerna publish to push appium package triggers other package releases in https://github.com/appium/appium/tree/2.0/packages as well. Other packages like base-driver in packages are also triggered in the appium package release. They have pinned dependencies. (So, users install the same dependencies versions). Is it correct?

I believe so, yes. I checked and the result of the latest publish I did using Lerna exhibited exactly this behavior.

> In this case, the appium 2.1 beta refers to base-driver 9.0.1 after the push command, or we also should bump appium 2.1 beta to appium 2.1.1 beta to refer to the base-driver 9.0.1? So, if installing appium@beta refers to appium 2.1 and it refers to base-driver 9.0, we should bump appium version to 2.1.1 to refer to the 9.0.1...?

I'm not *entirely* sure, but I did poke around in Lerna's [publish code](https://github.com/lerna/lerna/blob/main/commands/publish/index.js) a bit and it seems to take these kinds of situations into account. It may also depend on whether we've pinned specific deps within this repo, or are allowing a caret range for them. I think we'll probably just need to experiment and see what happens.

For this to all work, it seems like we would need to do 2 things:
1. Switch to pinning specific dep versions (except maybe internal versions)
2. Check back in `package-lock.json` to all the packages within this repo (so that we get guaranteed reproducible installs for the same version of Appium).

jlipps:
OK, I'm going to take the outcome of this conversation as that we should get rid of shrinkwrap and associated methods for Appium 2.0, and give things a try with deps pinning, package locks, and frequent lerna publishes. If it doesn't work for us in practice, we can go back to shrinkwrap.

jlipps:
(With @mykola-mokhnach's objections fully noted)

PR #15734, opened 2021-08-16
docs(appium): add note to service/cloud providers in migration doc

jlipps:
I think it's important we get eyes from at least a few active service providers on this PR. cc @KazuCocoa @boneskull @Dharin-shah

### DIFF
--- a/docs/en/advanced-concepts/migrating-to-appium-2.0.md
+++ b/docs/en/advanced-concepts/migrating-to-appium-2.0.md
@@ -14,7 +14,7 @@ Together these do introduce a few breaking changes to how Appium is installed, h
 
 Have a look at the [Appium 2.0 release notes](https://github.com/appium/appium/releases/tag/v2.0.0-beta) for a more comprehensive list of changes. Here we call out the breaking changes and what you need to do do account for them.
 
-#### Installing drivers during setup
+#### :bangbang: Installing drivers during setup
 
 When you installed Appium 1.x, all available drivers would be installed at the same time as the main Appium server. This is no longer the case. Simply installing Appium 2.0 (e.g., by `npm install -g appium`), will install the Appium server only, but no drivers. To install drivers, you must instead use the new [Appium extension CLI](../drivers/driver-cli.md). For example, to install the latest versions of the XCUITest and UiAutomator2 drivers, after installing Appium you would run the following commands:
 
@@ -32,7 +32,7 @@ npm install -g appium --drivers=xcuitest,uiautomator2
 
 This will install Appium and the two drivers for you in one go.
 
-#### Driver updates
+#### :bangbang: Driver updates
 
 In the past, to get updates to your iOS or Android drivers, you'd simply wait for those updates to be rolled into a new release of Appium, and then update your Appium version. With Appium 2.x, the Appium server and the Appium drivers are versioned and released separately. This means that drivers can be on their own release cadence and that you can get driver updates as they happen, rather than waiting for a new Appium server release. The way to check for driver updates is with the CLI:
 
@@ -48,13 +48,13 @@ appium driver update xcuitest
 
 To update the Appium server itself, you do the same thing as in the past: `npm install -g appium`. Now, installing new versions of the Appium server will leave your drivers intact, so the whole process will be much more quick.
 
-#### Protocol changes
+#### :bangbang: Protocol changes
 
 Appium's API is based on the [W3C WebDriver Protocol](https://www.w3.org/TR/webdriver/), and it has supported this protocol for years. Before the W3C WebDriver Protocol was designed as a web standard, several other protocols were used for both Selenium and Appium. These protocols were the "JSONWP" (JSON Wire Protocol) and "MSJONWP" (Mobile JSON Wire Protocol). The W3C Protocol differs from the (M)JSONWP protocols in a few small ways.
 
 Up until Appium 2.0, Appium supported both protocols, so that older Selenium/Appium clients could still communicate with newer Appium servers. Moving forward, support for older protocols will be removed.
 
-*Capabilities*
+:bangbang: *Capabilities*
 
 One significant difference between old and new protocols is in the format of capabilities. Previously called "desired capabilities", and now called simply "capabilities", there is now a requirement for a so-called "vendor prefix" on any non-standard capabilities. The list of standard capabilities is given in the [WebDriver Protocol spec](https://www.w3.org/TR/webdriver/#capabilities), and includes a few commonly used capabilities such as `browserName` and `platformName`.
 
@@ -84,7 +84,7 @@ To make everyone's lives a bit easier, we've also introduced the option of wrapp
 
 (Of course, each client will have a different way of creating structured capabilities like `appium:options` or other ones that you might have seen such as `goog:chromeOptions`). NB: capabilities that show up in `appium:options` will overwrite capabilities of the same name that show up at the top level of the object.
 
-*Removed Commands*
+:bangbang: *Removed Commands*
 
 Commands which were a part of the old JSON Wire Protocol and not a part of the W3C Protocol are no longer available:
 
@@ -92,7 +92,7 @@ Commands which were a part of the old JSON Wire Protocol and not a part of the W
 
 If you use a modern Appium or Selenium client, you should no longer have access to these anyway, so any breaking changes should appear on the client side first and foremost.
 
-#### Image analysis features moved to plugin
+#### :bangbang: Image analysis features moved to plugin
 
 One of the design goals for Appium 2.0 is to migrate non-core features into special extensions called [plugins](#TODO). This allows people to opt into features which require extra time to download or extra system setup. The various image-related features of Appium (image comparison, finding elements by image, etc...) have been moved into an officially supported plugin called [images](https://github.com/appium/appium-plugins/tree/master/packages/images).
 
@@ -103,19 +103,19 @@ If you use these image-related methods, to continue accessing them you will need
 
 Image-related commands will also be removed on the client side of things, which means you will need to follow the instructions on the plugin README for installing client-side plugins to access these features.
 
-#### Old drivers removed
+#### :bangbang: Old drivers removed
 
 The old iOS and Android (UiAutomator 1) drivers and related tools (e.g., `authorize-ios`) have been removed. They haven't been relevant for many years anyway.
 
-#### Internal packages renamed
+#### :warning: Internal packages renamed
 
 Some Appium-internal NPM packages have been renamed (for example, `appium-base-driver` is now `@appium/base-driver`). This is not a breaking change for Appium users, only for people who have built software that directly incorporates Appium's code.
 
-#### "WD" JavaScript client library no longer supported
+#### :warning: "WD" JavaScript client library no longer supported
 
 For many years, some of Appium's authors maintained the [WD](https://github.com/admc/wd) client library. This library has been deprecated and has not been updated for use with the W3C WebDriver protocol. As such, if you're using this library you'll need to move to a more modern one. We recommend [WebdriverIO](https://webdriver.io).
 
-#### Appium Inspector split out from Appium Desktop
+#### :warning: Appium Inspector split out from Appium Desktop
 
 The inspecting portion of Appium Desktop has been moved to its own app, Appium Inspector: [github.com/appium/appium-inspector](https://github.com/appium/appium-inspector). It's fully compatible with Appium 2.0 servers. Simply download it and run it on its own. You no longer need the GUI Appium Desktop server to inspect apps. The Appium Desktop server will continue to be supported at its original site: [github.com/appium/appium-desktop](https://github.com/appium/appium-desktop). It will simply no longer bundle the Inspector with it.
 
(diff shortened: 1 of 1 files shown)


docs/en/advanced-concepts/migrating-to-appium-2.0.md
```diff
@@ -127,3 +127,81 @@ TODO
...
+
+With Appium 2.0, we enter a new era where end users may wish to target various independent versions of drivers and plugins. With Appium 1.x, this was never the case, since any given version of Appium would contain one and only one version of each driver. It is of course up to each service provider how they wish to implement the discovery, installation, and availability of any official or third party drivers or plugins. But the Appium team wants to make a recommendation in terms of the capabilities service providers support, for consistency across the industry. This is a recommendation only, and not a standard, but adopting it will help users to navigate the increased complexity that working with Appium 2.0 in a cloud environment may bring.
+
+### Suggested capabilities
+
+In addition to the standard `platformName`, `appium:deviceName`, `appium:automationName`, and `appium:platformVersion`, we recommend adopting the following set:
```
KazuCocoa, line 138: What about replacing `appium:` with `<your vendor prefix>:` (or similar) to clarify they are not appium capabilities?

Appium expects `appium:` prefix capabilities are for Appium, but they are not for Appium capabilities. They will always print as "not appium capabilities" warning message in a new session request log in appium. This could be noise. (not issue tho)

I guess this main reason is to standardized appium stuff for end-users to reduce confusion.
So, like...:

`<your vendor prefix>:appiumVersion`
`<your vendor prefix>:appiumAutomationVersion`  (I wondered "driverVersion", but mm, actually "automationVersion" is probably better naming for "automationName".
`<your vendor prefix>:appiumAutomation`
`<your vendor prefix>:appiumPlugins`

KazuCocoa (commented):
The idea is nice to me 👍

docs/en/advanced-concepts/migrating-to-appium-2.0.md
jlipps: yeah I think that makes sense to disambiguate it from the things that appium itself uses. it does mean each cloud would have a separate set but i guess that's OK because they are in fact different vendors.

KazuCocoa (approved):

docs/en/advanced-concepts/migrating-to-appium-2.0.md
```diff
@@ -127,3 +127,81 @@ TODO
...
+* `version`: the version of the extension, e.g., the NPM package version or Git SHA.
+* (optional) `source`: a denotation of where the extension can be downloaded from. Recommended to support the following values: `appium`, `npm`, `git`, `github`. Here, `appium` means "Appium's own official list", and should be the default value if this key is not included.
+* (optional) `package`: when downloading extensions from git or github, the NPM package name of the extension must also be provided. This is optional for non-git sources.
+
+Since each session is handled by a single driver, the `$cloud:automation` capability could be used with an extension object value to denote this driver, for example:
+
+```json
+{
+    ...,
+    "$cloud:automation": {
```
KazuCocoa, line 180: btw, I think most case to use this parameter is non-appium official drivers. (or this can be 
If this is an official xcuitest, for example, then the capabilities should be enough with the below:

```
{
    "$cloud:appiumVersion": "2.0.0",
    "platformName": "iOS",
    "appium:platformVersion": "14.4",
    "appium:deviceName": "iPhone 11",
    "appium:app": "Some-App.app.zip",
    "appium:automationName": "XCUITest",
    "$cloud:automationVersion": "3.52.0"
}
```

So, as an example here, I think it would be nice with non-npm appium custom drivers rather than simple npm like plugin's example.

Dharin-shah (approved):

docs/en/advanced-concepts/migrating-to-appium-2.0.md
jlipps: yep, good idea

docs/en/advanced-concepts/migrating-to-appium-2.0.md
```diff
@@ -127,3 +127,82 @@ TODO
...
+
+### Suggested capabilities
+
+In addition to the standard `platformName`, `appium:deviceName`, `appium:automationName`, and `appium:platformVersion`, we recommend adopting the following set, where the label `$cloud` is not meant to be interpreted literally but instead should be replaced by your vendor prefix (so for HeadSpin it would be `headspin`, Sauce Labs it would be `sauce`, and BrowserStack it would be `browserstack`, to name just a few examples):
```
boneskull, line 138: Why?

docs/en/advanced-concepts/migrating-to-appium-2.0.md
```diff
@@ -127,3 +127,82 @@ TODO
...
+|`$cloud:automationVersion`|Designating which version of the specified driver should be used.|`1.55.2`|
+|`$cloud:automation`|Designating a custom driver to use (see below for more info). This would override `appium:automationName` and `$cloud:automationVersion`|`{"name": "org/custom-driver", "source": "github", "package": "custom-driver"}`|
+|`$cloud:appiumPlugins`|Designating the list of plugins (and potentially versions of plugins) to be activated (see below for more info).|`["images", "universal-xml"]`|
+
+### Basic example
+
+Appium extensions (drivers and plugins) have a set of properties that specify where they can be installed from. Cloud providers are obviously under no obligation to provide support for arbitrarily specified extensions, seeing as these may represent untrusted code running in a managed environment. In the case where arbitrary extensions are not supported, the `appium:automationName`, `$cloud:automationVersion`, and `$cloud:appiumPlugins` capabilities should be sufficient. See the following JSON object representing capabilities for a session:
```
boneskull, line 149: Given that "images" was previously included by default, perhaps cloud providers would want to include this plugin by default.

docs/en/advanced-concepts/migrating-to-appium-2.0.md
```diff
@@ -127,3 +127,82 @@ TODO
...
+    "$cloud:automationVersion": "3.52.0",
+    "$cloud:appiumPlugins": ["images"]
+}
+```
+
+This set of capabilities requests an Appium 2.0 server supporting the XCUITest driver at version `3.52.0`, and the `images` plugin active. This set is easy for a cloud provider to verify. The cloud provider can obviously do anything it wants in response to these capabilities, including downloading Appium and driver and plugin packages on the fly, or erroring out if the versions requested are not in a supported set, or if the plugin is not supported, etc...
+
+### Extension objects
+
+Some service providers may wish to dynamically allow access to all of the features of the Appium 2.0 CLI, including downloading arbitrary drivers and plugins. To represent these extensions, we can define special JSON "extension objects", with the following keys:
```
boneskull, line 168: Just want to make sure I understand:

Are these capabilities intended to be provided _before_ the cloud provider starts a server?

docs/en/advanced-concepts/migrating-to-appium-2.0.md
jlipps: Why what? Why use cloud-specific vendor prefixes?

docs/en/advanced-concepts/migrating-to-appium-2.0.md
jlipps: I don't think we want to make that part of this recommendation (since Appium 2.0 explicitly doesn't include 'images' by default), but yes I agree it'd be a good idea for cloud providers to think about stuff like this and provide useful defaults unless otherwise specified.

docs/en/advanced-concepts/migrating-to-appium-2.0.md
boneskull: Yes.

docs/en/advanced-concepts/migrating-to-appium-2.0.md
boneskull: fair enough

docs/en/advanced-concepts/migrating-to-appium-2.0.md
jlipps: That would be an implementation detail, but generally speaking, yes, cloud providers don't start internal Appium servers until their front end has received capabilities and processed them and decided which ones to pass on to the internal Appium server, etc...

docs/en/advanced-concepts/migrating-to-appium-2.0.md
jlipps: As @KazuCocoa pointed out, we don't want to use `appium:` because it could confuse users into thinking these are capabilities the Appium server understands. And so the next sensible option seemed to be the cloud provider's existing vendor prefix, which they're presumably already using for cloud-specific capabilities.

docs/en/advanced-concepts/migrating-to-appium-2.0.md
```diff
@@ -127,3 +127,82 @@ TODO
...
+{
+    "$cloud:appiumVersion": "2.0.0",
+    "platformName": "iOS",
+    "appium:platformVersion": "14.4",
+    "appium:deviceName": "iPhone 11",
+    "appium:app": "Some-App.app.zip",
+    "appium:automationName": "XCUITest",
+    "$cloud:automationVersion": "3.52.0",
+    "$cloud:appiumPlugins": ["images"]
+}
```
christian-bromann, line 161: I could imagine this to work better:
```js
{
    "platformName": "iOS",
    "$cloud:appiumOptions": {
        "version": "2.0.0",
        "platformVersion": "14.4",
        "deviceName": "iPhone 11",
        "app": "Some-App.app.zip",
        "automationName": "XCUITest",
        "automationVersion": "3.52.0",
        "plugins": ["images"]
    }
}
```

It is not only consistent with what browserdriver do (e.g. `'goog:chromeOptions'`) , it also seems less repetitive in terms of having appium in the key. Furthermore I don't really see the necessity to differentiate between `appium:automationName` and `$cloud:automationVersion` , why would the one be Appium specific and the other cloud vendor specific?

christian-bromann: After thinking a bit more I believe I understood now: locally vs running in the cloud. `automationVersion` doesn't really make sense locally as you only have the version that you have installed. I would still tend to the this format given it is less verbose and more aligned to other implementations.

jlipps: Yes, `appium:automationName` is a standard Appium capability that the Appium server understands, and it means the least amount of changes when adjusting caps from local to cloud. IMO clouds should support both `appium:automationName` and `$cloud:automationName`.

jlipps: the idea of putting it all into `$cloud:appiumOptions` is interesting though, let's think about that.

docs/en/advanced-concepts/migrating-to-appium-2.0.md
Dharin-shah: yep ^

docs/en/advanced-concepts/migrating-to-appium-2.0.md
christian-bromann: On the Sauce Labs side we start to follow exactly this pattern as it better allows to group certain domain settings and also to extend them with new capabilities, e.g. we have `"sauce:visual": { ... }` and soon something like `"sauce:network": { ... }`.

The only downside I see from this grouping is that it makes it more difficult to differentiate between native Appium capabilities and possible cloud vendor capabilities, e.g. here you see the difference:

```js
    "appium:automationName": "XCUITest",
    "$cloud:automationVersion": "3.52.0",
```

but with grouping they are thrown into one bucket:

```js
    "$cloud:appiumOptions": {
        "automationName": "XCUITest",
        "automationVersion": "3.52.0"
    }
```

That said, I am not sure if this can conflict if Appium extends their caps with one that is already being used by a vendor. Then grouping would not work anymore but it would be still possible to differentiate when using proposed suggestion, e.g.:

```js
    "appium:foobar": "...",
    "$cloud:foobar": "...",
```

Another way to look at this is to group cloud specific Appium options and native ones:

```js
{
    "platformName": "iOS",
    "$cloud:appiumOptions": {
        "version": "2.0.0",
        "automationVersion": "3.52.0",
        "plugins": ["images"]
     },
     "appium:options": {
        "platformVersion": "14.4",
        "deviceName": "iPhone 11",
        "app": "Some-App.app.zip",
        "automationName": "XCUITest",
    }
}
```

This even seems much cleaner to me. It allows extending both without conflicts and avoids duplication.

KazuCocoa: Personally. when appium gets group section as its capabilities, I like the below format as suggested example since users can keep appium options as-is between their local and cloud.  (of course, cloud vendors can accept all capabilities as `$cloud:appiumOptions` and convert appium specific capabilities into `appium:` or `appium:options` (when it is available) in their internal for their local appium. It depends on their decision).

```json
{
    "platformName": "iOS",
    "$cloud:appiumOptions": {
        "version": "2.0.0",
        "automationVersion": "3.52.0",
        "plugins": ["images"]
     },
     "appium:options": {
        "platformVersion": "14.4",
        "deviceName": "iPhone 11",
        "app": "Some-App.app.zip",
        "automationName": "XCUITest",
    }
}
```

Current appium clients (in appium repository) adds `appium:` prefix for each capability that has no vendor prefix and non-w3c standard capabilities to reduce end-user side changes. When we support this style, it would be better to put such caps into appium options. (Then, will work for this)

jlipps: I like this idea too. Two things to consider though:

1. If we want to recommend cloud providers support `appium:options`, then we should also support it in the Appium 2.0 server, otherwise it'd mean too much moving around of caps when deciding whether to run on local or remote.
2. Tools like Appium Inspector are not set up to handle objects within the cap structure. There's currently no way for something like `appium:options` to work in the Inspector without manually writing the JSON. So this would entail doing some work there.

But I agree, I like the clean separation of this.

jlipps: took a stab at implementing support for the capability within the appium server: https://github.com/appium/appium/pull/15765

KazuCocoa: 👍

jlipps: ok, that's been merged and this doc updated with the new style

docs/en/advanced-concepts/migrating-to-appium-2.0.md
```diff
@@ -127,19 +127,129 @@ Apart from the breaking changes mentioned above, in this section is a list of so
...
+    "appium:deviceName": "iPhone 11",
+    "appium:app": "Some-App.app.zip",
+    "appium:automationName": "XCUITest",
+    "$cloud:appiumOptions": {
+        "appiumVersion": "2.0.0",
+        "automationVersion": "3.52.0",
+        "plugins": ["images"],
+    }
+}
+```
```
christian-bromann, line 180: > `$cloud:automationVersion`, and `$cloud:appiumPlugins`

Am I assuming right that Appium understands and treats the following cap examples in the same way:

```js
{
    "platformName": "iOS",
    "appium:platformVersion": "14.4",
    "appium:deviceName": "iPhone 11",
    "appium:app": "Some-App.app.zip",
    "appium:automationName": "XCUITest",
    "$cloud:appiumVersion": "2.0.0",
    "$cloud:automationVersion": "3.52.0",
    "$cloud:plugins": ["images"]
}
```
vs
```js
{
    "platformName": "iOS",
     "appium:options": {
        "platformVersion": "14.4",
        "deviceName": "iPhone 11",
        "app": "Some-App.app.zip",
        "automationName": "XCUITest"
    },
    "$cloud:appiumOptions": {
        "appiumVersion": "2.0.0",
        "automationVersion": "3.52.0",
        "plugins": ["images"],
    }
}
```
If we talk about `appium:options` a bit further above it seems confusing to use the other style in this example. If both ways are supported (thinking about backwards compat to Appium 1) I suggest to specially mention the differences.

jlipps: Hmm, that is what I tried to do with:

> The previous example still looks a bit disorganized, so of course we also recommend that cloud providers support the `appium:options` capability as detailed above,

Can you suggest better phrasing?

christian-bromann: Re-reading it now clears up the confusion. I think I missed the fact that `appium:automationName` is currently the way people pass in this information.

christian-bromann (approved):
LGTM 👍

PR #15742, opened 2021-08-17
docs(appium): add FAQ to 1.x EOL plan

jlipps:
any other questions/answers we should add here?

### DIFF
--- a/docs/en/about-appium/1.x-eol.md
+++ b/docs/en/about-appium/1.x-eol.md
@@ -1,12 +1,11 @@
 # Appium version 1 End Of Life (EOL) Plan
 
-Appium 1 has been around for quite a while and now is the time to move
-project development forward and to provide even better automation
-experience with Appium version 2.
+Appium 1 has been around for quite a while and now is the time to move project development forward
+and to provide even better automation experience with Appium version 2.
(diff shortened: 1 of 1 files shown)


KazuCocoa:
I remember that Java, Python and Ruby client have/had `forceMjsonwp` client side preference to sent only MJSONWP protocol so that its session follows non-W3C. It helped to keep non-w3c protocol native context as non-w3c as well as non-w3c chromedriver era.
( e.g. currently python client print warning message https://github.com/appium/python-client/blob/e48085d27968a93d9cfb5d86964ac84198b52214/appium/webdriver/webdriver.py#L376-L377 tho)

So, it would be nice to address this client side preference like:


Java, Python and Ruby appium bindings have `forceMjsonwp` capability to keep non-W3C session. The option no longer works with Appium 2.0 since the option does not send W3C spec capabilities. Please remove the option from your test case and update your test case properly from now on Appium 1.x so that you will face protocol relevant errors in Appium 2.x.

jlipps:
@KazuCocoa should we put that in the migration doc instead?

KazuCocoa:
ah.. yea, it seems better

PR #15806, opened 2021-09-03
feat!(appium): support driver-specific args via --driver-args

jlipps:
Co-authored-by: Ayan Khan <ayan@headspin.io>

cleaning up the git history for https://github.com/appium/appium/pull/15638 and pulling it here so I can do more work on it.


docs/en/writing-running-appium/server-args.md
```diff
@@ -1,23 +1,24 @@
...
+## Server params
 
+All params are optional, but some are required in conjunction with certain others.
 
 <expand_table>
 
-|Flag|Default|Description|Example|
+|Param|Default|Description|Example|
 |----|-------|-----------|-------|
-|`--shell`|null|Enter REPL mode||
```
jlipps, line 16: took the opportunity to remove this since it doesn't work anyway.

docs/en/advanced-concepts/migrating-to-appium-2.0.md
```diff
@@ -32,6 +32,28 @@ npm install -g appium --drivers=xcuitest,uiautomator2
 
 This will install Appium and the two drivers for you in one go.
 
+#### :bangbang: Driver-specific command line options
```
mykola-mokhnach, line 35: 👍

docs/en/writing-running-appium/server-args.md
```diff
@@ -1,23 +1,24 @@
...
 |`-a`, `--address`|0.0.0.0|IP Address to listen on|`--address 0.0.0.0`|
 |`-p`, `--port`|4723|port to listen on|`--port 4723`|
 |`-pa`, `--base-path`|null|Initial path segment where the Appium/WebDriver API will be hosted. Every endpoint will be behind this segment.|`--base-path /wd/hub`|
+|`--driver-args`|`{}`|JSON object string (or path to JSON file) consiting of CLI arguments to send to specific drivers. Should be of the form `{"[driverName]": {"[driverArg1Name]": [driverArg1Value], ...}}`|`--driver-args '{"xcuitest": {"webkitDebugProxyPort": 1234}}'|
+|`--allow-cors`|false|Turn on CORS compatibility mode, which will allow connections to the Appium server from within websites hosted on any domain. Be careful when enabling this feature, since there is a potential security risk if you visit a website that uses a cross-domain request to initiate or introspect sessions on your running Appium server.||
```
mykola-mokhnach, line 21: I assume it makes sense to mention Appium inspector here

packages/appium/lib/utils.js
```diff
@@ -34,18 +33,42 @@ function inspectObject (args) {
...
   }
-  return extensionSpecificArgs;
+  return extensionArgs[extensionName];
 }
 
+
+function ensureNoUnknownArgs (extensionArgs, argsConstraints) {
+  const knownArgNames = Object.keys(argsConstraints);
+  for (const argName of Object.keys(extensionArgs)) {
+    if (!knownArgNames.includes(argName)) {
```
mykola-mokhnach, line 50: I would rather collect all unsupported values first. We could quickly this by calling `_.difference` lodash helper

packages/appium/lib/utils.js
```diff
@@ -34,18 +33,42 @@ function inspectObject (args) {
...
+ * Takes in a set of driver/plugin args passed in by user, and arg constraints
+ * and throws an error if any arg is unknown or of the incorrect type
+ *
+ * @param {object} extensionArgs - Driver or Plugin specific args
+ * @param {object} argsConstraints - Constraints for arguments
+*/
+function validateExtensionArgs (extensionArgs, argsConstraints) {
+  if (!_.isEmpty(extensionArgs) && !_.isEmpty(argsConstraints)) {
+    ensureNoUnknownArgs(extensionArgs, argsConstraints);
+    validateCaps(extensionArgs, argsConstraints);
```
mykola-mokhnach, line 67: we could probably rename this method on import to make it more readable, e.g. `import {validateCaps as validateArgs}`

packages/fake-driver/lib/driver.js
```diff
@@ -69,7 +77,7 @@ class FakeDriver extends BaseDriver {
 
   async getFakeDriverArgs () {
     await B.delay(1);
-    return this.driverArgs;
+    return this.cliArgs;
```
mykola-mokhnach, line 80: +1: cliArgs better reflects the variable purpose

jlipps:
Note that CI will fail until I publish https://github.com/appium/appium-plugins/pull/18

KazuCocoa:
[just a confirmation]
Will arguments addressed in docs/en/writing-running-appium/server-args.md like `--allow-insecure` keep existing usage?
Just wondered since the security argument also could have driver specific items.
(In my current understanding is options in the docs/en/writing-running-appium/server-args.md works as appium-wide options)

packages/appium/lib/appium.js
```diff
@@ -176,23 +176,38 @@ class AppiumDriver extends BaseDriver {
...
+
+      // Likewise, any driver-specific CLI args that were passed in should be assigned directly to
+      // the driver so that they cannot be mimicked by a malicious user sending in capabilities
+      const driverCliArgs = getExtensionArgs(this.args.driverArgs, driverName);
+      if (!_.isEmpty(driverCliArgs)) {
+        if (!_.has(InnerDriver, 'argsConstraints')) {
+          throw new Error(`You sent in CLI args for the ${driverName} driver, but the driver ` +
+                          `does not define any`);
+        }
+        // parse/verify CLI args and merge them to this.args if OK or throw an exception
```
mykola-mokhnach, line 202: is this comment still relevant?

packages/appium/lib/appium.js
```diff
@@ -406,11 +421,21 @@ class AppiumDriver extends BaseDriver {
...
-      const name = `${PluginClass.pluginName} (${sessionId})`;
-      const opts = {
-        pluginArgs: parseExtensionArgs(this.args.pluginArgs, PluginClass.pluginName),
-      };
-      return new PluginClass(name, opts);
+      const generalName = PluginClass.pluginName;
+      const name = `${generalName} (${sessionId})`;
+      const plugin = new PluginClass(name);
+      const pluginCliArgs = getExtensionArgs(this.args.pluginArgs, generalName);
+      if (!_.isEmpty(pluginCliArgs)) {
```
mykola-mokhnach, line 428: I can observe some code duplication here

packages/appium/lib/utils.js
```diff
@@ -34,18 +33,41 @@ function inspectObject (args) {
...
-function parseExtensionArgs (extensionArgs, extensionName) {
-  if (!_.isString(extensionArgs)) {
+function getExtensionArgs (extensionArgs, extensionName) {
+  if (!_.has(extensionArgs, extensionName)) {
     return {};
   }
-  const parsedExtensionArgs = parseJsonStringOrFile(extensionArgs);
-  const extensionSpecificArgs = parsedExtensionArgs[extensionName];
-  if (!_.isPlainObject(extensionSpecificArgs)) {
+  if (!_.isPlainObject(extensionArgs[extensionName])) {
```
mykola-mokhnach, line 40: Maybe TypeError would be more descriptiive

packages/appium/lib/utils.js
```diff
@@ -34,18 +33,41 @@ function inspectObject (args) {
...
     throw new Error(`Driver or plugin arguments must be plain objects`);
   }
-  return extensionSpecificArgs;
+  return extensionArgs[extensionName];
 }
 
+
+function ensureNoUnknownArgs (extensionArgs, argsConstraints) {
+  const knownArgNames = Object.keys(argsConstraints);
+  const unknownArgs = _.difference(Object.keys(extensionArgs), Object.keys(argsConstraints));
```
mykola-mokhnach, line 49: the second arg could be `knownArgNames`

packages/appium/lib/utils.js
```diff
@@ -34,18 +33,41 @@ function inspectObject (args) {
...
+  if (unknownArgs.length > 0) {
+    throw new Error(`Some arguments were not recognized: ${JSON.stringify(unknownArgs)}. ` +
+                    `Are you sure they are in the list of supported args? ${JSON.stringify(knownArgNames)}`);
+  }
+}
+
+/**
```
mykola-mokhnach, line 56: It would be useful to have docstrings for all newly added utility functions

packages/appium/lib/appium.js
```diff
@@ -176,23 +176,38 @@ class AppiumDriver extends BaseDriver {
...
+      // the driver so that they cannot be mimicked by a malicious user sending in capabilities
+      const driverCliArgs = getExtensionArgs(this.args.driverArgs, driverName);
+      if (!_.isEmpty(driverCliArgs)) {
+        if (!_.has(InnerDriver, 'argsConstraints')) {
+          throw new Error(`You sent in CLI args for the ${driverName} driver, but the driver ` +
+                          `does not define any`);
+        }
+        // parse/verify CLI args and merge them to this.args if OK or throw an exception
+        validateExtensionArgs(driverCliArgs, InnerDriver.argsConstraints);
+        driverInstance.cliArgs = driverCliArgs;
```
mykola-mokhnach, line 204: where do we want to put the logic that reflects cliArgs onto driver capabilities (basically rewrites or adds them if needed)?

jlipps: I don't think we do. Drivers need to be able to distinguish what has been passed in as a CLI arg and what has been passed in as a capability. I don't want someone to be able to send in a capability of the same name as a CLI arg and make the driver think that CLI arg has been used. CLI args are for sysadmins only (like setting the security flags etc).

So that is why we keep cli args on `driver.cliArgs` and capabilities get merged to `driver.opts`.

mykola-mokhnach: yes, this makes sense. I just wanted to know where exactly capabilities should be modified based on cliArgs. For example: 
There is --chromedriver-executable uia2 arg. what it does is basically adds/overrides the corresponding capability, so the executable is always takes from the corresponding server argument instead of user-provided caps. Now my questions is: we have `cliArgs`, where the value of the arg is stored. Where do we want to override capabilities with this value (and how) : in the uia2 driver code or in the umbrella driver code? I'm asking because we also need to verify caps and the override must happen before the verification.

docs/en/writing-running-appium/server-args.md
```diff
@@ -1,23 +1,24 @@
...
 
-|Flag|Default|Description|Example|
+|Param|Default|Description|Example|
 |----|-------|-----------|-------|
-|`--shell`|null|Enter REPL mode||
-|`--allow-cors`|false|Turn on CORS compatibility mode, which will allow connections to the Appium server from within websites hosted on any domain. Be careful when enabling this feature, since there is a potential security risk if you visit a website that uses a cross-domain request to initiate or introspect sessions on your running Appium server.||
 |`-a`, `--address`|0.0.0.0|IP Address to listen on|`--address 0.0.0.0`|
 |`-p`, `--port`|4723|port to listen on|`--port 4723`|
 |`-pa`, `--base-path`|null|Initial path segment where the Appium/WebDriver API will be hosted. Every endpoint will be behind this segment.|`--base-path /wd/hub`|
+|`--driver-args`|`{}`|JSON object string (or path to JSON file) consiting of CLI arguments to send to specific drivers. Should be of the form `{"[driverName]": {"[driverArg1Name]": [driverArg1Value], ...}}`|`--driver-args '{"xcuitest": {"webkitDebugProxyPort": 1234}}'|
```
youngfreeFJS, line 20: What is the difference between `--dc` and `--driver-args`? 
For example, I now use `--dc` to set uiautomator2ServerLaunchTimeout, but it seems a driver arg.

mykola-mokhnach: basically driver-args can be anything while default capabilities are, well, only capabilities. Also The latter could be overridden by user-provided values, while driver args have the highest priority and cannot be overridden.

youngfreeFJS: Thanks, got it. 👀
The doc seems to need to explain: `the driver args have the highest priority and cannot be overridden.`

jlipps: `--dc` sets default capabilities that will exist on any new session.  drivers might want to get information not from capabilities but from the command line, to ensure that only a system administrator can set it, and not just anyone. in some cases there can also be an overlap: some options can be set both from capabilities or the CLI.

jlipps: oops, didn't realize this conversation already happened when i replied. i think we probably need a special doc explaining the difference between capabilities and arguments, that we can link from here.

packages/appium/lib/appium.js
jlipps: why does the override need to happen before verification? do you think my solution in https://github.com/appium/appium-xcuitest-driver/pull/1331 would work?

jlipps:
@KazuCocoa 
> Will arguments addressed in docs/en/writing-running-appium/server-args.md like --allow-insecure keep existing usage? Just wondered since the security argument also could have driver specific items.

Yes. `--allow-insecure` is Appium wide, so drivers can already make use of insecure flags however they want via that CLI arg. We should document this behavior for driver creators so they know they can have users put in flags and access them via `this.allowInsecure`.

packages/appium/lib/appium.js
mykola-mokhnach: added my comments to that PR

PR #15820, opened 2021-09-08
ci: ensure commits made by ci map to github users

jlipps:
This way CI commits can work for the CLA.

### DIFF
--- a/ci-jobs/generate-docs.yml
+++ b/ci-jobs/generate-docs.yml
@@ -15,6 +15,8 @@ jobs:
         npm install
         brew install hub
         hub --version
+        git config --global user.name "Appium Maintainers"
+        git config --global user.email "maintainers@appium.io"
         git remote add triager https://triager:$GITHUB_TOKEN@github.com/appium/appium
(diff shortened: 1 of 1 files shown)


brianwarner:
LGTM. For context, this is fixing an issue where the commits co-authored by `runner@Mac-1628453453384.local` are not associated with a GitHub account, and cannot be pre-approved in EasyCLA. By associating these commits with maintainers@appium.io (which does have a GitHub account) we can add them to the pre-approved list and they won't fail the EasyCLA checks.

For the existing PRs that include commits co-authored by `runner@Mac-1628453453384.local`, these will likely fail the EasyCLA checks because there's no associated GitHub user to be pre-approved in EasyCLA. In the very narrow scope where a PR is failing solely because `runner@Mac-1628453453384.local` has not signed EasyCLA, a maintainer may bypass the failing check and merge the code. However, this only applies to this very limited situation.

jlipps:
No way to really test this without merging, so let's merge it and see what happens next run.

Issue #15566, opened 2021-06-22
Appium 2.0 docs planning


jlipps:
_This issue tracks the umbrella proposal for how docs of all kinds will work in Appium 2.0. It may be divided into smaller tasks_

## The big picture

At the present time, disregarding Appium 2.0, our docs are a bit confusing and outdated. Some of the main problems are:

* The marketing website hasn't been updated in probably 5-6 years. 
* We have a sophisticated docs generation system that has a bunch of great organization and information, but some of the information is outdated, and it proved to be too much to keep an exhaustive list of every command in every driver updated.
* We have docs translations but in the past we were told it was pretty bad (excepting perhaps the Chinese documentation, which is excellent Chinese but also probably very outdated).
* We have a sample code directory which contains all or mostly non functional code.
* It's difficult to find docs for a specific version of Appium
* Docs are in multiple places (the Appium docs site at appium.io as well as in various driver READMEs). There is a tension between desire not to duplicate and desire to have all the information in one place.

On top of all this, we are now moving to Appium 2.0, which potentially makes the problem worse in that Appium 2.0 allows for (and hopes for) an explosion of new drivers and plugins, each of which would require its own documentation. To list the problems more specifically:

* Some of these extensions might be "officially" supported, and many will not be, but all will need documentation.
* There is currently no documentation for developers that will enable them to build new extensions with confidence, and this is important for growing the ecosystem.
* So far none of Appium 2.0's new features (the driver/plugin CLI, for example) are documented.
* We need a 1.x -> 2.x migration doc that we can keep updated with the current situation, listing all breaking changes and workarounds.

## Design goals

* Prefer stable links over duplication of information (which could easily get stale).
* Clearly differentiate the types of documentation we have: overviews, conceptual articles, how-to guides, articles on specific complex features, API documentation, and contributing/developer docs.
* Keep documentation closer to the source code (i.e., drivers should publish their own documentation rather than syndicating it to the main docs site). This makes it possible to request/update documentation in the same PR as code changes, which in turn makes it more likely that we will keep our docs up to date with reality.
* Optimize for easy localization, using standard i18n frameworks to allow crowd-sourced translations or easy translation PRs

## The proposal

_TODO_

#### Working ideas

* Create a driver/plugin docs framework that any driver/plugin can use simply by implementing a bunch of YAML and MD files. Then each driver/plugin could simply run a script to get beautiful documentation in a style and format consistent with every other Appium driver/plugin.
* Make docs easily navigable
[...]

PR #15821, opened 2021-09-08
docs(appium): develop new TOC structure for appium 2.0

jlipps:
As part of https://github.com/appium/appium/issues/15566, I'm trying to put together a TOC that will make sense for Appium 2.0. The idea isn't to merge this yet, rather to use this PR as a basis for making sure we have identified all the aspects of documentation we want to be able to host at the main Appium docs site.

Everything that I'm removing here is either obsolete and should just be deleted, or is related to a specific driver or plugin and should be moved over there. This PR can also be a good checklist of docs to move and docs to create.

One note specifically about command/API documentation. I think all command docs should be in individual drivers. My hope is that we can come up with a standard way for drivers to document their commands, which could make this process easy and maintainable.


jlipps:
@mykola-mokhnach @KazuCocoa @christian-bromann @boneskull would appreciate any ideas here. what are the things we would want to see in the appium documentation?

mykola-mokhnach:
I'm fine with any solution, which could make our documentation better. It was always the part of the project that was suffering much.

mykola-mokhnach:
btw, is it expected that integration tests fail?

jlipps:
no, the error is due to an unpinned dev dep updating in a way that broke the tests. they will be fixed when the cli args PR is merged.

docs/toc.js
```diff
@@ -2,327 +2,42 @@ module.exports = {
...
+    ['Plugins', ['plugins',
+      ['The Plugin CLI', 'plugin-cli.md'],
+      ['List of Known Plugins', 'known-plugins.md'],
+      ['Client-side Plugins', 'client-plugins.md'],
+      ['Building and Sharing Plugins', 'building-plugins.md'],
     ]],
     ['Commands', ['commands',
       ['Status', 'status.md'],
-      ['Execute Mobile Command', 'mobile-command.md'],
-      ['Session', ['session',
```
KazuCocoa, line 30: I believe each `mobile` command can be each repository, but I'm been considering common apis such as `Create` session API will be...

Some descriptions can remove from each api (e.g. which driver/client support a API), but I still think the list of apis should be in this central doc. (e.g. the left side's list in https://w3c.github.io/webdriver/ as quick search/index commands)
It is easier to share commands, search commands for beginners. (Still I'm considering tho)

docs/toc.js
```diff
@@ -2,327 +2,42 @@ module.exports = {
...
+      ['CLI Arguments', 'server-args.md'],
+      ['Driver and Plugin Arguments', 'extension-args.md'],
       ['Desired Capabilities', 'caps.md'],
       ['The --default-capabilities flag', 'default-capabilities-arg.md'],
-      ['Finding Elements', 'finding-elements.md'],
-      ['Touch Actions', 'touch-actions.md'],
-      ['CLI Arguments', 'server-args.md'],
       ['Server Security', 'security.md'],
-      ['Web/Web Views', ['web',
-        ['Mobile Web Testing', 'mobile-web.md'],
```
KazuCocoa, line 240: How like this common-method/Tutorial/tips etc will be?
(Ideally) Move them into each driver repository...? (I'd like to know current idea.)

docs/toc.js
```diff
@@ -2,327 +2,42 @@ module.exports = {
...
-        ['Execute', 'execute.md'],
-      ]],
+      ['Driver-specific commands', 'driver-commands.md'],
     ]],
-    ['Writing & Running Tests', ['writing-running-appium',
-      ['Running Tests', 'running-tests.md'],
+    ['The Server', ['server',
+      ['CLI Arguments', 'server-args.md'],
+      ['Driver and Plugin Arguments', 'extension-args.md'],
       ['Desired Capabilities', 'caps.md'],
```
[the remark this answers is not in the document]
jlipps: yes, the idea behind this caps.md in the main repo is simply to talk about what capabilities are, and then to list any global capabilities (e.g., `newCommandTimeout`). Driver-specific capabilities would not be documented here.

docs/toc.js
jlipps: > I believe each mobile command can be each repository, but I'm been considering common apis such as Create session API will be...

Yes, any global commands should also be here, for example `Create Session` and `Delete Session`, or anything else which is handled by the main Appium server and not a specific driver.

> but I still think the list of apis should be in this central doc. 

I guess I disagree. I think it's just an opportunity for confusion. Maybe in the future, if we have an automatic way to determine which commands a driver supports (like I was proposing in #15566), then we could have a big matrix of commands and driver support here (for officially recognized drivers). But in the short term, I think it's better to ensure that drivers themselves document their commands in an easy-to-understand way. So here would be my suggestion for a path to follow:

1. Document commands for specific drivers in the driver repos themselves (necessary before Appium 2.0 release)
2. Remove non-global commands from main Appium docs, and simply link to driver docs (for Appium 2.0 release)
3. At some point in the future, when driver command docs/generation/extraction is totally standardized, we can plan a way to incorporate officially recognized driver docs as part of the main Appium docs site if we want.

docs/toc.js
jlipps: Yep, exactly, move them into each driver repository. I think each driver should have its own docs generation and docs site, or at least links between markdown files to start, so that all of this information is not lost, just moved to the appropriate drivers.

youngfreeFJS:
> Everything that I'm removing here is either obsolete and should just be deleted, or is related to a specific driver or plugin and should be moved over there. This PR can also be a good checklist of docs to move and docs to create.

Some immature ideas about Appium-plugin.  Should Appium have a third-party plugin market(Like [VSCode Extension Marketplace](https://code.visualstudio.com/docs/editor/extension-marketplace) )? And some rules to manage the plugin client and **plugin document**. 
@jlipps

PR #15830, opened 2021-09-10
chore(base-driver): start to document BaseDriver types for TS users

jlipps:
I decided to write a typescript plugin to see what we would need in terms of type definitions, so here's my first stab.

### DIFF
--- a/packages/base-driver/index.d.ts
+++ b/packages/base-driver/index.d.ts
@@ -1,0 +1,121 @@
+import { W3C_ELEMENT_KEY } from './lib/protocol/helpers'
+
+declare class BaseDriver {
+    // class variables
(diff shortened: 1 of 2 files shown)


jlipps:
going through this code is also helping me realize how much of an opportunity for some nice refactoring there is here :-)

KazuCocoa (commented):
nice 👍
Will this pr include TODO comment as well?

jlipps:
I think I will actually add the commands before merging

jlipps:
OK, I've added the rest of the things I found in BaseDriver. There are a few types that could probably be expanded later on, but this at the very list will be quite helpful for driver developers.

KazuCocoa (approved):

PR #15849, opened 2021-09-14
feat(appium): add env var to trigger reloading of extensions

jlipps:
this is useful during driver/plugin development, so the appium server does not always
need to be restarted, at least not when extensions are created on new sessions

### DIFF
--- a/packages/appium/lib/extension-config.js
+++ b/packages/appium/lib/extension-config.js
@@ -203,7 +203,13 @@ export default class ExtensionConfig {
 
   require (extName) {
     const {mainClass} = this.installedExtensions[extName];
-    return require(this.getExtensionRequirePath(extName))[mainClass];
+    const reqPath = this.getExtensionRequirePath(extName);
+    const reqResolved = require.resolve(reqPath);
+    if (process.env.APPIUM_RELOAD_EXTENSIONS && require.cache[reqResolved]) {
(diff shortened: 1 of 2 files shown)


jlipps:
@mykola-mokhnach any objections to this?

packages/appium/lib/extension-config.js
```diff
@@ -203,7 +203,13 @@ export default class ExtensionConfig {
 
   require (extName) {
     const {mainClass} = this.installedExtensions[extName];
-    return require(this.getExtensionRequirePath(extName))[mainClass];
+    const reqPath = this.getExtensionRequirePath(extName);
+    const reqResolved = require.resolve(reqPath);
+    if (process.env.APPIUM_RELOAD_EXTENSIONS && require.cache[reqResolved]) {
```
mykola-mokhnach, line 208: how about `APPIUM_FORCE_EXTENSIONS_REALOAD`?

mykola-mokhnach (approved):
it would be nice to have this feature documented

packages/appium/lib/extension-config.js
jlipps: hmm, that makes it longer, and it will never reload them under any circumstances otherwise, so I don't see the need for 'FORCE' here

KazuCocoa (approved):

PR #15858, opened 2021-09-16
feat(base-driver): allow drivers, and plugins to declare that certain…

jlipps:
… routes must never be proxied

pretty easy fix for https://github.com/appium/appium-plugins/pull/22#issuecomment-921061315 (cc @KazuCocoa )


KazuCocoa (approved):
oh, nice idea.

PR #15895, opened 2021-09-25
chore(base): Enable system unzip usage also in Windows by default

KazuCocoa:
https://github.com/appium/appium-base-driver/pull/510/ for 2.0 branch.

This requires https://github.com/appium/appium/pull/15882.
So, probably we need to do:

1. bump the support version
2. update this package.json

How should I do to release a new support version...? @jlipps


KazuCocoa:
oh, ok.
This PR requires appium/appium-base-driver#510 , so the next push can have both.

KazuCocoa:
https://github.com/appium/appium/pull/16082

PR #16059, opened 2021-11-09
refactor plugin handling to fix interaction with proxies

jlipps:
In the past, it was actually not possible to use plugins well with drivers that proxied commands. This was because Appium chose whether or not to proxy a request before it set up the plugin management chain. This pull request tries to make things work by deferring proxying until a later point if a plugin is going to handle the command. The net result is that if and when a plugin calls `await next()`, the proxying will happen at that point, and it will take place 'internally' to the driver/plugin system rather than writing directly to the HTTP response object. This increases the complexity and requires that drivers expose more than the current `proxyReqRes` function (which writes directly to the HTTP response object and as such doesn't work in our case). They now need to also expose a `proxyCommand` function, which is typically a matter of assigning a new alias to the jwproxy object.

Related PRs:
- [x] https://github.com/appium/appium-espresso-driver/pull/719 (espresso driver update)
- [x] https://github.com/appium/appium-android-driver/pull/715 (android driver update, for chrome webview)
- [x] https://github.com/appium/appium-uiautomator2-driver/pull/480 (uiauto2 driver update)
- [x] https://github.com/appium/appium-plugins/pull/64 (plugins repo update)

The XCUITest driver doesn't an update since it already exposes `proxyCommand` as required.

Other drivers might also need updates. I just focused on the core mobile ones right now. I added a nice error message so it should be clear if and when a user tries to use a plugin with a driver that needs to be updated.

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -391,19 +391,17 @@ class AppiumDriver extends BaseDriver {
    * @returns {Array} - array of plugin instances
    */
   pluginsForSession (sessionId = null) {
-    let plugins = [];
     if (sessionId) {
       if (!this.sessionPlugins[sessionId]) {
         this.sessionPlugins[sessionId] = this.createPluginInstances();
       }
-      plugins = this.sessionPlugins[sessionId];
-    } else {
-      if (_.isEmpty(this.sessionlessPlugins)) {
-        this.sessionlessPlugins = this.createPluginInstances();
-      }
-      plugins = this.sessionlessPlugins;
+      return this.sessionPlugins[sessionId];
     }
-    return plugins;
+
+    if (_.isEmpty(this.sessionlessPlugins)) {
+      this.sessionlessPlugins = this.createPluginInstances();
+    }
+    return this.sessionlessPlugins;
   }
 
   /**
@@ -423,16 +421,11 @@ class AppiumDriver extends BaseDriver {
       .filter((p) => _.isFunction(p[cmd]) || _.isFunction(p.handle));
   }
 
-  createPluginInstances (sessionId) {
-    if (!sessionId) {
-      sessionId = 'sessionless';
-    }
-    sessionId = _.truncate(sessionId, {length: 11});
+  createPluginInstances () {
     return this.pluginClasses.map((PluginClass) => {
-      const generalName = PluginClass.pluginName;
-      const name = `${generalName} (${sessionId})`;
+      const name = PluginClass.pluginName;
       const plugin = new PluginClass(name);
-      this.assignCliArgsToExtension('plugin', generalName, plugin);
+      this.assignCliArgsToExtension('plugin', name, plugin);
       return plugin;
     });
   }
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -3,7 +3,6 @@ import log from './logger';
 import { getBuildInfo, updateBuildInfo, APPIUM_VER } from './config';
 import { findMatchingDriver } from './drivers';
 import { BaseDriver, errors, isSessionCommand } from '@appium/base-driver';
-import B from 'bluebird';
 import AsyncLock from 'async-lock';
 import { parseCapsForInnerDriver, pullSettings } from './utils';
 import { util } from '@appium/support';
@@ -276,34 +275,27 @@ class AppiumDriver extends BaseDriver {
   }
 
   attachUnexpectedShutdownHandler (driver, innerSessionId) {
-    const removeSessionFromMasterList = (cause = new Error('Unknown error')) => {
-      log.warn(`Closing session, cause was '${cause.message}'`);
+    const onShutdown = (cause = new Error('Unknown error')) => {
+      log.warn(`Ending session, cause was '${cause.message}'`);
+
+      if (this.sessionPlugins[innerSessionId]) {
+        for (const plugin of this.sessionPlugins[innerSessionId]) {
+          if (_.isFunction(plugin.onUnexpectedShutdown)) {
+            log.debug(`Plugin ${plugin.name} defines an unexpected shutdown handler; calling it now`);
+            plugin.onUnexpectedShutdown(driver, cause);
+          } else {
+            log.debug(`Plugin ${plugin.name} does not define an unexpected shutdown handler`);
+          }
+        }
+      }
+
       log.info(`Removing session '${innerSessionId}' from our master session list`);
       delete this.sessions[innerSessionId];
       delete this.sessionPlugins[innerSessionId];
     };
 
-    // eslint-disable-next-line promise/prefer-await-to-then
-    if (_.isFunction((driver.onUnexpectedShutdown || {}).then)) {
-      // TODO: Remove this block after all the drivers use base driver above v 5.0.0
-      // Remove the session on unexpected shutdown, so that we are in a position
-      // to open another session later on.
-      driver.onUnexpectedShutdown
-        // eslint-disable-next-line promise/prefer-await-to-then
-        .then(() => {
-          // if we get here, we've had an unexpected shutdown, so error
-          throw new Error('Unexpected shutdown');
-        })
-        .catch((e) => {
-          // if we cancelled the unexpected shutdown promise, that means we
-          // no longer care about it, and can safely ignore it
-          if (!(e instanceof B.CancellationError)) {
-            removeSessionFromMasterList(e);
-          }
-        }); // this is a cancellable promise
-    } else if (_.isFunction(driver.onUnexpectedShutdown)) {
-      // since base driver v 5.0.0
-      driver.onUnexpectedShutdown(removeSessionFromMasterList);
+    if (_.isFunction(driver.onUnexpectedShutdown)) {
+      driver.onUnexpectedShutdown(onShutdown);
     } else {
       log.warn(`Failed to attach the unexpected shutdown listener. ` +
         `Is 'onUnexpectedShutdown' method available for '${driver.constructor.name}'?`);
@@ -539,7 +531,7 @@ class AppiumDriver extends BaseDriver {
       const sessionId = res.value[0];
       log.info(`Promoting ${this.sessionlessPlugins.length} sessionless plugins to be attached ` +
                `to session ID ${sessionId}`);
-      this.pluginsForSession[sessionId] = this.sessionlessPlugins;
+      this.sessionPlugins[sessionId] = this.sessionlessPlugins;
       this.sessionlessPlugins = [];
     }
 
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -2,7 +2,8 @@ import _ from 'lodash';
 import log from './logger';
 import { getBuildInfo, updateBuildInfo, APPIUM_VER } from './config';
 import { findMatchingDriver } from './drivers';
-import { BaseDriver, errors, isSessionCommand } from '@appium/base-driver';
+import { BaseDriver, errors, isSessionCommand,
+         CREATE_SESSION_COMMAND } from '@appium/base-driver';
 import AsyncLock from 'async-lock';
 import { parseCapsForInnerDriver, pullSettings } from './utils';
 import { util } from '@appium/support';
@@ -282,7 +283,11 @@ class AppiumDriver extends BaseDriver {
         for (const plugin of this.sessionPlugins[innerSessionId]) {
           if (_.isFunction(plugin.onUnexpectedShutdown)) {
             log.debug(`Plugin ${plugin.name} defines an unexpected shutdown handler; calling it now`);
-            plugin.onUnexpectedShutdown(driver, cause);
+            try {
+              plugin.onUnexpectedShutdown(driver, cause);
+            } catch (e) {
+              log.warn(`Got an error when running plugin ${plugin.name} shutdown handler: ${e}`);
+            }
           } else {
             log.debug(`Plugin ${plugin.name} does not define an unexpected shutdown handler`);
           }
@@ -492,12 +497,10 @@ class AppiumDriver extends BaseDriver {
         // in case it calls 'await next()'. This requires that the driver have defined
         // 'proxyCommand' and not just 'proxyReqRes'.
         if (!dstSession.proxyCommand) {
-          throw new Error(`The default behavior for this command was to proxy, but the driver ` +
-                          `did not have the 'proxyCommand' method defined. To fully support ` +
-                          `plugins, drivers should have 'proxyCommand' set to a jwpProxy object's ` +
-                          `'command()' method, in addition to the normal 'proxyReqRes'`);
+          throw new NoDriverProxyCommandError();
         }
-        return await dstSession.proxyCommand(reqForProxy.originalUrl, reqForProxy.method, reqForProxy.body);
+        return await dstSession.proxyCommand(reqForProxy.originalUrl, reqForProxy.method,
+          reqForProxy.body);
       }
 
       if (isGetStatus) {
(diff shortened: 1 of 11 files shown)


jlipps:
CI will fail until the plugin changes are merged and published, FYI

jlipps:
I did e2e testing with the uiauto2 driver and the universal xml plugin, this seems to work as expected

packages/appium/lib/appium.js
```diff
@@ -276,34 +275,27 @@ class AppiumDriver extends BaseDriver {
...
   attachUnexpectedShutdownHandler (driver, innerSessionId) {
-    const removeSessionFromMasterList = (cause = new Error('Unknown error')) => {
-      log.warn(`Closing session, cause was '${cause.message}'`);
+    const onShutdown = (cause = new Error('Unknown error')) => {
+      log.warn(`Ending session, cause was '${cause.message}'`);
+
+      if (this.sessionPlugins[innerSessionId]) {
+        for (const plugin of this.sessionPlugins[innerSessionId]) {
+          if (_.isFunction(plugin.onUnexpectedShutdown)) {
+            log.debug(`Plugin ${plugin.name} defines an unexpected shutdown handler; calling it now`);
```
boneskull, line 284: note that `plugin.name` can theoretically be an empty string. you might want `plugin.constructor.pluginName`?

packages/appium/lib/appium.js
```diff
@@ -492,6 +486,20 @@ class AppiumDriver extends BaseDriver {
       // if we make it here, we know that the default behavior is handled
       cmdHandledBy.default = true;
 
+      if (reqForProxy) {
+        // we would have proxied this command had a plugin not handled it, so the default behavior
+        // is to do the proxy and retrieve the result internally so it can be passed to the plugin
+        // in case it calls 'await next()'. This requires that the driver have defined
+        // 'proxyCommand' and not just 'proxyReqRes'.
+        if (!dstSession.proxyCommand) {
+          throw new Error(`The default behavior for this command was to proxy, but the driver ` +
```
boneskull, line 495: I would love it if we'd have specific subclasses of `Error`, each with their own unique `code` property. Never too late to start!!

boneskull: Also... this is rather verbose.  Maybe in the future we could provide a URL to view detailed explanations about each error (indexed by `code`, of course).

packages/appium/lib/appium.js
```diff
@@ -516,6 +524,17 @@ class AppiumDriver extends BaseDriver {
     // handling the command and which didn't
     plugins.length && this.logPluginHandlerReport({cmd, cmdHandledBy});
 
+    // And finally, if the command was createSession, we want to migrate any plugins which were
+    // previously sessionless to use the new sessionId, so that plugins can share state between
+    // their createSession method and other instance methods
+    if (cmd === 'createSession' && this.sessionlessPlugins.length && !res.error) {
+      const sessionId = res.value[0];
```
boneskull, line 531: since you used `_.last()`, what about `_.first()`? :wink:

packages/appium/lib/appium.js
```diff
@@ -516,6 +524,17 @@ class AppiumDriver extends BaseDriver {
...
+      log.info(`Promoting ${this.sessionlessPlugins.length} sessionless plugins to be attached ` +
```
boneskull, line 532: btw, if you hate having to split template strings, there's a [lib](https://npm.im/common-tags) containing tagged template functions which make this less ornery

packages/appium/lib/appium.js
```diff
@@ -516,6 +524,17 @@ class AppiumDriver extends BaseDriver {
     // handling the command and which didn't
     plugins.length && this.logPluginHandlerReport({cmd, cmdHandledBy});
 
+    // And finally, if the command was createSession, we want to migrate any plugins which were
+    // previously sessionless to use the new sessionId, so that plugins can share state between
+    // their createSession method and other instance methods
+    if (cmd === 'createSession' && this.sessionlessPlugins.length && !res.error) {
```
boneskull, line 530: I might want to create a constant for this string

packages/appium/lib/appium.js
```diff
@@ -516,6 +524,17 @@ class AppiumDriver extends BaseDriver {
     // handling the command and which didn't
     plugins.length && this.logPluginHandlerReport({cmd, cmdHandledBy});
 
+    // And finally, if the command was createSession, we want to migrate any plugins which were
+    // previously sessionless to use the new sessionId, so that plugins can share state between
+    // their createSession method and other instance methods
```
boneskull, line 529: I think I could use some more explanation (though not necessarily in the comment).

- What's a sessionless plugin?
- Does this "migration" of plugins happen once, or multiple times?  
- Why wouldn't this happen in the `createSession()` implementation?

packages/appium/lib/appium.js
```diff
@@ -516,6 +524,17 @@ class AppiumDriver extends BaseDriver {
     // handling the command and which didn't
     plugins.length && this.logPluginHandlerReport({cmd, cmdHandledBy});
```
boneskull, line 525: tangential, but maybe `logPluginHandlerReport` should handle "no plugins" instead of `plugins.length &&`

packages/base-driver/lib/protocol/protocol.js
```diff
@@ -239,9 +239,17 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       // this command should never be proxied (which is useful for plugin developers who add
       // commands and generally would not want that command to be proxied instead of handled by the
       // plugin)
+      let pluginOverrodeProxy = false;
       if (isSessCmd && !spec.neverProxy && driverShouldDoJwpProxy(driver, req, spec.command)) {
-        await doJwpProxy(driver, req, res);
-        return;
+        if (!driver.pluginsToHandleCmd ||
```
boneskull, line 244: I wonder if `pluginsToHandleCmd` should be a noop by default?  I do not love with all the checking-for-existence -of-functions which seems to happen often 😛

packages/base-driver/lib/protocol/protocol.js
```diff
@@ -283,12 +291,15 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
         `${driver.constructor.name}.${spec.command}() with args: ` +
         _.truncate(JSON.stringify(args), {length: MAX_LOG_BODY_LENGTH}));
 
-      if (driver.executeCommand) {
-        driverRes = await driver.executeCommand(spec.command, ...args);
-      } else {
-        driverRes = await driver.execute(spec.command, ...args);
+      if (pluginOverrodeProxy) {
```
boneskull, line 294: What's the most direct way of understanding wtf a JSONWP server is and why we'd proxy commands to it?

packages/base-driver/lib/protocol/protocol.js
```diff
@@ -283,12 +291,15 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
...
+        // TODO for now we add this information on the args list, but that's mixing purposes here.
+        // We really should add another 'options' parameter to 'executeCommand', but this would be
+        // a breaking change for all drivers so would need to be handled carefully.
+        args.push({reqForProxy: req});
       }
 
+      driverRes = await driver.executeCommand(spec.command, ...args);
```
boneskull, line 301: What is `driver.execute()`?

packages/base-driver/test/basedriver/driver-tests.js
```diff
@@ -335,8 +335,8 @@ function baseDriverUnitTests (DriverClass, defaultCaps = {}) {
         it('should have a #canProxy method', function () {
           d.canProxy.should.be.an.instanceof(Function);
         });
-        it('should return false from #canProxy', function () {
-          d.canProxy(sessId).should.be.false;
+        it('should return a boolean from #canProxy', function () {
+          [true, false].should.include(d.canProxy(sessId));
```
boneskull, line 339: I'd probably write this as

```suggestion
          d.canProxy(sessId).should.be.a('boolean');
```

but that's just me

packages/fake-driver/lib/driver.js
```diff
@@ -24,6 +25,27 @@ class FakeDriver extends BaseDriver {
     };
   }
 
+  proxyActive () {
```
boneskull, line 28: So this function _doesn't_ actually need a `sessionId`?

packages/appium/lib/appium.js
boneskull: Is the order of plugins relevant?

Should `sessionPlugins` really be a `Set`?

jlipps: the order is relevant. they do a russian nesting doll sort of thing so depending on the order in which they are declared on the CLI, the ultimate behaviour might be different.

jlipps: and here plugin.name has been assigned by us so we should be able to trust it

packages/appium/lib/appium.js
```diff
@@ -276,34 +275,27 @@ class AppiumDriver extends BaseDriver {
...
-    const removeSessionFromMasterList = (cause = new Error('Unknown error')) => {
-      log.warn(`Closing session, cause was '${cause.message}'`);
+    const onShutdown = (cause = new Error('Unknown error')) => {
+      log.warn(`Ending session, cause was '${cause.message}'`);
+
+      if (this.sessionPlugins[innerSessionId]) {
+        for (const plugin of this.sessionPlugins[innerSessionId]) {
+          if (_.isFunction(plugin.onUnexpectedShutdown)) {
+            log.debug(`Plugin ${plugin.name} defines an unexpected shutdown handler; calling it now`);
+            plugin.onUnexpectedShutdown(driver, cause);
```
mykola-mokhnach, line 285: should it be awaited?

packages/appium/lib/appium.js
```diff
@@ -276,34 +275,27 @@ class AppiumDriver extends BaseDriver {
...
+          } else {
+            log.debug(`Plugin ${plugin.name} does not define an unexpected shutdown handler`);
+          }
+        }
+      }
+
       log.info(`Removing session '${innerSessionId}' from our master session list`);
       delete this.sessions[innerSessionId];
```
mykola-mokhnach, line 293: shall we put these statements into a `finally` block in case the supplied handler throws some unexpected error?

packages/base-driver/lib/protocol/protocol.js
```diff
@@ -239,9 +239,17 @@ function buildHandler (app, method, path, spec, driver, isSessCmd) {
       // this command should never be proxied (which is useful for plugin developers who add
       // commands and generally would not want that command to be proxied instead of handled by the
       // plugin)
+      let pluginOverrodeProxy = false;
```
mykola-mokhnach, line 242: I usually prefix boolean vars with `did` or `is`

mykola-mokhnach:
Is it expected that CI is failing?

packages/appium/lib/appium.js
jlipps: that's a good idea. i think that would best be its own project of just grepping the code and systematically creating these things. in general i don't mind a bit of verbosity in my error messages, they also serve as good context for developers in the code.

packages/appium/lib/appium.js
jlipps: ok fine :-)

packages/appium/lib/appium.js
jlipps: yes, i like that! i still find js tagged template functions kind of ugly, and don't want to add an extra lib in this PR but could be a good thing to consider making standard or optional in the future.

packages/appium/lib/appium.js
jlipps: yeah, it was already constantified in base driver so i'll just hook up the import/export pipes.

packages/appium/lib/appium.js
jlipps: > What's a sessionless plugin?

a plugin which exists outside the context of a session. for example, a plugin which is handling the `createSession` command is by definition not associated with any particular session (but then, once the session is created, we want to associate the plugin with that particular session, and not any other sessions that might be happening on the server at once.) a sessionless plugin might also handle `getStatus` or `getSessions`, which are likewise server-level commands and not session-level commands.

> Does this "migration" of plugins happen once, or multiple times?

It should happen on every new session. Which makes me realize there's a bug here, in that sessionless plugins are not "backfilled". I'll need to address that.

> Why wouldn't this happen in the createSession() implementation?

There's no guarantee that Appium's `createSession` will be called, since plugins can override it.

jlipps: > Which makes me realize there's a bug here, in that sessionless plugins are not "backfilled". I'll need to address that.

No, I was wrong, sessionless plugins are recreated as necessary based on the fact that we empty out the list after we 'migrate' them.

packages/appium/lib/appium.js
jlipps: yeah reasonable

packages/base-driver/lib/protocol/protocol.js
jlipps: then this would break when used with a driver that doesn't rely on the latest version of basedriver. so we'll be stuck with the check anyway if we want compatibility with older drivers.

jlipps: oh, also, i don't really want the concept of plugins to apply to drivers other than the umbrella appium driver. it's sort of a special case. i don't like that we have to mention it in the protocol at all, there's probably some massive rearchitecture that's possible to get proxying and plugins working together more generically.

all that to say, i only want `driver.pluginsToHandleCmd` to exist when the protocol is working with `AppiumDriver`, not any other driver. if i trusted instanceof checks on custom classes I guess we could do it that way

packages/base-driver/lib/protocol/protocol.js
jlipps: JSONWP is just the old name of the WebDriver protocol. long ago we created the jsonwp-proxy package which allows us to proxy webdriver commands to another webdriver compatible server. the jsonwp protocol itself is defunct, but the package has been updated to use the w3c webdriver protocol.

so the strings "jsonwp" or "jwp" (same thing) are still in our code and our mostly vestigial. at some point we need to go through and actually remove all efforts to handle the (actual) jsonwp, and at that point could rename everything that's left as "wd" or something, so "wdproxy" instead of "jwproxy" or whatever.

jlipps: but to answer 'why we'd proxy commands to it', many of our drivers are set up as a node.js shim which talks to a webdriver server on the device (e.g. WDA or the UiAuto 2 server). because we have this proxy mode, it means we don't have to write any code at all in the node.js portion of the driver and commands will be transparently proxied to the server on the device).

packages/base-driver/lib/protocol/protocol.js
jlipps: it was an old way of doing what `executeCommand` does that we left for backwards compatibility with older appium drivers. this probably could have been removed years ago and just doing it now since i read the TODO.

packages/base-driver/test/basedriver/driver-tests.js
jlipps: that's what I tried to do, with `.should.be.a(Boolean)`, but it broke and I didn't know there was a better way!

packages/fake-driver/lib/driver.js
jlipps: Nope, one is required in the umbrella driver since it can handle multiple sessions and so disambiguation is required. but an instance of fake driver only ever represents its own session.

packages/appium/lib/appium.js
jlipps: i thought about this, but that would break the signature of this method and require us to put `awaits` in the other drivers. better to require the plugin author to use appropriate error handling if they wish it to be an `async` function. we are any rate not going to do anything if there is an error here.

packages/appium/lib/appium.js
jlipps: yeah

jlipps: actually opted for just logging a warning and catching the only possible source of exceptions.

packages/base-driver/lib/protocol/protocol.js
jlipps: 👍

jlipps:
> Is it expected that CI is failing?

yes, until the plugin PR was merged and published. it has been done so so I will push all my PR updates and hopefully see CI turn green.

jlipps:
Yay, CI passed.

jlipps:
@mykola-mokhnach @boneskull any last comments? would like to merge this and publish a new appium 2 beta

packages/appium/lib/appium.js
mykola-mokhnach: putting await won't break anything. It's simply ignored if the returned result is not a promise

mykola-mokhnach (approved):

packages/base-driver/lib/protocol/protocol.js
boneskull: the server on the device... is this natively supported, or?

packages/base-driver/lib/protocol/protocol.js
boneskull: ok, I'll ask more about this later

boneskull (approved):
I'm good

packages/base-driver/lib/protocol/protocol.js
jlipps: no. by "server on the device" i mean "some software we wrote that runs on a device, usually a server, usually launched inside a test case because what the device thinks we are doing is running an XCTest or a UiAutomator test". for an example of "server on the device" see WebDriverAgent. "server on the device" is exactly what it would be great if Apple and Google made instead of us.

PR #16071, opened 2021-11-10
fix(appium): remove extra logging from config-file

boneskull:
Logging was happening here before the logger was properly initialized, which looked weird.  Just remove logs for now until we find a better solution.

In addition, fixed some typing problems:
- we need `@types/node` for node globals (like errors that have a `code` property)
- set various typings as globals: `node`, `mocha`, `chai` & its ilk


jlipps (approved):

jlipps:
LGTM, seems like tests need updating though

PR #16072, opened 2021-11-10
fix(appium): add missing support for schema-as-object

boneskull:
I had intended to allow extensions to define a schema via `package.json` instead of a path to a file, and apparently forgot to implement this.  So, now it is implemented.


boneskull:
Essentially, this allows [this changes](https://github.com/appium/appium-xcuitest-driver/pull/1353/files) to actually work.  In the case of `fake-driver`, the value of `appium.schema` is just `./build/lib/fake-driver-schema.js` instead of an object.

jlipps (approved):

PR #16080, opened 2021-11-12
feat(appium): support for deprecated arguments

boneskull:
This PR adds the `appiumDeprecated` keyword which can be used to mark an argument as deprecated in the `--help` text.

In addition:

- fix a bug where `--help` would cause Appium to exit with a nonzero exit code (it would essentially `throw Error(undefined)`)
- remove unused `lib/cli/argparse-actions` and stuff that used it, since that stuff was also unused
- implement `appiumCliIgnore`, though it's not actually being used atm.
- extract helpers for running `appium` into `test/cli/cli-helpers.js`

* * *

Note that this does _not_ support all of the functionality as the original deprecation system which used a custom `argparse` `Action` and did a few fancy things.  If we need fancy things in the future, we can implement them in the future.  Basically, the only thing this does is prepend `[DEPRECATED]` to the help text description.


boneskull:
A newer version of the JSON schema draft has support for a `deprecated` keyword which would allow validation of config files in this manner.  Unfortunately, we can only handle deprecations in CLI arguments until we decide to upgrade the base metaschema to draft-09.  We could do that, but `Ajv` claims there's a performance impact, so I'd rather wait until it's absolutely needed.  It should not be too bad, as I believe it is backwards-compatible with draft-07 (which is what we're using).

KazuCocoa (approved):
lgtm.
mm, node 12 and 14 got unhappy CI 👀

docs/en/advanced-concepts/configuration-system.md
```diff
@@ -271,3 +271,5 @@ Keywords are defined in `lib/schema/keywords.js`.
 - `appiumCliDest`: allows a schema to specify a custom property name in the post-`argprase` arguments objects. If not set, this becomes a camelCased string.
 - `appiumCliDescription`: allows a schema to override the description of the argument when displayed on the command-line. This is useful paired with `appiumCliTransformer` (or `array`/`object`-typed properties), since there's a substantial difference between what a CLI-using user can provide vs. what a config-file-using user can provide.
 - `appiumCliTransformer`: currently a choice between `csv` and `json`. These are custom functions which post-process a value. They are not used when loading & validating config files, but the idea should be that they result in the same object you'd get if you used whatever the config file wanted (e.g., an array of strings). `csv` is for comma-delimited strings and CSV files; `json` is for raw JSON strings and `.json` files.
+- `appiumCliIgnore`: If `true`, do not support this property on the CLI.
```
mykola-mokhnach, line 274: I usually add `is` to better reflect a variable is of boolean type

jlipps (approved):

boneskull:
Looks like an `npm link` failure.  I wonder if this has something to do with permissions and how `npm link` behaves in those older node versions...  anyone seen anything like this?

docs/en/advanced-concepts/configuration-system.md
boneskull: I agree, but in this case, it's not a variable--it's a schema keyword.  the convention doesn't use "is", and we also want to be sure we are namespacing the keyword.

but could do `appiumShouldCliIgnore` or `appiumCliShouldIgnore` or something as @jlipps suggests

boneskull: @jlipps I changed it to `appiumCliIgnored`, which is more obviously a boolean (precedent: the `deprecated` keyword is a boolean, and uses the same tense).  That work?

jlipps: yep!

boneskull:
This might be solved by #16118

boneskull:
Rebased onto `2.0` after #16118 merged

boneskull:
I should buy a lottery ticket

jlipps (approved):

PR #16088, opened 2021-11-12
chore: update FOSSA workflow

boneskull:
This PR moves the FOSSA check from the Node workflow to its own.

Due to https://github.com/fossas/spectrometer/issues/346 this will now be a nightly check, and will _fail_ if there are unaddressed license issues (whereas before it would ignore them).  It will likely fail the first time it runs, and we'll need to go into the FOSSA app and manually set the issues to be ignored since they seem to only occur in multi-license repos or sample code.

It will check the `2.0` branch nightly.


boneskull:
Note that the path check isn't happening yet, so I can test.

boneskull:
either FOSSA is hosed or I have hosed something. hosing all around.  marking as draft until I can resolve this with FOSSA

boneskull:
(see updated description)

KazuCocoa (approved):

boneskull:
upstream issue is now https://github.com/fossas/fossa-cli/issues/694

jlipps (approved):

boneskull:
Actually... I think I'll just merge this and see what happens.  That's really the only way to test a scheduled action anyhow...

PR #16100, opened 2021-11-15
chore: fix lerna publish

boneskull:
hopefully.

here's what I did:

1. ran `lerna link convert` which converts all packages which reference _other_ packages to `file:..` urls and adds the same kind of urls to these packages as "dependencies" in the root monorepo.  the upshot here is that we no longer need `lerna bootstrap` because `npm install` from the root will get everything.
2. update packages whose `bin` or `main` fields reference a build artifact; these now use a file in the root which just re-exports whatever the built file was.  otherwise, npm would refuse to install these packages w/o having built the projects first... which would put a monkeywrench in 1.
3. remove `lerna bootstrap` from `postinstall`

so, now, _NO_ `package-lock.json` files should ever be generated in `packages/*`; only the root one.

and hopefully, `lerna publish` will figure out the names of the dependencies (converting `file:` urls to the actual package name) when it's time to do that.  that'd be pretty broken if it didn't!

### DIFF
--- a/packages/fake-driver/package.json
+++ b/packages/fake-driver/package.json
@@ -21,15 +21,16 @@
   },
   "license": "Apache-2.0",
   "author": "https://github.com/appium",
-  "main": "./build/index.js",
+  "main": "fake-driver.js",
   "bin": {
-    "appium-fake-driver": "./build/index.js"
+    "appium-fake-driver": "fake-driver.js"
   },
   "directories": {
     "lib": "lib"
   },
   "files": [
     "index.js",
+    "fake-driver.js",
     "lib",
     "build/index.js",
     "build/lib",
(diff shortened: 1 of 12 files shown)


packages/fake-driver/package.json
```diff
@@ -21,24 +21,25 @@
   },
   "license": "Apache-2.0",
   "author": "https://github.com/appium",
-  "main": "./build/index.js",
+  "main": "fake-driver.js",
   "bin": {
```
jlipps, line 25: i'm surprised we have a bin dir for fake driver, i don't think we need it?

jlipps (approved):

packages/fake-driver/package.json
boneskull: idk.  it was there before I got here 😄

boneskull:
this is too deep of a hole

PR #16105, opened 2021-11-16
chore(support): Revert "chore(support): update dependency klaw to v4"

KazuCocoa:
Reverts appium/appium#15864

It seems the version expect NodeJS 14.14.0+
https://github.com/appium/appium/discussions/16103#discussioncomment-1649877


mykola-mokhnach (approved):

mykola-mokhnach:
I think for Appium 2 we could set node 14 as the minimum version

jlipps:
I agree, setting v14 to minimum for Appium 2.0 is a good idea

KazuCocoa:
oh, i thought we followed https://github.com/nodejs/Release#release-schedule 's eol (so, currently we support nodejs 12+) by considering enterprise use as well in slack before.

---

anyway, this PR should be merged to support v14 entirely as well.

PR #16108, opened 2021-11-16
docs(appium): --plugins is now --use-plugins

jlipps:


packages/appium/test/fixtures/args-for-schema.js
```diff
@@ -185,9 +185,9 @@ export default [
     },
   ],
   [
-    ['--plugins'],
+    ['--use-plugins'],
```
jlipps, line 188: @boneskull is this file used anymore? i was surprised that whatever it was for didn't break since it missed the change to the new arg.

KazuCocoa (approved):

packages/appium/test/fixtures/args-for-schema.js
boneskull: Hmm.. yes it looks like it can be removed.

PR #16118, opened 2021-11-18
disproportionate changes in reaction to lerna problems

boneskull:
This is a continuation of #16100.

There was some sort of disagreement between lerna, lockfiles and npm which was causing `lerna publish` to fail.  We are still not sure exactly _what_ the problem is here, but the maintainer of Lerna makes a reference to using _relative paths_ for package dependencies.  For example, if `appium` depends on `@appium/base-driver`, instead of a _version_ for the value in the `dependencies` property, it would contain `file:../base-driver`.  I am under the impression (!) that Lerna will modify `package.json` for each package on the way out to reflect the actual version number of the local package dependency.  Anyway, we'll find out what happens when we attempt to publish; I give this a 50/50 chance of needing to be reverted.

This "relative path" thing is what I wanted to do originally, since it's recommended.  However, I could not do so because the `main` (and `bin`) fields of each package pointed to files in `build/`, which are of course build artifacts.  Upon installation, `npm` will attempt to resolve `main` and `bin`, and if it cannot do so, it will fail.  Because the files hadn't been built yet, it would always fail.  To avoid the problem, I have created commonjs wrappers in each package (I think `doctor` had one already) which simply re-export whatever the old ESM `index.js` re-exported.  The old `index.js` files have generally moved to `lib/index.js`.

in the case of `appium`, I needed to add a `require.main` check, because the one in `main.js` will no longer work unless that file is referenced directly (which some tests do, but maybe they shouldn't. unclear if consumers depend on the location of `main.js`).

This also means that some relative paths within the projects themselves (generally in test files) have changed; requiring `..` from the `test/` dir no longer works; you'll need to require `../lib` instead.

Further, there were problems around ESLint.  ESLint wants its reusable configs to ask for `eslint` and whatever else it needs (plugins, other configs, etc.) as _peer_ dependencies.  Our config was not doing so, and I think it's mostly just dumb luck that it worked at all.  **This is a breaking change**, at least for users of npm pre-v7 which installs peer deps automatically.

The boilerplate gulp task needed to change as well, because it should no longer automatically transpile a root `index.js` by default (because they are now CJS files with references to build artifacts).  In fact, there should not be any ESM files in the root of each package; they should _all_ be in `lib` or `test`.  This is also a breaking change (because changing defaults _is_ generally a breaking change), if we care.

So...yeah.  This is the best I got.  I am hoping this positions us to avoid more weird problems around npm/lerna in the future.  If it works out, perhaps `appium-plugins` should do the same.

Note: I had previously abandoned this because `npm run install-fake-driver` was failing; maybe it works now?


KazuCocoa:
maybe I haven't caught up with well recent publish's issue, but I confirmed this change and dependencies structure. (with npm 6.14.11).

e.g.
current:
```
% ls -all node_modules/@appium
total 0
drwxr-xr-x     6 kazuaki  staff    192 Nov 17 22:20 .
drwxr-xr-x  1031 kazuaki  staff  32992 Nov 17 22:20 ..
drwxr-xr-x     8 kazuaki  staff    256 Nov 17 22:20 base-plugin
lrwxr-xr-x     1 kazuaki  staff     35 Nov 17 22:20 eslint-config-appium -> ../../packages/eslint-config-appium
drwxr-xr-x     7 kazuaki  staff    224 Nov 17 22:20 fake-plugin
drwxr-xr-x     9 kazuaki  staff    288 Nov 17 22:20 support
```

will be:
```
% ls node_modules/@appium
base-driver          doctor               fake-driver          gulp-plugins         test-support
base-plugin          eslint-config-appium fake-plugin          support
kazuaki@kazus-MB-pro appium % ls -all node_modules/@appium
total 0
drwxr-xr-x    11 kazuaki  staff    352 Nov 17 22:11 .
drwxr-xr-x  1130 kazuaki  staff  36160 Nov 17 22:12 ..
lrwxr-xr-x     1 kazuaki  staff     26 Nov 17 22:11 base-driver -> ../../packages/base-driver
drwxr-xr-x     9 kazuaki  staff    288 Nov 17 22:11 base-plugin
lrwxr-xr-x     1 kazuaki  staff     21 Nov 17 22:11 doctor -> ../../packages/doctor
lrwxr-xr-x     1 kazuaki  staff     35 Nov 17 22:11 eslint-config-appium -> ../../packages/eslint-config-appium
lrwxr-xr-x     1 kazuaki  staff     26 Nov 17 22:11 fake-driver -> ../../packages/fake-driver
drwxr-xr-x     8 kazuaki  staff    256 Nov 17 22:11 fake-plugin
lrwxr-xr-x     1 kazuaki  staff     27 Nov 17 22:11 gulp-plugins -> ../../packages/gulp-plugins
lrwxr-xr-x     1 kazuaki  staff     22 Nov 17 22:11 support -> ../../packages/support
lrwxr-xr-x     1 kazuaki  staff     27 Nov 17 22:11 test-support -> ../../packages/test-support
```

package.json
```diff
@@ -32,7 +32,7 @@
     "e2e-test": "lerna exec \"gulp e2e-test\"",
     "generate-docs": "lerna run --scope=appium generate-docs",
     "generate-schema-declarations": "node ./scripts/generate-schema-declarations.mjs",
-    "postinstall": "lerna bootstrap && lerna exec --parallel \"gulp prepublish\"",
+    "postinstall": "lerna exec --parallel \"gulp prepublish\"",
```
jlipps, line 35: omg were we running bootstrap on user installs as well as dev installs?

package.json
```diff
@@ -127,5 +130,15 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
```
jlipps, line 134: why do we now add prod deps for the monorepo project?

packages/eslint-config-appium/package.json
```diff
@@ -28,7 +28,7 @@
     "eslint:find:unused-rules": "eslint-find-rules -u -n",
     "lint": "eslint index.js"
   },
-  "dependencies": {
+  "peerDependencies": {
```
jlipps, line 31: this sucks. i wonder why we got away with this for so long.

jlipps (approved):
amazing work. i know it's a lot but i think this is probably the setup we want anyway

package.json
boneskull: there are no "user installs" of `appium-monorepo`, thankfully. 😉

package.json
boneskull: ¯\\_(ツ)_/¯ 

It's what lerna did.  tbh `dependencies` vs `devDependencies` don't matter too much in the monorepo root.

packages/eslint-config-appium/package.json
boneskull: dumb luck, I guess.

boneskull:
Unfortunately this is running into the same problems as #16100 did--some tests in `appium` which perform `appium driver install --source local /path/to/fake-driver` are busted.  It's unclear to me what exactly the problem is, but I am certain it involves these relative paths.  The best part: it only fails in CI.

boneskull:
```
Error: Command 'npm link --no-package-lock /home/runner/work/appium/appium/packages/fake-driver' exited with code 254
npm WARN tar ENOENT: no such file or directory, open '/home/runner/work/appium/appium/packages/fake-driver/node_modules/.staging/async-8004b3c1/internal/setImmediate.js'
npm WARN tar ENOENT: no such file or directory, open '/home/runner/work/appium/appium/packages/fake-driver/node_modules/.staging/async-8004b3c1/internal/withoutIndex.js'
(cont'd)
```

This is what fails in CI.  I will note that installing locally, these modules (`async`, `image-q`, etc.) do not exist in `packages/fake-driver/node_modules`, but rather the _root_ `node_modules`.  `packages/*/node_modules` _should not exist at all_.  Hmm.

boneskull:
@KazuCocoa What did you install (how?) to get those directory trees?

jlipps:
> packages/*/node_modules should not exist at all

Really? I think each package has its own node_modules dir (at least that's the way it's always been), just no dev deps are put there since dev deps are in the monorepo root.

jlipps:
I think we ran into something in another PR recently where it turned out github ci's management of global node modules is weird or broken. i think it might make npm link not work since npm link uses global node modules as an intermediary.

jlipps:
What is that `.staging/async-8004b3c1` nonsense? Why would we expect that to exist in node_modules?

boneskull:
Ooh, I can reproduce locally, sort of!

Running `npm install` in `packages/fake-driver` causes the problem (you shouldn't do this), but _only_ on a clean cache and with npm@6.

This sort of thing--running `npm link` from the `packages/fake-driver` dir in the monorepo--won't be encountered by end users.  I think we can probably just mitigate it by upgrading npm.

boneskull:
> What is that .staging/async-8004b3c1 nonsense? Why would we expect that to exist in node_modules?

These are npm internals that we should never need to worry about.  I'm guessing it's just a work directory of some sort.

boneskull:
@jlipps 

> I think we ran into something in another PR recently where it turned out github ci's management of global node modules is weird or broken. i think it might make npm link not work since npm link uses global node modules as an intermediary.

Yes, we ran into that.   Maybe upgrading `npm` will help?  Still, can never really be sure; we're at the mercy of the team managing GitHub Actions' VMs.

boneskull:
Looks like upgrading npm worked!

boneskull:
Here we go...

jlipps:
huzzah!

PR #16126, opened 2021-11-18
upgrade ESlint & drop gulp-eslint; closes #15952

### LINKED ISSUE #15952: Upgrade to ESLint v8.x
Appium should upgrade to the new ESLint v8 when possible, as this adds support for all manner of new syntax (including private fields).  Given some of our plugins have peer dependencies on v7.x, this won't be possible until those plugins upgrade, or we drop them.

The plugins preventing this upgrade include:

- `@babel/eslint-parser` - We may not even need this at all.  AFAICT It's only necessary if Appium is using experimental (e.g., Stage-3 or less) proposals, and I don't believe it is.
- `eslint-plugin-import` - This provides several safeguards which are useful (e.g., guards against missing modules)
- `eslint-find-rules` - I have no idea what this is or why it's here.  @jlipps?

I think we should give `eslint-plugin-import` some time (say, a month), then just go ahead with the upgrade without it.  It's useful, but not so useful it should hold us back from upgrading ESLint.

Note: this issue affects the `eslint-config-appium` package

boneskull:
This will probably fail CI because `eslint-plugin-promise` still does not have `eslint@8` support.  They seem to be working on it, though.

This PR _completely removes_ ESLint from `@appium-gulp-plugins`.  [gulp-eslint](https://npm.im/gulp-eslint) hasn't been published in 2 years, and is unlikely to offer `eslint@8` support.  Further, the peer dependencies it contains cause `npm@7` to automatically install `eslint@7` and use it; there is no way to use `eslint@8` via `gulp-eslint` unless we manually remove `eslint` from `node_modules/gulp-eslint/node_modules` after `npm install`.

`@babel/eslint-parser` still seems required due to syntax issues, however, it may not take much to drop it.


boneskull:
Yeah, we'd have to use `npm install --force` or some other such nonsense to make this work for now

jlipps (approved):

boneskull:
given #16459 I'm going to close this.  we can do this piecemeal (package at-a-time)

Issue #16073, opened 2021-11-11
better errors


boneskull:
Appium can throw exceptions.  Lots of exceptions.  Naturally, exceptions are presented to the end-user.  Appium should carefully consider how it throws exceptions, because those exceptions _may not be actionable._. 

Regardless, Appium should do its best to help or _help the user get help_ in both of these cases.  To that end, I'd like to propose:

1. Errors which a user can receive during normal usage of Appium (e.g., an incorrectly-typed CLI argument, or an unknown driver name):
    - _Must not_ display a stack trace!!
    -  _Should_ display a concise, human-readable message
    -  _Should_ display any context-specific information (e.g., config file line numbers, usage for a specific argument, etc.)
    - _Should_ provide a URL to documentation which explains the error in more detail
    - _May_ use color and/or emojis to better convey information
1. Errors which a _developer_ can receive (e.g., you called `someFunc()` incorrectly, or your driver implementation is hosed, uncaught exception, etc.):
    - _Must_ display a stack trace
    - _Should_ display a excruciatingly thorough message
1. Further, _all_ errors _must_ display a unique, easily-searchable error code, regardless of language

Some rationale:

- Actionable errors do not need stack traces; users do not care about them.  We should not show stack traces unless we think it will help whoever is seeing it.  Stack traces are an implementation detail, and an end-user shouldn't really care if Appium is written in JavaScript or Malbolge.
- External documentation on specific errors and how to resolve them will be helpful, especially to beginners
- Using unique codes to identify each error will help the user find more information via the forums, Google, etc.
- Emojis typically aren't my taste, but they make the CLI seem friendly and "modern"
- Ultimately, _we should see less issues in the tracker around common errors_

(To be clear, I'm talking about "fatal" errors, which will cause the Appium process to exit. Anything else falls into the realm of "logs", which is another can of worms.)

* * *

Implementation-wise, it's probably Good to have named error subclasses for each type of error.  These classes should be exported to any consumers/well-typed, etc.  In other words, the use-cases for `throw new Error()` are few, if any.  If we do end up using emoji, we should not display them where they don't work. Likewise, assume "no color" if not run in a tty--but have an override--etc.

I realize the "link to docs" part definitely needs its own issue, but I'd like to get there someday.

Thoughts?

mykola-mokhnach:
We do throw specific error where it is possible. There errors are defined by webdriver spec itself and could be found in https://github.com/appium/appium-base-driver/blob/master/lib/protocol/errors.js

[...]

PR #16135, opened 2021-11-19
print npm output upon error

boneskull:
I'm not sure if this is desirable or not from a user perspective.  In general, `npm` should not _expectedly_ fail for an end-user when using it.  But I can tell you that eating its output makes `npm` issues painful to debug!  And we've been having a few of those...

Also added some type checks to `npm.js`

This is not a `fix`, so I'll need to amend the commit message.

Related: #16073


jlipps (approved):
yeah, makes sense

PR #16157, opened 2021-11-23
fix(appium): make object dumps less weird

boneskull:
The `inspectObject()` function in `utils` was doing... stuff... to display both the default caps and non-default arguments.  Arrays, in particular, wound up looking like objects.  I found this objectionable; I hope there isn't a good reason it's this way.

`inspectObject()` was used for both dumping non-default args and default capabilities.

- Remove `utils.inspectObject()` and replace it with `inspect()`, which delegates to Node.js' built-in `util.inspect()` function (which does a bang-up job of dumping objects, including color!)
- Remove the `parser` argument (and return value!) from various functions where it was used.  e.g., `init()` in `main.js` was returning the `parser` instance, but now we don't need it, since it was only needed for `getNonDefaultServerArgs()`, which doesn't need it.  `parser` was being passed through several functions this way.
- Add a dumb test for `utils.inspect()`
- Add another dumb test for `getNonDefaultServerArgs()`
- Add `@types/express` since it was missing.
- More types in various associated places; fix other types
- Consistent naming in impl of `getNonDefaultServerArgs()`

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "@types/bluebird": "3.5.36",
     "@types/chai": "4.2.22",
     "@types/chai-as-promised": "7.1.4",
+    "@types/express": "4.17.13",
     "@types/fancy-log": "1.3.1",
     "@types/gulp": "4.0.9",
     "@types/json-schema": "7.0.9",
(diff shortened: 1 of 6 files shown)


boneskull:
I'm going to try my new stacked PR tool with this one

boneskull:
* 2.0
  * **PR #16157**  (this PR)
    * **PR #16158** 

boneskull:
neat!

jlipps (approved):

PR #16158, opened 2021-11-23
fix(appium): fix behavior of ReadonlyMap to be compatible with Map

boneskull:
`Map.prototype.delete` returns a `boolean`, so `ReadonlyMap.prototype.delete` should also return a `boolean`.

### DIFF
--- a/packages/appium/lib/utils.js
+++ b/packages/appium/lib/utils.js
@@ -235,8 +235,8 @@ class ReadonlyMap extends Map {
     return super.set(key, value);
   }
 
-  delete (key) {
-    throw new Error(`${key} cannot be deleted`);
+  delete () {
+    return false;
   }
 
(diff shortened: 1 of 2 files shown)


boneskull:
* 2.0
  * **PR #16157** 
    * **PR #16158**  (this PR)

jlipps (approved):

jlipps:
even though deleting something in a readonly map is against the point of a readonly map?

boneskull:
@jlipps It doesn't delete anything; it returns `false` which means "didn't delete anything" in `Map.prototype.delete`.  It just doesn't throw anymore.

boneskull:
ah neat, now this is retargeted to `2.0`.

jlipps:
ah ok cool.

PR #16168, opened 2021-11-28
refactor(base-driver): Make configureApp helper more flexible

mykola-mokhnach:
## Proposed changes

The configureApp helper was created to only support .apks/.ipa packages and is not very flexible about adding new formats.
This PR allows to change it, so a custom post processing could be added and the platform-specific stuff (like app bundle lookup inside .ipa/.zip) may be extracted to appropriate drivers. Backward compatibility is still preserved.

## Types of changes

### DIFF
--- a/packages/base-driver/lib/basedriver/helpers.js
+++ b/packages/base-driver/lib/basedriver/helpers.js
@@ -181,6 +181,8 @@ async function isAppIntegrityOk (currentPath, expectedIntegrity = {}) {
  * cached and a fresh copy of it is downloaded each time. If this function returns an object
  * containing `appPath` property then the integrity of it will be verified and stored into
  * the cache.
+ * @property {!string[]} supportedExtensions List of supported application extensions (
+ * including starting dots)
  */
 
 /**
--- a/packages/base-driver/lib/basedriver/helpers.js
+++ b/packages/base-driver/lib/basedriver/helpers.js
@@ -152,7 +152,7 @@ async function isAppIntegrityOk (currentPath, expectedIntegrity = {}) {
 
 /**
  * @typedef {Object} PostProcessOptions
- * @property {Object} cachedAppInfo The information about the previously cached app instance (if exists):
+ * @property {?Object} cachedAppInfo The information about the previously cached app instance (if exists):
  *    - packageHash: SHA1 hash of the package if it is a file and not a folder
  *    - lastModified: Optional Date instance, the value of file's `Last-Modified` header
  *    - immutable: Optional boolean value. Contains true if the file has an `immutable` mark
@@ -186,8 +186,8 @@ async function isAppIntegrityOk (currentPath, expectedIntegrity = {}) {
  * cached and a fresh copy of it is downloaded each time. If this function returns an object
  * containing `appPath` property then the integrity of it will be verified and stored into
  * the cache.
- * @property {!string[]} supportedExtensions List of supported application extensions (
- * including starting dots)
+ * @property {string[]} supportedExtensions List of supported application extensions (
+ * including starting dots). This property is mandatory and must not be empty.
  */
 
 /**
--- a/packages/base-driver/lib/basedriver/helpers.js
+++ b/packages/base-driver/lib/basedriver/helpers.js
@@ -18,13 +18,14 @@ const CACHED_APPS_MAX_AGE = 1000 * 60 * 60 * 24; // ms
 const APPLICATIONS_CACHE = new LRU({
   maxAge: CACHED_APPS_MAX_AGE, // expire after 24 hours
   updateAgeOnGet: true,
-  dispose: async (app, {fullPath}) => {
-    if (!await fs.exists(fullPath)) {
-      return;
-    }
-
-    logger.info(`The application '${app}' cached at '${fullPath}' has expired`);
-    await fs.rimraf(fullPath);
+  dispose: (app, {fullPath}) => {
+    logger.info(`The application '${app}' cached at '${fullPath}' has ` +
+      `expired after ${CACHED_APPS_MAX_AGE}ms`);
+    setTimeout(async () => {
+      if (fullPath && await fs.exists(fullPath)) {
+        await fs.rimraf(fullPath);
+      }
+    });
   },
   noDisposeOnSet: true,
 });
@@ -66,54 +67,56 @@ async function retrieveHeaders (link) {
   return {};
 }
 
-function getCachedApplicationPath (link, currentAppProps = {}) {
+function getCachedApplicationPath (link, currentAppProps = {}, cachedAppInfo = {}) {
   const refresh = () => {
     logger.debug(`A fresh copy of the application is going to be downloaded from ${link}`);
     return null;
   };
 
-  if (APPLICATIONS_CACHE.has(link)) {
-    const {
-      lastModified: currentModified,
-      immutable: currentImmutable,
-      // maxAge is in seconds
-      maxAge: currentMaxAge,
-    } = currentAppProps;
-    const {
-      // Date instance
-      lastModified,
-      // boolean
-      immutable,
-      // Unix time in milliseconds
-      timestamp,
-      fullPath,
-    } = APPLICATIONS_CACHE.get(link);
-    if (lastModified && currentModified) {
-      if (currentModified.getTime() <= lastModified.getTime()) {
-        logger.debug(`The application at ${link} has not been modified since ${lastModified}`);
-        return fullPath;
-      }
-      logger.debug(`The application at ${link} has been modified since ${lastModified}`);
-      return refresh();
-    }
-    if (immutable && currentImmutable) {
-      logger.debug(`The application at ${link} is immutable`);
+  if (!_.isPlainObject(cachedAppInfo) || !_.isPlainObject(currentAppProps)) {
+    return refresh();
+  }
+
+  const {
+    lastModified: currentModified,
+    immutable: currentImmutable,
+    // maxAge is in seconds
+    maxAge: currentMaxAge,
+  } = currentAppProps;
+  const {
+    // Date instance
+    lastModified,
+    // boolean
+    immutable,
+    // Unix time in milliseconds
+    timestamp,
+    fullPath,
+  } = cachedAppInfo;
+  if (lastModified && currentModified) {
+    if (currentModified.getTime() <= lastModified.getTime()) {
+      logger.debug(`The application at ${link} has not been modified since ${lastModified}`);
       return fullPath;
     }
-    if (currentMaxAge && timestamp) {
-      const msLeft = timestamp + currentMaxAge * 1000 - Date.now();
-      if (msLeft > 0) {
-        logger.debug(`The cached application '${path.basename(fullPath)}' will expire in ${msLeft / 1000}s`);
-        return fullPath;
-      }
-      logger.debug(`The cached application '${path.basename(fullPath)}' has expired`);
+    logger.debug(`The application at ${link} has been modified since ${lastModified}`);
+    return refresh();
+  }
+  if (immutable && currentImmutable) {
+    logger.debug(`The application at ${link} is immutable`);
+    return fullPath;
+  }
+  if (currentMaxAge && timestamp) {
+    const msLeft = timestamp + currentMaxAge * 1000 - Date.now();
+    if (msLeft > 0) {
+      logger.debug(`The cached application '${path.basename(fullPath)}' will expire in ${msLeft / 1000}s`);
+      return fullPath;
     }
+    logger.debug(`The cached application '${path.basename(fullPath)}' has expired`);
   }
   return refresh();
 }
 
 function verifyAppExtension (app, supportedAppExtensions) {
-  if (supportedAppExtensions.includes(path.extname(app))) {
+  if (supportedAppExtensions.map(_.toLower).includes(_.toLower(path.extname(app)))) {
     return app;
   }
   throw new Error(`New app path '${app}' did not have ` +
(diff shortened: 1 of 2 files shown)


KazuCocoa (approved):

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -18,13 +18,14 @@ const CACHED_APPS_MAX_AGE = 1000 * 60 * 60 * 24; // ms
...
-    if (!await fs.exists(fullPath)) {
-      return;
-    }
-
-    logger.info(`The application '${app}' cached at '${fullPath}' has expired`);
-    await fs.rimraf(fullPath);
+  dispose: (app, {fullPath}) => {
+    logger.info(`The application '${app}' cached at '${fullPath}' has ` +
+      `expired after ${CACHED_APPS_MAX_AGE}ms`);
+    setTimeout(async () => {
```
boneskull, line 24: what is this for?

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -18,13 +18,14 @@ const CACHED_APPS_MAX_AGE = 1000 * 60 * 60 * 24; // ms
...
+      if (fullPath && await fs.exists(fullPath)) {
+        await fs.rimraf(fullPath);
+      }
```
boneskull, line 27:
```suggestion
      if (fullPath) {
        await fs.rimraf(fullPath);
      }
```

checking for file existence is not necessary, since `rimraf` shouldn't care either way.

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -66,73 +67,157 @@ async function retrieveHeaders (link) {
...
-        return fullPath;
-      }
-      logger.debug(`The application at ${link} has been modified since ${lastModified}`);
-      return refresh();
-    }
-    if (immutable && currentImmutable) {
-      logger.debug(`The application at ${link} is immutable`);
+  if (!_.isPlainObject(cachedAppInfo) || !_.isPlainObject(currentAppProps)) {
+    return refresh();
+  }
```
boneskull, line 78: so the logic here is, "if someone passes an argument of an unexpected type, just return `null` so the cache misses?"  I feel like there might want to be a comment explaining the intent here.

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -66,73 +67,157 @@ async function retrieveHeaders (link) {
   return {};
 }
 
-function getCachedApplicationPath (link, currentAppProps = {}) {
+function getCachedApplicationPath (link, currentAppProps = {}, cachedAppInfo = {}) {
   const refresh = () => {
     logger.debug(`A fresh copy of the application is going to be downloaded from ${link}`);
     return null;
```
boneskull, line 73: (opinion) `null` has precious few use-cases in JS, and this isn't one of them.  given the usage, `return false` or just `return` would work just as well.

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -66,73 +67,157 @@ async function retrieveHeaders (link) {
...
+  return await fs.hash(filePath);
+}
+
+async function isAppIntegrityOk (currentPath, expectedIntegrity = {}) {
+  if (!await fs.exists(currentPath)) {
+    return false;
+  }
+
+  const isDir = (await fs.stat(currentPath)).isDirectory();
+  // Folder integrity check is simple:
```
boneskull, line 141: - are the files in `folderPath` provided by the user? if so, we can't trust them. is it possible `folderPath` will contain symlinks?  there should be a flag to `glob` which disallows symlink traversal.
- the globbing operation in `calculateFolderIntegrity` may be slow; we should keep an eye on it.  if it is, consider forking a child process, or just quickly tarballing the whole dir and taking the hash of that.

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -66,73 +67,157 @@ async function retrieveHeaders (link) {
...
+  // Ofc, validating the hash sum of each file (or at least of file path) would be much
+  // more precise, but we don't need to be very precise here and also don't want to
+  // overuse RAM and have a performance drop.
+  if (isDir && await calculateFolderIntegrity(currentPath) >= expectedIntegrity?.folder) {
+    return true;
+  }
+  if (!isDir && await calculateFileIntegrity(currentPath) === expectedIntegrity?.file) {
+    return true;
+  }
+  return false;
```
boneskull, line 154:
```suggestion
  if (isDir) {
    return await calculateFolderIntegrity(currentPath) >= expectedIntegrity?.folder;
  }
  return await calculateFileIntegrity(currentPath) === expectedIntegrity?.file;
```

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -66,73 +67,157 @@ async function retrieveHeaders (link) {
...
+    supportedAppExtensions = options.supportedExtensions;
+  }
+  if (_.isEmpty(supportedAppExtensions)) {
+    throw new Error(`One or more supported app extensions must be provided`);
   }
 
   let newApp = app;
   let shouldUnzipApp = false;
-  let archiveHash = null;
+  let packageHash = null;
```
boneskull, line 219: (opinion) don't use `null`

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -234,19 +322,24 @@ async function configureApp (app, supportedAppExtensions) {
       throw new Error(errorMessage);
     }
 
-    if (shouldUnzipApp) {
+    const isPackageAFile = !(await fs.stat(newApp)).isDirectory();
```
boneskull, line 325:
```suggestion
    const isPackageAFile = (await fs.stat(newApp)).isFile();
```

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -265,24 +358,43 @@ async function configureApp (app, supportedAppExtensions) {
...
-    if (app !== newApp && (archiveHash || _.values(remoteAppProps).some(Boolean))) {
-      if (APPLICATIONS_CACHE.has(app)) {
-        const {fullPath} = APPLICATIONS_CACHE.get(app);
-        // Clean up the obsolete entry first if needed
-        if (fullPath !== newApp && await fs.exists(fullPath)) {
-          await fs.rimraf(fullPath);
-        }
+    const storeAppInCache = async (appPathToCache) => {
+      const cachedFullPath = cachedAppInfo?.fullPath;
+      if (cachedFullPath && cachedFullPath !== appPathToCache && await fs.exists(cachedFullPath)) {
```
boneskull, line 363: don't need to check for existence

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -265,24 +358,43 @@ async function configureApp (app, supportedAppExtensions) {
...
+        await fs.rimraf(cachedFullPath);
+      }
+      const integrity = {};
+      if ((await fs.stat(appPathToCache)).isDirectory()) {
```
boneskull, line 367: some of these conditionals could be avoided by creating a `calculateIntegrity` function that will check if it's a file or folder and do the right thing.

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -66,73 +67,157 @@ async function retrieveHeaders (link) {
...
   if (!_.isString(app)) {
     // immediately shortcircuit if not given an app
     return;
   }
-  if (!_.isArray(supportedAppExtensions)) {
-    supportedAppExtensions = [supportedAppExtensions];
+
+  let supportedAppExtensions;
+  const {
+    onPostProcess,
```
boneskull, line 204: post-processing of what?  maybe this should be something like `onAppConfigured`

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -66,73 +67,157 @@ async function retrieveHeaders (link) {
...
+  }
+  if (!isDir && await calculateFileIntegrity(currentPath) === expectedIntegrity?.file) {
+    return true;
+  }
+  return false;
+}
+
+/**
+ * @typedef {Object} ConfigureAppOptions
+ * @property {?Function} onPostProcess Optional function, which should be applied
```
boneskull, line 159:
```suggestion
 * @property {(obj: Foo) => (Promise<Retval|undefined>|Retval|undefined)} onPostProcess - Optional function, which should be applied
```

where `Foo` is

```js
/**
 * @typedef {Object} Foo
 * @property {boolean} isUrl - Desc of isUrl
 * @property etc
 */
```

likewise a `@typedef` for whatever `Retval` is supposed to be.

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -66,73 +67,157 @@ async function retrieveHeaders (link) {
...
+ * - headers: Optional headers object. Only present if `isUrl` is true and if the server
+ * responds to HEAD requests.
+ * - appPath: A string containing full path to the preprocessed application package (either
+ * downloaded or a local one)
+ *
+ * The function is expected to either return a falsy value, which means the app must not be
+ * cached and a fresh copy of it is downloaded each time. If this function returns an object
+ * containing `appPath` property then the integrity of it will be verified and stored into
+ * the cache.
+ * @property {!string[]} supportedExtensions List of supported application extensions (
```
boneskull, line 184: I'm unsure what the leading `!` is.  does TS understand this?

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: to run the async snippet

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: ok

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: added a comment

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: from my perspective returning null is more strict from typing perspective. returning `false` there is not obvious there from the method name and I don't really like `undefined` as the original meaning of it is `var is not defined` while `null` means `it's defined, but empty`

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: by default glob does not check for symlinked dirs: 
```
Note that symlinked directories are not crawled as part of a **, though their contents may match against subsequent portions of the pattern. This prevents infinite loops and duplicates and the like
```

> the globbing operation in calculateFolderIntegrity may be slow; we should keep an eye on it. if it is, consider forking a child process, or just quickly tarballing the whole dir and taking the hash of that.

It's pretty fast, usually takes less that a second for thousands of recursive items.

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: removed

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: post processing of the app. Basically this stage is executed after we ensure the package is on the local FS or has been successfully downloaded

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: changed

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: Not sure about that. I was following JSDoc there

packages/base-driver/lib/basedriver/helpers.js
boneskull: no, I mean, why is `setTimeout` here?

packages/base-driver/lib/basedriver/helpers.js
boneskull: OK.  it'd make sense then to use a more descriptive name anyway

packages/base-driver/lib/basedriver/helpers.js
boneskull: OK; what it means is `non-null`.  This may make sense in jsdoc, but in typescript, this doesn't make sense (since "non-null" is essentially the default behavior).  imo we should remove `!` and all occurrences thereof

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: rimraf is safe to call, it should not create any exceptions

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: I cannot think about some name that would match better, maybe you'd have some proposals there. 
`onAppConfigured` does not describe the stage properly,  we'd need something more descriptive

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: I just wanted to emphasise this property is mandatory if an object is passed as an argument. I'm going to mention this in the docstring though

packages/base-driver/lib/basedriver/helpers.js
jlipps: i generally don't like the `setTimeout` method of calling async functions in a non-async context. I'd rather do something like:

```js
(async () => { /* async function logic */ })().catch(log.error.bind(log));
```

or to make it more readable, define the function and call it in 2 steps

```js
async function foo () {
 /* ... */
}
foo().catch(log.error)
```

Adding `.catch` is just a safeguard against unhandled rejections. By running async code in a non async context we are already agreeing that we don't care if something bad happens, but we might as well log it just in case something does happen, however unlikely.

mykola-mokhnach:
Are there any more comments to this PR? I'm going to merge it if there are none

jlipps (approved):
i'm still not a fan of the `setTimeout` strategy but everything looks ok

Issue #15963, opened 2021-10-14
Appium2.0: JWP responses within a W3C session


rillgen-saucelabs:
## The problem

When running tests using ChromeDriver some commands such as `/url` and `/title` return a JWP response (i.e. containing the `status` field) although the session is created as W3C.

## Environment

* Appium version (or git revision) that exhibits the issue: 2.0.0-beta.18
* Last Appium version that did not exhibit the issue (if applicable):
* Desktop OS/version used to run Appium: macOS
* Node.js version (unless using Appium.app|exe): v12.22.7
* Npm or Yarn package manager:
* Mobile platform/version under test: Android
* Real device or emulator/simulator: Simulator
* Appium CLI or Appium.app|exe: cli

## Details

The session creation and session close commands return as W3C, seems to be affecting only commands ran by the chromedriver. Same version of chromedriver works as expected with appium server v1.18.3.

## Link to Appium logs

https://gist.github.com/rillgen-saucelabs/d2beb99c8d9aec4cde3d4a77ec915222

## Code To Reproduce Issue [ Good To Have ]

Here is a dump of the requests and responses in HAR format: https://gist.github.com/rillgen-saucelabs/dbd5f347c8154d25cd23289983c5064b

rillgen-saucelabs:
I'm using the same version of chromedriver `83.0.4103.39` in appium server v1.18.3 and it returns W3C responses as expected. The binary I'm using for the v1.18.3 is the one embeded in the Appium app, while for the Appium 2.0 test I've installed the CLI and I'm passing the chromedriver location [downloaded from googleapis.com](https://chromedriver.storage.googleapis.com/chromedriver-83.0.4103.39/chromedriver_mac64.zip) using the `appium:chromedriverChromeMappingFile` and `appium:chromedriverExecutableDir` as default capabilities.

Here are the logs for the appium 1.18.3 server running the same test which returns all responses as W3C: https://gist.github.com/rillgen-saucelabs/a372329c385327e3ed94b4a4558b9d2f

I also tried passing the `{"appium:chromeOptions": {"w3c": true}}` capability (which I think it is default anyways) but it made no difference. Maybe the appium server v2.0.0 is not passing this to the chrome driver?

rillgen-saucelabs:
[...]

PR #16175, opened 2021-11-29
refactor(base-driver): Always assume the returned responses are in W3C format

mykola-mokhnach:
## Proposed changes

In Appium 2 we don't support JWP sessions anymore, so it makes sense to always assume that returned responses require W3C format. Also, this is supposed to address https://github.com/appium/appium/issues/15963, but only after we bump base driver version for upstream drivers

## Types of changes


KazuCocoa (approved):

jlipps (approved):
nice!

PR #16179, opened 2021-11-29
fix(appium): fix incorrect handling of delete session

jlipps:
…with regard to plugin driver assignment

also take the opportunity to constantize a few more strings

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -3,7 +3,8 @@ import log from './logger';
 import { getBuildInfo, updateBuildInfo, APPIUM_VER } from './config';
 import { findMatchingDriver } from './drivers';
 import { BaseDriver, errors, isSessionCommand,
-         CREATE_SESSION_COMMAND } from '@appium/base-driver';
+         CREATE_SESSION_COMMAND, DELETE_SESSION_COMMAND, GET_STATUS_COMMAND
+} from '@appium/base-driver';
 import AsyncLock from 'async-lock';
 import { parseCapsForInnerDriver, pullSettings } from './utils';
 import { util } from '@appium/support';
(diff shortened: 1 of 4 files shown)


packages/appium/lib/appium.js
```diff
@@ -443,9 +444,10 @@ class AppiumDriver extends BaseDriver {
     // The tricky part is that because we support command plugins, we need to wrap any of these
     // cases with plugin handling.
 
-    const isGetStatus = cmd === 'getStatus';
+    const isGetStatus = cmd === GET_STATUS_COMMAND;
+    const isDeleteSession = cmd === DELETE_SESSION_COMMAND;
     const isUmbrellaCmd = !isGetStatus && isAppiumDriverCommand(cmd);
```
boneskull, line 449: what's an "umbrella command"?

packages/base-driver/lib/index.js
```diff
@@ -21,15 +21,17 @@ const {
...
+  DELETE_SESSION_COMMAND, GET_STATUS_COMMAND,
 } = protocol;
 
 export {
   Protocol, routeConfiguringFunction, errors, isErrorType, PROTOCOLS,
   errorFromMJSONWPStatusCode, errorFromW3CJsonCode, determineProtocol,
   errorFromMJSONWPStatusCode as errorFromCode, ALL_COMMANDS, METHOD_MAP,
   routeToCommandName, NO_SESSION_ID_COMMANDS, isSessionCommand,
-  DEFAULT_BASE_PATH, normalizeBasePath, CREATE_SESSION_COMMAND, DELETE_SESSION_COMMAND
+  DEFAULT_BASE_PATH, normalizeBasePath, CREATE_SESSION_COMMAND,
```
boneskull, line 33: I'm curious about all these exports.  Why do we need all these exports?

boneskull (approved):

packages/appium/lib/appium.js
mykola-mokhnach: this means the command must be executed by the specific downstream platform driver rather than the AppiumDriver one

mykola-mokhnach (approved):

packages/base-driver/lib/index.js
jlipps: so that we can import them in Appium. Wanted to define magic strings in one place.

packages/appium/lib/appium.js
jlipps: it's the other way around, @mykola-mokhnach . the "umbrella" driver is the one defined here, AppiumDriver in appium.js. the "session" driver is the one that will handle most commands. There's a distinction because some commands (createSession, deleteSession) are handled by Appium initially. They don't get executed by the session driver immediately because Appium itself manages its own list of session drivers and this needs maintenance. Once a session exists, commands (like findElement or whatever) are executed directly by the session driver; this AppiumDriver class doesn't define logic for them at all.

PR #16180, opened 2021-11-29
chore(appium): make some appium plugins available via short name

jlipps:


packages/appium/lib/plugins.js
```diff
@@ -4,6 +4,8 @@
 // npm package. I.e., these are the officially recognized plugins.
 const KNOWN_PLUGINS = {
   images: '@appium/images-plugin',
+  'execute-driver': '@appium/execute-driver-plugin',
```
boneskull, line 7: worth mentioning that it's not ideal to need to update Appium when we add a plugin or a driver or rename something -- but at least it won't happen too often.  hopefully 😄

boneskull (approved):

packages/appium/lib/plugins.js
mykola-mokhnach: maybe we could just move this mapping to a separate (yaml) config file?

jlipps: i'm open to better ideas. the rate at which we create new officially supported plugins is so slow that this is not likely to be an annoyance for now. but by all means improve this if you can think of a better way!

PR #16181, opened 2021-11-30
fix(appium): correctly apply extension defaults

### LINKED ISSUE #16165: Undefined values should not be used as CLI arg values in drivers
## The problem

1. Install Appium2+xcuitest driver
2. Start appium server without arguments 
3. Start a driver session
4. Check logs

Actual result:
There are log lines about using undefined values as CLI arg values

Expected result:
CLI arg values must only be considered if they are defined

## Environment

* Appium version (or git revision) that exhibits the issue: 2.0-beta23
* Last Appium version that did not exhibit the issue (if applicable):
* Desktop OS/version used to run Appium:
* Node.js version (unless using Appium.app|exe):
* Npm or Yarn package manager:
* Mobile platform/version under test:
* Real device or emulator/simulator:
* Appium CLI or Appium.app|exe:

## Details

```
[HTTP] --> POST /wd/hub/session
[HTTP] {"capabilities":{"firstMatch":[{}],"alwaysMatch":{"appium:app":"/Users/elf/code/java-client/build/resources/test/apps/TestApp.app.zip","appium:automationName":"XCuiTest","appium:commandTimeouts":"240000","appium:deviceName":"iPhone 12","appium:platformVersion":"15.2","appium:wdaLaunchTimeout":240000,"platformName":"ios"}}}
[debug] [W3C] Calling AppiumDriver.createSession() with args: [null,null,{"firstMatch":[{}],"alwaysMatch":{"appium:app":"/Users/elf/code/java-client/build/resources/test/apps/TestApp.app.zip","appium:automationName":"XCuiTest","appium:commandTimeouts":"240000","appium:deviceName":"iPhone 12","appium:platformVersion":"15.2","appium:wdaLaunchTimeout":240000,"platformName":"ios"}}]
[debug] [BaseDriver] Event 'newSessionRequested' logged at 1637910174511 (08:02:54 GMT+0100 (Central European Standard Time))
[Appium] Attempting to find matching driver for automationName 'XCuiTest' and platformName 'ios'
[Appium] The 'xcuitest' driver was installed and matched caps.
[Appium] Will require it at /Users/elf/.appium/appium-xcuitest-driver/node_modules/appium-xcuitest-driver
[Appium] Appium v2.0.0-beta.23 creating new XCUITestDriver (v3.57.0) session
[Appium] Checking BaseDriver versions for Appium and XCUITestDriver
[Appium] Appium's BaseDriver version is 8.2.2
[Appium] Could not determine XCUITestDriver's BaseDriver version
[debug] [BaseDriver] Creating session with W3C capabilities: {
[debug] [BaseDriver]   "alwaysMatch": {
[debug] [BaseDriver]     "platformName": "ios",
[debug] [BaseDriver]     "appium:app": "/Users/elf/code/java-client/build/resources/test/apps/TestApp.app.zip",
[debug] [BaseDriver]     "appium:automationName": "XCuiTest",
[debug] [BaseDriver]     "appium:commandTimeouts": "240000",
[debug] [BaseDriver]     "appium:deviceName": "iPhone 12",
[debug] [BaseDriver]     "appium:platformVersion": "15.2",
[debug] [BaseDriver]     "appium:wdaLaunchTimeout": 240000
[debug] [BaseDriver]   },
[debug] [BaseDriver]   "firstMatch": [
[debug] [BaseDriver]     {}
[debug] [BaseDriver]   ]
[debug] [BaseDriver] }
[BaseDriver] Session created with session id: c0dc45cd-8460-4852-b356-0bd0aed00e93
>>>> [XCUITest] Using webkitDebugProxyPort value of 'undefined' from CLI
>>>> [XCUITest] Using wdaLocalPort value of 'undefined' from CLI
```

boneskull:
Extensions were not correctly receiving the default values as specified in their schemas.  For example, in `xcuitest` driver, the `wdaLocalPort` option has a default value of `8100`, but this value would never make it to the driver.

The root cause was that `getDefaultsForSchema()` returns a _flattened_ object (which is useful for displaying to users as part of `getNonDefaultServerArgs()`), but by the time defaults should get assigned to parsed arguments in `main.js`, the parsed arguments object has already been expanded to a deeply-nested, non-flattened object.  So `"driver.driverName.argName": "value"` would be set in the args, but  not `{"driver": {"driverName": {"argName": "value"}}}`.

This change fixes the issue by iterating over the result of `getDefaultsForSchema()` and using `_.set()` to build a deeply-nested object from the defaults, which is then applied to the parsed arguments.

Closes #16165.

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -7,6 +7,7 @@ import { BaseDriver, errors, isSessionCommand,
 import AsyncLock from 'async-lock';
 import { parseCapsForInnerDriver, pullSettings } from './utils';
 import { util } from '@appium/support';
+import { getDefaultsForExtension } from './schema';
 
 const desiredCapabilityConstraints = {
   automationName: {
@@ -118,14 +119,22 @@ class AppiumDriver extends BaseDriver {
    * Validate and assign CLI args for a driver or plugin
    *
    * If the extension has provided a schema, validation has already happened.
+   *
+   * Any arg which is equal to its default value will not be assigned to the extension.
    * @param {import('./ext-config-io').ExtensionType} extType 'driver' or 'plugin'
    * @param {string} extName the name of the extension
    * @param {Object} extInstance the driver or plugin instance
    */
   assignCliArgsToExtension (extType, extName, extInstance) {
-    const cliArgs = this.args[extType]?.[extName];
-    if (!_.isEmpty(cliArgs)) {
-      extInstance.cliArgs = cliArgs;
+    const allCliArgsForExt = this.args[extType]?.[extName];
+    if (!_.isEmpty(allCliArgsForExt)) {
+      const defaults = getDefaultsForExtension(extType, extName);
+      const cliArgs = _.isEmpty(defaults)
+        ? allCliArgsForExt
+        : _.omitBy(allCliArgsForExt, (value, key) => _.isEqual(defaults[key], value));
+      if (!_.isEmpty(cliArgs)) {
+        extInstance.cliArgs = cliArgs;
+      }
     }
   }
 
--- a/packages/appium/lib/main.js
+++ b/packages/appium/lib/main.js
@@ -18,7 +18,7 @@ import { DRIVER_TYPE, PLUGIN_TYPE } from './extension-config';
 import registerNode from './grid-register';
 import logger from './logger'; // logger needs to remain first of imports
 import { init as logsinkInit } from './logsink';
-import { getDefaultsFromSchema, validate } from './schema/schema';
+import { getDefaultsForSchema, validate } from './schema/schema';
 import { inspect } from './utils';
 
 /**
@@ -219,14 +219,9 @@ async function init (args) {
     parsedArgs = _.defaultsDeep(
       parsedArgs,
       configResult.config?.server,
-      getDefaultsFromSchema()
+      getDefaultsForSchema(false)
     );
   }
-
-  parsedArgs = _.defaultsDeep(
-    parsedArgs,
-    configResult.config ?? {},
-  );
 
   await logsinkInit(parsedArgs);
 
--- a/packages/appium/test/parser-specs.js
+++ b/packages/appium/test/parser-specs.js
@@ -20,11 +20,11 @@ describe('parser', function () {
     });
 
     it('should accept only server and driver subcommands', function () {
-      p.parse_args([]);
-      p.parse_args(['server']);
-      p.parse_args(['driver', 'list']);
-      (() => p.parse_args(['foo'])).should.throw();
-      (() => p.parse_args(['foo --bar'])).should.throw();
+      p.parseArgs([]);
+      p.parseArgs(['server']);
+      p.parseArgs(['driver', 'list']);
+      (() => p.parseArgs(['foo'])).should.throw();
+      (() => p.parseArgs(['foo --bar'])).should.throw();
     });
   });
 
@@ -35,12 +35,12 @@ describe('parser', function () {
       });
 
       it('should return an arg parser', function () {
-        should.exist(p.parse_args);
-        p.parse_args([]).should.have.property('port');
+        should.exist(p.parseArgs);
+        p.parseArgs([]).should.have.property('port');
       });
       it('should default to the server subcommand', function () {
-        p.parse_args([]).subcommand.should.eql('server');
-        p.parse_args([]).should.eql(p.parse_args(['server']));
+        p.parseArgs([]).subcommand.should.eql('server');
+        p.parseArgs([]).should.eql(p.parseArgs(['server']));
       });
       it('should keep the raw server flags array', function () {
         should.exist(p.rawArgs);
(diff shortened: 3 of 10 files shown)


packages/appium/test/schema/schema-specs.js
```diff
@@ -33,11 +33,6 @@ describe('schema', function () {
     sandbox = sinon.createSandbox();
 
     mocks = {
-      '../../lib/extension-config': {
-        DRIVER_TYPE: 'driver',
-        PLUGIN_TYPE: 'plugin',
-      },
```
boneskull, line 39: this mock was just not needed

packages/appium/test/parser-specs.js
```diff
@@ -20,11 +20,11 @@ describe('parser', function () {
     });
 
     it('should accept only server and driver subcommands', function () {
-      p.parse_args([]);
-      p.parse_args(['server']);
-      p.parse_args(['driver', 'list']);
-      (() => p.parse_args(['foo'])).should.throw();
-      (() => p.parse_args(['foo --bar'])).should.throw();
+      p.parseArgs([]);
```
boneskull, line 23: `parseArgs()` is the canonical name of this method.  `parse_args()` is deprecated (and should probably just be removed). vestige from calling into `argparse` directly

jlipps (approved):

mykola-mokhnach:
Seems like tests are quite unhappy

Also, I'm not quite sure about how default values are processed. The idea is that if an argument is not provided then it should not be passed to a driver at all, otherwise there might be an issue where a user passes a capability to a test session and then it still gets overridden by CLI arg, because it did accept a default value.

packages/appium/lib/main.js
```diff
@@ -216,18 +216,18 @@ async function init (args) {
   // 2. config file
   // 3. defaults from config file.
   if (parsedArgs.subcommand === SERVER_SUBCOMMAND) {
+    const flatDefaults = getDefaultsFromSchema();
+    const defaults = {};
+    for (const [key, value] of Object.entries(flatDefaults)) {
```
mykola-mokhnach, line 221: can we use Object.assign?

mykola-mokhnach (approved):

jlipps:
>  The idea is that if an argument is not provided then it should not be passed to a driver at all, otherwise there might be an issue where a user passes a capability to a test session and then it still gets overridden by CLI arg, because it did accept a default value.

Yeah this is an interesting point. It's nice to have driver args with defaults in a schema. But we want to make sure not to override capabilities if the args were the defaults. Not sure if there's any way for a driver to tell whether args were defaults? Is cliArgs populated with defaults or only if a user actually specifies something on the CLI?

packages/appium/lib/main.js
boneskull: I'm not sure how.  Some keys in this object are of the format `extType.extName.argName` which would get ugly.

boneskull:
> Yeah this is an interesting point. It's nice to have driver args with defaults in a schema. But we want to make sure not to override capabilities if the args were the defaults. Not sure if there's any way for a driver to tell whether args were defaults? Is cliArgs populated with defaults or only if a user actually specifies something on the CLI?

If we only assign to `cliArgs` if the user provided non-default values, then drivers would have to assign default values manually, which is not currently a thing that happens (and, imo, is undesirable behavior).

So I think what I could do is gather the default values and only overwrite the caps if the CLI args differ from those defaults.

I hope I'm making sense.

boneskull:
OK, I am confused.

>> The idea is that if an argument is not provided then it should not be passed to a driver at all, otherwise there might be an issue where a user passes a capability to a test session and then it still gets overridden by CLI arg, because it did accept a default value.

> Yeah this is an interesting point. It's nice to have driver args with defaults in a schema. But we want to make sure not to override capabilities if the args were the defaults. Not sure if there's any way for a driver to tell whether args were defaults? Is cliArgs populated with defaults or only if a user actually specifies something on the CLI?

It _sounds like_ if a driver wants to override caps with CLI args, this must be implemented in the driver itself.

Is this correct?  If so, I can provide an API to do so in this PR, and send another PR to `xcuitest` to use it.

boneskull:
(And: should the API be specific to the use-case, e.g., `overrideCapsWithCliArgs()` instead of building blocks for such a method?)

mykola-mokhnach:
> (And: should the API be specific to the use-case, e.g., overrideCapsWithCliArgs() instead of building blocks for such a method? I think it'd be useful to provide facilities for a driver to "query" its own defaults, fwiw)

I'd say we go with the current implementation and then I'll spend some time testing the actual implementation after the changes are published

boneskull:
@mykola-mokhnach We can't go with this implementation, because it's failing the build (for good reason).  

I'm not entirely sure how drivers are supposed to handle defaults right now (since `argsConstraints` did not have any notion of defaults).  But I'll make it so, at minimum, those default values do not wind up in `cliArgs`.

boneskull:
@jlipps @mykola-mokhnach This is now updated to reflect the requested behavior (and _should_ pass the build).

From the commit message:

> 
> Extensions were not correctly receiving the default values as specified in their schemas.  For example, in `xcuitest` driver, the `wdaLocalPort` option has a default value of `8100`, but this value would never make it to the driver.
> 
> The root cause was that `getDefaultsForSchema()` returns a _flattened_ object (which is useful for displaying to users as part of `getNonDefaultServerArgs()`), but by the time defaults should get assigned to parsed arguments in `main.js`, the parsed arguments object has already been expanded to a deeply-nested, non-flattened object.  So `"driver.driverName.argName": "value"` would be set in the args, but  not `{"driver": {"driverName": {"argName": "value"}}}`.  Furthermore, the properties of `cliArgs` are not prefixed with `<extType>.<extName>.`, but are just the arg name itself (which makes sense).
> 
> To address this, two main changes:
> 
> 1. `getDefaultsForSchema(flatten = true)` (renamed from `getDefaultsFromSchema()`) now accepts a `flatten` flag, which will return a flat object using `<extType>.<extName>.`-style prefixes. This is useful (and less computationally intensive) for comparing defaults across the whole of Appium and its extensions.  This is _not_ useful when merging defaults into the parsed arguments object, however, as that object will be deeply-nested.  If `flatten` is `false`, a deeply-nested object will be returned, which is consumed by the `main` module when post-processing parsed CLI arguments and config file values.
> 2. Added `getDefaultsForExtension(extType, extName)` which acts like `getDefaultsForSchema(flatten = true)` _but_ only returns defaults for a particular extension _and_ strips the leading prefixes.  This is used by `assignCliArgsToExtension` in the `AppiumDriver` class.
> 
> Other changes:
> 
> 1. `ArgSpec` now keeps a `rawDest` property which is like `dest` w/o the prefixes.
> 2. Make more types
> 3. Re-organize `driver-specs` unit test somewhat, and update to reflect changes.

boneskull:
To be clear: extensions will still need to apply default values manually--which I believe is a) the status quo and b) less-than-ideal--and I hope we can address that soon

packages/appium/lib/appium.js
```diff
@@ -118,14 +119,22 @@ class AppiumDriver extends BaseDriver {
...
   assignCliArgsToExtension (extType, extName, extInstance) {
-    const cliArgs = this.args[extType]?.[extName];
-    if (!_.isEmpty(cliArgs)) {
-      extInstance.cliArgs = cliArgs;
+    const allCliArgsForExt = this.args[extType]?.[extName];
+    if (!_.isEmpty(allCliArgsForExt)) {
+      const defaults = getDefaultsForExtension(extType, extName);
+      const cliArgs = _.isEmpty(defaults)
+        ? allCliArgsForExt
+        : _.omitBy(allCliArgsForExt, (value, key) => defaults[key] === value);
```
boneskull, line 134: This will probably fail if the default is, say, an empty array.  I'll take care of that.

boneskull: OK, this is addressed

jlipps:
OK great. Haven't reviewed the code yet but how you've described it is I think where we should land for this PR. I agree with you that it might make sense for there to be an appium-sponsored API for driver authors to deal with cli args and caps more ergonomically. In general, this is one problem with the fact that within appium and its driver ecosystem we have chosen to conflate everything into one object `this.opts`. It's convenient because drivers don't need to think about where parameters are coming from (CLI, appium itself, or users via caps). But it has all the problems you have now seen. Some other interface entirely (maybe a proxy object on basedriver that looks like `this.opts` but isn't actually a merged object, rather something that pulls from e.g. `this.cliArgs`, `this.driverOpts`, and `this.capabilities` with appropriate priority) might be a better idea, and maybe we want to tackle that for Appium 2.0. Not sure, and will be a big project since it may involve evaluating each use of `this.opts` throughout the official drivers.

boneskull:
* 2.0
  * **PR #16181**  (this PR)
    * **PR #16194** 

boneskull:
@jlipps I'll wait for you to review before merging.

jlipps (approved):

PR #16182, opened 2021-11-30
chore(fake-driver): Bump xmldom version

mykola-mokhnach:
## Proposed changes

The original xmldom NPM package has been abandoned and is not maintained anymore, but rather was moved under `@xmldom/xmldom`


jlipps (approved):
seems like fossa failed with some api key error

PR #16194, opened 2021-12-01
fix(appium): disallow unsupported schemas

boneskull:
Some theoretical schemas are not in any sort of shape Appium can use, so we now catch them as soon as possible (at the registration phase).

- adds `AppiumSchema.isUnsupportedSchemaType()` type guard
- adds `SchemaUnsupportedSchemaError` error


boneskull:
* 2.0
  * **PR #16181** 
    * **PR #16194**  (this PR)

jlipps (approved):

PR #16207, opened 2021-12-04
feat(appium): add --show-config

### LINKED ISSUE #15672: "--show-config" doesn't show config
`appium server --show-config` prints build info.  I do not know the "role" that needs build info, but I'm sure there's a reason that it is accessible.

That's fine, but _actually showing the configuration_ would be pretty darn helpful, especially since Appium configuration spans the server and multiple drivers and plugins.  For example, ESLint has a flag that will dump its entire configuration to the console for any given file.  I've found this useful if something unexpected is happening.  Anyway... mo config mo problems.

This will become _more_ useful if/when #15664 is merged, because Appium will then merge flags, config file(s) and defaults to determine what it should do.  

Example use case: if you're not sure why Appium is tossing gizzards, this should help consumers/contributors debug; they may use this to confirm whether Appium is _actually configured_ to toss gizzards.

@jlipps suggests perhaps `--show-config` becomes `--build-info` and `--show-config` would actually just dump the config to `STDOUT`.

boneskull:
This adds a new CLI flag: `--show-config`.  When called, this will dump the values of a) arguments to the `init()` function of `main.js` _or_ args passed thru CLI, b) any config pulled from a config file, and c) the defaults as defined in schemas.

Each section is separated, and `undefined` values are omitted.

Instead of a JSON blob, the output just uses `console.dir()`.

To make this work, some refactors were needed:

- Use of `store_true` "action" for `argparse` in are now `store_const` with `const: true` because `store_true` sets the value to `false` if the flag is not present.  This essentially assigns the default value of the flag (`false`) before it should; default values must come from the schemas.
- Along those lines, removed default values from the "extra" boolean args in `args.js`
- The CLI parser now ignores `undefined` values of arguments

Other:

- Created some types for extension classes and added the `server` prop to `AppiumDriver`.
- Enabled TS checks on `lib/config.js` (where `showConfig()` lives) and `test/config-specs.js`
- Removed unnecessary call to `validate()` in `preflightChecks()`; CLI arguments are validated individually (defined in `lib/schema/cli-args.js`) and the config file is validated separately (defined in `lib/config-file.js`).
- Add more types to `main.js`; rewrite `getServerUpdaters` in lodashese for fun (which I am happy to revert)
- Fix missing type aliases in `test/cli/cli-helpers.js`
- Misc types/refactors in `config-specs.js`
- Update parser tests to reflect that `undefined` values and empty objects are eliminated from the parsed arguments

Notes:

- The `showBuildInfo()` test in `config-specs.js` (which is ostensibly a unit test) forks `git` and also seems to require network connectivity, at least on my box. This stuff should be stubbed out or moved into an E2E test.

Resolves #15672

* * *

BREAKING CHANGE:

The old `--show-config` is now `--show-build-info`.

### DIFF
--- a/packages/appium/lib/cli/parser.js
+++ b/packages/appium/lib/cli/parser.js
@@ -146,7 +146,7 @@ class ArgParser {
     return _.reduce(
       args,
       (unpacked, value, key) => {
-        if (hasArgSpec(key)) {
+        if (!_.isUndefined(value) && hasArgSpec(key)) {
           const {dest} = /** @type {import('../schema/arg-spec').ArgSpec} */(getArgSpec(key));
           _.set(unpacked, dest, value);
         } else {
--- a/packages/appium/lib/config.js
+++ b/packages/appium/lib/config.js
@@ -144,7 +144,7 @@ function warnNodeDeprecations () {
   // }
 }
 
-async function showConfig () {
+async function showBuildInfo () {
   await updateBuildInfo(true);
   console.log(JSON.stringify(getBuildInfo())); // eslint-disable-line no-console
 }
@@ -219,7 +219,7 @@ async function validateTmpDir (tmpDir) {
 }
 
 export {
-  getBuildInfo, checkNodeOk, showConfig,
+  getBuildInfo, checkNodeOk, showBuildInfo,
   warnNodeDeprecations, validateTmpDir, getNonDefaultServerArgs,
   getGitRev, APPIUM_VER, updateBuildInfo
 };
--- a/packages/appium/lib/config.js
+++ b/packages/appium/lib/config.js
@@ -159,8 +159,8 @@ async function showBuildInfo () {
 
 /**
  * Returns k/v pairs of server arguments which are _not_ the defaults.
- * @param {import('../types/types').ParsedArgs} args
- * @returns {Partial<import('../types/types').ParsedArgs>}
+ * @param {ParsedArgs} args
+ * @returns {Partial<ParsedArgs>}
  */
 function getNonDefaultServerArgs (args) {
   // hopefully these function names are descriptive enough
@@ -227,14 +227,18 @@ function getNonDefaultServerArgs (args) {
 const compactConfig = _.partial(
   _.omitBy,
   _,
-  (value, key) => key === 'subcommand' || _.isUndefined(value) || (!_.isBoolean(value) && _.isEmpty(value))
+  (value, key) => key === 'subcommand' || _.isUndefined(value) || (_.isObject(value) && _.isEmpty(value))
 );
 
 /**
  * Shows a breakdown of the current config after CLI params, config file loaded & defaults applied.
- * @param {import('../types/types').ParsedArgs} preConfigParsedArgs - Parsed CLI args (or param to `init()`) before config & defaults applied
+ *
+ * The actual shape of `preConfigParsedArgs` and `defaults` does not matter for the purposes of this function,
+ * but it's intended to be called with values of type {@link ParsedArgs} and `DefaultValues<true>`, respectively.
+ *
+ * @param {object} preConfigParsedArgs - Parsed CLI args (or param to `init()`) before config & defaults applied
  * @param {import('./config-file').ReadConfigFileResult} configResult - Result of attempting to load a config file
- * @param {import('./schema/schema').DefaultValues<true>} defaults - Configuration defaults from schemas
+ * @param {object} defaults - Configuration defaults from schemas
  */
 function showConfig (preConfigParsedArgs, configResult, defaults) {
   console.log('Appium Configuration\n');
@@ -267,3 +271,7 @@ export {
   warnNodeDeprecations, validateTmpDir, getNonDefaultServerArgs,
   getGitRev, APPIUM_VER, updateBuildInfo, showConfig
 };
+
+/**
+ * @typedef {import('../types/types').ParsedArgs} ParsedArgs
+ */
--- a/packages/appium/lib/config.js
+++ b/packages/appium/lib/config.js
@@ -1,3 +1,6 @@
+// @ts-check
+
+/* eslint-disable no-console */
 import _ from 'lodash';
 import { mkdirp, system, fs } from '@appium/support';
 import axios from 'axios';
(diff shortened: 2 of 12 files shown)


boneskull:
Sample output:

image

boneskull:
e97f454 is just removing stuff that became unused via 9dfc3e3 (and fixing an odd type check or two).  nice to be able to delete code!

packages/appium/lib/config.js
```diff
@@ -209,6 +217,42 @@ function getNonDefaultServerArgs (args) {
...
+ * Compacts an object for {@link showConfig}:
+ * 1. Removes `subcommand` key/value
+ * 2. Removes `undefined` values
+ * 3. Removes empty objects (but not `false` values)
+ * Does not operate recursively.
+ */
+const compactConfig = _.partial(
+  _.omitBy,
+  _,
+  (value, key) => key === 'subcommand' || _.isUndefined(value) || (!_.isBoolean(value) && _.isEmpty(value))
```
mykola-mokhnach, line 230: isEmpty is also true for raw numbers

packages/appium/lib/main.js
```diff
@@ -12,13 +13,12 @@ import { AppiumDriver } from './appium';
 import { driverConfig, pluginConfig, USE_ALL_PLUGINS } from './cli/args';
 import { runExtensionCommand } from './cli/extension';
 import { default as getParser, SERVER_SUBCOMMAND } from './cli/parser';
-import { APPIUM_VER, checkNodeOk, getGitRev, getNonDefaultServerArgs, showConfig, validateTmpDir, warnNodeDeprecations } from './config';
+import { APPIUM_VER, checkNodeOk, getGitRev, getNonDefaultServerArgs, showBuildInfo, validateTmpDir, warnNodeDeprecations, showConfig } from './config';
```
mykola-mokhnach, line 16: this line seems too long to me

packages/appium/lib/config.js
mykola-mokhnach: cannot we simply reuse the lodash's compact method here?

packages/appium/test/config-specs.js
```diff
@@ -27,11 +60,7 @@ describe('Config', function () {
       // need to be able to write to process.version
       // but also to have access to process methods
       // so copy them over to a writable object
-      let tempProcess = {};
-      for (let [prop, value] of _.toPairs(process)) {
-        tempProcess[prop] = value;
-      }
-      process = tempProcess; // eslint-disable-line no-global-assign
+      process = {...process}; // eslint-disable-line no-global-assign
```
mykola-mokhnach, line 63: not sure I understand what happens here

packages/appium/test/config-specs.js
```diff
@@ -1,22 +1,55 @@
+// @ts-check
+
 // transpile:mocha
-import _ from 'lodash';
 import sinon from 'sinon';
 import getParser from '../lib/cli/parser';
-import { checkNodeOk, getBuildInfo, getNonDefaultServerArgs, showConfig, validateTmpDir, warnNodeDeprecations } from '../lib/config';
+import { checkNodeOk, getBuildInfo, getNonDefaultServerArgs, showBuildInfo, showConfig, validateTmpDir, warnNodeDeprecations } from '../lib/config';
```
mykola-mokhnach, line 6: same as above

packages/appium/lib/config.js
boneskull: `_.isEmpty()` is true for booleans.  I'll fix it for numbers.  this probably needs a test

packages/appium/lib/main.js
boneskull: imo if it didn't fail eslint it's not worth fiddling with.  if we need formatting, `prettier` does a good job

packages/appium/test/config-specs.js
boneskull: global process is assigned to a shallow copy of the original global process.  essentially just that the line I added did the same thing as the previous 5 or whatever

boneskull: fwiw: this is hard to test because `process.version` is read-only.  that suggests to me that we should not actually be interfacing with it directly, and instead provide, say, a function which wraps `process.version`.  we can then stub that function for test purposes.

packages/appium/lib/config.js
boneskull: ok, I've added a test here and fixed the issue with numeric values.  using `_.compact([value])` doesn't work, because `_.compact` does not recognize an empty object as something it should omit.

packages/appium/test/config-specs.js
boneskull: OTOH... I am not sure if we actually even need to be checking node versions?  I suppose if you miss the engine warnings and install anyway you _might_ see the warning out of Appium, _if_ the syntax is intelligible by the runtime.

packages/appium/lib/cli/parser.js
```diff
@@ -146,7 +146,7 @@ class ArgParser {
     return _.reduce(
       args,
       (unpacked, value, key) => {
-        if (hasArgSpec(key)) {
+        if (!_.isUndefined(value) && hasArgSpec(key)) {
```
mykola-mokhnach, line 149: is it possible to set `null` as an arg value? I assume the check should look like `!_.isNil(value)`

packages/appium/lib/config.js
```diff
@@ -209,6 +217,46 @@ function getNonDefaultServerArgs (args) {
...
+ * Compacts an object for {@link showConfig}:
+ * 1. Removes `subcommand` key/value
+ * 2. Removes `undefined` values
+ * 3. Removes empty objects (but not `false` values)
+ * Does not operate recursively.
+ */
+const compactConfig = _.partial(
+  _.omitBy,
+  _,
+  (value, key) => key === 'subcommand' || _.isUndefined(value) || (_.isObject(value) && _.isEmpty(value))
```
mykola-mokhnach, line 230: I assume `_.isObject(value) && _.isEmpty(value)` should filter out undefined values as well

packages/appium/lib/main.js
mykola-mokhnach: It is hard to read such code. We write code for people, not for linters

mykola-mokhnach (approved):

packages/appium/lib/cli/parser.js
boneskull: no.

packages/appium/lib/config.js
boneskull: no, because `undefined` is not an object

packages/appium/test/config-specs.js
boneskull: ping @jlipps  ^

PR #16303, opened 2022-01-05
fix!(gulp-plugins): index.js should probably be transpiled if it exists

jlipps:
not sure of the consequences for downstream packages of this change, hopefully it wouldn't hurt?

### DIFF
--- a/packages/gulp-plugins/lib/boilerplate.js
+++ b/packages/gulp-plugins/lib/boilerplate.js
@@ -10,7 +10,7 @@ if (process.env.TRAVIS || process.env.CI) {
 }
 
 const DEFAULT_OPTS = {
-  files: ['lib/**/*.js', 'test/**/*.js', '!gulpfile.js'],
+  files: ['*.js', 'lib/**/*.js', 'test/**/*.js', '!gulpfile.js'],
   transpile: true,
   transpileOut: 'build',
   babelOpts: {},


boneskull (changes requested):
This is on purpose; all of the `index.js` files in this repo are written in CJS. see 85cd55bc2c54e2091dec69ead1462c5f022e590b

boneskull:
(unsure if transpiling it would break anything, but it's certainly not needed)

mykola-mokhnach:
@boneskull Not transpiling index.js does break it completely: https://github.com/appium/appium-mac2-driver/issues/109

boneskull:
@mykola-mokhnach ...in that project, yes, but not in this one.

jlipps:
ok, so in all the other projects where it breaks, we need to override the `files` key. that means that the appium gulp plugins update was actually a breaking change

boneskull:
the change to `@appium/gulp-plugins` didn't break any packages that were using it.  even if I'd had bumped major, if you wanted to migrate from `appium-gulp-plugins` to the `@appium/gulp-plugins`, you'd have to fix the problem or hold back to an old version.  I figure nbd

jlipps:
ah right, i forgot it was a new package. ok. just something for us to remember for when we update the package to the new one.

PR #16327, opened 2022-01-11
fix(support): allow @u4/opencv4nodejs in addition to the old one

jlipps:
and promote it in doctor instead

### DIFF
--- a/packages/support/lib/image-util.js
+++ b/packages/support/lib/image-util.js
@@ -8,6 +8,10 @@ import log from './logger';
 import { requirePackage } from './node';
 
 
+const OCV_OLD = 'opencv4nodejs';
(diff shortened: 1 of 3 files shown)


packages/support/lib/image-util.js
```diff
@@ -128,15 +132,19 @@ async function initOpenCV () {
 
   log.debug(`Initializing opencv`);
   try {
-    cv = await requirePackage('opencv4nodejs');
-  } catch (err) {
-    log.warn(`Unable to load 'opencv4nodejs': ${err.message}`);
+    cv = await requirePackage(OCV_OLD);
```
mykola-mokhnach, line 135: I'd rather do it in a loop. Nested try's are hard to read

mykola-mokhnach (approved):

PR #16339, opened 2022-01-14
fix!(support): migrate to opencv-bindings and away from opencv4nodejs

jlipps:
This package is 7mb of wasmed code but at least it avoids global requires and building stuff from source.

So far I've just migrated the `getImageOccurrences` function, so the others would also have to be updated. There's lots of discrepancies in API surface, but so far by spelunking around the OpenCV docs I've been able to make most of it work (having to implement a bit of matrix logic myself in some places instead of using their methods).

### DIFF
--- a/packages/support/lib/image-util.js
+++ b/packages/support/lib/image-util.js
@@ -5,15 +5,10 @@ import { PNG } from 'pngjs';
 import B from 'bluebird';
 import { hasValue } from './util';
 import log from './logger';
-import { requirePackage } from './node';
 
 
-const OCV_OLD = 'opencv4nodejs';
-const OCV_NEW = '@u4/opencv4nodejs';
-
-
+let cv;
 const { MIME_JPEG, MIME_PNG, MIME_BMP } = Jimp;
-let cv = null;
 
 /**
  * @typedef {Object} Region
@@ -42,26 +37,25 @@ const SCANLINE_FILTER_METHOD = 4;
 const DEFAULT_MATCH_THRESHOLD = 0.5;
 const MATCH_NEIGHBOUR_THRESHOLD = 10;
 
-const AVAILABLE_DETECTORS = [
-  'AKAZE',
-  'AGAST',
-  'BRISK',
-  'FAST',
-  'GFTT',
-  'KAZE',
-  'MSER',
-  'SIFT',
-  'ORB',
-];
+const AVAILABLE_DETECTORS = {
+  AKAZE: 'AKAZE',
+  AGAST: 'AgastFeatureDetector',
+  BRISK: 'BRISK',
+  FAST: 'FastFeatureDetector',
+  GFTT: 'GFTTDetector',
+  KAZE: 'KAZE',
+  MSER: 'MSER',
+  ORB: 'ORB',
+};
 
-const AVAILABLE_MATCHING_FUNCTIONS = [
-  'FlannBased',
-  'BruteForce',
-  'BruteForceL1',
-  'BruteForceHamming',
-  'BruteForceHammingLut',
-  'BruteForceSL2',
-];
+const AVAILABLE_MATCHING_FUNCTIONS = {
+  FlannBased: 'FlannBased',
+  BruteForce: 'BruteForce',
+  BruteForceL1: 'BruteForce-L1',
+  BruteForceHamming: 'BruteForce-Hamming',
+  BruteForceHammingLut: 'BruteForce-HammingLUT',
+  BruteForceSL2: 'BruteForce-SL2',
+};
 
 const MATCHING_METHODS = [
   'TM_CCOEFF',
@@ -123,34 +117,28 @@ async function getJimpImage (data) {
 }
 
 /**
- * @throws {Error} If opencv4nodejs module is not installed or cannot be loaded
+ * Spins until the opencv-bindings module is fully loaded
  */
-async function initOpenCV () {
-  if (cv) {
-    return;
+async function initOpenCv () {
+  cv = require('opencv-bindings');
+  let waited = false;
+  while (!cv.getBuildInformation) {
+    log.info(`OpenCV module not fully loaded, waiting...`);
+    await B.delay(500);
+    waited = true;
   }
-
-  log.debug(`Initializing opencv`);
-  for (const ocvPackage of [OCV_OLD, OCV_NEW]) {
-    try {
-      log.debug(`Attempting to load '${ocvPackage}'`);
-      cv = await requirePackage(ocvPackage);
-      break;
-    } catch (err) {
-      log.warn(`Unable to load '${ocvPackage}': ${err.message}`);
-    }
-  }
-
-  if (!cv) {
-    throw new Error(`An opencv node module is required to use OpenCV features. ` +
-                    `Please install one first (e.g., 'npm i -g ${OCV_NEW}') and restart Appium. ` +
-                    'Read https://github.com/UrielCh/opencv4nodejs#how-to-install for more details on this topic.');
+  // opencv-bindings sets a global unhandledRejection handler of an abort, which we don't want, so
+  // undo it here. https://github.com/opencv/opencv/issues/21481
+  process.removeAllListeners('unhandledRejection');
+  if (waited) {
+    log.info(`OpenCV module successfully loaded`);
   }
 }
 
+
 /**
  * @typedef {Object} MatchComputationResult
- * @property {cv.DescriptorMatch} desciptor - OpenCV match descriptor
+ * @property {cv.DescriptorMatch} descriptor - OpenCV match descriptor
  * @property {Array<cv.KeyPoint>} keyPoints - The array of key points
  */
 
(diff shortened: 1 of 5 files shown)


jlipps:
maybe we should make `@appium/image-support` its own package? really only the images plugin uses it now, and some other stuff i'm working on. but that would keep the 7mb size out of generic installs.

mykola-mokhnach:
This looks great so far. If opecv-bindings support all matching methods (aka ORB, AKAZE, etc) that opencv4nodejs does then we rock.

mykola-mokhnach:
and yeah, I think image util methods are not used anywhere except of that particular plugin, so it would be fine to extract that. Then we could bump the major version of `@appium/support` and remove the stuff from there

StephanB75:
great news :) that makes my live with opencv much easier :)

jlipps:
OK, got `getImagesMatches` converted as well. that just leaves `getImagesSimilarity`. will work on that soon.

packages/support/lib/image-util.js
```diff
@@ -44,23 +38,23 @@ const MATCH_NEIGHBOUR_THRESHOLD = 10;
 
 const AVAILABLE_DETECTORS = [
   'AKAZE',
-  'AGAST',
+  //'AGAST',
```
mykola-mokhnach, line 41: What was the reason to exclude these detectors? In theory we could have them all in the list and let the code fail if the particular OpenCV deployment does not support it

packages/support/lib/image-util.js
```diff
@@ -293,23 +265,24 @@ async function getImagesMatches (img1Data, img2Data, options = {}) {
...
-  const detector = new cv[`${detectorName}Detector`]();
+  const detector = new cv[detectorName]();
   const [img1, img2] = await B.all([
-    cv.imdecodeAsync(img1Data),
-    cv.imdecodeAsync(img2Data)
-  ]);
-  const [result1, result2] = await B.all([
-    detectAndCompute(img1, detector),
-    detectAndCompute(img2, detector)
+    cvMatFromImage(img1Data),
```
mykola-mokhnach, line 270: oh these bindings also support async?

packages/support/lib/image-util.js
```diff
@@ -293,23 +265,24 @@ async function getImagesMatches (img1Data, img2Data, options = {}) {
...
   let matches = [];
-  try {
-    matches = await cv[`match${matchFunc}Async`](result1.descriptor, result2.descriptor);
-  } catch (e) {
-    throw new Error(`Cannot find any matches between the given images. Try another detection algorithm. ` +
-                    ` Original error: ${e}`);
+  matcher.match(result1.descriptor, result2.descriptor, matchesVec);
+  const totalCount = matchesVec.size();
+  if (totalCount < 1) {
+    throw new Error(`Could not find any matches`);
```
mykola-mokhnach, line 281: Would it be possible to have an error message here, that contains some hint for a user on how to solve this error?

packages/support/lib/image-util.js
```diff
@@ -513,26 +490,32 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
     method = DEFAULT_MATCHING_METHOD,
   } = options;
 
-  const [fullImg, partialImg] = await B.all([
-    cv.imdecodeAsync(fullImgData),
-    cv.imdecodeAsync(partialImgData)
-  ]);
+  const fullImg = await cvMatFromImage(fullImgData);
+  const partialImg = await cvMatFromImage(partialImgData);
```
mykola-mokhnach, line 494: images load could be done in parallel

packages/support/lib/image-util.js
jlipps: this code will only ever run against opencv-bindings. And I was unable to find these detectors in this package anywhere. so there's no reason to include them if we know they won't work

packages/support/lib/image-util.js
jlipps: no, this is a helper method i wrote

packages/support/lib/image-util.js
jlipps: yea

packages/support/lib/image-util.js
jlipps: true

jlipps:
OK @mykola-mokhnach i've finished converting all the opencv stuff. please do a full review, and then:

1. I can remove all this code and put it into its own package (want to wait until after review so that review is easier).
2. Please test with all your own test cases you can think of. The test suite itself is pretty minimal. I wrote out the visualizations and inspected them by hand to make sure they work, but didn't try other detectors or matching algorithms since I've never used any of them.

Also @KazuCocoa if you have any image related test cases that you could test on this branch, would be very helpful as well!

packages/support/lib/image-util.js
jlipps: actually I found all except SIFT. their feature detector naming scheme is not consistent.

KazuCocoa:
okay, will run image tests in https://github.com/appium/ruby_lib_core at least, and do some tests.
(hopefully this week/weekend...)

packages/support/lib/image-util.js
```diff
@@ -4,16 +4,10 @@ import { Buffer } from 'buffer';
 import { PNG } from 'pngjs';
 import B from 'bluebird';
 import { hasValue } from './util';
-import log from './logger';
-import { requirePackage } from './node';
-
-
-const OCV_OLD = 'opencv4nodejs';
-const OCV_NEW = '@u4/opencv4nodejs';
+import cv from 'opencv-bindings';
```
mykola-mokhnach, line 7: This module is 8.4 MB in size. Not sure we want to always fetch it no matter what

packages/support/lib/image-util.js
```diff
@@ -44,23 +38,22 @@ const MATCH_NEIGHBOUR_THRESHOLD = 10;
 
 const AVAILABLE_DETECTORS = [
   'AKAZE',
-  'AGAST',
+  'AgastFeatureDetector',
   'BRISK',
-  'FAST',
-  'GFTT',
+  'FastFeatureDetector',
```
mykola-mokhnach, line 43: Can we just rewrite alises in the client code to not break the backward compatibility (e.g. AGAST -> AgastFeatureDetector) instead of forcing clients to change their code?

packages/support/lib/image-util.js
```diff
@@ -44,23 +38,22 @@ const MATCH_NEIGHBOUR_THRESHOLD = 10;
...
 ];
 
 const AVAILABLE_MATCHING_FUNCTIONS = [
   'FlannBased',
   'BruteForce',
-  'BruteForceL1',
-  'BruteForceHamming',
-  'BruteForceHammingLut',
-  'BruteForceSL2',
+  'BruteForce-L1',
```
mykola-mokhnach, line 53: same as above

packages/support/lib/image-util.js
```diff
@@ -345,16 +318,25 @@ async function getImagesMatches (img1Data, img2Data, options = {}) {
...
       x: img1.cols + rect2.x,
       y: rect2.y,
       width: rect2.width,
       height: rect2.height
     });
-    result.visualization = await cv.imencodeAsync('.png', visualization);
+    result.visualization = await jimpImgFromCvMat(visualization).getBufferAsync(Jimp.MIME_PNG);
   }
+
+  img1.delete(); img2.delete(); detector.delete(); result1.keyPoints.delete(); result1.descriptor.delete();
```
mykola-mokhnach, line 338: I assume this stuff must be moved to `finally` section. Otherwise we might get quite a few resource leaks if this function does not reach to its end successfully

mykola-mokhnach: same below

mykola-mokhnach:
I'm not quite sure about https://github.com/wesuuu/opencvjs#example-usage. The comment says `// set a timeout, it takes some time for opencv.js to load since it's just one massive file`. Shall we somehow handle this delay in the client code? Otherwise I assume Appium would always take a while to start if we have these imports in the static context.

jlipps:
> Shall we somehow handle this delay in the client code? Otherwise I assume Appium would always take a while to start if we have these imports in the static context.

I did some tests to figure out what is happening. Looks like the module is loaded asynchronously in an IIFE which we can't await. To be safe I'll write a wrapper around it to make sure we wait until it's ready before we use it.

packages/support/lib/image-util.js
jlipps: That is why I will put it into its own package that will only ever be required when it will definitely be used (e.g. in the images plugin)

packages/support/lib/image-util.js
jlipps: there will potentially be lots of nested try/finallys then, as many of these objects are not defined and used in the same parts of functions. i'll see what i can do though.

packages/support/lib/image-util.js
jlipps: yeah ok good idea

jlipps:
Also, I realized that the non-CV image stuff (JIMP functionality) is used by the MJPEG server, maybe other things. So now I'm thinking to just move the CV-related functions into @appium/opencv, so anything that needs cv-specific stuff can rely on it.

jlipps:
@mykola-mokhnach any more thoughts? i'll probably merge this and then cut the new package from it in a different PR, before publishing either

mykola-mokhnach:
I'd prefer to not merge this code into appium-support. The main reason for that is that it statically imports opencv-bindings, which is going to slow down the server startup for all scenarios

mykola-mokhnach:
I'd say we simply close this PR in favour of https://github.com/appium/appium/pull/16360

jlipps:
> I'd prefer to not merge this code into appium-support. The main reason for that is that it statically imports opencv-bindings, which is going to slow down the server startup for all scenarios

Sure thing, we never would have made a publish with this code in appium/support anyhow.

PR #16360, opened 2022-01-21
refactor!(support): move opencv related functionality into its own package

jlipps:
here's the follow up to https://github.com/appium/appium/pull/16339, which could just be merged into that PR, which takes the opencv stuff out of support and makes it its own package.


packages/opencv/lib/index.js
```diff
@@ -0,0 +1,629 @@
...
+      multiple: results
+    };
+  } finally {
+    try {
+      fullImg.delete(); partialImg.delete(); matched.delete();
+    } catch (ign) {}
+  }
+}
+
+function jimpImgFromCvMat (mat) {
```
mykola-mokhnach, line 581: it would be nice to add docstrings to these new functions

mykola-mokhnach (approved):
👍

packages/support/lib/image-util.js
```diff
@@ -3,86 +3,10 @@ import Jimp from 'jimp';
 import { Buffer } from 'buffer';
 import { PNG } from 'pngjs';
 import B from 'bluebird';
-import { hasValue } from './util';
```
mykola-mokhnach, line 6: shall we also update images plugin?

packages/opencv/lib/index.js
jlipps: ok, added.

KazuCocoa:
👍

PR #16408, opened 2022-02-03
refactor: Allow any of createSession arguments to be W3C capabilities object

mykola-mokhnach:
## Proposed changes

Historically the first two arguments were reserved for JSONWP capabilities.
Appium 2 has dropped the support of these, so now we only accept capability objects in W3C format and thus allow any of the three arguments to represent the latter.

## Types of changes


KazuCocoa (approved):
seems reasonable

jlipps (commented):
yep, looks like a good idea

PR #16420, opened 2022-02-07
feat(base-driver): Add the size validation of the passed settings object

mykola-mokhnach:
## Proposed changes

In theory Appium allows to pass data chunks up to 1GB as POST request body. Also, every time these settings are added they get appended into the existing object and stored inside the driver. This change prevents settings to reach 20MB limit and potentially abuse driver memory usage.

## Types of changes

### DIFF
--- a/packages/support/lib/node.js
+++ b/packages/support/lib/node.js
@@ -1,7 +1,14 @@
 import { isWindows } from './system';
 import log from './logger';
+import _ from 'lodash';
 import { exec } from 'teen_process';
 import path from 'path';
+
+const ECMA_SIZES = Object.freeze({
+  STRING: 2,
+  BOOLEAN: 4,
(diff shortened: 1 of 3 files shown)


jlipps (approved):

packages/support/lib/node.js
```diff
@@ -1,8 +1,15 @@
 import { isWindows } from './system';
 import log from './logger';
+import _ from 'lodash';
 import { exec } from 'teen_process';
 import path from 'path';
 
+const ECMA_SIZES = Object.freeze({
```
KazuCocoa, line 7: 👀

KazuCocoa (approved):
I don't know of much about sizing stuff (just read https://github.com/miktam/sizeof and learned in this pr), but the code looks good

PR #16443, opened 2022-02-11
allow installation of extensions via package managers

### LINKED ISSUE #16442: allow installation of extensions via package managers
For a JS project which depends upon `appium`, installation of extensions should be allowable via idiomatic means (e.g., npm, yarn, etc) instead of requiring consumers to learn a new workflow with the `appium` CLI.   Both methods of extension management should be supported; if we don't, it will cause significant negative UX.

boneskull:
## Summary

This enables installation of extensions via `npm`; e.g., `npm install appium-xcuitest-driver -D`.  Because `appium` usually manages the `extensions.yaml` (the "manifest"), this is not possible when installing modules via other means.  To handle this situation, if we have an `appium`-dep-containing `package.json` in `APPIUM_HOME`, a hash is taken of the `package.json`.  This hash is checked at runtime; if it has changed, `appium` will scan `node_modules` (_not_ `package.json`; we only care about extensions that are _actually installed_, which is different from how we handle an `appium` dep) for extensions, and the hash will be updated.  Extensions are defined as "modules having a `package.json` which contains an `appium` field which is an object of a particular shape".  When these are found, they are appended to `extensions.yaml`.  In theory, this allows a user to use `appium` _or_ `npm` (at their convenience) to manage extensions.

See the "when appium is a dependency" suite in `test/cli/cli-e2e-specs.js` for more context.

## BREAKING CHANGES

(potentially, anyway)

1. Installation of extensions via `appium` now treats `$APPIUM_HOME` _mostly_ like a project containing a `node_modules` subfolder, with extensions under that.  Previously, each extension would have its own folder, with a `node_modules` and empty `package.json`, and finally the extension.  The intent was to inhibit deduping and conflicts with any `package.json` files in ancestors of `$APPIUM_HOME`.  I showed this was unnecessary because the same thing can be accomplished via the flag `--global-style` for `npm install`.  Ultimately, this simplifies the directory structure of `$APPIUM_HOME`.
2. `$APPIUM_HOME/extensions.yaml` now lives at `$APPIUM_HOME/node_modules/.cache/appium/extensions.yaml`.  The `node_modules/.cache/<pkg-name>` convention is fairly common for packages having temporary files which need to persist while a module is installed (no more, no less).
3. If the _current working directory_ contains a `package.json` which has `appium` as any sort of (non-`file:`) dependency _or_ `$CWD/node_modules/appium` is a resolvable module, `APPIUM_HOME` defaults to `cwd`.
4. "Install" and "link" lockfiles (see `lib/cli/npm`) have also moved from `$APPIUM_HOME` to `$APPIUM_HOME/node_modules/.cache/appium`.
5. The `runExtensionCommand()` function in the `cli/extension` module no longer accepts a `type` parameter, which can be derived from the `configObject` parameter anyway (redundant)
6. Loading of extensions (i.e. reading `extensions.yaml`) has been moved into its own function, `loadExtensions()` in `extension/index.js`.
7. `getActivePlugins` and `getActiveDrivers` have been moved here from `lib/main`.  Changed the signatures of these, since they don't actually need the entire `ParsedArgs` object.
8. `ext-config-io` module, which handles reading/writing `extensions.yaml` has been moved to `extension/manifest.js`.
9. The `installPath` property in `extensions.yaml` has been removed and is unneeded due to 1.
10. `read()` has been removed from `extension-config` and its subclasses.
11. Moved `ReadonlyMap` from `utils` to `schema/schema` module.  Probably so I didn't have to stub it in a test.
12. Moved `rootDir` from `utils` to `config`.
13. Removed `drivers` module; moved `findMatchingDriver` to `DriverConfig#findMatchingDriver`.  Removed `AppiumDriver#_findMatchingDriver`; calls `DriverConfig#findMatchingDriver` directly instead.
14. Removed `plugins` module; its contents now live in `constants`.
15. Dynamic determination of `APPIUM_HOME` now happens in the `env.resolveAppiumHome()` method of `@appium/support` (introduced in a previous commit).  It should not happen anywhere else, because it will be wrong.
16. Because the `lib` dir was becoming rather unwieldy, I moved modules associated with extension loading and config into an `extension` subdir
17. `getParser()` in `cli/parser` is no longer async; the async stuff it was doing moved into `loadExtensions()`.  `getParser()` can be used without calling `loadExtensions()` first (no extensions will be loaded)

## Notes

- Added a `constants` module, which provides a centralized place for various constants.  This makes finding constants less cognitively demanding. 🙂  You may not like this; if you don't, the constants can be moved to more appropriate places.
- Added `extension/package-changed` module which is used to determine if a `package.json` has changed
- Many types are added/changed throughout.  See especially `extension/manifest` for many type definitions involving the shape of `extensions.yaml` and the `appium` prop of extensions' `package.json`.
- Reorganized some tests based on structure of `lib`.
- Split `extension-config-specs.js` into `driver-config-specs.js` and `plugin-config-specs.js`.
- Many refactors in now-fully-typed `cli-e2e-specs`; introduced some more helpers in `test/cli/cli-helpers.js` and used these.  Makes calling out to the `appium` executable from E2E tests a little easier.
- Introduced `test/extension/mocks` for DRY mocks, since many of the unit tests here now use the same ones.
- Consolidated fixtures into `test/fixtures`. Fixtures can always be found via the `resolveFixture()` method of `test/helpers.js`.

## TODO (before this lands)

- Instead of using `$CWD` for `APPIUM_HOME` if `appium` is a dep, we probably want to use `find-pkg-up` or something.  This should be okay, since if we find a `package.json` and `appium` _isn't_ a dep, we will just be using the default `APPIUM_HOME`.
- We _may_ also need to search `package.json` for known extensions, since it's possible to use them standalone.  However, I'm not sure if this use case needs `APPIUM_HOME` at all?

* * *

First, sorry for another huge PR.  

I want to answer those TODO items before moving this out of draft.   If there's something anyone wishes I pull out of this PR, please lmk.


packages/appium/lib/cli/npm.js
```diff
@@ -169,13 +187,13 @@ export default class NPM {
     // everything got installed ok. Remember, pkgName might end up with a / in it due to an npm
     // org, so if so, that will get correctly exploded into multiple directories, by path.resolve here
     // (even on Windows!)
-    const pkgJson = path.resolve(pkgDir, 'node_modules', pkgName, 'package.json');
+    const pkgJsonPath = resolveFrom(this.appiumHome, `${pkgName}/package.json`);
```
mykola-mokhnach, line 190: based on the comment above is it ok to explicitly use slash as a path separator? would that still work on Windows?

mykola-mokhnach: same below

packages/appium/lib/cli/parser.js
```diff
@@ -2,21 +2,16 @@
 
 import { fs } from '@appium/support';
 import { ArgumentParser } from 'argparse';
-import B from 'bluebird';
 import _ from 'lodash';
 import path from 'path';
-import { DRIVER_TYPE, PLUGIN_TYPE } from '../extension-config';
+import { DRIVER_TYPE, PLUGIN_TYPE, SERVER_SUBCOMMAND} from '../constants';
```
mykola-mokhnach, line 7: missing space after SERVER_SUBCOMMAND ?

packages/appium/lib/cli/parser.js
```diff
@@ -258,19 +253,16 @@ class ArgParser {
...
+ * Creates a {@link ArgParser} instance; finalizes the config schema.
  *
  * @constructs ArgParser
  * @param {boolean} [debug] - If `true`, throw instead of exit upon parsing error
- * @returns {Promise<ArgParser>}
+ * @returns {ArgParser}
  */
-async function getParser (debug = false) {
-  await B.all([driverConfig.read(), pluginConfig.read()]);
+function getParser (debug) {
```
mykola-mokhnach, line 262: why the default value has been removed?

packages/appium/lib/extension/driver-config.js
```diff
@@ -0,0 +1,215 @@
...
+import { DRIVER_TYPE } from '../constants';
+import log from '../logger';
+import { ExtensionConfig } from './extension-config';
+
+/**
+ * @extends {ExtensionConfig<DriverType>}
+ */
+export class DriverConfig extends ExtensionConfig {
+  /**
+   * A mapping of `APPIUM_HOME` values to {@link DriverConfig} instances.
```
mykola-mokhnach, line 13: so this assumes we could have multiple appium homes with different configs on the same machine and switch between them?

packages/appium/lib/extension/driver-config.js
```diff
@@ -0,0 +1,215 @@
...
+      this.validate(extData);
+    }
+  }
+
+  /**
+   * Checks extensions for problems
+   * @param {ExtRecord<DriverType>} exts
+   */
+  validate (exts) {
+    this.knownAutomationNames.clear();
```
mykola-mokhnach, line 41: is that a side effect?

packages/appium/lib/extension/driver-config.js
```diff
@@ -0,0 +1,215 @@
...
+  static create (io, {extData, logFn} = {}) {
+    const instance = new DriverConfig(io, {logFn, extData});
+    DriverConfig._instances[io.appiumHome] = instance;
+    return instance;
+  }
+
+  /**
+   * Gets an existing instance of {@link DriverConfig} based value of `io.appiumHome`
+   * @param {import('./manifest').Manifest} io - IO object
+   * @returns {DriverConfig}
```
mykola-mokhnach, line 62: could the result be undefined?

packages/appium/lib/extension/driver-config.js
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /**
+   * Given capabilities, find a matching driver within the config. Load its class and return it along with version and driver name.
+   * @param { { automationName: string, platformName: string } } caps
+   * @returns { { driver: import('./manifest').DriverClass, version: string, driverName: string } }
+   */
+  findMatchingDriver ({automationName, platformName}) {
+    if (!_.isString(platformName)) {
+      throw new Error('You must include a platformName capability');
```
mykola-mokhnach, line 129: not sure if we need to throw a generic error here or more specific one from the list of available W3C error codes

packages/appium/lib/extension/extension-config.js
```diff
@@ -0,0 +1,458 @@
...
+      ];
+    }
+
+    const problemSummaries = [];
+    for (const [extName, problems] of _.toPairs(foundProblems)) {
+      if (_.isEmpty(problems)) {
+        continue;
+      }
+      // remove this extension from the list since it's not valid
+      delete exts[extName];
```
mykola-mokhnach, line 97: is it ok to implicitly mutate the argument?

packages/appium/lib/extension/extension-config.js
```diff
@@ -0,0 +1,458 @@
...
+
+  /**
+   * @param {ExtData<ExtType>} extData
+   * @param {ExtName<ExtType>} extName
+   * @returns {Problem[]}
+   */
+  getSchemaProblems (extData, extName) {
+    const problems = [];
+    const {schema: argSchemaPath} = extData;
+    if (ExtensionConfig.extDataHasSchema(extData)) {
```
mykola-mokhnach, line 131: I prefer to keep such methods on module level and only make them static if we need this data in subclasses

packages/appium/lib/extension/extension-config.js
```diff
@@ -0,0 +1,458 @@
...
+  configKey;
+
+  /** @type {ExtRecord<ExtType>} */
+  installedExtensions;
+
+  /** @type {ExtensionLogFn} */
+  log;
+
+  /** @type {import('./manifest').Manifest} */
+  io;
```
mykola-mokhnach, line 46: is there any reason to call this variable `io`? I cannot really understand what it contains

packages/appium/lib/extension/extension-config.js
```diff
@@ -0,0 +1,458 @@
...
+  getInstallPath (extName) {
+    return path.join(this.appiumHome, 'node_modules', this.installedExtensions[extName].pkgName);
+  }
+
+  /**
+   * Loads extension and returns its main class (constructor)
+   * @param {ExtName<ExtType>} extName
+   * @returns {ExtClass<ExtType>}
+   */
+  require (extName) {
```
mykola-mokhnach, line 317: mayt this method shadow the original `require` global?

packages/appium/lib/extension/extension-config.js
```diff
@@ -0,0 +1,458 @@
...
+    log.debug(`Requiring ${this.extensionType} at ${reqPath}`);
+    return require(reqPath)[mainClass];
+  }
+
+  /**
+   * @param {string} extName
+   * @returns {boolean}
+   */
+  isInstalled (extName) {
+    return _.includes(Object.keys(this.installedExtensions), extName);
```
mykola-mokhnach, line 335: you could also use _.keys

packages/appium/lib/extension/extension-config.js
```diff
@@ -0,0 +1,458 @@
...
+      : moduleObject;
+    registerSchema(extType, extName, schema);
+    return schema;
+  }
+
+  /**
+   * Returns `true` if a specific {@link ExtData} object has a `schema` prop.
+   * The {@link ExtData} object becomes a {@link ExtDataWithSchema} object.
+   * @template {ExtensionType} ExtType
+   * @param {ExtData<ExtType>} extData
```
mykola-mokhnach, line 374: is extData optional?

packages/appium/lib/extension/index.js
```diff
@@ -0,0 +1,87 @@
...
+    try {
+      log.info(`Attempting to load plugin ${pluginName}...`);
+      const PluginClass = pluginConfig.require(pluginName);
+
+      PluginClass.pluginName = pluginName; // store the plugin name on the class so it can be used later
+      return PluginClass;
+    } catch (err) {
+      log.error(`Could not load plugin '${pluginName}', so it will not be available. Error ` +
+                   `in loading the plugin was: ${err.message}`);
+      log.debug(err.stack);
```
mykola-mokhnach, line 53: 👍

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,564 @@
...
+
+/**
+ * Module containing {@link Manifest} which handles reading & writing of extension config files.
+ */
+
+import {env, fs, mkdirp} from '@appium/support';
+import _ from 'lodash';
+import path from 'path';
+import YAML from 'yaml';
+import {DRIVER_TYPE, PLUGIN_TYPE} from '../constants';
```
mykola-mokhnach, line 11: we probably need to follow one style here - either we keep spaced after and before object curly brackets or we don't

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,564 @@
...
+
+/**
+ * Current configuration schema revision!
+ */
+const CONFIG_SCHEMA_REV = 2;
+
+/**
+ * @type {`${typeof DRIVER_TYPE}s`}
+ */
+const CONFIG_DATA_DRIVER_KEY = `${DRIVER_TYPE}s`;
```
mykola-mokhnach, line 41: what is the magic `s` suffix?

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,564 @@
...
+    )) {
+      if (filepath !== this._appiumHome && stats.isDirectory()) {
+        try {
+          const pkg = await env.readPackageInDir(filepath);
+          if (pkg && isExtension(pkg)) {
+            this.addExtensionFromPackage(
+              pkg,
+              path.join(filepath, 'package.json'),
+            );
+            didChange = true;
```
mykola-mokhnach, line 182: can we return right here?

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,564 @@
...
+import log from '../logger';
+import {INSTALL_TYPE_NPM} from './extension-config';
+import {packageDidChange} from './package-changed';
+
+/**
+ * Default depth to search in directory tree for whatever it is we're looking for.
+ *
+ * It's 4 because smaller numbers didn't work.
+ */
+const DEFAULT_SEARCH_DEPTH = 4;
```
mykola-mokhnach, line 21: why not 5 or 6?

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,564 @@
...
+   *
+   * If the manifest does not exist on disk, an {@link INITIAL_MANIFEST_DATA "empty"} manifest file will be created.
+   *
+   * If `APPIUM_HOME` contains a `package.json` with an `appium` dependency, then a hash of the `package.json` will be taken. If this hash differs from the last hash, the contents of `APPIUM_HOME/node_modules` will be scanned for extensions that may have been installed outside of the `appium` CLI.  Any found extensions will be added to the manifest file, and if so, the manifest file will be written to disk.
+   *
+   * Only one read operation should happen at a time.  This is controlled via the {@link Manifest._reading} property.
+   * @returns {Promise<ManifestData>} The data
+   */
+  async read () {
+    if (this._reading) {
```
mykola-mokhnach, line 288: I prefer to use bluebird promises. they have convenience methods like http://bluebirdjs.com/docs/api/ispending.html, which helps to avoid nullifying the value

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,564 @@
...
+      /** @type {ManifestData} */
+      let data;
+      let isNewFile = false;
+      await this._setManifestPath();
+      try {
+        log.debug(`Reading ${this._manifestPath}...`);
+        const yaml = await fs.readFile(this._manifestPath, 'utf8');
+        data = YAML.parse(yaml);
+      } catch (err) {
+        if (err.code === 'ENOENT') {
```
mykola-mokhnach, line 303: I would rather prefer to explicitly check if the file exists first. Exception-driven flow is not very obvious

packages/appium/lib/extension/package-changed.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * If they have, we need to sync them with the `extensions.yaml` manifest.
+ *
+ * _Warning: this makes a blocking call to `writeFileSync`._
+ * @param {string} appiumHome
+ * @returns {Promise<boolean>} `true` if `package.json` `appiumHome` changed
+ */
+export async function packageDidChange (appiumHome) {
+  if (!appiumHome || !_.isString(appiumHome)) {
+    throw new TypeError('appiumHome must be a string');
```
mykola-mokhnach, line 21: such error might be confusing for a user if they see it.

packages/appium/lib/extension/package-changed.js
```diff
@@ -0,0 +1,72 @@
...
+  let writeHash;
+  /** @type {string} */
+  let hash;
+  /** @type {string|undefined} */
+  let oldHash;
+
+  // first mkdirp the target dir.
+  const hashFilenameDir = path.dirname(hashFilename);
+  try {
+    log.debug(`Creating hash file directory: ${hashFilenameDir}`);
```
mykola-mokhnach, line 39: I usually try to reduce the scope of try/catch blocks as much as possible to not catch errors which we don't want to catch.

packages/appium/lib/extension/package-changed.js
```diff
@@ -0,0 +1,72 @@
...
+      `Appium could not create the directory for hash file: ${hashFilenameDir}. Original error: ${err.message}`,
+    );
+  }
+
+  try {
+    const result = await isPackageChanged({
+      cwd: appiumHome,
+      hashFilename: HASHFILE_RELATIVE_PATH,
+    });
+    isChanged = result.isChanged;
```
mykola-mokhnach, line 52: this could be simplified to
```
({isChanged, writeHash, oldHash, hash} = result);
```

packages/appium/lib/extension/package-changed.js
```diff
@@ -0,0 +1,72 @@
...
+    writeHash = result.writeHash;
+    oldHash = result.oldHash;
+    hash = result.hash;
+  } catch {
+    return true;
```
mykola-mokhnach, line 57: should we return true if an exception has happened?

packages/appium/lib/extension/package-changed.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  if (isChanged) {
+    try {
+      writeHash();
```
mykola-mokhnach, line 62: is this call synchronous?

packages/appium/lib/extension/plugin-config.js
```diff
@@ -0,0 +1,105 @@
...
+   * @returns {PluginConfig}
+   */
+  static create (io, {extData, logFn} = {}) {
+    const instance = new PluginConfig(io, {logFn, extData});
+    PluginConfig._instances[io.appiumHome] = instance;
+    return instance;
+  }
+
+  /**
+   * Creates or gets an instance of {@link PluginConfig} based value of `appiumHome`
```
mykola-mokhnach, line 49: hm, not sure this method creates anything, but IMHO it makes sense to combine it with `create` one

packages/appium/lib/schema/schema.js
```diff
@@ -4,11 +4,40 @@ import Ajv from 'ajv';
...
+ * A Map where you can set properties, but only once. And you can't remove anything. So there.
+ * @template K,V
+ * @extends {Map<K,V>}
+ */
+export class ReadonlyMap extends Map {
+  /**
+   * @param {K} key
+   * @param {V} value
+   */
+  set (key, value) {
```
mykola-mokhnach, line 22: IMHO it's confusing to have `set` in an immutable map. I'd rather prefer an implementation where a map can only be initialised upon creation like in https://immutable-js.com/#javaScript-first-api

packages/appium/lib/schema/schema.js
```diff
@@ -113,7 +144,7 @@ class AppiumSchema {
    * @returns {boolean} If registered
    */
   hasRegisteredSchema (extType, extName) {
-    return this._registeredSchemas[extType].has(extName);
+    return Boolean(this._registeredSchemas[extType].has(extName));
```
mykola-mokhnach, line 147: isn't `has` already returning a boolean value?

packages/support/lib/env.js
```diff
@@ -0,0 +1,156 @@
...
+ * `DEFAULT_APPIUM_HOME`.
+ *
+ * To be clear: if `appium` is a non-`file:` dependency, we don't care if it
+ * hasn't been installed yet; we use `cwd` as `APPIUM_HOME` regardless.
+ */
+export const resolveAppiumHome = _.memoize(
+  /**
+   * @param {string} [cwd] - Current working directory.  _Must_ be absolute, if specified.
+   */
+  async (cwd) => {
```
mykola-mokhnach, line 110: I prefer to use non anonymous functions there, because it looks more clean in the exception stack trace

packages/support/lib/env.js
```diff
@@ -0,0 +1,156 @@
...
+    if (cwd && !path.isAbsolute(cwd)) {
+      throw new TypeError('`cwd` parameter must be an absolute path');
+    }
+
+    if (process.env.APPIUM_HOME) {
+      return process.env.APPIUM_HOME;
+    }
+
+    try {
+      cwd = cwd ?? process.cwd();
```
mykola-mokhnach, line 120: not sure about javascript, but in Java I always try to [avoid](https://softwareengineering.stackexchange.com/questions/245767/is-modifying-an-incoming-parameter-an-antipattern) reassigning of arg values and create a new variable instead

packages/support/lib/fs.js
```diff
@@ -23,13 +22,13 @@ const findRootCached = _.memoize(findRoot);
 const fs = {
   async hasAccess (path) {
     try {
-      await this.access(path, _fs.R_OK);
+      await fs.access(path, _fs.R_OK);
```
mykola-mokhnach, line 25: is `fs` really known at this stage?

packages/support/lib/util.js
```diff
@@ -288,16 +291,14 @@ async function isSameDestination (path1, path2, ...pathN) {
...
- * @param {string|number} ver - Version string to coerce
- * @param {boolean} strict [true] - If true then an exception will be thrown
- * if `ver` cannot be coerced
+ * @param {string} ver - Version string to coerce
  * @returns {string} Coerced version number or null if the string cannot be
  * coerced and strict mode is disabled
  * @throws {Error} if strict mode is enabled and `ver` cannot be coerced
  */
-function coerceVersion (ver, strict = true) {
+function coerceVersion (ver) {
```
mykola-mokhnach, line 299: this is a breaking change, which might render other depending code to fail. Are you sure you want to keep it?

packages/appium/lib/cli/npm.js
boneskull: so, yes, because `resolveFrom` uses `require.resolve()` under the hood (which, in turn, is used by `require()`) which normalizes paths.  

If I wanted to build a path but _not_ attempt to resolve it as a module, I'd need to use e.g., `path.join()` or `path.resolve()` and pass multiple args without slashes.

packages/appium/lib/cli/parser.js
boneskull: it was redundant; `getParser()` is the only code which calls `new ArgParser()` and the signature for the `ArgParser` constructor is `constructor(debug = false)`.  So the default gets set there.  This prevents us from needing to maintain two sets of the same default value.

packages/appium/lib/config.js
```diff
@@ -296,10 +295,12 @@ async function validateTmpDir (tmpDir) {
   }
 }
 
+const rootDir = fs.findRoot(__dirname);
```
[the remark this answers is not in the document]
boneskull: Yes. It's re-exported from the `find-root` module.

packages/appium/lib/extension/driver-config.js
boneskull: our tests could do this sort of thing, if they don't already.  but we needed a "key" to ensure we're not creating multiple copies of this class, and `APPIUM_HOME` was obvious, but it doesn't need to be that.  It could be an instance of a `Manifest` object, and we could use a `WeakMap`.

packages/appium/lib/extension/driver-config.js
```diff
@@ -0,0 +1,215 @@
...
+ * @extends {ExtensionConfig<DriverType>}
+ */
+export class DriverConfig extends ExtensionConfig {
+  /**
+   * A mapping of `APPIUM_HOME` values to {@link DriverConfig} instances.
+   * Each `APPIUM_HOME` should only have one associated `DriverConfig` instance.
+   * @type {Record<string,DriverConfig>}
+   * @private
+   */
+  static _instances = {};
```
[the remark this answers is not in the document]
boneskull: we could! this is mostly a stylistic choice, but I'm exposing it here for debugging/testing purposes.  I know you've mentioned you don't love `static`, but I think it's a fine place to put it. 😄

packages/appium/lib/extension/driver-config.js
boneskull: Yes.  Not my fault. 😄

packages/appium/lib/extension/driver-config.js
boneskull: No.  If one does not exist, it will create a new one.

packages/appium/lib/extension/driver-config.js
```diff
@@ -0,0 +1,215 @@
...
+
+    return problems;
+  }
+
+  /**
+   * @param {ExtName<DriverType>} driverName
+   * @param {ManifestDriverData} extData
+   * @returns {string}
+   */
+  extensionDesc (driverName, {version, automationName}) {
```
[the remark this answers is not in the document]
boneskull: this is old code and only appears here since I moved it.

packages/appium/lib/extension/driver-config.js
boneskull: old code

packages/appium/lib/extension/extension-config.js
boneskull: old code.  but... since nothing other than `ExtensionConfig` or its subclasses should ever call this function, it's _probably_ OK.

packages/appium/lib/extension/extension-config.js
boneskull: noted

packages/appium/lib/extension/extension-config.js
boneskull: Yes, that is old. I'll change it.

packages/appium/lib/extension/extension-config.js
boneskull: Only if you dereference it, I suppose.  old code

packages/appium/lib/extension/extension-config.js
boneskull: old code

packages/appium/lib/extension/extension-config.js
boneskull: No

packages/appium/lib/extension/manifest.js
boneskull: right, so, the reason this _sometimes_ happens is that we don't seem to have anything that automates the spaces around the imports.  I try to remember to do it, but my formatter will sometimes remove them.

packages/appium/lib/extension/manifest.js
boneskull: Well... it's an `s`.  `driver` becomes `drivers`.  This is a key in the root of `extensions.yaml`.

packages/appium/lib/extension/manifest.js
boneskull: Because 4 was large enough :D

packages/appium/lib/extension/manifest.js
boneskull: I don't really need anything bluebird provides here and am not using the global `Promise`...  Anyway, if we want to enforce an explicit cast of all promises to bluebird, we should put that in our eslint config.

packages/appium/lib/extension/manifest.js
boneskull: I learned the exact opposite.  checking if a file exists first is more I/O overhead

packages/appium/lib/extension/package-changed.js
boneskull: if a user calls `packageDidChange` they have other problems 😄  

this exception is really for a contributor to appium who may call the function with something weird.

I could also just remove the conditional/exception and trust people will mind the types.

packages/appium/lib/extension/package-changed.js
boneskull: hm, yes, I can do that here

packages/appium/lib/extension/package-changed.js
boneskull: ahh, that syntax had eluded me.  parentheses!! who knew

packages/appium/lib/extension/package-changed.js
boneskull: This case is probably rare, but I think yes. A `true` return value causes appium to look for extensions.  If we're here, we've already determined that we're in a dir with a `package.json` and that `package.json` contains appium.  if we don't return `true` we may miss an extension that the user added.  I think it's better to do (potentially) extra work in this case.

packages/appium/lib/extension/driver-config.js
mykola-mokhnach: but this particular call does not check if `_instances` mapping contains `io.appiumHome`, does it?

packages/appium/lib/extension/manifest.js
mykola-mokhnach: maybe a small explanation comment would make it cleaner. For me this magic char looked quite cryptic

packages/appium/lib/extension/manifest.js
mykola-mokhnach: I usually don't really care what type of promise it is. If we need special bluebird features we could always safely call B.resolve(...) on the initial promise and then safely use its features.

packages/appium/lib/extension/package-changed.js
mykola-mokhnach: if this is intended to be used by devs only then I assume they would be fine with default exception they'd get if a wrong method is called on appiumHome instance. The docs clearly states the var is expected to be a string

packages/appium/lib/extension/package-changed.js
boneskull: yes

packages/appium/lib/extension/plugin-config.js
boneskull: yes, this is confusing and one of these functions should probably go away.

packages/appium/lib/schema/schema.js
boneskull: I think this is a misnomer. I probably need to rename it.  It's _not_ an immutable Map.  It's a readable map which only supports "inserts"; no delete and no update.    I don't really know what to call that.

packages/appium/lib/schema/schema.js
boneskull: Yes, good catch.

packages/support/lib/env.js
boneskull: probably a good idea

packages/support/lib/env.js
boneskull: I hear what you're saying, but I'm not really convinced by many of the arguments in that blog post.  L120 is no different than assigning a default value in the function signature.  Which.. I think I can do here, so I'm going to do that.

packages/support/lib/fs.js
boneskull: yes.

packages/support/lib/util.js
boneskull: no, there is no other code calling `coerceVersion` afaik.

mykola-mokhnach: There is. A simple search through the appium org looks like

coerce

packages/appium/lib/extension/package-changed.js
boneskull: you're right, I'm not sure why I did that.

packages/appium/lib/extension/driver-config.js
boneskull: As you mentioned here and below, these were confusing.  I rewrote them (pushing in a few), so:

- `DriverConfig.create(manifest, opts)` only creates an instance, and throws if one is already associated with a `Manifest`.
- `DriverConfig.getInstance(manifest)` returns an existing instance if it exists, or `undefined` if it does not.

(same for `PluginConfig`)

The mapping is now a `WeakMap` based on the `Manifest` instance instead of `appiumHome`.

packages/appium/lib/extension/driver-config.js
boneskull: I guess we could do `this.knownAutmationNames = new Set()`

boneskull: On second thought, maybe it is my fault.  Just not _recently_ my fault. :D

So, when we validate, we're validating the entire collection of drivers found in `extensions.yaml`.

`knownAutomationNames` tracks if there's a collision in automation names, since there can only be one driver per name.

`validate()` in `ExtensionConfig` calls `getWhateverProblems()`, and _that_ function begins adding automation names to the `Set`.  

So, what we could do to avoid the side effect is to create a new `Set` in `validate()`, then explicitly pass it into each call to `getWhateverProblems()`.  This is more verbose, but probably a better idea.

boneskull: But, since this is a subclass, and `ExtensionConfig#validate` does the work, we break the abstraction by passing something driver-specific into the abstract `getConfigProblems(..)`.

The refactoring required to make this work due to the inheritance at play may not be worth the bother.

packages/support/lib/util.js
boneskull: oh... ugh.

boneskull: I think I thought this was a different, non-public utils...

boneskull: Anyway, I've reverted this and fixed the type.

packages/appium/lib/schema/schema.js
boneskull: I've renamed it 😄

packages/appium/lib/extension/driver-config.js
boneskull: `loadExtensions()` now tries `getInstance()` before calling `create()`.

boneskull:
status: 

- changing logic to search in parent dirs for a `package.json` containing an `appium` dependency. if I don't do this, running appium from any subdirectory of a working copy containing a `package.json` depending on `appium` will likely result in using the _default_ `APPIUM_HOME` instead of the parent dir of `package.json` like it should. 
- e2e tests

I'm finding the line between what goes in `@appium/support` and what doesn't kind of fuzzy. questions:

- should the `env` module actually live in `@appium/support`? I'm thinking yes, because other packages or scripts may have need for resolving `APPIUM_HOME`.  
- do all of the functions in `env` actually belong there? for instance, the functions which deal with reading/finding `package.json`.
- likewise, should `package-changed` (currently in `appium/lib/extension`) move somewhere else?  it can trivially be made generic

@jlipps @mykola-mokhnach

jlipps:
should the env module actually live in @appium/support?

yeah, putting some of those things in `@appium/support` if we think other packages will use them isn't a bad idea.

> do all of the functions in env actually belong there?

seems ok

> likewise, should package-changed (currently in appium/lib/extension) move somewhere else? it can trivially be made generic

up to you

boneskull:
that failure is a good head-scratcher since I didn't change anything there.  this will be fun to debug

boneskull:
* 2.0
  * **PR #16443** <a href="https://app.graphite.dev/github/pr/appium/appium/16443" target="_blank">Graphite</a>  (this PR)
    * **PR #16482** <a href="https://app.graphite.dev/github/pr/appium/appium/16482" target="_blank">Graphite</a> 
    * **PR #16459** <a href="https://app.graphite.dev/github/pr/appium/appium/16459" target="_blank">Graphite</a> 

boneskull:
@jlipps this is now passing the build, so ready for your perusal.

I'll do a proper squash when it looks ok.

packages/appium/lib/extension/driver-config.js
boneskull: ... that being said, I think it _should_ be refactored.  Just not here.

jlipps:
ok i should be ready to review this, shall i wait for all the depended PRs to be finalized in case that triggers changes here?

boneskull:
@jlipps, no, those PRs are deps on _this_ one.  so once this is merged, those directly targeting this one will be retargeted to 2.0.

packages/appium/test/setup.js
```diff
@@ -0,0 +1,6 @@
+// This configures @babel/register to look for a babel config in parent dir(s) of
```
jlipps, line 1: helpful comment 👍

jlipps (commented):
uhhh that took a long time. i think there were one or two minor changes i requested and a few clarifying questions. overall looks good.

package.json
```diff
@@ -124,6 +130,7 @@
     "sync-monorepo-packages": "0.3.4",
     "through2": "4.0.2",
     "type-fest": "2.8.0",
+    "typescript": "4.6.0-dev.20220210",
```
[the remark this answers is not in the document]
boneskull: yeah, last time I checked, "it doesn't work".  I'll try again though with 4.5.x

packages/appium/lib/cli/args.js
```diff
@@ -87,9 +82,9 @@ function makeInstallArgs (type) {
     [['--source'], {
       required: false,
       default: null,
-      choices: INSTALL_TYPES,
+      choices: INSTALL_TYPES_ARRAY,
```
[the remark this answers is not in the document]
boneskull: yeah. I thought I added a comment to that effect somewhere. `argparse` does naughty things.

packages/appium/lib/cli/extension-command.js
```diff
@@ -1,47 +1,47 @@
...
-   * @return {ExtensionCommand}
+   * @param {ExtensionCommandOptions<ExtType>} opts
    */
   constructor ({config, json, type}) {
+    /** @type {ExtensionConfig<ExtType>} */
     this.config = config;
+
+    /** @type {ExtType} */
     this.type = type;
+    /**
```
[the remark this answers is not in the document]
boneskull: oh, uh, no reason.  I guess we could actually document these variables

packages/appium/lib/cli/extension-command.js
```diff
@@ -121,7 +121,7 @@ export default class ExtensionCommand {
           typeTxt = '(NPM)';
       }
       const installTxt = installed ?
-        `@${version.yellow} ${('[installed ' + typeTxt + ']').green}` :
+        `@${/** @type {string} */(version).yellow} ${('[installed ' + typeTxt + ']').green}` :
```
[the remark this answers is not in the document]
boneskull: This isn't _truly_ a JS-in-TS thing.  `version` may be `undefined`, per the type we've given it.  the equivalent in TS would be

```ts
`@${(<string>version).yellow}`
```

...so it's still bothersome.   Anyway, whenever we see this sort of inline typing, it's a coercion.  It means one of two things: 

1. TS simply cannot understand the type.
2. The type is wrong

For example, here's a situation where TS just doesn't get it:

```ts
const foo: Set<string> = new Set([' bar ']);
if (foo.has('bar')) {
  const bar: string = foo.get('bar'); // <-- ERROR, `bar` may be `undefined`
}
```

In TS, you need to use type `string!` (iirc; the `!` means "not undefined" I think).  but it's not possible to do this through type inference alone; you have to tell the compiler that it _really_ is defined.  likewise with JS... but the only tool we have is inline types, which is more verbose but also easier to understand.

Ideally, these inline types should be kept to a bare minimum.  In this case, I think `version` _may always be a `string`_,  so I'll need to double check.

IMO the effort required to migrate to TS is not going to be worth it.

packages/appium/lib/extension/driver-config.js
```diff
@@ -0,0 +1,230 @@
...
+    *
+    * @param {Manifest} manifest
+    * @param {DriverConfigOptions} [opts]
+    * @throws If `manifest` already associated with a `DriverConfig`
+    * @returns {DriverConfig}
+    */
+   static create (manifest, {extData, logFn} = {}) {
+     const instance = new DriverConfig(manifest, {logFn, extData});
+     if (DriverConfig.getInstance(manifest)) {
+       throw new Error(`Manifest with APPIUM_HOME ${manifest.appiumHome} already has a DriverConfig; use DriverConfig.getInstance() to retrieve it.`);
```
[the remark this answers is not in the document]
boneskull: I'd say yes, because the "someone" is likely a contributor/developer who did a thing wrong. 

I revisited this code a couple times.  The problem is the options object in second parameter.  A developer could be tempted to call `DriverConfig.create('/root', myOpts)` and then `DriverConfig.create('/root', otherOpts)` for some reason.  if the code just returned the `Manifest` for `/root`, what of `otherOpts`?  This would be unexpected behavior; it'd return a `Manifest` having opts `myOpts`.  

That's why there's a `create()` _and_ a `getInstance()`.

packages/appium/lib/extension/extension-config.js
```diff
@@ -0,0 +1,459 @@
...
+      await this.manifest.write();
+    }
+  }
+
+  /**
+   * @param {ExtName<ExtType>[]} [activeNames]
+   * @returns {void}
+   */
+  // eslint-disable-next-line no-unused-vars
+  print (activeNames) {
```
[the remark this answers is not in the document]
boneskull: yes.

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,570 @@
...
+  /**
+   * Searches `APPIUM_HOME` for installed extensions and adds them to the manifest.
+   * @param {SyncWithInstalledExtensionsOpts} opts
+   * @returns {Promise<boolean>} `true` if any extensions were added, `false` otherwise.
+   */
+  async syncWithInstalledExtensions ({depthLimit = DEFAULT_SEARCH_DEPTH} = {}) {
+    const walkOpts = _.defaults({depthLimit}, DEFAULT_FIND_EXTENSIONS_OPTS);
+    // this could be parallelized, but we can't use fs.walk as an async iterator
+    let didChange = false;
+    for await (const {stats, path: filepath} of fs.walk(
```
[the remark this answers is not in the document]
boneskull: is this an interview question? 😄   this is my best guess; I've never seen the second syntax, but it looks valid.

```js
for await (x of y) { // y is an async iterator object. y is _not_ treated like a `Promise`
  // the promise is here. the loop does not continue until it resolves
}
```

```js
for (x of await y) {  // y may or may not be a promise, but the loop doesn't begin until it resolves to a synchronous iterator.
  // no promise in here
}
```

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,570 @@
...
+      }
+    } else if (isPlugin(pkgJson)) {
+      if (!this._data.plugins[pkgJson.appium.pluginName]) {
+        this._data.plugins[pkgJson.appium.pluginName] = {
+          ..._.omit(pkgJson.appium, 'pluginName'),
+          ...internal,
+        };
+      }
+    } else {
+      throw new TypeError(
```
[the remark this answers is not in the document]
boneskull: This function (`addExtensionFromPackage`) is called by `syncWithInstalledExtensions`, which is called by `read()` under certain conditions.  `read()` is called by `init()`, which is called by `main()`, and this `TypeError` is going to bubble up to `main`'s rejection handler.

I'd say the people likely to encounter this error are extension developers. They should know if their extension is invalid. For that reason, I probably wouldn't want to just warn; it'd be too easy to miss the fact that their extension isn't loaded, and this would be a confusing debugging experience.

packages/appium/lib/extension/manifest.js
```diff
@@ -0,0 +1,570 @@
...
+    const internal = {
+      pkgName: pkgJson.name,
+      version: pkgJson.version,
+      installType: INSTALL_TYPE_NPM,
+      installSpec: `${pkgJson.name}@${pkgJson.version}`,
+    };
+
+    if (isDriver(pkgJson)) {
+      if (!this._data.drivers[pkgJson.appium.driverName]) {
+        this._data.drivers[pkgJson.appium.driverName] = {
```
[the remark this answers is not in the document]
boneskull: good catch

packages/appium/lib/extension/plugin-config.js
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @extends {ExtensionConfig<PluginType>}
+ */
+export class PluginConfig extends ExtensionConfig {
+
+  /**
+   * A mapping of {@link Manifest} instances to {@link PluginConfig} instances.
+   *
+   * `Manifest` and {@link ExtensionConfig} have a one-to-many relationship; each `Manifest` should be associated with a `DriverConfig` and a `PluginConfig`; no more, no less.
```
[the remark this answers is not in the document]
boneskull: It's...well, that depends if you would define "two" as "many".

packages/appium/lib/grid-register.js
```diff
@@ -12,9 +12,9 @@ const hubUri = (config) => {
 /**
  * Registers a new node with a selenium grid
  * @param {string|object} data - Path or object representing selenium grid node config file
- * @param {string} addr - Bind to this address
- * @param {number} port - Bind to this port
- * @param {string} basePath - Base path for the grid
+ * @param {string} [addr] - Bind to this address
```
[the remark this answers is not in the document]
boneskull: Yes.  If `data` is a `string`, they are required. If `data` is an `object`, they are not.  I'll make sure there's a comment to that effect.

packages/appium/lib/schema/schema.js
```diff
@@ -4,12 +4,42 @@ import Ajv from 'ajv';
...
 import { APPIUM_CONFIG_SCHEMA_ID, ArgSpec, SERVER_PROP_NAME } from './arg-spec';
 import { keywords } from './keywords';
 
+/**
+ * Key/value pairs go in... but they don't come out.
+ *
+ * @template K,V
+ * @extends {Map<K,V>}
+ */
+export class RoachHotelMap extends Map {
```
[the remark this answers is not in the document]
boneskull: I renamed `ReadonlyMap` to `RoachHotelMap` because @mykola-mokhnach pointed out that it was a misnomer.  It's _not_ "read-only", since you can write to it.  But you _can't_ take things out or change what's already there.

packages/appium/test/setup.js
boneskull: yes, my verbosity increases with the size of the headache that the problem gave me

packages/appium/lib/cli/extension-command.js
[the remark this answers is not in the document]
boneskull: OK, the `version` _can_ be empty if the item was pulled out of `knownExtensions`.  The type is not _wrong_ per se, but it's not specific enough; if `installed` is `true`, this _implies_ that `version` is a `string`.  So I'll see if I can work that out.

packages/appium/lib/extension/driver-config.js
```diff
@@ -0,0 +1,230 @@
...
+       const {
+         driverName,
+         mainClass,
+         version,
+       } = this._getDriverBySupport(automationName, platformName);
+       log.info(`The '${driverName}' driver was installed and matched caps.`);
+       log.info(`Will require it at ${this.getInstallPath(driverName)}`);
+       const driver = this.require(driverName);
+       if (!driver) {
+         throw new Error(`MainClass ${mainClass} did not result in a driver object`);
```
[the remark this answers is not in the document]
boneskull: err, actually, it's not instantiating the class, it's just returning the class/constructor itself. but I can make it better

packages/appium/lib/cli/extension-command.js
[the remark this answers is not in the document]
jlipps: I was making a simpler point, which is that ``@${/** @type {string} */(version).yellow}` is much uglier and harder to read in the code than `@${(version as string).yellow}`.

packages/appium/lib/extension/driver-config.js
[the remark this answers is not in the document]
jlipps: 👍

package.json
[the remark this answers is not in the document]
jlipps: ok. should we add a comment in here like:

```
"devDependencies": {
  ...,
  "//": {
    "typescript": "4.5.x doesn't work because reasons"
  }
}
```

packages/appium/lib/extension/manifest.js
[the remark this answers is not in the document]
jlipps: ok. i think you thought i was being smarter than i am. but your comment helped me to see the answer is "async iterators".

i've only ever used the second syntax. i guess i thought we needed it because it was a promise that resolved to an array. but instead it is apparently an async iterator. so we want the first syntax.

packages/appium/lib/extension/manifest.js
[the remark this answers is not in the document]
jlipps: 👍

packages/appium/lib/extension/plugin-config.js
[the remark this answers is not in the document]
jlipps: it just made it hard for me to grok this comment, as if there was more structure possible here than there is. it would have read more straightforward to me as something like just: "Each `Manifest` has two  `ExtensionConfig`s; one for plugins and one for drivers"

boneskull:
I am the king of git and I am going to merge this now

jlipps:
\o/ woooo!!!

PR #16458, opened 2022-02-15
test(base-driver): replace direct calls to sinon with a sandbox

boneskull:
This is safer, because test failures can happen due to lack of isolation and order in which tests are executed


jlipps (commented):
👍

PR #16478, opened 2022-02-21
refactor: Make loggers connected to the particular driver instances

mykola-mokhnach:
## Proposed changes

The current issue is that driver loggers are created on per-module basis. this is probably ok for utility functions, but is not very helpful for driver methods, where we have instance specific data, like a session identifier and a unique instance identifier itself. This refactor add a new `log` property to the base driver, whose purpose it to properly setup log prefix based on the current driver instance state. This makes it easier to read logs especially if multiple driver instances are being executed within one nodejs process.

## Types of changes


KazuCocoa (approved):
lgtm (i like this). It would be nice to wait for jlips as well

packages/base-driver/lib/protocol/protocol.js
```diff
@@ -38,13 +34,28 @@ function extractProtocol (driver, sessionId = null) {
...
   // Extract the protocol for the current session if the given driver is the umbrella one
-  return dstDriver ? dstDriver.protocol : SESSIONS_CACHE.getProtocol(sessionId);
+  return dstDriver?.protocol ?? PROTOCOLS.W3C;
 }
 
 function isSessionCommand (command) {
   return !_.includes(NO_SESSION_ID_COMMANDS, command);
 }
 
+function getLogger (driver, sessionId = null) {
```
jlipps, line 44: seems like we've defined this logger name generator logic in several places. should we consolidate it into one helper?

packages/support/lib/node.js
```diff
@@ -150,4 +151,19 @@ function getObjectSize (obj) {
...
+
+/**
+ * Calculates a unique object identifier
+ *
+ * @param {object} object Any valid ECMA object
+ * @returns {string} A uuidV4 string that uniquely identifies given object
+ */
+function getObjectId (object) {
+  if (!OBJECTS_MAPPING.has(object)) {
+    OBJECTS_MAPPING.set(object, uuidV4());
```
jlipps, line 164: do we care about the size of this object map getting out of hand?

jlipps (approved):

packages/support/lib/node.js
mykola-mokhnach: it should not because it only contains references. And the map object itself is quite small

packages/base-driver/lib/protocol/protocol.js
mykola-mokhnach: not sure what exactly could be extracted into a common function here

PR #16482, opened 2022-02-22
mkdirp => fs.mkdirp; use built-in mkdir recursive

boneskull:
This PR:

1. Removes `mkdirp` dep from `@appium/support`
2. Creates the `fs.mkdirp` function in `@appium/support`
3. Deprecates the `mkdirp` module (re-exports `fs.mkdirp`)
4. Updates `appium` to use `fs.mkdirp`


boneskull:
* 2.0
  * **PR #16443** <a href="https://app.graphite.dev/github/pr/appium/appium/16443" target="_blank">Graphite</a> 
    * **PR #16459** <a href="https://app.graphite.dev/github/pr/appium/appium/16459" target="_blank">Graphite</a> 
    * **PR #16482** <a href="https://app.graphite.dev/github/pr/appium/appium/16482" target="_blank">Graphite</a>  (this PR)

packages/support/lib/fs.js
```diff
@@ -16,14 +18,15 @@ import log from './logger';
...
+const mkdirAsync = /** @type {(filepath: string, opts: _fs.MakeDirectoryOptions|undefined) => B<void>} */(B.promisify(_fs.mkdir));
+const ncpAsync = /** @type {(source: string, dest: string, opts: ncp.Options|undefined) => B<void>} */(B.promisify(ncp));
 const findRootCached = _.memoize(pkgDir.sync);
 
 const fs = {
   async hasAccess (path) {
     try {
-      await fs.access(path, _fs.R_OK);
+      // as of recent-ish node versions, `R_OK` moved into the `constants` property of `fs`
+      await fs.access(path, _fs.constants.R_OK);
```
mykola-mokhnach, line 29: shall we use `_fs.constants.R_OK ?? _fs.R_OK` ?

packages/support/lib/fs.js
```diff
@@ -16,14 +18,15 @@ import log from './logger';
 import Timer from './timing';
 import pkgDir from 'pkg-dir';
 
-const mkdirAsync = B.promisify(_fs.mkdir);
-const ncpAsync = B.promisify(ncp);
+const mkdirAsync = /** @type {(filepath: string, opts: _fs.MakeDirectoryOptions|undefined) => B<void>} */(B.promisify(_fs.mkdir));
```
mykola-mokhnach, line 21: can we use https://nodejs.org/api/fs.html#fspromisesmkdirpath-options, which returns a promise?

packages/support/lib/fs.js
```diff
@@ -45,15 +48,15 @@ const fs = {
    * Copies files _and entire directories_
    * @param {string} source - Source to copy
    * @param {string} destination - Destination to copy to
-   * @param {...any} args - Additional arguments to pass to `ncp`
+   * @param {ncp.Options} [opts] - Additional arguments to pass to `ncp`
```
mykola-mokhnach, line 51: would that be a breaking change? are these args used anywhere else in appium org?

packages/support/lib/fs.js
```diff
@@ -32,9 +35,9 @@ const fs = {
   exists (path) { return fs.hasAccess(path); },
   rimraf: B.promisify(rimraf),
   rimrafSync: rimraf.sync.bind(rimraf),
-  async mkdir (...args) {
+  async mkdir (filepath, opts = {}) {
```
mykola-mokhnach, line 38: shall we check other org modules for possible usages of special args?

packages/support/lib/fs.js
boneskull: I guess the comment is rather ambiguous.  I'll fix that, but the changed one is correct

packages/support/lib/fs.js
boneskull: I was just keeping things the way they already are. that said: we can replace all of the promisified `fs` methods with native ones instead of doing it ourselves, unless there's a reason to wrap them in Bluebird promises...

packages/support/lib/fs.js
boneskull: no. `ncp` accepts four parameters, one of which is a callback which is promisified away.  `source`, `dest`, and `options`.  any further arguments are discarded, so it is not necessary to use a rest param here.

packages/support/lib/fs.js
boneskull: much like `ncp` above, `mkdir` only accepts three params (`dirname`, `opts`, and a `callback`).  so we only need the first two.

packages/support/lib/fs.js
boneskull: I just removed the comment.  `fs.R_OK` is deprecated, afaict.  guess I don't need to add a comment to explain code that no longer exists, huh

packages/support/lib/fs.js
mykola-mokhnach: there was a reason for it before.:the promisified mkdir only appeared in node 10. Now, when we set node14 to be the minimum supported version it is totally safe to drop promisifying.

mykola-mokhnach (approved):

packages/support/lib/fs.js
boneskull: I think we're at Node.js v12 minimum now, so we can just use the builtins w/o promisifying.  I'll take care of that now.

boneskull: @mykola-mokhnach I've made those changes.

KazuCocoa (approved):

jlipps (commented):
👍

PR #16488, opened 2022-02-23
refactor(base-driver): Tune the logic for the destination logger extraction

mykola-mokhnach:
## Proposed changes

It is easier to read logs when a destination driver name is present in the prefix rather than the umbrella one

## Types of changes

### DIFF
--- a/packages/base-driver/lib/protocol/protocol.js
+++ b/packages/base-driver/lib/protocol/protocol.js
@@ -42,13 +42,16 @@ function isSessionCommand (command) {
 }
 
 function getLogger (driver, sessionId = null) {
-  if (_.isFunction(driver?.log?.info)) {
-    return driver.log;
+  const dstDriver = _.isFunction(driver.driverForSession)
+    ? (driver.driverForSession(sessionId) ?? driver)
+    : driver;
+  if (_.isFunction(dstDriver.log?.info)) {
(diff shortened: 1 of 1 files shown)


packages/base-driver/lib/protocol/protocol.js
```diff
@@ -42,14 +42,16 @@ function isSessionCommand (command) {
...
+    return dstDriver.log;
   }
 
-  let logPrefix = 'AppiumDriver';
-  if (driver && driver.constructor) {
-    logPrefix = driver.constructor.name;
-  }
+  let logPrefix = dstDriver.constructor
+    ? `${dstDriver.constructor.name}@${node.getObjectId(dstDriver).substring(0, 8)}`
+    : 'AppiumDriver';
```
KazuCocoa, line 54: or `BaseDriver` as this package name? (maybe `Appium` prefix might be redundant)

KazuCocoa (approved):

packages/base-driver/lib/protocol/protocol.js
mykola-mokhnach: I use this one, because it is aligned with other loggers in this module. Anyway, I don't think there ever might be a situation where the destination driver does not have a constructor

PR #16540, opened 2022-03-09
chore: lint the code once per build action

boneskull:
~~This adopts the `bahmutov/npm-install` action, which also does some caching for us~~

This may speed up builds.

**UPDATE:** Unfortunately, we're going to need to hand-roll our caching if we want it. The action I tried before (which I've used elsewhere) won't work well because `package-lock.json` is not under version control.  Without this file, the action will be unable to determine if the dependencies of a package (e.g., `@appium/support`) have changed, because it will only look at the root `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,8 @@
     "test": "lerna run test",
     "test:appium": "lerna run --scope=appium test",
     "test:base-driver": "lerna run --scope=@appium/base-driver test",
+    "pretest:ci": "npm run build",
+    "test:ci": "lerna run test",
     "test:doctor": "lerna run --scope=@appium/doctor test",
     "test:e2e:appium": "lerna run --scope=appium test:e2e",
     "test:e2e:base-driver": "lerna run --scope=@appium/base-driver test:e2e",
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,13 +1,13 @@
 # This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
-name: Node.js CI
+name: Appium Build
 on: [push, pull_request]
 
 jobs:
-  build:
+  test:
+    name: Test
     runs-on: ubuntu-latest
-
     strategy:
       matrix:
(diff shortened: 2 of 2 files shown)


jlipps:
now that we're pinning deps, i don't know that there's much to be gained in terms of developer experience by _not_ including package-lock in git. maybe we should start?

boneskull:
@jlipps I think we could go either way.  

The reason why we wouldn't is "we won't install the same transitive dependencies end-users get when they `npm install` our packages".  Which is perhaps true.  However, if you don't commit `package-lock.json` and don't religiously run `npm run reinstall`, you're going to be in the same boat.  `npm upgrade` might work, too--but only if you've removed `package-lock.json` first.

My opinion: let's commit it, but be aware that doing so could be a source of unreproducible bugs.  If we encounter such a situation, we must be sure to `npm run reinstall`.   We may be able to create a scheduled GH action to automatically regenerate `package-lock.json` (and commit it) with the latest transitive deps, as well.

jlipps:
> We may be able to create a scheduled GH action to automatically regenerate package-lock.json (and commit it) with the latest transitive deps, as well.

Yeah I think we should do this. I think probably Renovate will end up doing this every night for us as well, every time it updates deps. Although it will also create a cascade of conflicts for all its open PRs in doing so. But yeah, anything that makes it more likely we'll be up to date when running `git pull` is good. I tend to `git pull && npm run reinstall` before I work anyway, just to have a clean slate.

And in projects that are monorepos with lots of package-locks, it's also a bit annoying there tends to be more noise in PRs and more conflicts when rebasing etc.

package.json
```diff
@@ -48,6 +48,8 @@
     "test": "lerna run test",
     "test:appium": "lerna run --scope=appium test",
     "test:base-driver": "lerna run --scope=@appium/base-driver test",
+    "pretest:ci": "npm run build",
+    "test:ci": "lerna run test",
     "test:doctor": "lerna run --scope=@appium/doctor test",
     "test:e2e:appium": "lerna run --scope=appium test:e2e",
     "test:e2e:base-driver": "lerna run --scope=@appium/base-driver test:e2e",
```
1108469527, line 55: عبدالله

PR #16542, opened 2022-03-10
chore: add directory prop to repository prop of package.json

boneskull:
I think npm uses this to link to the correct path of the package in a monorepo.


jlipps (approved):
nice, i didn't know they had this, and npm packages in monorepos were always a bit wonky on npmjs.com

boneskull:
@jlipps Yeah, but I think that's mostly a "user error" sort of issue.  npmjs.com supports monorepos, but people don't know how to configure them properly.  Missing READMEs...

PR #16548, opened 2022-03-10
fix(base): address zip/unzip in the docstring of configureApp

KazuCocoa:
## Proposed changes

`appPath` in https://github.com/appium/appium-espresso-driver/pull/750/files#diff-ed3c013edbc896bd87ef62539626fcdc8930a31f3cee7ecb37438e172af2c8e0R236 is still `.zip`.
It results https://github.com/appium/appium-espresso-driver/pull/750/files#diff-ed3c013edbc896bd87ef62539626fcdc8930a31f3cee7ecb37438e172af2c8e0R247 is always `false`. It caused https://github.com/appium/ruby_lib_core/issues/370

At least this change allows appium to unzip `.zip` file first before calling `onPostProcess`. Then, the `appPath` is after unzipped one.

I haven't dug deeply yet, so this might cause another issue...

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/base-driver/lib/basedriver/helpers.js
+++ b/packages/base-driver/lib/basedriver/helpers.js
@@ -193,6 +193,8 @@ async function isAppIntegrityOk (currentPath, expectedIntegrity = {}) {
 /**
  * Prepares an app to be used in an automated test. The app gets cached automatically
  * if it is an archive or if it is downloaded from an URL.
+ * If the downloaded app has `.zip` extension, this method will unzip it.
+ * The unzip does not work when `onPostProcess` is provided.
  *
  * @param {string} app Either a full path to the app or a remote URL
  * @param {string|string[]|ConfigureAppOptions} options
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
I believe unzip logic must be then implemented in espresso driver then. We should not do any assumptions about file formats in the base driver

mykola-mokhnach:
> Okay, only espresso driver uses this feature. Probably the docstring should address the unzip does not work when `onPostProcess` is given, and add it in espresso driver.
> 
> Will do it

I mean you could easily address unzipping in espresso driver code. Although I don't really find it useful in that since .apk is already a zipped file (unlike .app bundle for iOS)
Updating the documentation would also work.

KazuCocoa:
> since .apk is already a zipped file (unlike .app bundle for iOS)

yes, i know it. The ruby's test code, i found this, expected a siautation when a user stored an app as zip compressed `.apk` on a remote host. We've supported `.zip` so the aim was to prevent regressions.

KazuCocoa:
changed only docstring update as this pr

mykola-mokhnach (approved):

PR #16553, opened 2022-03-10
chore: commit package-lock.json

boneskull:
In #16540, I ran into trouble optimizing our CI build because we do not commit `package-lock.json`.

There are tradeoffs, of course.

The main issue with committing this file is that end-users which `npm install` our packages will not necessarily install the same transitive dependencies which we get in our development environment.  This _can_ be a source of issues, though likely rare.  We must be aware of it.

This problem can be mitigated to a degree by creating a scheduled GH action which automatically updates `package-lock.json` and commits it.  I _think_ `npm upgrade` might work.  We do not need to automatically upgrade the transitive devdeps--only the production ones--since end-users do not install our devdeps.  That'd be ideal, otherwise we will likely get lots of cache misses in CI (when we start using a cache...see below).

On the other hand, doing so will enable our CI runs to be cached and thus run faster.  Without `package-lock.json`, it will be exceedingly difficult to cache `node_modules` (and any `packages/*/node_modules`, which contain version conflicts with packages already installed in the root `node_modules`). Likewise, CI can use `npm ci` instead of `npm install` which is more performant.

This also "enables" deep security audits.  Right now, since we only have `package.json` committed, automated tools cannot find vulnerabilities in transitive dependencies unless they are pinned to our direct dependencies.  Granted, most statically-analyzable vuln warnings are complete garbage, but automatic updating of `package-lock.json` may also mitigate the level of noise here.  _Or_ it may result in yet more noisy warnings that we cannot fix because no fix has yet been published (TBD)!

Anyway.  What do we think?


mykola-mokhnach:
What if we keep the lock file separately for ci env (btw, did you know there is `npm ci` command, which speeds up package install significantly) ? This lock file may be regenerated automatically every we do a new package release, so dev env always corresponds to the one, which is being currently used by early adopters.

boneskull:
@mykola-mokhnach Yes this would enable `npm ci`.

I'm not sure about your suggestion... or mine.  If we automate updates, we're going to get new `package-lock.json` files pretty much daily--which means more changesets to pull down.  We'd also likely end up in a place where Renovate is sending PRs which quickly become out-of-date, or worse, conflict.  

Another route around this may be `overrides`. npm v8.3 introduced this property, which allows a package to override the version of a transitive dependency; it could be locked to an old version, changed to allow major versions, or even point to a different package/version/tarball entirely.  We could essentially _abuse_ that feature to lock our transitive dependencies:

1.  delete `package-lock.json`
2. Run `npm install` to generate a new one
3. Run a script which parses `package-lock.json` and inserts overrides into `packages/*/package.json` corresponding to the versions found in the lockfile

This would kind of be like shrinkwrapping; we'd force our transitive deps to be pinned too.  

This is also probably a bad idea.  We should not do this (but we could).

I don't know.  Unless we shrinkwrap `appium`, users _will_ get new versions of transitive dependencies.  Pinning our own dependencies only mitigates a fraction of the potential delta.  In my mind, committing a lockfile doesn't look too different from the status quo.  The current state of things is just easier for us to ignore entirely.

jlipps:
As for Renovate, it does generate package-lock.json dependencies. The net result is that if that if it opens N dep update PRs, then we run (N^2 + N)/2 builds (since it will auto-merge one, generate a conflict for all the others, all of which re-run, then it will auto-merge one, generating a conflict for the rest, and so on). so at the end of the day it slows things down a bit and uses more resources, but isn't an active pain for developers.

boneskull:
@jlipps Right, but what I'm saying is because Renovate will _also_ start looking at `package-lock.json` we will get _many_ more PRs to review.  Right now we only get PRs for our direct dependencies.

jlipps:
@boneskull there's gotta be a way to configure renovate so that it only proposes updates when our direct deps have updates. renovate has seemed infinitely configurable so far.

jlipps:
what are we thinking of doing with this? I'm pretty sure that Renovate's default behaviour doesn't trigger when a subdep updates (only when a direct dep is published).

boneskull:
@jlipps I'll update this PR to configure renovate appropriately.  If we don't like the result, or if we run into other problems in the future, we can always undo this.

boneskull:
errm, wait, can't do that in _this_ PR, since `renovate.json` lives in `master`.

boneskull:
* 2.0
  * **PR #16892** <a href="https://app.graphite.dev/github/pr/appium/appium/16892" target="_blank">Graphite</a> 
    * **PR #16553** <a href="https://app.graphite.dev/github/pr/appium/appium/16553" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16553?utm_source=stack-comment).

jlipps:
I went to review this PR but it looks like it has changes from other PRs in it; can we resolve that?

jlipps (approved):
there go the good old days of no package lock conflicts!

boneskull:
I think I'm gonna try to merge this, but I need to manually update `package-lock.json`

boneskull:
wait what, the "update" button appends a merge changeset? crappy crappiness

PR #16594, opened 2022-03-16
add @appium/types package

boneskull:
This PR does two things:

1. Provides some type declarations which will be used across other packages and consumers in `@appium/types`
2. New TS configuration
    This changes the TS configuration to to a) emit declarations for supported packages, and b) build declarations incrementally by package.

    To begin, we are targeting `appium`, `@appium/base-driver` and `@appium/support` for declarations; these three are intended to be published with declarations generated via their JS.

    Both `@appium/support` and `appium` are fully typechecked (sans implicit `any` types), but `@appium/base-driver` is mostly not.  Regardless, declarations are generated for all three.

    Each package will have its own `tsconfig.json` which "inherits" from `config/tsconfig.base.json`.  Each will also need to declare relative paths to dependencies. e.g., the `appium` package must configure its `tsconfig.json` so that it depends on the declarations of `@appium/support` and also where to find `@appium/support`. Essentially, a small dependency tree gets built, and those at the "root" get their declarations built first; then the consumers use those declarations.


boneskull:
* 2.0
  * **PR #16594** <a href="https://app.graphite.dev/github/pr/appium/appium/16594" target="_blank">Graphite</a>  (this PR)

packages/types/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@appium/types",
+  "version": "0.0.0",
```
jlipps, line 3: this will actually be published to NPM, correct? should we start at 0.0.1?

jlipps (approved):

packages/types/package.json
boneskull: I figured we could just update that when publishing.

boneskull:
Build failure here is expected.

packages/types/index.d.ts
```diff
@@ -0,0 +1,610 @@
+import { Capabilities } from '@wdio/types';
```
jlipps, line 1: is this file hand-rolled or auto-generated?

jlipps (approved):

packages/types/index.d.ts
jlipps: it's nice that we can reuse some of the work wdio has done, but internally we really don't care. internally it should just be:

```ts
{[key: string]: any}
```

maybe i've missed it and it's there somewhere already.

what if we add a new capability that wdio doesn't know about yet? what if a third party driver author adds a capability that wdio will never add to their list?

jlipps (changes requested):

packages/types/index.d.ts
boneskull: Then I'd say we should provide somewhere to define those things that do not occur in wdio's capabilities.

But in general, more specificity is better.  Any _optional_ properties should be typed, because they won't prevent someone from providing the wrong type, otherwise.  Further, you get inline documentation on the properties, where provided.

So I guess I'm in disagreement about using `any` here.  Even if we did avoid reusing wdio's types, there's still stuff that is absolutely required that both consumers and contributors should be aware of (e.g. W3C caps have a certain shape, `platformName` is required, etc.).  It may be that at some point in the control flow we don't really care any more, but it matters for public APIs.

I think we can reuse the W3C caps in a way which gives us more flexibility.  First, we can redefine what, e.g., `AppiumCapabiltiies` are by doing something like:

```ts
// untested
import type {AppiumCapabilities as BaseAppiumCapabilities} from '@wdio/types/capabilities';

interface OtherCapabilities {
 // whatever we need to add
 'appium:foo': string
}

export type AppiumCapabilities = BaseAppiumCapabilities & OtherCapabilities;
```

We can also use a generic to allow drivers to optionally specify their own capabilities:

```ts
export interface Driver<C extends AppiumCapabilities = AppiumCapabilities> {
  validateDesiredCaps(caps: C): boolean;
  // etc
}
```

An implementing class could then do something like:

```ts
interface MyCaps extends AppiumCapabilities {
  'stuff': 'things'
}

class MyMainClassDriver extends BaseDriver implements Driver<MyCaps> {

}
```

Of course, minding the types is completely optional.

What do you think?

jlipps: Maybe we should put some effort into defining our own capability types then. I'm just not confident that WDIO's current ones are correct or will remain so even if they are.

Another issue to consider is that when we deal with capabilities internally, they don't have `appium:` in front, because this is stripped out in capabilities processing, but when users send capabilities to an Appium server (in which case they probably shouldn't be interfacing with these type definitions because they'll be using a client anyway, like WDIO, and consuming that client's type interface instead of ours), they do.

If we want to take this on, we'll need to hunt through the code to find the following types of capabilities mentioned so we can do the right thing with them:

- capabilities used by BaseDriver
- capabilities used by AppiumDriver
- capabilities used by a platform driver (which should be moved to a type definition in the platform driver instead)

And yes, in our internal code we'll always need to assume that the capabilities object could have other random things that the user has sent in. There are probably very few capabilities that we can definitely assume will always be present (I think platformName, automationName might be the only ones we can universally assume).

boneskull: I've come up with something I think will be workable here.  `Capabilities` for internal use, and `W3CCapabilities` for external use.  We build these types off of wdio's types, but they are very loose, and we are free to extend them as we wish.  This provides autocompletion and _only_ errors if the type is actually _wrong_ (e.g., a `number` when a `string` was expected).  `W3CCapabilities` must correspond (roughly) to the standard.

boneskull:
(@jlipps 's review still says "changes requested" but this was approved otherwise via slack msg)

boneskull:
will merge these four PRs into 2.0 imminently (once build passes)

jlipps:
\o/ woohoo!

PR #16596, opened 2022-03-16
feat(base-driver): generate declaration files

boneskull:
This was not as easy as the other two.  The main issue involves the `BaseDriver` class and is threefold:

1. `BaseDriver` and a valid external subclass thereof (a `Driver`) are not the same, because the former is essentially _abstract_. There is no notion of an abstract class in JS nor in TS for JS.
2. Splitting `BaseDriver` into several modules and "mixing in" some modules is just not well-understood by TS.
3. `AppiumDriver` inherits _most_ of `BaseDriver`, but overrides some functions with differing signatures. This is unsupported by TS.

To work around these problems, I had to:

1. Move the `Driver` interface into `@appium/types`, since other packages may depend on it.  We do not want circular dependencies, which is likely to occur otherwise.  Further, there is no way to tell TS for JS that a class _implements_ an interface or type.
2. Refactor the mixins (in `basedriver/commands/`) into "mixin classes", as understood by typescript. These are dynamic classes returned by functions which accept a class parameter, and _extend_ that class. The result is marginally better, but the implementation is ... well, see for yourself.  It is not possible to use declaration merging to just "trick" TS into thinking that `BaseDriver` has all of these extra methods; classes cannot be merged, and even if you could, you can't use declaration merging to "add" methods afaict.
3. `BaseDriver` is now just a class in a variable after the mixins have been applied. Mixins are applied to `BaseDriverCore`, and `BaseDriverCore` inherits from `DriverCore`. `AppiumDriver` _also_ inherits from `DriverCore`, since it does not override `BaseDriver` "properly".  `BaseDriverCore` itself is pretty small, containing only those functions which `AppiumDriver` defines as differing.

Otherwise:

- `BaseDriver` does not inherit from `Protocol`, which is just an empty class. `Protocol` was removed.
- All of the error classes are exported from `protocol/errors`, because TS cannot see them otherwise.
- Exported missing `GET_STATUS_COMMAND` from `protocol/index`
- Re-exported `normalizeBasePath` from `server` instead of `protocol`, since that's where it's defined
- Removed `index.d.ts`
-


boneskull:
* 2.0
      * **PR #16596** <a href="https://app.graphite.dev/github/pr/appium/appium/16596" target="_blank">Graphite</a>  (this PR)

packages/base-driver/lib/basedriver/commands/find.js
```diff
@@ -1,51 +1,108 @@
...
-    return await this.findElOrEls(strategy, selector, mult, context);
-  } catch (err) {
-    if (this.opts.printPageSourceOnFindFailure) {
-      const src = await this.getPageSource();
-      this.log.debug(`Error finding element${mult ? 's' : ''}: ${err.message}`);
-      this.log.debug(`Page source requested through 'printPageSourceOnFindFailure':`);
-      this.log.debug(src);
+/**
+ *
+ * @param {ReturnType<import('./event').EventMixin>} Base
```
jlipps, line 8: oh wow so we have to apply these in order?

packages/base-driver/lib/basedriver/commands/find.js
```diff
@@ -1,51 +1,108 @@
...
+ * @returns {import('../driver').BaseDriverBase<import('@appium/types').TimeoutCommands & import('@appium/types').EventCommands & import('@appium/types').FindCommands>}
```
jlipps, line 9: this is awful

jlipps: can't we alias this type variable somewhere so it's not so ugly?

packages/base-driver/lib/basedriver/commands/find.js
```diff
@@ -1,51 +1,108 @@
...
+    }
+    // Override the following function for your own driver, and the rest is taken
+    // care of!
+    // Returns an object which adheres to the way the JSON Wire Protocol represents elements:
+    // { ELEMENT: # }    eg: { ELEMENT: 3 }  or { ELEMENT: 1.023 }
+    /**
+     * @template {boolean} Mult
+     * @param {string} strategy
+     * @param {string} selector
+     * @param {Mult} mult
```
jlipps, line 62: mult *should* just be a boolean (whether or not we are looking for multiple elements)

packages/base-driver/lib/basedriver/commands/find.js
```diff
@@ -1,51 +1,108 @@
...
+     * @param {string} [context]
+     * @returns {Promise<Mult extends true ? Element[] : Element>}
```
jlipps, line 64: oh I see, yeah it's a tricky return type

packages/base-driver/lib/basedriver/commands/find.js
boneskull: Sort of.  It'd be one thing if these were _pure_ mixins, but they aren't.  They don't work standalone; they call out to other methods in `DriverCore` or maybe even other mixin classes.  You can't just apply them to any given class.  

It sucks, but I couldn't find another way to do this without making things so generic we don't know what's in the final class.  TS doesn't handle our use case very well.

If none of these actually call out to other mixins (I can't recall), then the order itself isn't important, but the types still are.  It's rather arbitrary right now.

packages/base-driver/lib/basedriver/commands/find.js
boneskull: We can create a bunch of typedefs.  In TS this would be a bunch of `import` statements.

packages/base-driver/lib/basedriver/commands/find.js
boneskull: Yeah.  I think I hate it `findElOrEls`, but too late now.

boneskull:
Failure is expected; #16597 should be the build to pass.

packages/base-driver/lib/basedriver/commands/find.js
boneskull: another avenue I explored was just trying to trick TS into thinking that `BaseDriver` had a bunch of methods that were implemented elsewhere.  declaration merging (e.g., defining an `interface BaseDriver`) didn't work because TS generated broken declarations. module augmentation (`declare module './driver' { }`) didn't work because you cannot "add" stuff to a typed class.  so idk.  I was pulling my hair out.

all of this goes away, of course, if we just make a huge class.

boneskull: also, nothing in `DriverCore` or `BaseDriverCore` is allowed to call out to a method defined in any of these mixins.  that--and the discrepancies introduced in `AppiumDriver` was why I had to move stuff around.

boneskull: if typescript understood prototypes at all (it doesn't), I think this would be way easier.  but it's fundamentally based on classical OO and inheritance and all that garbage.  it makes some things difficult, and we found one of them.

```ts
class Foo {}

Foo.prototype = {bar: () => 'baz'};

let foo = new Foo();
foo.bar(); // TS238974239 error wtf no such method `bar`
```

packages/base-driver/lib/basedriver/commands/session.js
```diff
@@ -1,153 +1,51 @@
...
-import { errors } from '../../protocol';
-import { util } from '@appium/support';
-import {
-  processCapabilities, promoteAppiumOptions, APPIUM_OPTS_CAP, PREFIXED_APPIUM_OPTS_CAP,
-} from '../capabilities';
-import { isW3cCaps } from '../../helpers/capabilities';
 
-const commands = {};
-
-commands.createSession = async function createSession (w3cCapabilities1, w3cCapabilities2, w3cCapabilities) {
```
jlipps, line 12: where did createSession, deleteSession, et al go?

packages/base-driver/lib/basedriver/driver.js
```diff
@@ -382,116 +115,267 @@ class BaseDriver extends Protocol {
...
+     * @param {W3CCapabilities} w3cCapabilities1
+     * @param {W3CCapabilities} [w3cCapabilities2]
+     * @param {W3CCapabilities} [w3cCapabilities]
+     * @param {DriverData[]} [driverData]
+     * @returns {Promise<[string,object]>}
+     */
+  async createSession (
+    w3cCapabilities1,
+    w3cCapabilities2,
+    w3cCapabilities,
```
jlipps, line 225: is it a bit strange this one doesn't have a number? or that if it doesn't, that it's not first in the list?

packages/base-driver/lib/basedriver/driver.js
```diff
@@ -382,116 +115,267 @@ class BaseDriver extends Protocol {
...
-      }
-      if (!_.isRegExp(avoidPathRegex)) {
-        throw new Error('Proxy avoidance path must be a regular expression');
-      }
-      let normalizedUrl = url.replace(new RegExp(`^${_.escapeRegExp(this.basePath)}`), '');
-      if (avoidMethod === method && avoidPathRegex.test(normalizedUrl)) {
-        return true;
+    this.originalCaps = _.cloneDeep(originalCaps);
+    this.log.debug(
+         `Creating session with W3C capabilities: ${JSON.stringify(
```
jlipps, line 252: too much indentation

packages/base-driver/lib/basedriver/driver.js
```diff
@@ -382,116 +115,267 @@ class BaseDriver extends Protocol {
...
+     * Appium 2 has dropped the support of these, so now we only accept capability
+     * objects in W3C format and thus allow any of the three arguments to represent
+     * the latter.
+     * @param {W3CCapabilities} w3cCapabilities1
+     * @param {W3CCapabilities} [w3cCapabilities2]
+     * @param {W3CCapabilities} [w3cCapabilities]
+     * @param {DriverData[]} [driverData]
+     * @returns {Promise<[string,object]>}
+     */
+  async createSession (
```
jlipps, line 222: ah, here's createSession! i see where you put them now

packages/base-driver/lib/helpers/capabilities.js
```diff
@@ -20,6 +20,50 @@ function isW3cCaps (caps) {
   return false;
 }
 
+function fixCaps (originalCaps, desiredCapConstraints = {}, log) {
```
jlipps, line 23: no type definitions in this function? is there a reason we're not ts-checking this file?

jlipps (approved):

packages/base-driver/lib/basedriver/commands/find.js
jlipps: it's my fault, i guess i should have made driver authors implement both findEl and findEls instead /shrug

packages/base-driver/lib/basedriver/commands/find.js
jlipps: yeah, but imports are on their own up at the top, this is "in the middle" of code. it's ugly. typedefs are probably prettier.

jlipps (approved):

PR #16597, opened 2022-03-16
feat(appium): generate declaration files

boneskull:
Configure TS to generate declaration files for `appium` package.

Most of these changes are just types getting added or changed via docstrings; in other places, typechecking has caught problems and I had to refactor code to address them.

- Removed `// @ts-check` pragma from all files in `lib/`, since now `checkJs: true` is in `tsconfig.json`.  It needs to stay in place elsewhere (I think) because only sources in `lib` have declarations generated.
- The `install`, `uninstall`, `update` and `run` functions in the base `ExtensionCommand` class were renamed and prefixed with a `_`; TS does not allow overridden functions with differing signatures.   It no longer exports a default.
- Removed `type` prop of options for `DriverCommand` and `PluginCommand` constructors, respectively (this value is derived from the `ExtensionConfig` `config` parameter instead)
- `AppiumDriver` now extends `DriverCore` instead of `BaseDriver`, since some of its function signatures differ from `BaseDriver`'s.
- Small refactor of `deleteSession` to make typechecker happy.
- `executeCommand` now calls to `BaseDriver` directly if `isUmbrellaCmd` is truthy.  This is one place where inheritance just kind of breaks, since `BaseDriver#executeCommand` returns something entirely different than `AppiumDriver#executeCommand` would otherwise.
- Removed `--delete-dir-on-start` from build command, since it may blast generated declarations.


boneskull:
* 2.0
        * **PR #16597** <a href="https://app.graphite.dev/github/pr/appium/appium/16597" target="_blank">Graphite</a>  (this PR)

packages/appium/lib/cli/extension-command.js
```diff
@@ -307,27 +313,31 @@ export default class ExtensionCommand {
     const {appium, name, version} = pkgJsonData;
     this.validateExtensionFields(appium);
 
-    return {...appium, pkgName: name, version};
+    // we need to omit `driverName` and `pluginName`. `_.omit()` does the job, but it isn't really
+    // type-safe.  so we're fudging it here.
+    delete appium[/** @type {string} */(`${this.type}Name`)];
```
jlipps, line 318: can ts not detect that `${this.type}Name` is a string?

packages/appium/lib/cli/extension-command.js
```diff
@@ -307,27 +313,31 @@ export default class ExtensionCommand {
...
+    return /** @type {ExtensionFields<ExtType>} */(/** @type {unknown} */({...appium, pkgName: name, version}));
```
jlipps, line 319: this one's a mind bender

packages/appium/lib/cli/extension.js
```diff
@@ -30,7 +35,7 @@ async function runExtensionCommand (args, configObject) {
   const logFn = (msg) => log(json, msg);
   let config = configObject;
   config.log = logFn;
-  const CommandClass = type === DRIVER_TYPE ? DriverCommand : PluginCommand;
+  const CommandClass = /** @type {ExtType extends import('../extension/manifest').DriverType ? import('@appium/types').Class<DriverCommand> : import('@appium/types').Class<PluginCommand>} */(commandClasses[type]);
```
jlipps, line 38: can we put this in a typedef or something so it doesn't impact readability so much?

packages/appium/lib/main.js
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 
 // transpile:main
```
jlipps, line 3: btw, with the removal of gulp, i think we can get rid of all of these `//transpile` comments?

packages/appium/lib/utils.js
```diff
@@ -1,4 +1,3 @@
-// @ts-check
```
jlipps, line 1: what is the consequence of removing `// @ts-check` from files?

packages/appium/types/cli.d.ts
```diff
@@ -0,0 +1,137 @@
...
+  PLUGIN_TYPE as PLUGIN_SUBCOMMAND,
+  SERVER_SUBCOMMAND,
+} from '../lib/constants';
+import {DriverOpts} from '@appium/types';
+
+/**
+ * Possible subcommands for the `appium` CLI.
+ */
+type CliSubcommand =
+  | typeof SERVER_SUBCOMMAND
```
jlipps, line 17: weird, you can put a `|` as a no-op prefix operator?

packages/appium/types/cli.d.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  /**
+   * Subcommands of `plugin` subcommand
+   */
+  pluginCommand?: CliExtensionSubcommand;
+
+  /**
+   * Possible subcommands
+   */
+  subcommand:
```
jlipps, line 104: how is this different from `CliSubcommand`?

boneskull:
this is failing b/c I changed something I thought was safe to change, but it wasn't.

packages/appium/lib/appium.js
```diff
@@ -21,48 +25,78 @@ const desiredCapabilityConstraints = {
...
+   * map of sessions to actual plugin instances per session
+   * @type {Record<string,InstanceType<import('../types').Extension.PluginClass>[]>}
+   */
+  sessionPlugins = {};
+
+  /**
+   * some commands are sessionless, so we need a set of plugins for them
+   * @type {InstanceType<import('../types').Extension.PluginClass>[]}
+   */
+  sessionlessPlugins = [];;
```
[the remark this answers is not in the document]
boneskull: for some definition of 'unnecessary' ;)

packages/appium/lib/cli/extension-command.js
boneskull: it can now, apparently.

boneskull: pretty sure there was a bug with string template types

boneskull: actually, nevermind. no, it doesn't work. no, I don't know why.

packages/appium/lib/cli/extension-command.js
boneskull: yeah. like, you can't just `delete` prop `foo` from `baz` of type `Bar` if `foo` is not optional.  var `baz` remains `Baz`, which is no longer true?  I don't know.  even a bare type casting didn't work, so you first cast to `unknown` then to whatever you want.

I suppose I could rewrite this by explicitly pulling out each prop we want from `appium` but that seems really tedious.

packages/appium/lib/utils.js
boneskull: `// @ts-check` tells typescript "check this file" regardless if it is configured to check the file.  previously, typescript was _not_ configured to typecheck _any_ javascript, and we enabled it on a per-file basis instead.  now that is no longer true--at least for the `appium` package--as we've added `checkJs: true` to its `tsconfig.json`.  thus the `// @ts-check`s are redundant.

packages/appium/lib/cli/extension-command.js
jlipps: i'm kinda less concerned about type gymnastics here than the fact that I can't even see what is returned on one line because of the type comment.

packages/appium/types/cli.d.ts
boneskull: It's not. good catch

packages/appium/lib/cli/extension-command.js
```diff
@@ -237,6 +242,9 @@ export default class ExtensionCommand {
       extData = await this.installViaNpm({ext, pkgName, pkgVer});
     }
 
+    /**
```
jlipps, line 245: jeez does this have to be 3 lines?

jlipps (approved):

jlipps (approved):

packages/appium/lib/cli/extension-command.js
boneskull: it doesn't need to exist at all

packages/appium/lib/cli/extension-command.js
boneskull: I'll make it more legible

PR #16605, opened 2022-03-18
chore(fake-driver): gulp is dead, long live scripts

boneskull:
This PR removes Gulp from `@appium/fake-driver` as already done in a couple other packages.

Also: moved `screen.png` into `lib/` since `@babel/cli` can just copy it over into `build/`.


jlipps (approved):
looks like a package.json conflict but otherwise LGTM

PR #16607, opened 2022-03-18
chore(base-driver): ungulpify

boneskull:
This PR removes Gulp from `@appium/base-driver` and reorganizes the test files.

The build needs to be a little different since other modules depend on its `build/test/basedriver` module.  IMO this should probably get moved elsewhere, but this does not break compatibility.


packages/base-driver/test/basedriver/README.md
```diff
@@ -0,0 +1,5 @@
+# Driver Test Helpers
```
jlipps, line 1: 👍

jlipps (approved):

PR #16636, opened 2022-03-23
fix(base-driver): Use WeakRef to reference the driver instance in the log prefix generator

mykola-mokhnach:
## Types of changes


jlipps (approved):
looks good. also fwiw semantic commits should have a scope for intelligent releasing, e.g. `fix(appium):` instead of just `fix:`.

KazuCocoa (approved):

PR #16643, opened 2022-03-23
fix(support): add missing resolve-from dependency

boneskull:
see https://github.com/appium/java-client/issues/1667

### DIFF
--- a/packages/support/package.json
+++ b/packages/support/package.json
@@ -66,6 +66,7 @@
     "pluralize": "8.0.0",
     "pngjs": "6.0.0",
     "read-pkg": "5.2.0",
+    "resolve-from": "5.0.0",
     "rimraf": "3.0.2",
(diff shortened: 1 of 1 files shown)


boneskull:
so, yeah, we need some smoke tests.  essentially this would be a "test" which runs on the package as it would be installed by the user (no dev deps). it surfaces missing dependencies, among other things.

I've had this sort of problem before... others have tried to come up with a solution. I'll ask around if they came up with anything.  a naive implementation would simply call `npm pack` in each package to create a tarball, then chdir into a temp directory and `npm install /path/to/tarball`.  finally, it would run a basic test (without use of a test framework), e.g., just `require()` the package and/or assert the exports are correct.

boneskull:
I'm going to merge this, using my exquisite judgement

jlipps:
agree that some kind of smoke test like that would be nice as part of our CI

PR #16670, opened 2022-03-27
fix(base-driver): Make sure we never mutate incoming args

mykola-mokhnach:
## Types of changes

### DIFF
--- a/packages/base-driver/lib/basedriver/core.js
+++ b/packages/base-driver/lib/basedriver/core.js
@@ -155,7 +155,7 @@ class DriverCore {
      if (!this._log) {
        const instanceName = `${this.constructor.name}@${node
          .getObjectId(this)
-         .substring(0, 8)}`;
+         .substring(0, 4)}`;
        this._log = logger.getLogger(() =>
          this.sessionId
            ? `${instanceName} (${this.sessionId.substring(0, 8)})`
--- a/packages/base-driver/lib/basedriver/driver.js
+++ b/packages/base-driver/lib/basedriver/driver.js
@@ -233,11 +233,11 @@ export class BaseDriverCore extends DriverCore {
 
     this.log.debug();
 
-    const originalCaps = [
+    const originalCaps = _.cloneDeep([
       w3cCapabilities,
       w3cCapabilities1,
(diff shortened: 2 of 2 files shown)


KazuCocoa (approved):

boneskull (approved):
This seems fine to me, but I'm curious:

1. Why do we mutate the value?
2. What bug does this cause?

mykola-mokhnach:
> This seems fine to me, but I'm curious:
> 
> 1. Why do we mutate the value?

We should not. This is what the fix does

> 2. What bug does this cause?

It should not be a real deal when Appium is running as http-server, but internal tests are failing nevertheless:

https://github.com/appium/appium-android-driver/pull/735
https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=20424&view=logs&j=d4d48b56-56f2-5159-6052-c4831402bb90&t=89e5db65-2b81-5ecf-de1e-097733dadfbd

mykola-mokhnach:
Thanks for the review @boneskull 
I'm still not quite sure what to do with `core.js` file, which is a duplicate of what exists in `driver.js`. Which of them is a true one?

boneskull:
@mykola-mokhnach `core.js` and `driver.js` are different.  you might want to try to merge again...

`core.js` contains `DriverCore` which is now the base class for `BaseDriver` (initially class `BaseDriverCore`) and `AppiumDriver`.  `BaseDriverCore` contains stuff that either a) `AppiumDriver` doesn't need or b) conflicts with `AppiumDriver`.

boneskull:
@mykola-mokhnach 

> Why do we mutate the value?
> > We should not. This is what the fix does

Errr, yes, but rather I mean: why was this happening in the first place?  Where exactly is the object mutated, and why?

mykola-mokhnach:
> @mykola-mokhnach
> 
> > Why do we mutate the value?
> > > We should not. This is what the fix does
> 
> Errr, yes, but rather I mean: why was this happening in the first place? Where exactly is the object mutated, and why?

It looks like its mutated in one of helper methods that strip out capability prefixes inside of createSession

PR #16681, opened 2022-03-28
fix installation of multiple extensions

### LINKED ISSUE #16674: (v2) appium install driver deletes previous installed driver(s) when installing multiple drivers.
## The problem

I installed appium 2.0.0-beta.27 via npm, 
first installed uiautomator2 driver - run test with uiautomator2 -> all ok
then installed espresso driver - run test with uiautomator2 again -> error:
```
Could not start a new session. Response code 500. Message: An unknown server-side error occurred while processing the command. Original error: Could not find a driver for automationName 'UIAutomator2' and platformName android'. Have you installed a driver that supports those capabilities? Run 'appium driver list --installed' to see. (Lower-level error: Cannot find module '/home/cmaganu/.appium/node_modules/appium-uiautomator2-driver
```

## Environment

* Appium version (or git revision) that exhibits the issue: 2.0.0-beta.27
* Last Appium version that did not exhibit the issue (if applicable): -
* Desktop OS/version used to run Appium: Ubunntu 20.10
* Node.js version (unless using Appium.app|exe): v16.14.2
* Npm or Yarn package manager: npm 8.5.5
* Mobile platform/version under test: Android 10
* Real device or emulator/simulator: emulator
* Appium CLI or Appium.app|exe: cli

## Details
- after `appium driver install uiautomator2`:
```
cmaganu@cma-x1:~/git/lt-appium-mobile-tests(main)$ ls -Al ~/.appium/node_modules/
total 128
drwxrwxr-x 5 cmaganu cmaganu   4096 mar 28 16:45 appium-uiautomator2-driver
drwxrwxr-x 3 cmaganu cmaganu   4096 mar 28 16:45 .cache
-rw-rw-r-- 1 cmaganu cmaganu 121877 mar 28 16:45 .package-lock.json
```
```
cmaganu@cma-x1:~/git/lt-appium-mobile-tests(main)$ cat  ~/.appium/node_modules/.cache/appium/extensions.yaml 
drivers:
  uiautomator2:
    automationName: UiAutomator2
    platformNames:
      - Android
    mainClass: AndroidUiautomator2Driver
    pkgName: appium-uiautomator2-driver
    version: 2.0.5
    installType: npm
    installSpec: uiautomator2
plugins: {}
schemaRev: 2
```
```
cmaganu@cma-x1:~/git/lt-appium-mobile-tests(main)$ appium driver list --installed
✔ Listing installed drivers
- uiautomator2@2.0.5 [installed (NPM)]
```
- after `appium driver install espresso`:
```
cmaganu@cma-x1:~/git/lt-appium-mobile-tests(main)$ appium driver list --installed
✔ Listing installed drivers
- uiautomator2@2.0.5 [installed (NPM)]
- espresso@2.2.0 [installed (NPM)]
```
```
cmaganu@cma-x1:~/git/lt-appium-mobile-tests(main)$ ls -Al ~/.appium/node_modules/
total 708
drwxrwxr-x 6 cmaganu cmaganu   4096 mar 28 16:49 appium-espresso-driver
drwxrwxr-x 3 cmaganu cmaganu   4096 mar 28 16:45 .cache
-rw-rw-r-- 1 cmaganu cmaganu 714864 mar 28 16:49 .package-lock.json
```
```
cmaganu@cma-x1:~/git/lt-appium-mobile-tests(main)$ cat  ~/.appium/node_modules/.cache/appium/extensions.yaml 
drivers:
  uiautomator2:
    automationName: UiAutomator2
    platformNames:
      - Android
    mainClass: AndroidUiautomator2Driver
    pkgName: appium-uiautomator2-driver
    version: 2.0.5
    installType: npm
    installSpec: uiautomator2
  espresso:
    automationName: Espresso
    platformNames:
      - Android
    mainClass: EspressoDriver
    pkgName: appium-espresso-driver
    version: 2.2.0
    installType: npm
    installSpec: espresso
plugins: {}
schemaRev: 2
```
`.appium/package.json` is empty

boneskull:
Closes #16674

the problem here was the existence of the empty `package.json`.  it should exist, because if it doesn't, then `npm` may end up using a `package.json` in a parent directory.  but it can't be _empty_, because subsequent `npm install`s will essentially remove any package that it's currently not installing!  `npm` likes to "fix" `node_modules`, which is undesirable here.

to avoid this, the behavior has changed to persisting the dependency version to `package.json`.  behavior differs if we're using a local `APPIUM_HOME` or the default one.  with the default, we have more control, and choose to use `--save-dev --save-exact --no-package-lock --global-style` which gives us what we need.  if `APPIUM_HOME` is a local package, all we can do is `--save-dev` (since it's _likely_ that an extension will be a dev dependency).  in the future, we may choose to actually determine _what kind_ of dependency `appium` is and tell `npm` to use the same one.  not sure if that's needed, though.

further, in the case of a local appium dep (local `APPIUM_HOME`), updating `package.json` should cause the package hash (used by the thing that syncs `extensions.yaml` with installed packages) to update as well, so we're doing that.

modified the logic in `ExtensionCommand` to treat `--source=local` to pass the `installSpec` thru ~~verbatim~~ as an absolute path

also fixed `BuildInfo` type to reflect reality and use `@types/teen_process`.  fixed some other broken types in `test` (files in `test` are not checked by typescript yet; we'll need a separate `tsconfig.json` for this, since we do not want to emit declarations)

* * *

@jlipps I'm unclear on the intent of avoiding writing to the "dummy" `package.json` (the presence of `--no-save` in the `npm install` command, which I needed to remove).  If we have to keep that behavior, then I need to go back to the drawing board.  Unfortunately, the result will be more complex than it already is, as I'll need to revert "do not create an empty directory for each extension", and keep that behavior separate from how it behaves when run in a dir with an `appium` dep.  I am not sure how to tell `npm` "do not look for a parent `package.json` above `cwd`", or if that's even possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "clean:distfiles": "npx rimraf \"packages/*/build\"",
     "clean:monorepo": "npx rimraf node_modules package-lock.json",
     "clean:packages": "lerna clean -y && npx rimraf \"packages/*/package-lock.json\"",
-    "clean:distfiles": "npx rimraf \"packages/*/build\"",
     "clean:types": "tsc -b --clean",
     "dev": "lerna run --parallel --prefix --concurrency=8 dev",
     "dev:types": "tsc -b --watch",
--- a/package.json
+++ b/package.json
@@ -138,6 +138,7 @@
     "@types/shell-quote": "1.7.1",
     "@types/sinon": "10.0.11",
     "@types/sinon-chai": "3.2.8",
+    "@types/teen_process": "1.16.0",
     "@types/through2": "2.0.36",
     "@types/uuid": "8.3.4",
     "@wdio/types": "7.19.1",
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -1,11 +1,13 @@
 /* eslint-disable no-console */
 
 import _ from 'lodash';
-import { npm, fs, util } from '@appium/support';
+import path from 'path';
+import { npm, fs, util, env } from '@appium/support';
 import { log, spinWith, RingBuffer } from './utils';
 import { SubProcess } from 'teen_process';
 import { INSTALL_TYPE_NPM, INSTALL_TYPE_GIT, INSTALL_TYPE_GITHUB,
          INSTALL_TYPE_LOCAL } from '../extension/extension-config';
+import { packageDidChange } from '../extension/package-changed';
 
 const UPDATE_ALL = 'installed';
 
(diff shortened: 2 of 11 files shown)


boneskull:
build fails because now local source paths need to be converted as relative to `process.cwd()`.

boneskull:
Also note the changes in `@appium/support`: we are no longer using `npm link` for anything because it's bad

mykola-mokhnach (approved):

boneskull:
going to wait on @jlipps to merge this

jlipps (approved):

PR #16683, opened 2022-03-29
fix(base-driver): Update/simplify the logic for logger prefix

mykola-mokhnach:
## Proposed changes

This restores https://github.com/appium/appium/pull/16636 with some slight changes to make ts linter happy

### DIFF
--- a/packages/base-driver/lib/basedriver/core.js
+++ b/packages/base-driver/lib/basedriver/core.js
@@ -102,9 +102,9 @@ class DriverCore {
   eventEmitter = new EventEmitter();
 
   /**
-  * @type {AppiumLogger?}
+  * @type {AppiumLogger}
   */
-  _log = null;
+  _log = logger.getLogger(helpers.generateDriverLogPrefix(this));
 
   /**
   * @protected
@@ -152,29 +152,6 @@ class DriverCore {
   }
 
   get log () {
-    if (!this._log) {
-      const instanceName = `${this.constructor.name}@${node.getObjectId(this).substring(0, 4)}`;
-      // We don't want the self reference to be captured inside the below closure
-      // to avoid possible memory leaks,
-      // but NodeJS started supporting WeakRef only since v. 14.6
-      /**
-       * @type {WeakRef<DriverCore>|DriverCore}
-       */
-      const self = global.WeakRef ? new global.WeakRef(this) : this;
-      this._log = logger.getLogger(() => {
-        let sessionId;
-        if (self instanceof DriverCore) {
-          sessionId = self.sessionId;
-        } else if (global.WeakRef && self instanceof global.WeakRef) {
-          const ref = self.deref();
-          if (ref?.sessionId) {
-            sessionId = ref.sessionId;
-          }
-        }
-        return sessionId ? `${instanceName} (${sessionId.substring(0, 8)})` : instanceName;
-      });
-    }
-
     return this._log;
   }
 
--- a/packages/base-driver/lib/basedriver/core.js
+++ b/packages/base-driver/lib/basedriver/core.js
@@ -153,14 +153,26 @@ class DriverCore {
 
   get log () {
     if (!this._log) {
-      const instanceName = `${this.constructor.name}@${node
-        .getObjectId(this)
-        .substring(0, 4)}`;
-      this._log = logger.getLogger(() =>
-        this.sessionId
-          ? `${instanceName} (${this.sessionId.substring(0, 8)})`
-          : instanceName,
-      );
+      const instanceName = `${this.constructor.name}@${node.getObjectId(this).substring(0, 4)}`;
+      // We don't want the self reference to be captured inside the below closure
+      // to avoid possible memory leaks,
+      // but NodeJS started supporting WeakRef only since v. 14.6
+      /**
+       * @type {WeakRef<DriverCore>|DriverCore}
+       */
+      const self = global.WeakRef ? new global.WeakRef(this) : this;
+      this._log = logger.getLogger(() => {
+        let sessionId;
+        if (self instanceof DriverCore) {
+          sessionId = self.sessionId;
+        } else if (global.WeakRef && self instanceof global.WeakRef) {
+          const ref = self.deref();
+          if (ref?.sessionId) {
+            sessionId = ref.sessionId;
+          }
+        }
+        return sessionId ? `${instanceName} (${sessionId.substring(0, 8)})` : instanceName;
+      });
     }
 
     return this._log;
--- a/packages/base-driver/lib/basedriver/core.js
+++ b/packages/base-driver/lib/basedriver/core.js
@@ -104,7 +104,7 @@ class DriverCore {
   /**
   * @type {AppiumLogger}
   */
-  _log = logger.getLogger(helpers.generateDriverLogPrefix(this));
+  _log;
 
   /**
   * @protected
@@ -134,6 +134,8 @@ class DriverCore {
     opts = /** @type {DriverOpts} */ ({}),
     shouldValidateCaps = true,
   ) {
+    this._log = logger.getLogger(helpers.generateDriverLogPrefix(this));
+
     // setup state
     this.opts = opts;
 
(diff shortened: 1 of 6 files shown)


boneskull:
Can you explain what this avoids?

mykola-mokhnach:
This avoid possible memory leak in case the logger object itself is preserved somewhere. Without WeakRef the actual driver instance would be preserved with it as it is captured in the corresponding closure.

packages/base-driver/lib/basedriver/core.js
```diff
@@ -153,14 +153,26 @@ class DriverCore {
...
+        if (self instanceof DriverCore) {
+          sessionId = self.sessionId;
+        } else if (global.WeakRef && self instanceof global.WeakRef) {
+          const ref = self.deref();
+          if (ref?.sessionId) {
+            sessionId = ref.sessionId;
+          }
+        }
+        return sessionId ? `${instanceName} (${sessionId.substring(0, 8)})` : instanceName;
+      });
```
boneskull, line 175: I don't think `WeakRef` is necessary here.  In fact, we should probably [avoid use](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef#avoid_where_possible) unless there is no other possible way to do so.

What we're trying to do here is hand a prefix to the logger.  Currently, we're passing a function, but it can just be a string.  This prefix may be dynamic, based upon the current `sessionId` (if any).

What if, upon session creation/destruction (in `createSession` and `deleteSession` of `BaseDriverCore`), we did something like this:

```js
// this would also probably need to be in AppiumDriver's createSession(), which is *not the same* function
async createSession(stuff) {
  this.sessionId = getNewSessionId();
  const logPrefix = createLogPrefix() // does the instanceName stuff w/ the ternary operator, etc.

  // we could do this:
  this._log = logger.getLogger(logPrefix);

  // OR refactor the logger to expose a `setPrefix` method:
  this._log.setPrefix(logPrefix);
}
```

We could create a setter for `sessionId` which did the log-prefix dance as well.

This would only create a potential leak if a driver implementation held on to the old value `this._log` for too long.  But a driver could create a potential leak in myriad ways which we cannot prevent.

boneskull (changes requested):
I don't think we need `WeakRef` here... please see comment.

packages/base-driver/lib/basedriver/core.js
boneskull: If we do this, we should also add a test that shows the prefix gets updated.

mykola-mokhnach: changed to add/remove the prefix in create/deleteSession

mykola-mokhnach:
I have updated the implementation, but now I have no idea about how to update unit tests after _log object gets reassigned. I also don't have a good idea on how to make prefix mutable on the logger itself

boneskull:
@mykola-mokhnach Can I push to your branch?

mykola-mokhnach:
> @mykola-mokhnach Can I push to your branch?

sure, not a problem at all

mykola-mokhnach:
Seems like I got some progress as of now. Please review once more @boneskull

jlipps (approved):

packages/types/src/index.ts
```diff
@@ -365,6 +365,7 @@ export interface DriverHelpers {
   isPackageOrBundle: (app: string) => boolean;
   duplicateKeys: <T>(input: T, firstKey: string, secondKey: string) => T;
   parseCapsArray: (cap: string | string[]) => string[];
+  generateDriverLogPrefix: (obj: Object, sessionId?: string) => string;
```
boneskull, line 368: This is weird, and I had to look it up.

`Object` allows primitives, but not `null` nor `undefined`.  While that may actually work, it's not the intent of this function.  You probably mean `object`, which means an "actual" object.  

And again, while that might actually work, the only thing we really expect `obj` to be is an object implementing `Core`.  So I think `Object` should be `Core`.

I'm also not sure why some things go in "driver helpers" and not others.  But that's neither here nor there.

> However--and I view this as a major shortcoming of TypeScript--neither `object` nor `Object` understand if the object in question extends `Object.prototype`.  This means, given `const obj: object = Object.create(null)` _or_ `const obj: Object = Object.create(null)`  and then `generateDriverLogPrefix(obj)`, TS will _not_ complain that the `constructor` property does not exist on `obj` (it doesn't), and you will get a runtime error.

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -546,6 +546,22 @@ function parseCapsArray (cap) {
...
+ */
+function generateDriverLogPrefix (obj, sessionId = null) {
+  const instanceName = `${obj.constructor.name}@${node.getObjectId(obj).substring(0, 4)}`;
+  return sessionId ? `${instanceName} (${sessionId.substring(0, 8)})` : instanceName;
+}
+
 /** @type {import('@appium/types').DriverHelpers} */
-export default {configureApp, isPackageOrBundle, duplicateKeys, parseCapsArray};
-export {configureApp, isPackageOrBundle, duplicateKeys, parseCapsArray};
+export default {
```
boneskull, line 562: We should probably move these "double" exports across the codebase; they are not needed anywhere except maybe on the public API edges.  (just saying)

packages/types/src/index.ts
```diff
@@ -481,10 +482,15 @@ export interface EventHistoryCommand {
 }
 export type HTTPMethod = _Method;
 
+export type PrefixCallback = () => string;
+
+export type Prefix = PrefixCallback|string;
```
boneskull, line 487: If we _do_ keep `PrefixCallback`, I think it's fine to shorten this to just

```suggestion
export type Prefix = string | () => string;
```

and

boneskull (commented):
This is good and was basically what I was suggesting, so thanks.  

For a test I think all you really need to do is have something that compares `this.log.prefix` before and after calling `createSession()`, and again for `deleteSession()`.  We can assert the prefix looks a certain way, or we can just assert it's not the same.

You'd probably want to put that in the `baseDriverUnitTests` function of `test/basedriver/driver-tests.js`

packages/types/src/index.ts
mykola-mokhnach: changed type to Core

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: it was like that before, so it does not make sense to me to change it in this PR

packages/types/src/index.ts
mykola-mokhnach: removed prefixcallback

packages/base-driver/lib/basedriver/core.js
```diff
@@ -102,9 +102,9 @@ class DriverCore {
   eventEmitter = new EventEmitter();
 
   /**
-  * @type {AppiumLogger?}
+  * @type {AppiumLogger}
   */
-  _log = null;
+  _log = logger.getLogger(helpers.generateDriverLogPrefix(this));
```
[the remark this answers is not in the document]
mykola-mokhnach: makes sense. Moved the initializer to the class constructor

packages/types/src/index.ts
```diff
@@ -481,10 +482,15 @@ export interface EventHistoryCommand {
...
+
 export interface AppiumLogger {
   unwrap(): Logger;
   level: string;
   levels: string[];
+  prefix?: Prefix,
```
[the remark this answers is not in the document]
mykola-mokhnach: it was like that before. Not sure I need to change that now

packages/base-driver/test/basedriver/driver-tests.js
```diff
@@ -55,6 +55,12 @@ function baseDriverUnitTests (DriverClass, defaultCaps = {}) {
       sessId.length.should.be.above(5);
     });
 
+    it('should setup log prefix after createSession', async function () {
+      const previousPrefix = d.log.prefix;
+      await d.createSession(null, null, w3cCaps);
+      previousPrefix.should.not.eql(d.log.prefix);
+    });
```
boneskull, line 62: Don't we need to reset the prefix after `deleteSession` is called?  Are we doing that?  If not, we need to do that, I think. And add a test.

It'd be more valuable to make an assertion about the expected value (since there's logic involved in determining the prefix) instead of just asserting it has changed (in the above test case and also one for `deleteSession`)

mykola-mokhnach: updated the test

PR #16689, opened 2022-03-29
chore: fix reinstall script and clean ordering

jlipps:
The ordering of `clean:*` matters since we use packages in clean.


boneskull (approved):

boneskull:
Thanks for removing the dupe script... I think I did it in another PR too

PR #16704, opened 2022-03-31
chore(appium,types,schema): create @appium/schema

### LINKED ISSUE #16685: v2: move schema to its own package
`@appium/types` exposes various types (`DriverOpts`, `AppiumConfig`, and `NormalizedAppiumConfig`, in particular), which are based upon the schema definition.  `appium` and `@appium/base-driver` need these types.  The schema definition itself lives in `appium`.  That's a cyclic dependency and it is bad.

Currently, this is being hacked around via overly-clever scripts.  

Instead, the schema definition should live in its own package (`@appium/schema`) and `@appium/types` and `appium` should depend on it.

boneskull:
`@appium/schema` is a package which was extracted from `appium`.  It _only_ contains the schema.  The impetus here was to avoid cyclical dependencies, since `appium` depends on `@appium/types`, and `@appium/types` depends on the schema.  This package breaks the cycle.

The `scripts/generate-schema-json.js` (which converts the "real" JS schema file into JSON) has moved into this package, and the generated JSON file is now under version control.  This is because the `generate-schema-types.js` script (previously `generate-schema-declarations`) depends on it, and `@appium/types` depends on the generated types.  The generated types are _also_ under version control.  I do not expect the schema to change often, so I believe the tradeoff to be worth it.

## @appium/schema

- `lib/appium-config-schema.mjs` is the "single source of truth" for the schema.
- This file is `.mjs` because the `generate-schema-json.js` script needs to read it.  It (read: Node.js) cannot read an ESM file if that file does not have a `.mjs` extension, because this package does not publish a ESM entrypoint.  Babel still compiles this file down to CommonJS.  I had considered just using TypeScript for this file, but the `generate-schema-json.js` script would also need to be TypeScript, and we'd either need to compile it or use `ts-node`, and both of those sound distasteful.  To be clear: we are (still) _not_ using TypeScript for compiling anything except declaration files.

## appium

- Removed schema source
- Removed namespaces from internal types because they suck (see change to `types/index.d.ts`)
- Removed unused `continuation-local-storage`
- Add dep for `@appium/schema`
- Added a TS project dependency on `@appium/schema`

## @appium/types

- The "sources" have moved from `src` to `lib` for consistency.
- `lib/appium-config.ts` (the generated types) are now under version control.
- `lib/schema/` has been removed (it previously held the JSON schema artifact)
- `scripts/generate-schema-declarations.js` becomes `scripts/generate-schema-types.js`
- Updated scripts in `package.json`
- Added dep for `@appium/schema`
- Added a TS project dependency on `@appium/schema`

## Other

- Reorganize root `tsconfig.json`; add `schema` reference; remove unused `noEmit: true`
- Sort root `package.json`
- Remove schema JSON and schema declarations scripts, since those are now both handled in their respective packages
- Remove `lint-staged` config for schema file, since it didn't work right anyway (still need to do something about this, but probably doesn't need to be in a precommit hook)
- Remove stuff from `.gitignore`


boneskull:
It _looks like_ conventional commits accepts "comma-delimited scopes" and I hope it does.

packages/types/README.md
```diff
@@ -14,6 +14,7 @@ npm install @appium/types -D
 
 - The sources are `.ts` files, _not_ `.d.ts` files. This allows other packages in the root TypeScript "project" to define a dependency upon this one, and enables incremental builds an "watch" mode.
 - If there is a way to switch to `.d.ts` files and configure this package to work in our "project" context _without_ needing to actually "emit" anything, then we should do that instead.  Help accepted!
+`lib/appium-config.ts` is generated by this package from `@appium/schema`, but is under version control to avoid chicken-or-egg build problems.
```
jlipps, line 17: did you intend for this to be a part of the list?

jlipps (approved):

boneskull:
* 2.0
  * **PR #16704** <a href="https://app.graphite.dev/github/pr/appium/appium/16704" target="_blank">Graphite</a>  (this PR)
    * **PR #16722** <a href="https://app.graphite.dev/github/pr/appium/appium/16722" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16704?utm_source=stack-comment).

PR #16722, opened 2022-04-06
migrate entire plugins repo into this one

boneskull:
This PR takes all changesets and packages from the [appium-plugins](/appium/appium-plugins) monorepo and brings them into this monorepo.

### Major Changes

There's one: see 6084946 (which may become outdated if I rebase again, but the commit message title is: **fix(base-driver): do not throw exception if updateSettings is not provided**.  This was necessary to get the plugin tests passing.  I'm not sure why this wasn't breaking in an `appium-plugins` context unless the `appium` dep (which doesn't seem to be an actual dep??) was old.

### Build Normalization

In `appium-plugins`, each package was seemingly compiled with TypeScript.  Now--like the existing packages in this monorepo--`tsc` is only responsible for creating declarations, and compilation is done via `babel`.  Likewise, we're using the Mocha configuration already present.

For each package, declarations (and declaration maps) will be published in the `build` dir along with the compiled sources.

The `main` module of each package points to an `index.js` which is a _non-transpiled_ CJS script.  The reader may or may not recall that this is necessary to bootstrap a working copy via `npm install`; if the `main` module _does not exist_, as it wouldn't if it were compiled, installation will fail.  For `universal-xml-plugin` in particular, the `main` function (intended to be run via CLI) was moved from the root `index.js` into `lib/index.js`.  I am not sure what that script is for, but there it is!

## Other

- Created an ESLint rule to enforce CJS in `./packages/*/index.js` and `./test/*.js` (the latter of which is just a `setup.js` file).
- Modified the `test/helpers` module in `base-plugin` to work with a local Appium and local "fake" extensions instead of via `npx`. It also accepts an `appiumHome` option.  We're currently using a directory called `local_appium_home`, which is not under version control.  This path should probably just be changed to a temp directory.
- While debugging, I created some stub types for `@appium/opencv` (types for `opencv-bindings` should be created and submitted to DefinitelyTyped).
- Note that `package-lock.json` is no longer under VCS for each new package. There should only be _one_ `package-lock.json`, and it is in the monorepo root.  See #16553


jlipps:
i don't think we care about keeping history if all the commits aren't modified to use the semantic commit scope for the correct packages in the new repo?

jlipps:
instead of reviewing this PR i kind of just want you to tell me that it all works, that all unit and e2e tests are being successfully run in CI, etc... presumably you didn't make any consequential changes to the plugins themselves right?

boneskull:
> i don't think we care about keeping history if all the commits aren't modified to use the semantic commit scope for the correct packages in the new repo?

I'd prefer to keep it--there's more to commit messages than just the semantic message.  don't want to lose context around intent

boneskull:
@jlipps 

> instead of reviewing this PR i kind of just want you to tell me that it all works, that all unit and e2e tests are being successfully run in CI, etc... presumably you didn't make any consequential changes to the plugins themselves right?

There's one "[consequential change](https://github.com/appium/appium/pull/16722/commits/6084946805f2f2900b0db733767be47357bbbb66)" I had to make to get the tests passing (and it could be considered a breaking change).  It may have been the incorrect, but there was a clear disagreement between the plugins and the base driver.  Please tell me if I need to back this change out ...

boneskull:
@jlipps okay, this is ready...

jlipps (approved):
i did not review every specific thing but if all the moved plugin tests exist and are being run (unit and e2e), should be good!

boneskull:
alright, here it goes.  see appium/appium-plugins#104

PR #16733, opened 2022-04-07
chore: fix dev scripts

boneskull:
These weren't working correctly.

Currently unable to run `generate-schema-json` in "watch" mode w/o some third-party tool. nbd


jlipps (commented):
:+

Issue #16736, opened 2022-04-08
Error: The session with id 'undefined' does not exist


StephanB75:
## The problem

Since yesterday evening my tests throwing this error

## Environment

* Appium version (or git revision) that exhibits the issue: beta27/beta28
* Last Appium version that did not exhibit the issue (if applicable): beta27
* Desktop OS/version used to run Appium: Monterey
* Node.js version (unless using Appium.app|exe): 14
* Npm or Yarn package manager: 6.14
* Mobile platform/version under test: iOS
* Real device or emulator/simulator: Simulator
* Appium CLI or Appium.app|exe: CLI

## Details

The server is started via AppiumServiceBuilder in java

## Link to Appium logs

```
2022-04-08 08:24:35:798 - [Appium] Attempting to load plugin images...
2022-04-08 08:24:35:800 - [debug] [Appium] Requiring plugin at /Users/jenkins/HomeOfQA/workspace/RMA-iOS-Versions/uitesting/.appium/node_modules/@appium/images-plugin
2022-04-08 08:24:37:030 - [Appium] Welcome to Appium v2.0.0-beta.28 (REV 28603c381745d77b6e6efb8b70e8de870638eb29)
2022-04-08 08:24:37:032 - [Appium] Non-default server args:
2022-04-08 08:24:37:034 - [Appium] { address: '127.0.0.1',
2022-04-08 08:24:37:034 - [Appium]   allowInsecure: [ 'adb_shell' ],
2022-04-08 08:24:37:035 - [Appium]   localTimezone: true,
2022-04-08 08:24:37:035 - [Appium]   logTimestamp: true,
2022-04-08 08:24:37:035 - [Appium]   port: 4727,
2022-04-08 08:24:37:035 - [Appium]   usePlugins: [ 'images' ] }
2022-04-08 08:24:37:035 - [Appium] Default capabilities, which will be added to each request unless overridden by desired capabilities:
2022-04-08 08:24:37:035 - [Appium] { 'appium:noReset': 'false' }
2022-04-08 08:24:37:036 - [Appium] Attempting to load driver xcuitest...
2022-04-08 08:24:37:038 - [debug] [Appium] Requiring driver at /Users/jenkins/HomeOfQA/workspace/RMA-iOS-Versions/uitesting/.appium/node_modules/appium-xcuitest-driver
2022-04-08 08:24:38:485 - [Appium] Appium REST http interface listener started on 127.0.0.1:4727
[...]

PR #16737, opened 2022-04-08
fix(appium): Properly detect if a command is a session command

mykola-mokhnach:
## Proposed changes

Related to https://github.com/appium/appium/issues/16736

## Types of changes

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -516,7 +516,7 @@ class AppiumDriver extends DriverCore {
     // cases with plugin handling.
 
     const isGetStatus = cmd === GET_STATUS_COMMAND;
-    const isUmbrellaCmd = !isGetStatus && isAppiumDriverCommand(cmd);
+    const isUmbrellaCmd = isAppiumDriverCommand(cmd);
     const isSessionCmd = isSessionCommand(cmd);
 
     // if a plugin override proxying for this command and that is why we are here instead of just
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -115,7 +115,6 @@ class AppiumDriver extends DriverCore {
     return this._log;
   }
 
-
   /**
    * Cancel commands queueing for the umbrella Appium driver
    */
@@ -138,9 +137,8 @@ class AppiumDriver extends DriverCore {
     };
   }
 
-  async getSessions () {
-    const sessions = await sessionsListGuard.acquire(AppiumDriver.name, () => this.sessions);
-    return _.toPairs(sessions)
+  async getSessions () { // eslint-disable-line require-await
+    return _.toPairs(this.sessions)
       .map(([id, driver]) => ({id, capabilities: driver.caps}));
   }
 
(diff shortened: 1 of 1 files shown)


packages/appium/lib/appium.js
```diff
@@ -522,9 +516,8 @@ class AppiumDriver extends DriverCore {
     // cases with plugin handling.
 
     const isGetStatus = cmd === GET_STATUS_COMMAND;
-    const isDeleteSession = cmd === DELETE_SESSION_COMMAND;
     const isUmbrellaCmd = !isGetStatus && isAppiumDriverCommand(cmd);
-    const isSessionCmd = !isUmbrellaCmd || isDeleteSession;
+    const isSessionCmd = isSessionCommand(cmd);
```
jlipps, line 520: does this evaluate to exactly the same value as the previous one? i remember having to tune this boolean to get plugin handling to work properly.

jlipps (approved):

packages/appium/lib/appium.js
mykola-mokhnach: according to the previous implementation it should be. Do you see any potential issues there?

jlipps: hmm the only thing I can think is that I introduced it in https://github.com/appium/appium/commit/7b3893a36202018de7c2124c2028bfbbd8a9d7fd. And maybe no one has tried to call GET /status without a session for a while? but this code has been published for longer than the report says the issue has existed, so i am not sure. maybe some interaction with this change and the driver refactor?

but yeah i think this change is good. i'm realizing we can also change the above line to simply:

```js
const isUmbrellaCmd = isAppiumDriverCommand(cmd);
```

because that function already returns true if it's not get status.

packages/appium/lib/appium.js
```diff
@@ -138,9 +137,8 @@ class AppiumDriver extends DriverCore {
     };
   }
 
-  async getSessions () {
-    const sessions = await sessionsListGuard.acquire(AppiumDriver.name, () => this.sessions);
-    return _.toPairs(sessions)
+  async getSessions () { // eslint-disable-line require-await
+    return _.toPairs(this.sessions)
```
boneskull, line 141: this whole function can be written as 

```js
// untested
return _.mapValues((driver, id) => ({id, capabilities: driver.caps})
```

packages/appium/lib/appium.js
```diff
@@ -377,16 +373,14 @@ class AppiumDriver extends DriverCore {
...
    */
-  async curSessionDataForDriver (InnerDriver) {
-    const sessions = await sessionsListGuard.acquire(AppiumDriver.name, () => this.sessions);
-    const data = _.compact(_.values(sessions)
-                   .filter((s) => s.constructor.name === InnerDriver.name)
-                   .map((s) => s.driverData));
-    for (let datum of data) {
+  async curSessionDataForDriver (InnerDriver) { // eslint-disable-line require-await
+    const data = _.compact(_.values(this.sessions)
+      .filter((s) => s.constructor.name === InnerDriver.name)
```
boneskull, line 378: 1. why not use `s instanceof InnerDriver` instead of `s.constructor.name` comparison?
2. we can remove the `_.compact` (I may have even added this?) and change the `filter` callback instead:

```js
// untested
.filter((s) => s.constructor.name === InnerDriver.name && s.driverData)
```

boneskull (approved):

packages/appium/lib/appium.js
mykola-mokhnach: I remember there were some issues with instanceof call if a class name was imported from a different package. Maybe @jlipps knows more about it

> we can remove the _.compact (I may have even added this?) and change the filter callback instead:

I don't want to change too much here, since this is not the intention of that PR

jlipps: yeah `instanceof` checks don't work for classes that come from two different versions of the same package. not sure if that is a possibility here but it's something we ran into a lot in earlier versions of appium (we couldn't do `instanceof` checks on appium errors because we might be dealing with different versions of BaseDriver imported by a driver vs imported by Appium)

PR #16742, opened 2022-04-10
fix(base-driver): Make sure proxyReqRes helper never throws any exceptions

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/appium/appium-espresso-driver/issues/756#issuecomment-1094091844

## Types of changes

### DIFF
--- a/packages/base-driver/lib/jsonwp-proxy/proxy.js
+++ b/packages/base-driver/lib/jsonwp-proxy/proxy.js
@@ -329,7 +329,7 @@ class JWProxy {
           : err
       );
     }
-    res.set('content-type', 'application/json');
+    res.set('content-type', 'application/json; charset=utf-8');
     if (!_.isPlainObject(resBodyObj)) {
       const error = new errors.UnknownError(
         `The downstream server response with the status code ${statusCode} is not a valid JSON object: ` +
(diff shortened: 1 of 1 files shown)


packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -312,10 +313,31 @@ class JWProxy {
...
+      res.headers = response.headers;
+      statusCode = response.statusCode;
+    } catch (err) {
+      [statusCode, resBodyObj] = getResponseForW3CError(
+        errors.isErrorType(err, errors.ProxyRequestError)
+          ? err.getActualError()
+          : err
+      );
+    }
+    res.set('content-type', 'application/json');
```
KazuCocoa, line 332: does here need `charset=utf-8` as well?
Probably we provide `application/json; charset=utf-8` regularly.

KazuCocoa (approved):
otherwise, lgtm

packages/base-driver/lib/jsonwp-proxy/proxy.js
mykola-mokhnach: makes sense, fixed

PR #16748, opened 2022-04-11
feat(base-driver): Add a route for new window creation

mykola-mokhnach:
## Proposed changes

https://w3c.github.io/webdriver/#new-window

## Types of changes

cc @titusfortner

### DIFF
--- a/packages/types/lib/index.ts
+++ b/packages/types/lib/index.ts
@@ -113,7 +113,7 @@ export interface ExternalDriver extends Driver {
   maximizeWindow?(): Promise<Rect>;
   minimizeWindow?(): Promise<Rect>;
   fullScreenWindow?(): Promise<Rect>;
-  createNewWindow?(): Promise<NewWindow>;
+  createNewWindow?(type?: NewWindowType): Promise<NewWindow>;
   active?(): Promise<Element>;
   elementSelected?(elementId: string): Promise<boolean>;
   getAttribute?(name: string, elementId: string): Promise<string | null>;
@@ -406,9 +406,11 @@ export interface Rect {
   height: number;
 }
 
+export type NewWindowType = 'tab' | 'window';
+
 export interface NewWindow {
   handle: string;
-  type: string;
+  type: NewWindowType;
 }
 
 export interface Cookie {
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
Linking https://github.com/appium/appium/issues/16749

KazuCocoa (approved):

packages/types/lib/index.ts
```diff
@@ -405,6 +406,11 @@ export interface Rect {
   height: number;
 }
 
+export interface NewWindow {
```
boneskull, line 409: is this type unique to Appium or is it standardized somewhere? (could we borrow the type from wdio?)

packages/types/lib/index.ts
```diff
@@ -405,6 +406,11 @@ export interface Rect {
...
+  handle: string;
+  type: string;
```
boneskull, line 411: could `type` be anything, or is it only a handful of different strings?  if the latter, list them in a union:

```ts
  type: 'foo' | 'bar' | 'baz'
```

boneskull (approved):
lgtm, just a couple questions.

I'm curious if TS warned you that `createNewWindow` must exist in the `ExternalDriver` type?

packages/types/lib/index.ts
mykola-mokhnach: not sure how/where to check this

mykola-mokhnach:
> lgtm, just a couple questions.
> 
> I'm curious if TS warned you that `createNewWindow` must exist in the `ExternalDriver` type?

yes, the linter was complaining to a missing type declaration after only the route itself has been added

packages/types/lib/index.ts
boneskull: I'm not sure either.  I was just grepping around.  it doesn't look like this response is typed.

packages/types/lib/index.ts
```diff
@@ -406,9 +406,11 @@ export interface Rect {
   height: number;
 }
 
+export type NewWindowType = 'tab' | 'window';
```
boneskull, line 409: nice, thanks

PR #16756, opened 2022-04-12
chore(ci): only build on PRs and pushes to 2.0

boneskull:
For every push to a PR, when the PR branch exists on `appium/appium`, a "push" build happens and a "pr" build happens.  The "push" build tests the code in isolation, and the "pr" build tests it as it would be merged into the base branch.  mostly this is just extra work, as the latter type is much more important.  however, the `2.0` branch should be tested every time code is merged into it.

What do we think of this (assuming it works)?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,11 @@
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
 name: Appium Build
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - '2.0'
 
 jobs:
   test:


mykola-mokhnach:
I'd rather keep it as is for now until we move `2.0` to master

boneskull:
as you can see from the checks, no "push" events there. so this seems to do what it's supposed to, but just want to know if someone can think of a reason why we shouldn't.

jlipps (approved):

KazuCocoa (approved):
as my understanding, master branch's actions will keep in https://github.com/appium/appium/tree/master/.github/workflows . This is only for 2.0 branch's actions.
lg

mykola-mokhnach:
> @mykola-mokhnach what's the reasoning?

To avoid doing extra work after the branch switch. Priorities...

boneskull:
@mykola-mokhnach ah.  well, maybe we should just sit on this until we move `2.0` to `main`.

boneskull:
I'm going to close this, as I am doubting this does what I think it does.  I did something similar in another repo, and checks were not getting triggered on pushes to PR branches.

PR #16757, opened 2022-04-12
chore: fix duplicate preversion scripts

boneskull:
- `prepublishOnly` now cleans all build artifacts
- `preversion` does the same, except runs `sync-pkgs` also


package.json
```diff
@@ -30,9 +30,10 @@
     "build:types": "tsc -b",
     "clean": "run-s clean:*",
     "clean:distfiles": "npx rimraf \"packages/*/build\"",
+    "clean:monorepo": "npx rimraf node_modules package-lock.json",
```
jlipps, line 33: this will break `npm run clean` because it will delete the node_modules before running `clean:packages`.

the order here matters; we either need to make sure it stays the way it is, or just spell out which clean scripts to run (i.e., don't use `run-s clean:*`)

package.json
```diff
@@ -44,8 +45,7 @@
     "lint:fix": "npm run lint -- --fix",
     "precommit-lint": "lint-staged",
     "precommit-msg": "echo 'Pre-commit checks...'",
-    "prepublishOnly": "tsc -b --clean && npm run build",
-    "preversion": "tsc -b --clean && npm run build",
```
jlipps, line 48: oops! sorry

jlipps (changes requested):

package.json
boneskull: ah, I had just moved this line, but I've fixed this anyway

jlipps (approved):

package.json
```diff
@@ -28,11 +28,12 @@
     "build:distfiles": "lerna run --parallel --prefix --concurrency=8 build",
     "build:loose": "run-s -c build:distfiles build:types || true",
     "build:types": "tsc -b",
-    "clean": "run-s clean:*",
-    "clean:distfiles": "npx rimraf \"packages/*/build\"",
+    "clean": "run-s clean:artifacts clean:packages clean:monorepo",
+    "clean:distfiles": "npx rimraf \"packages/*/build/**/*.js\"",
```
jlipps, line 32: why are we only getting rid of js and not other things that might be in build? don't we want a totally clean build dir after we run this command?

boneskull: well, because `babel` makes `.js` files and typescript makes `.d.ts` files, and they both go in `build`. neither one depends on the other, so it could be a source of dev-workflow frustration.  `clean:artifacts` removes both.

boneskull: (and `clean:artifacts` is what `clean` calls--these two are more for the case where you just want to clean one or the other)

PR #16758, opened 2022-04-12
chore: dev script builds types

boneskull:
- `npm run dev` will build all artifacts in "watch" mode (distfiles and types)
- also updated the default build task for vscode


boneskull:
eh, w/e, it's fine.

jlipps (commented):
👍  yeah i was wanting something like this, thanks

Issue #16747, opened 2022-04-11
appium should be a peer dep of extentensions


JuanCarlos-Arceredillo:
## The problem

I have made some tries to run tests on iOS using XCUiTest and on Android using Espresso with Appium 2.0, but each time that I try to create the session, Appium shows an internal error.

## Environment

* Appium version v2.0.0-beta 28:
* Node.js version: v14.17.0
* Npm or Yarn package manager: NPM
* Mobile platform/version under test: iOS & Android
* Real device or emulator/simulator: Simulator & Device
* Appium CLI or Appium.app|exe: Appium command line

## Details

I have installed the drivers:
- xcuitest version 4.3.5
- uiautomator2 version 2.0.7
- espresso verision 2.2.1

iOS
`
[debug] [AppiumDriver@94a2] Event 'newSessionRequested' logged at 1649674844863 (13:00:44 GMT+0200 (Central European Summer Time))
[Appium] Attempting to find matching driver for automationName 'XCUITest' and platformName 'iOS'
[Appium] The 'xcuitest' driver was installed and matched caps.
[Appium] Will require it at /Users/SYS_CIBER_ADMIN/.appium/node_modules/appium-xcuitest-driver
[debug] [Appium] Requiring driver at /Users/SYS_CIBER_ADMIN/.appium/node_modules/appium-xcuitest-driver
[AppiumDriver@94a2] Appium v2.0.0-beta.28 creating new XCUITestDriver (v4.3.5) session
[AppiumDriver@94a2] Checking BaseDriver versions for Appium and XCUITestDriver
[AppiumDriver@94a2] Appium's BaseDriver version is 8.4.0
[AppiumDriver@94a2] XCUITestDriver's BaseDriver version is 8.3.1
[debug] [AppiumDriver@94a2] Event 'newSessionStarted' logged at 1649674844867 (13:00:44 GMT+0200 (Central European Summer Time))

[debug] [AppiumDriver@94a2] Encountered internal error running command: TypeError: driverInstance.assignServer is not a function
[debug] [AppiumDriver@94a2]     at AppiumDriver.createSession (/usr/local/lib/node_modules/appium/lib/appium.js:282:22)
[debug] [AppiumDriver@94a2]     at commandExecutor (/usr/local/lib/node_modules/appium/node_modules/@appium/base-driver/lib/basedriver/driver.js:74:9)
[...]

PR #16797, opened 2022-04-18
fix(appium,types): avoid call to assignServer()

boneskull:
Due to the way which drivers currently include `@appium/base-driver`, it's possible for a driver to become out-of-date with respect to `AppiumDriver` and fail to implement the expected interface.  This change partially reverts the addition of `Driver.assignServer()`, which is a new function that old versions of `@appium/base-driver` do not implement.

Ref: #16747 (does not fix)


boneskull:
(essentially `assignServer()` is still defined & declared, but it is not called at the moment

jlipps (approved):

PR #16829, opened 2022-04-19
fix(base-driver): supportedLogTypes does not get overwritten

### LINKED ISSUE #16738: supportedLogTypes does not get assigned properly on driver init since beta27
## The problem

After base driver update to 8.4.0 `getLogTypes` and `getLog` commands of appium-android-driver stopped working properly because `this.supportedLogTypes` property does not get initialized. I assume this is related to the recent typing refactoring in `base-driver/lib/basedriver/commands/log.js`. Previously this variable has been assigned by defining `extensions.supportedLogTypes`, which was equal to `{}` by default and is overridden in each driver if necessary. Now it always equals to `{}`. See https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=20705&view=logs&j=12e12a94-6f82-50bd-49c9-566fb2052675&t=7ba1c276-5935-513a-f217-d07020732771 for more details.

## Environment

* Appium version (or git revision) that exhibits the issue: base-driver 8.3.0
* Last Appium version that did not exhibit the issue (if applicable): base-driver 8.4.0
* Desktop OS/version used to run Appium:
* Node.js version (unless using Appium.app|exe):
* Npm or Yarn package manager:
* Mobile platform/version under test:
* Real device or emulator/simulator:
* Appium CLI or Appium.app|exe:

boneskull:
Closes #16738

There's a difference between assigning things to a prototype and public class fields, evidently.

To test this would be best done via interaction between drivers and `base-driver`, but affected drivers do not necessarily run the reusable test suite.  Fundamentally, a unit test for this would be testing Babel.  Unsure of best way to add an E2E test

Once it gets merged, we should publish as soon as possible, since @mykola-mokhnach is blocking on this.

### DIFF
--- a/packages/base-driver/lib/basedriver/commands/log.js
+++ b/packages/base-driver/lib/basedriver/commands/log.js
@@ -13,11 +13,12 @@ export function LogMixin (Base) {
    * @implements {ILogCommands}
    */
   class LogCommands extends Base {
-    /**
-     * XXX: dubious
-     * @type {Record<string,LogType<Driver>>}
-     */
-    supportedLogTypes;
+
+    constructor (...args) {
+      super(...args);
+      /** @type {Record<string, LogType<Driver>>} */
+      this.supportedLogTypes = this.supportedLogTypes ?? {};
+    }
 
     async getLogTypes () {
       this.log.debug('Retrieving supported log types');
@@ -26,19 +27,13 @@ export function LogMixin (Base) {
 
     /**
      * @this {Driver}
+     * @param {string} logType
      */
     async getLog (logType) {
       this.log.debug(`Retrieving '${logType}' logs`);
 
       if (!(await this.getLogTypes()).includes(logType)) {
-        const logsTypesWithDescriptions = _.reduce(
-          this.supportedLogTypes,
-          (acc, value, key) => {
-            acc[key] = value.description;
-            return acc;
-          },
-          {},
-        );
+        const logsTypesWithDescriptions = _.mapValues(this.supportedLogTypes, 'description');
         throw new Error(
           `Unsupported log type '${logType}'. ` +
             `Supported types: ${JSON.stringify(logsTypesWithDescriptions)}`,
--- a/packages/types/lib/index.ts
+++ b/packages/types/lib/index.ts
@@ -72,7 +72,7 @@ export interface Driver
   startNewCommandTimeout(): Promise<void>;
   reset(): Promise<void>;
 
-  assignServer(
+  assignServer?(
     server: AppiumServer,
     host: string,
     port: number,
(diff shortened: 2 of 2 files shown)


packages/base-driver/lib/basedriver/commands/log.js
```diff
@@ -13,11 +13,12 @@ export function LogMixin (Base) {
...
-    /**
-     * XXX: dubious
-     * @type {Record<string,LogType<Driver>>}
-     */
-    supportedLogTypes;
+
+    constructor (...args) {
+      super(...args);
+      /** @type {Record<string, LogType<Driver>>} */
+      this.supportedLogTypes = this.supportedLogTypes ?? {};
```
mykola-mokhnach, line 20: I'm not quite sure about this expression. Do we assign the variable to itself?

boneskull: it's this:

```js
if (this.supportedLogTypes === null || this.supportedLogTypes === undefined) {
  this.supportedLogTypes = {};
}
```

mykola-mokhnach (approved):

packages/base-driver/lib/basedriver/commands/log.js
mykola-mokhnach: I was just not quite sure how "this.supportedLogTypes" gets assigned before the constructor

jlipps (approved):

packages/base-driver/lib/basedriver/commands/log.js
jlipps: seems like it's just a safety mechanism because we can't rule out that it might have been set in some other "mixin"'s constructor and if it has, we leave it alone?

boneskull: No, it doesn't have anything to do with mixins potentially setting the value.  `this.supportedLogTypes` is assigned before the constructor if it is on the constructor's `prototype`.  In the case of `AndroidDriver`, this is the case.  

What Babel seems to do (and this may be correct behavior) is a class field, e.g.,:

```js
class Foo {
  bar = 'baz'
}
```

actually results in this:

```
class Foo {
  constructor() {
     this.bar = 'foo';
  }
}
```

Now, that's fine if something subclasses `Foo` and then assigns its own value:

```js
class Baz extends Foo
  constructor() {
    super();
    this.bar = 'baz';
  }
}
```

or uses a field itself:

```js
class Baz extends Foo {
  bar = 'baz';
}
```

...but the above code is not the same as this (which is what `AndroidDriver` does, and maybe other drivers as well):

```js
class Baz extends Foo {}
Baz.prototype.bar = 'baz';
```

In the above case, the constructor for `Foo` will overwrite `Baz.prototype.bar` on any newly-instantiated `Baz` object.

What this means is we need to be careful with class fields, because unless a subclass _also_ uses class fields (or assignments in its constructor), we'll run into this problem.

boneskull: actually, the android driver does something _really_ weird, which is assigns to its own prototype in its constructor (why?).  however, changing it to assign to the prototype after the `class` definition didn't fix the issue.

boneskull: Here's a playground link to see for yourself how this happens:

https://www.typescriptlang.org/play?filetype=js#code/MYGwhgzhAEBiD29oG8BQ0PQEZgE7QF5oByAM0WIG5V1MAXeAZTtwEsA7AcwAoBKFWpky4ApnQCuudtDoALVhAB0OXNSEBfVJtShIMAEJgAXtBEAPOiPYATGAiRohKwiRxHiWmrqjQAiuPEzUwsrWzhEAU9-QMUAB1x4BjoAT1iRZTwXYgBHALMqL3h2CHgQdJB4HgADQyMALmgAEmR2EQB3aFq+dSreamAikrLFCuroswbm1o7x7t7KaAB6RegAdQAJAE1oAElGaAAVdb3oKvJ4KoB+S5ogA

jlipps:
I think we should probably fix the android driver and any other driver that assigns to prototypes to do something different, yeah?

PR #16833, opened 2022-04-20
feat(eslint-config-appium): adds a warning if code attempts to assign to object prototype

boneskull:
See #16829

When using ES6 `class` syntax, direct assignment to `prototype` can be a source of bugs.  Class properties seem to exacerbate the issue.  While in many cases this code will work fine, there is not always a straightforward workaround, which is why the rule just emits a warning.

The new syntax rule considers assignment to the `prototype` of any object, as well as using `Object.assign()` to do the same.

### DIFF
--- a/packages/eslint-config-appium/index.js
+++ b/packages/eslint-config-appium/index.js
@@ -98,6 +98,17 @@ module.exports = {
     'require-atomic-updates': 0,
     'no-prototype-builtins': 1,
     'no-redeclare': 1,
+    'no-restricted-syntax': [
+      'warn',
+      {
+        selector: 'AssignmentExpression[left.object.property.name="prototype"]',
+        message: 'Avoid assignment to prototype; use class fields, methods or mixins instead.',
+      },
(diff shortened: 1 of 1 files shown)


boneskull:
(the rule is not exhaustive; there are ways to get around this. referencing `SomeObject.prototype` directly also has its uses, so we cannot just use the `no-restricted-properties`, which would grouse about _any_ appearance of `SomeObject.prototype` in the code)

mykola-mokhnach (approved):

mykola-mokhnach:
I assume it was always working like that in drivers. Would it make sense to rather update the implementation there first?

boneskull:
@mykola-mokhnach That will take much longer 😄

mykola-mokhnach:
yes, but would be much more effective than just having a warning

boneskull:
true, but it will also not avoid it in the future.  so we need this one way or the other

jlipps (approved):

PR #16834, opened 2022-04-20
test(docutils): run mike test against sources

boneskull:
referencing the project root pulls the root `index.js`, which requires `build/lib`, which is not what we want.

I wonder if `eslint-plugin-import` would allow us to avoid this... hm.


jlipps (approved):
looks good, i'm  used to doing it this way because in the past tests were transpiled as well, i forgot now they're run separately.

PR #16835, opened 2022-04-20
chore: add prettier

boneskull:
- As a pre-commit hook, `.js` files will be run thru `eslint --fix` and rewritten using `prettier`.
- Added `eslint-config-prettier` to disable rules incompatible with `prettier`.
- Added a `.gitattributes` to enforce `lf` line endings.

Files will not be formatted with `prettier` _unless_ they have been added to the stage; nothing is going to go reformat the whole codebase.

* * *

The idea with Prettier, of course, is to automatically format code.

You made not like some of the choices it makes.  Heck, I don't like some of the choices it makes.  But the point is twofold:

1. Consistent formatting across codebase
2. Avoiding the inevitable discussion about formatting in pull requests

Prettier has relatively few options.  There's a lot of minutiae that it simply won't let you control.  That said, I have set a few options which it _does_ have:

1. `bracketSpacing: false` - we don't use bracket spacing in general, but we _do_ use it in `import` statements.  It's all-or-nothing.
2. `singleQuote: true` - use single quotes for formatting strings, unless we'd end up with too many escapes. does not affect template literals

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/package.json
+++ b/package.json
@@ -29,11 +29,11 @@
     "build:loose": "run-s -c build:distfiles build:types || true",
     "build:types": "tsc -b",
     "clean": "run-s clean:artifacts clean:packages clean:monorepo",
+    "clean:artifacts": "run-s clean:distfiles clean:types && npx rimraf \"packages/*/build\"",
     "clean:distfiles": "npx rimraf \"packages/*/build/**/*.js\"",
     "clean:monorepo": "npx rimraf node_modules package-lock.json",
     "clean:packages": "lerna clean -y && npx rimraf \"packages/*/package-lock.json\"",
     "clean:types": "tsc -b --clean",
-    "clean:artifacts": "run-s clean:distfiles clean:types && npx rimraf \"packages/*/build\"",
     "dev": "run-p dev:*",
     "dev:distfiles": "lerna run --parallel --prefix --concurrency=8 dev",
     "dev:types": "tsc -b --watch",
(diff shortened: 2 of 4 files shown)


jlipps:
is there a way to see what it will do before showing the adjusted commit? i kind of want to know what i'm opting in to in most cases and potentially to override if i disagree on something that we can't customize.

boneskull:
> is there a way to see what it will do before showing the adjusted commit? i kind of want to know what i'm opting in to in most cases and potentially to override if i disagree on something that we can't customize.

@jlipps like a diff? the formatting should not be overridden, but you can do mark sections with a pragma to avoid formatting.

I'm not sure about vim or whatever you're using, but at least in vscode it's trivial to setup format-on-save, so you can see what it'll do before you commit.  likewise `prettier --write ./path/to/file` will format a single file, and `prettier ./path/to/file` will dump the formatted file to stdout, which you can probably pipe into a diff tool

boneskull:
if we like this, prettier can format yaml, markdown, json, etc., which we could configure in the future

KazuCocoa:
i haven't used prettier but i like like this autoformat.

KazuCocoa (approved):

jlipps (approved):
ok let's give it a go! if i become extremely disgruntled with it i can find a way to add rules to prettier or we can fight about it later.

boneskull:
@jlipps I don't really love the sound of that, but sure.  We are all going to need to worry a little less about how things are formatted.

jlipps:
I'm just saying without knowing what prettier will do it's hard for me to gauge how happy I will be with it. I'm all for having a formatter that eliminates conversations about style, of course, in principle.

boneskull:
@jlipps fwiw, every new `.js` file I've created in this repo was formatted with prettier (the only difference previously being that I run a secondary formatter to put spaces around the named `import`s)

jlipps:
oh ok, that's a good sign then!

PR #16848, opened 2022-04-22
chore: "lerna link convert"; replace siblings with relative "file:" links

boneskull:
not sure what went wrong with the last round of publishes, but this reverts the change.

during publish, lerna replaces these relative `file:..` dependencies with actual version numbers using a temp file.  I imagine if something breaks, that does not get restored properly.

the command to run is `lerna link convert` or `npm exec lerna link convert`

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "generate-docs": "lerna run --scope=appium generate-docs",
     "postinstall": "npm run build:loose",
     "install-fake-driver": "npm start -- driver install --source=local ./packages/fake-driver",
+    "install-mjpeg-consumer": "npm install mjpeg-consumer --no-save",
     "lint": "eslint .",
     "lint:fix": "npm run lint -- --fix",
     "precommit-lint": "lint-staged",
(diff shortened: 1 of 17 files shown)


jlipps (approved):
huh yeah i reverted the staged changes, but  i guess in the subsequent publishes that actually worked, lerna still did this swap. so it seems we need to configure it not to?

packages/base-driver/package.json
```diff
@@ -67,9 +67,6 @@
     "source-map-support": "0.5.21",
     "validate.js": "0.13.1"
   },
-  "devDependencies": {
```
mykola-mokhnach, line 70: ws is used to run e2e tests. What is the reason to remove it?

packages/support/package.json
```diff
@@ -85,9 +85,6 @@
     "access": "public"
   },
   "types": "./build/lib/index.d.ts",
-  "devDependencies": {
```
mykola-mokhnach, line 88: not sure this dev dependency is used anywhere

packages/base-driver/package.json
boneskull: all dev deps live in the monorepo root `package.json`

packages/support/package.json
boneskull: or rather set the version manually

boneskull:
weird that `mjpeg-consumer` tests just started breaking...

package.json
```diff
@@ -41,6 +41,7 @@
     "generate-docs": "lerna run --scope=appium generate-docs",
     "postinstall": "npm run build:loose",
     "install-fake-driver": "npm start -- driver install --source=local ./packages/fake-driver",
+    "install-mjpeg-consumer": "npm install mjpeg-consumer --no-save",
```
jlipps, line 44: why do we install this separately instead of using it as a dependency?

PR #16871, opened 2022-04-28
fix(eslint-config-appium): disable space-before-function-paren

boneskull:
BREAKING CHANGE

This explicitly-set rule conflicts with prettier and needs to be disabled.


jlipps (approved):
so prettier will create files without the space?

boneskull:
@jlipps yeah, it removes the space.

PR #16887, opened 2022-05-02
chore(appium): refactor some variable/parameter names for clarity

boneskull:
renamed parameters named `ext` to `installSpec`, since these correspond to the values of `installSpec` in the manifest

### DIFF
--- a/packages/appium/lib/cli/driver-command.js
+++ b/packages/appium/lib/cli/driver-command.js
@@ -1,60 +1,85 @@
 import _ from 'lodash';
 import ExtensionCommand from './extension-command';
-import { KNOWN_DRIVERS } from '../constants';
+import {KNOWN_DRIVERS} from '../constants';
 import '@colors/colors';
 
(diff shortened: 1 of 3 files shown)


boneskull:
* 2.0
  * **PR #16887** <a href="https://app.graphite.dev/github/pr/appium/appium/16887" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16887?utm_source=stack-comment).

boneskull:
mainly trying to reduce confusion about what kind of "name" we are dealing with.  `installSpec` is not `extName`, for example (though this could also be more clear), but `ext` doesn't tell me enough

jlipps (approved):
was hard to tell what was an actual change and what was what appears to be prettier reformatting. so i'll trust you that the rename is safe.

boneskull:
this also adds `installSpec` as a parameter to `validateExtensionFields` so that the subclasses can display more relevant information about which extension actually had a problem

PR #16901, opened 2022-05-05
extensions should have a peer dependency on appium

### LINKED ISSUE #16474: Memory error after bulid Success. also xcode will not launch the simulator at all
## The problem

After finished the config when I try to bulid WDA to the simulator, there are some memory error happend.
merroy error

I am not sure if i need to change this define  like "" or empty Array?
define

## Environment
* iPhone8  simulator
* IOS version 15.2
* Xcode 13.2.1
* Build version 13C100

## Details

Like the pic shows

boneskull:
In #16474, we found that since drivers depended directly on `@appium/base-driver`, it is difficult to make any internal additions or changes to `BaseDriver` without potentially breaking all drivers (since the API is used by the internal `AppiumDriver`).

A solution to this problem--which is common amongst "platform" tools (which support plugins) in the JS ecosystem--is for Appium's _extensions_ to have a peer dependency on Appium. The version specified can be considered the _latest version_ of Appium on which the plugin will run.

In practice, this means there can no longer be a mismatch between the version of e.g., `BaseDriver` that a driver extends, and the `BaseDriver` which Appium expects to use itself. In particular, #16474 would not have occurred had we been using this PR.

For most extensions, a peer dependency of `^2.0.0` will be sufficient (we should not recommend using `*`). For others, a more specific version may be required. It is the responsibility of each extension to ensure its compatibility with whatever version of Appium it wants to support.

As far as new behavior is concerned, this is mostly just a _check_ for a peer dependency version in the extension's `package.json` which satisfies the _current_
version of Appium. The peer dependency version is stored in the `appiumVersion` property of the `extensions.yaml` manifest, for internal use.

## For End-Users

If the peer dependency is missing or invalid Appium will emit a warning to the end-user, but it will attempt to use the extension anyway (since, again, most of the time, it should "just work", and newer versions of Appium won't break anything).

Further, if an upgrade of the extension is available, Appium will suggest a version (or versions) to upgrade to. This does not guarantee that the newer version of the driver has the correct peer dependency, but instead just wants to point the user in the right direction. If no upgrade is available, Appium will emit a different suggestion.

## For Extension (Plugin/Driver) Developers

Instead of listing `@appium/base-driver`, `@appium/base-plugin` or `@appium/support` as a direct dependency of your extension, replace this with a `peerDependency` on `appium`.

While it's likely that referencing `@appium/base-driver` will _work_, it's not recommended, as it is no longer a direct dependency of your extension. Instead, make changes like in these examples:

- Change `import BaseDriver from '@appium/base-driver'` to `import {BaseDriver} from 'appium/driver'`.
- Change `import BasePlugin from '@appium/base-plugin'` to `import {BasePlugin} from 'appium/plugin'`.
- Change `import {fs, logger, util} from '@appium/support'` to `import {fs, logger, util} from 'appium/support'`.

In summary:

- The entirety of `@appium/base-driver` is exposed as `appium/driver`
- The entirety of `@appium/base-plugin` is exposed as `appium/plugin`
- The entirety of `@appium/support` is exposed as `appium/support`

Closes #16474.


boneskull:
* 2.0
  * **PR #16901** <a href="https://app.graphite.dev/github/pr/appium/appium/16901" target="_blank">Graphite</a>  (this PR)
    * **PR #16903** <a href="https://app.graphite.dev/github/pr/appium/appium/16903" target="_blank">Graphite</a> 
    * **PR #16904** <a href="https://app.graphite.dev/github/pr/appium/appium/16904" target="_blank">Graphite</a> 
    * **PR #16905** <a href="https://app.graphite.dev/github/pr/appium/appium/16905" target="_blank">Graphite</a> 
    * **PR #16906** <a href="https://app.graphite.dev/github/pr/appium/appium/16906" target="_blank">Graphite</a> 
    * **PR #16907** <a href="https://app.graphite.dev/github/pr/appium/appium/16907" target="_blank">Graphite</a> 
    * **PR #16908** <a href="https://app.graphite.dev/github/pr/appium/appium/16908" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16901?utm_source=stack-comment).

mykola-mokhnach:
This PR is too huge to review 🤷

mykola-mokhnach:
> In summary:
> 
> The entirety of @appium/base-driver is exposed as appium/driver
> The entirety of @appium/base-plugin is exposed as appium/plugin
> The entirety of @appium/support is exposed as appium/support

Will it work for drivers installed into custom folders, e.g. `$HOME/.appium`. Are there any requirements to where/how the monorepo must be installed in order for this strategy to work properly?

boneskull:
It's easier to review if you hide whitespace from the diff:

image

boneskull:
> Will it work for drivers installed into custom folders, e.g. $HOME/.appium. Are there any requirements to where/how the server monorepo must be installed in order for this strategy to work properly?

By "server monorepo", I assume you mean `appium`.  I should have mentioned this, but a) npm (as run via `@appium/support`)  does not try to install `appium` into `APPIUM_HOME`.  From my testing, it works--I think it's because the extensions are loaded by Appium itself, and thus the `appium` module is already in the require cache.

boneskull:
> Further, if an upgrade of the extension is available, Appium will suggest a version (or versions) to upgrade to. This does not guarantee that the newer version of the driver has the correct peer dependency, but instead just wants to point the user in the right direction. If no upgrade is available, Appium will emit a different suggestion.

I don't really love the implementation here, since it seems weird for `ExtensionConfig` to pull in `ExtensionCommand` to do this, but that's what it does.

boneskull:
The "tree" comment Graphite left is incorrect.  Everything which appears to target #16909 actually targets this PR (#16901).

StephanB75:
Sorry that i ask again here, but it was not really clear for me.
if i use a custom APPIUM_HOME location, i need to install appium also there ?
And if yes, do i install appium itself than directly into the .appium folder ?

boneskull:
@jlipps Yeah, that's basically correct.  The caveat being that if you're using `package.json`, you will need to ensure npm can resolve any peer dependency problems.  For example, if you have a `package.json` with:

```json
{
"dependencies": {
  "appium": "2.2.0",
  "appium-foo-driver": "1.0.0",
  "appium-bar-driver": "3.0.0"
}
```

... and `appium-foo-driver` has a peer dependency of e.g., `appium@2.1.0` but `appium-bar-driver` needs `appium@2.2.0`, then npm will be unable to solve the conflict.  There is no such problem if you're using `APPIUM_HOME`/`appium driver install`--the peer dependencies of the drivers do not prevent installation.  

A user may still get a warning from Appium server at runtime (there are no peer dependency checks at the time of `appium driver install`, but maybe there should be?), but the Appium server will attempt to load the driver anyway.

jlipps:
Fair enough. Good motivation for extension authors to keep their peering requirements as loose as possible.

packages/appium/lib/extension/extension-config.js
```diff
@@ -113,11 +124,94 @@ export class ExtensionConfig {
...
+  /**
+   * Displays non-"fatal" warnings for the extension
+   *
+   * This method uses Appium's logger, since the default extension logger uses the `error` log level.
+   *
    * @param {ExtManifest<ExtType>} extData
    * @param {ExtName<ExtType>} extName
+   * @returns {Promise<void>}
+   */
+  async displayConfigWarnings(extData, extName) {
```
jlipps, line 152: so are we going to be polling npm for possible package updates on every boot of the server? seems like it may be a waste of time just for letting the person know that there's an extension upgrade available.

packages/appium/lib/extension/extension-config.js
```diff
@@ -113,11 +124,94 @@ export class ExtensionConfig {
...
+          log.warn(
+            `${_.capitalize(
+              this.extensionType
+            )} "${extName}" (package \`${pkgName}\`) may be incompatible with the current version of Appium (v${APPIUM_VER}) due to its peer dependency on older Appium v${appiumVersion}. Please ask the developer of \`${pkgName}\` to update the peer dependency on Appium to ${APPIUM_VER}.`
+          );
+        }
+      }
+    } else if (!_.isString(appiumVersion)) {
+      log.warn(
+        `${_.capitalize(
```
jlipps, line 200: can't say i love prettier's choice of line breaks here

packages/appium/lib/main.js
```diff
@@ -398,7 +398,7 @@ export {main, init, resolveAppiumHome};
 
 /**
  * @typedef ServerInitData
- * @property {AppiumDriver} appiumDriver - The Appium driver
+ * @property {import('./appium').AppiumDriver} appiumDriver - The Appium driver
  * @property {import('appium/types').ParsedArgs} parsedArgs - The parsed arguments
```
jlipps, line 402: almost everywhere else you updated this to `@appium/types`, is there a reason you didn't here?

packages/appium/lib/main.js
```diff
@@ -384,10 +384,10 @@ export {finalizeSchema, getSchema, validate} from './schema/schema';
...
 /**
- * @typedef {import('appium/types').DriverType} DriverType
- * @typedef {import('appium/types').PluginType} PluginType
- * @typedef {import('appium/types').DriverClass} DriverClass
- * @typedef {import('appium/types').PluginClass} PluginClass
+ * @typedef {import('@appium/types').DriverType} DriverType
+ * @typedef {import('@appium/types').PluginType} PluginType
+ * @typedef {import('@appium/types').DriverClass} DriverClass
+ * @typedef {import('@appium/types').PluginClass} PluginClass
  * @typedef {import('appium/types').WithServerSubcommand} WithServerSubcommand
```
jlipps, line 391: is there a reason you didn't update this type reference to `@appium/types`?

packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -130,7 +147,12 @@ class JWProxy {
       // we have something like /session/:id/foobar, so we need to replace
       // the session id
       const match = sessionBaseRe.exec(remainingUrl);
-      remainingUrl = remainingUrl.replace(match[1], this.sessionId);
+      // TODO: if `requiresSessionId` is `false` and `sessionId` is `null`, this is a bug.
+      // are we sure `sessionId` is not `null`?
```
jlipps, line 151: i think the `.test()` just above guarantees this?

packages/base-plugin/lib/plugin.js
```diff
@@ -1,68 +1,53 @@
 import {logger} from '@appium/support';
 
-export default class BasePlugin {
-  // plugins can define new methods for the Appium server to map to command names, of the same
```
jlipps, line 4: i don't see this helpful comment reproduced in the `Plugin` interface

packages/base-plugin/lib/plugin.js
```diff
@@ -1,68 +1,53 @@
...
-   *
-   * @return {object} - the result to pass to the user
+   * @param {import('express').Express} expressApp - the Express 'app' object used by Appium for route handling
+   * @param {import('@appium/types').AppiumServer} httpServer - the node HTTP server that hosts the app
    */
-  /*async <cmdName> (next, driver, ...args) {
-    return await next();
-  }*/
+  // eslint-disable-next-line no-unused-vars
+  static async updateServer(expressApp, httpServer) {}
```
jlipps, line 33: since this is now defined, does it trigger the server update logic when loading plugins? we probably don't want it to do that unless plugins implement this?

jlipps (approved):

packages/appium/lib/extension/extension-config.js
boneskull: Only if the peer deps aren't satisfied.

As whether or not to do this, or just ask the user to check for updates themselves, I could go either way--both have benefits and drawbacks.

packages/appium/lib/main.js
boneskull: just putting it here for discoverability: `appium/types` is not `@appium/types`. the former is an alias to the `types` dir in the `appium` package; the latter is the `@appium/types` package itself.

packages/base-driver/lib/jsonwp-proxy/proxy.js
boneskull: It's a bug because you can't (read: _shouldn't_, especially as per TypeScript) attempt to replace a substring with `null`. e.g., `'dull'.replace('dull', null)`

packages/appium/lib/extension/extension-config.js
boneskull: yeah it is pretty aggressive about linebreaks in template literals when there's a function call.  it kind of makes sense, but I try not to care too much

packages/base-plugin/lib/plugin.js
boneskull: ah, whoops.  I've added it to `PluginStatic`.

packages/base-plugin/lib/plugin.js
boneskull: I don't know.  I'll find out.

packages/appium/lib/extension/extension-config.js
jlipps: ok. as long as it only happens in the case where peer deps aren't satisfied, i think that's worth it.

packages/base-plugin/lib/plugin.js
jlipps: ok. that was the reason i originally left it commented out--the plugin loading harness checks for the existence of the function and calls it if it exists; it seemed safer to do it that way so we can explicitly log when a plugin is updating the server in this way to the user. with your change i worry that we wouldn't be able to know when a plugin actually is or isn't doing anything to the server.

packages/base-driver/lib/jsonwp-proxy/proxy.js
jlipps: yeah i was thinking of something else but you are right, maybe there's not a guarantee that `this.sessionId` is set here. at any rate it doesn't appear to be a common bug or we'd have run into it. maybe we should add a check and throw an exception so if it does happen in the future we at least know what's up?

jlipps (approved):

packages/base-driver/lib/jsonwp-proxy/proxy.js
boneskull: My best guess is that we haven't hit the bug because given the happy path, `sessionBaseRe.test(remainingUrl)` is never `true` if there's no session.  That may just mean someone hasn't tried to do something weird with an endpoint--like specify a session ID in the URL when `this.sessionId` is `null`.

So I think then it makes sense to throw if `sessionBaseRe.test(remainingUrl)` is `true` and `sessionId` is `null`.

packages/base-plugin/lib/plugin.js
boneskull: OK.  I've just removed `updateServer` (and `handle`‚ from `BasePlugin`, since a) it's a no-op, and b) it's not necessary.  Any useful comments have been moved into `packages/types/lib/plugin.ts`.

jlipps (approved):

jlipps:
LGTM once lint is happy

boneskull:
hmm..something weird with `lru-cache`

boneskull:
@jlipps Looks like the problem was temporary.  Perhaps in a transitive dep?  A lockfile would have told us more.  😄

PR #16905, opened 2022-05-05
feat(images-plugin): use peer deps

boneskull:
BREAKING CHANGE:

`@appium/images-plugin` now expects to be installed alongside `appium`.

PR #16904, opened 2022-05-05
feat(fake-plugin): use peer deps

boneskull:
BREAKING CHANGE:

`@appium/fake-plugin` now expects to be installed alongside `appium`.

PR #16902, opened 2022-05-05
feat(execute-driver-plugin): use peer deps

boneskull:
BREAKING CHANGE:

`@appium/execute-driver-plugin` now expects to be installed alongside `appium`.

PR #16907, opened 2022-05-05
relocate extension test fixtures into @appium/test-support

boneskull:
This PR moves various test helpers from `@appium/base-driver`, `@appium/base-plugin`, and `appium` into `@appium/test-support`. Building on the new "peer dependency" code, Appium also exports `@appium/test-support` as a `appium/test`, so that extensions can use it.

- `@appium/test-support` now exports the e2e setup for plugins, and e2e/unit suites for drivers from `@appium/base-driver`. it also generates declarations.
- The `capability.spec.js` unit test for basedriver now avoids spying on the "global" logger, due to test flake. It now spies directly on the function which calls the global logger. (@mykola-mokhnach)
- removed homebrew ansi-stripping code from `test-support` and replaced with `strip-ansi`
- type fixes for `@appium/test-support`
- `@appium/fake-driver` consumes `appium/test`


boneskull:
* 2.0
  * **PR #16901** <a href="https://app.graphite.dev/github/pr/appium/appium/16901" target="_blank">Graphite</a> 
    * **PR #16903** <a href="https://app.graphite.dev/github/pr/appium/appium/16903" target="_blank">Graphite</a> 
    * **PR #16904** <a href="https://app.graphite.dev/github/pr/appium/appium/16904" target="_blank">Graphite</a> 
    * **PR #16905** <a href="https://app.graphite.dev/github/pr/appium/appium/16905" target="_blank">Graphite</a> 
    * **PR #16906** <a href="https://app.graphite.dev/github/pr/appium/appium/16906" target="_blank">Graphite</a> 
    * **PR #16907** <a href="https://app.graphite.dev/github/pr/appium/appium/16907" target="_blank">Graphite</a>  (this PR)
    * **PR #16908** <a href="https://app.graphite.dev/github/pr/appium/appium/16908" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16907?utm_source=stack-comment).

packages/test-support/package.json
```diff
@@ -45,20 +45,27 @@
     "test:unit": "mocha \"./test/unit/**/*.spec.js\""
   },
   "dependencies": {
-    "@appium/support": "file:../support",
     "@babel/runtime": "7.17.9",
     "bluebird": "3.7.2",
+    "get-port": "5.1.1",
```
mykola-mokhnach, line 50: shell we also review base driver's deps and remove obsolete stuff?

mykola-mokhnach (approved):

mykola-mokhnach:
Should tests pass first?

packages/test-support/package.json
boneskull: I'm not sure exactly what you mean, but neither `@appium/base-driver` nor `@appium/base-plugin` have peer dependencies on Appium, since that introduces a cyclic dependency.  Appium depends on both of these deps directly.

mykola-mokhnach: I mean we could probably remove some unused dependencies from the base driver after this code has been extracted from it

boneskull:
Marking this as a draft: in its current state, it actually depends on #16902, #16903, #16904, #16905, #16906 and #16908.  😄

This is still usable without the whole "peer dependency" thing, but would need to be reworked some.

packages/test-support/package.json
boneskull: ah, hmm, maybe.  I'll take a look, but I suspect all that stuff wound up in the root devDependencies

boneskull: yeah I'm not seeing anything in there that jumps out at me

```json
  "dependencies": {
    "@appium/support": "file:../support",
    "@babel/runtime": "7.17.9",
    "@colors/colors": "1.5.0",
    "async-lock": "1.3.1",
    "asyncbox": "2.9.2",
    "axios": "0.27.2",
    "bluebird": "3.7.2",
    "body-parser": "1.20.0",
    "es6-error": "4.1.1",
    "express": "4.18.1",
    "http-status-codes": "2.2.0",
    "lodash": "4.17.21",
    "lru-cache": "7.8.1",
    "method-override": "3.0.0",
    "morgan": "1.10.0",
    "serve-favicon": "2.5.0",
    "validate.js": "0.13.1"
  }
```

boneskull: wait, maybe axios?

boneskull: nope, it's used for the proxy stuff.

boneskull:
yes, I'm still working on this

boneskull:
`@appium/test-support` (and thus `appium/test`) exports a function `createSessionHelpers()` which driver E2E tests can use to make various types of requests.

```js
// example

import {createSessionHelpers} from 'appium/test';

const helpers = createSessionHelpers(12345, 'localhost');

const {sessionId} = await helpers.startSession({capabilities: {/* ... */}});
const capabilities = await helpers.getSession(sessionId);
const els = await helpers.postCommand(sessionId, 'elements');
const status = await helpers.getCommand('status');
await helpers.endSession(sessionId);
```

this is the shape of `helpers`:

```ts

/**
 * Object returned by {@linkcode createSessionHelpers }
 */
export type SessionHelpers = {
    /**
     * - URL to create a new session
     */
    newSessionURL: string;
    /**
     * - Begin a session
     */
    startSession: (caps: Capabilities) => Promise<any>;
    /**
     * - End a session
     */
    endSession: (sessionId: string) => Promise<any>;
    /**
     * - Get info about a session
     */
    getSession: (sessionId: string) => Promise<any>;
};
```

boneskull:
Just so we're clear: **this is a breaking change**. Any extension leveraging the harnesses currently in `@appium/base-driver` or `@appium/base-plugin` will likely fail builds until they update their imports.  Extensions shouldn't have to change their tests otherwise, however.

mykola-mokhnach:
> Just so we're clear: **this is a breaking change**. Any extension leveraging the harnesses currently in `@appium/base-driver` or `@appium/base-plugin` will likely fail builds until they update their imports. Extensions shouldn't have to change their tests otherwise, however.

@jlipps @boneskull @KazuCocoa 
Lets agree on one important thing regarding breaking changes. We have a dozen of drivers plus different helper packages that use shared Appium modules. After breaking changes are published it is necessary to update all of them plus keep the conversation going in Issues thread if necessary. It is a lot of work. That is why my proposal would be the one who is making any breaking change in the common code is automatically assigned as its owner and should be responsible for updating the dependant stuff or, at least, on finding someone to bump the dependencies (and also explain the correct way it should be done/reviewing the changes). In case dependencies are still out of date and failing for, let say, more than one week, I feel obliged to revert breaking changes that caused the instability and/or unpublish them.

I am open to any other proposals on this topic. Thanks.

boneskull:
@mykola-mokhnach 

What do you mean by "automatically assigned as its owner"?  What sort of automation are you proposing?

IMO reverting changes should be considered on a case-by-case basis.  As a project, I don't think we're at the "maturity level" to add any sort of process/guarantee like what you're suggesting.

I _do_ agree that whoever is responsible for the breaking change should follow up with external changes. I know I haven't been on the ball about this, and don't want to cause you or anyone else more work that I should be doing.

mykola-mokhnach:
> What do you mean by "automatically assigned as its owner"? What sort of automation are you proposing?

No automation, it's just about taking responsibility for end-to-end features implementation (e.g. if I start it then I know how to finish it and what needs to be done to finish the implementation in a foreseeable time period). We already had some moments in the past when most of the depending modules where broken/impossible to run tests on/etc because of breaking changes in one or more fundamental classes and it was quite frustrating, especially if such state continues for longer time period.

Also, having the project in such state does not add confidence to users as the stability of existing features obviously has the highest priority for them

jlipps:
Good conversation. What about this: the person proposing the breaking change, before it is merged, must (a) have completed PRs for updates in depending modules (ideally all, but say 2-3 at a minimum, if we want to share the work of making updates to a large number of depending modules), and (b) have a plan with buy-in for others as to how we are complete the required updates.

I agree that we haven't been great about this in the past. I think we shouldn't be afraid of breaking changes but we do need to take more responsibility about the downstream effects.

So to role-play how this might have worked with the "Appium is now a peer dep" breaking change:

1. @boneskull proposes the idea, and gets sufficient buy-in for the idea in principle
2. @boneskull shows a PR for the appium monorepo changes
3. @boneskull shows 2-3 other PRs in depending driver repos so the team is aware how to make updates with this kind of breaking change
4. An issue is created to track every known place where we need to make an update (a list of all drivers or depending packages drawn from NPM)
5. @boneskull gets clarity on who will help tackle updating the other packages and in what timeframe (depending on the kind of change, we might set a tight timeline or a looser timeline)
6. Now we merge the initial PRs, publish the updates, and begin the work of rolling updates to other packages.

Definitely a little more process than we are used to, but probably for the best when it comes to breaking changes.

What do you all think?

boneskull:
jumping back into this since it's relevant to the tests I'm working on now...

@jlipps that sounds fine.  I do think though that the breaker should probably open PRs for _all_ affected modules--assuming they have the knowledge to do so themselves (though, if they _don't_, maybe they shouldn't be breaking anything in the first place).  I am not sure, however, how we'd expect CI to pass in any of those PRs until the change was merged.

boneskull:
You know what would make this easier?  All drivers live in this monorepo. 😝

jlipps:
> You know what would make this easier? All drivers live in this monorepo. 😝

yeah well there are pros and cons :-)

boneskull:
While this is technically breaking, I'm not finding any usages of these reusable suites in Appium v2 outside of this monorepo.

boneskull:
IMO this is ready-to-go once CI passes.

jlipps (approved):

PR #16911, opened 2022-05-05
chore: reformat everything

boneskull:
this reformats all `.js`, `.ts`, and `.d.ts` files.

there were some lint fixes necessary too after formatting, as it moved some "eslint ignore" comments around.  

after creating `.prettierignore`, I went into the root and ran `prettier --write .` and that's it.

you're not going to want to look at the diffs.


jlipps:
obviously not going to review. i'm fine to merge but should also get +1 from @mykola-mokhnach and @KazuCocoa so we all know what's going on. is there anyone with some WIP that this would cause unmanageable conflicts for, for example?

mykola-mokhnach (approved):

mykola-mokhnach:
It took my browser some seconds to even switch the radiobox. I'm ok with the formatting in general. Only the 80 chars line length limit seems too strict to me (I'd allow more).

KazuCocoa (approved):
i'm fine, too. I checked a few files, but no objections :)

jlipps:
> 80 chars

yeah i'd personally go for a bit more these days as well, like 100

jlipps (commented):
@boneskull want to bump up the line length and then rebase to clear conflicts? i think we'll be ready to merge then.

boneskull:
I can do that, but from Prettier's docs:

> In code styleguides, maximum line length rules are often set to 100 or 120. However, when humans write code, they don’t strive to reach the maximum number of columns on every line. Developers often use whitespace to break up long lines for readability. In practice, the average line length often ends up well below the maximum.

> Prettier’s printWidth option does not work the same way. It is not the hard upper allowed line length limit. It is a way to say to Prettier roughly how long you’d like lines to be. Prettier will make both shorter and longer lines, but generally strive to meet the specified printWidth.

> Remember, computers are dumb. You need to explicitly tell them what to do, while humans can make their own (implicit) judgements, for example on when to break a line.

> In other words, don’t try to use printWidth as if it was ESLint’s [max-len](https://eslint.org/docs/rules/max-len) – they’re not the same. max-len just says what the maximum allowed line length is, but not what the generally preferred length is – which is what printWidth specifies.

boneskull:
Just going to wait on rebasing my other PRs until this lands.

boneskull:
pullin' the trigger

jlipps:
🎉

PR #16925, opened 2022-05-09
fix(appium): fix extension autoinstall postinstall script

### LINKED ISSUE #16924: appium's postinstall is broken
It looks like it was not updated correctly as files were moved and migrated. 

Repro:
```
cd packages/appium/scripts
npm_config_drivers=xcuitest,uiautomator2 node ./postinstall.js
```

boneskull:
Resolves #16924

* * *

- This is a little more involved than I would have liked.  Maybe it'd make more sense to just spawn e.g., `appium driver install` child processes?
- Since this script exits with a non-zero code, it can prevent installation of Appium altogether, which seems undesirable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,8 +97,7 @@
     "@appium/test-support": "file:packages/test-support",
     "@appium/types": "file:packages/types",
     "@appium/universal-xml-plugin": "file:packages/universal-xml-plugin",
-    "appium": "file:packages/appium",
-    "lru-cache": "7.9.0"
+    "appium": "file:packages/appium"
   },
   "devDependencies": {
     "@appium/fake-plugin": "1.3.1",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
   "scripts": {
     "build": "run-s build:distfiles build:types",
     "build:distfiles": "lerna run --parallel --prefix --concurrency=8 build",
-    "build:loose": "run-s -c build:distfiles build:types || true",
     "build:types": "tsc -b",
     "clean": "run-s clean:artifacts clean:packages clean:monorepo",
     "clean:artifacts": "run-s clean:distfiles clean:types && npx rimraf \"packages/*/build\"",
@@ -39,7 +38,7 @@
     "dev:types": "tsc -b --watch",
     "doctor": "node ./packages/doctor",
     "generate-docs": "lerna run --scope=appium generate-docs",
-    "postinstall": "npm run build:loose",
+    "prepare": "npm run build:types || true",
     "install-fake-driver": "npm start -- driver install --source=local ./packages/fake-driver",
     "install-mjpeg-consumer": "npm install mjpeg-consumer --no-save",
     "lint": "eslint .",
@@ -77,8 +76,8 @@
   },
   "prettier": {
     "bracketSpacing": false,
-    "singleQuote": true,
-    "printWidth": 100
+    "printWidth": 100,
+    "singleQuote": true
   },
   "dependencies": {
     "@appium/base-driver": "file:packages/base-driver",
@@ -98,7 +97,8 @@
     "@appium/test-support": "file:packages/test-support",
     "@appium/types": "file:packages/types",
     "@appium/universal-xml-plugin": "file:packages/universal-xml-plugin",
-    "appium": "file:packages/appium"
+    "appium": "file:packages/appium",
+    "lru-cache": "7.9.0"
   },
   "devDependencies": {
     "@appium/fake-plugin": "1.3.1",
--- a/packages/appium/scripts/postinstall.js
+++ b/packages/appium/scripts/postinstall.js
@@ -1,51 +1,127 @@
 #!/usr/bin/env node
+
 /* eslint-disable no-console, promise/prefer-await-to-then */
 
-async function main() {
-  const driverEnv = process.env.npm_config_drivers;
-  const pluginEnv = process.env.npm_config_plugins;
+/** @type {import('../lib/cli/extension').runExtensionCommand} */
+let runExtensionCommand;
+/** @type {import('../lib/constants').DRIVER_TYPE} */
+let DRIVER_TYPE;
+/** @type {import('../lib/constants').PLUGIN_TYPE} */
+let PLUGIN_TYPE;
+/** @type {import('../lib/extension').loadExtensions} */
+let loadExtensions;
 
-  if (!driverEnv && !pluginEnv) {
-    console.log('Not auto-installing any drivers or plugins');
+const {env, util, logger} = require('@appium/support');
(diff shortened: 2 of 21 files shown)


packages/appium/scripts/postinstall.js
```diff
@@ -14,20 +17,34 @@ async function main () {
...
+  const {loadExtensions} = require('../build/lib/extension');
   const {runExtensionCommand} = extension;
-  const appiumHome = process.env.npm_config_appium_home || DEFAULT_APPIUM_HOME;
-  const specs = [[DRIVER_TYPE, driverEnv], [PLUGIN_TYPE, pluginEnv]];
+  const specs = [
+    [DRIVER_TYPE, driverEnv],
+    [PLUGIN_TYPE, pluginEnv],
+  ];
+
+  const appiumHome = process.env.npm_config_appium_home || (await env.resolveAppiumHome());
```
mykola-mokhnach, line 33: shouldn't `npm_config_appium_home` be also considered in `resolveAppiumHome` call?

packages/appium/scripts/postinstall.js
```diff
@@ -36,35 +53,41 @@ async function main () {
...
-    suppressOutput: true,
-  }, type);
+  await runExtensionCommand(
+    {
+      appiumHome,
+      [`${type}Command`]: 'install',
+      suppressOutput: true,
+      [type]: ext,
+    },
+    type === 'driver' ? driverConfig : pluginConfig
```
mykola-mokhnach, line 85: magic string?

packages/appium/scripts/postinstall.js
boneskull: I'm not sure what you're asking.  Are you asking if `resolveAppiumHome()` should consider that environment variable in addition to `APPIUM_HOME`?

mykola-mokhnach: yes, exactly

boneskull:
I've updated the `postinstall.js` script to _never_ exit with a non-zero code; errors will be displayed, but the installation will continue.

packages/appium/scripts/postinstall.js
jlipps: i think we settled on just leaning on APPIUM_HOME. i think this code came from a time when we had the `--appium-home` command line flag, and it mirrors that. but we no longer have that flag.

packages/appium/scripts/postinstall.js
```diff
@@ -10,23 +23,15 @@ async function main() {
...
-  const {DEFAULT_APPIUM_HOME, DRIVER_TYPE, PLUGIN_TYPE} = require('./build/lib/extension-config');
-  const {runExtensionCommand} = extension;
-  const appiumHome = process.env.npm_config_appium_home || DEFAULT_APPIUM_HOME;
   const specs = [
     [DRIVER_TYPE, driverEnv],
     [PLUGIN_TYPE, pluginEnv],
   ];
 
+  const appiumHome =
+    process.env.npm_config_appium_home || (await env.resolveAppiumHome());
```
jlipps, line 32: i *think* we can do away with this appium home npm install flag and just rely on APPIUM_HOME or the standard env discovery.

jlipps (approved):
i'm not sure if the right thing is actually to prevent installation of appium? i mean if someone requests drivers as part of the install, usually in a CI environment, wouldn't they want to know if those drivers werent installed before running their tests?

boneskull:
@jlipps 

> i'm not sure if the right thing is actually to prevent installation of appium? i mean if someone requests drivers as part of the install, usually in a CI environment, wouldn't they want to know if those drivers werent installed before running their tests?

How about a compromise: fail if `process.env.CI`?

packages/appium/scripts/postinstall.js
boneskull: OK.  Though ... I see a problem I hadn't thought of before.

This script is only applicable if Appium _isn't_ a direct dependency within the closest `package.json`. If that's the case, then it's expected these drivers/plugins will _also_ be direct dependencies.

I'll address this

boneskull:
I made it nicer.   Here's some screenshots, but there's also _spinners_:

image

image

image

packages/appium/scripts/postinstall.js
```diff
@@ -1,83 +1,162 @@
...
 
-  let extension;
-  try {
-    extension = require('../build/lib/cli/extension');
-  } catch (e) {
-    throw new Error(
-      `Could not load extension CLI file; has the project been transpiled? ` + `(${e.message})`
+  if (!driverEnv && !pluginEnv) {
+    spinner.succeed(
+      `No drivers or plugins to automatically install. If desired, set the "npm_config_drivers" or "npm_config_plugins" environment variable(s) to a comma-separated list, and they will be installed automatically with Appium.`
```
jlipps, line 62: NPM sets these env vars. What the user does is add them as flags to npm install, e.g.,

```
npm install -g appium --drivers=fake,xcuitest
```

then NPM magically turns `--drivers` into `process.env.npm_config_drivers`.

all that to say, we should probably refer to the `--drivers` and `--plugins` way of doing things because that is the intended user facing feature.

jlipps (changes requested):
looks good, one comment

packages/appium/scripts/postinstall.js
boneskull: Ah, k.

jlipps (approved):
LGTM when CI passes

boneskull:
ugh what have I done

boneskull:
OK, so basically: we can't require anything from `@appium/support` here because it won't have been built yet.  The `postinstall` scripts in the subpackages run before the postinstall script in the monorepo.  I think what we can do is add `prepare` scripts to each subpackage which call `npm run build`. These `prepare` scripts only get called before publish and on an `npm install` in a working copy.

boneskull:
* 2.0

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16925?utm_source=stack-comment).

jlipps (approved):

package.json
```diff
@@ -98,7 +97,8 @@
     "@appium/test-support": "file:packages/test-support",
     "@appium/types": "file:packages/types",
     "@appium/universal-xml-plugin": "file:packages/universal-xml-plugin",
-    "appium": "file:packages/appium"
+    "appium": "file:packages/appium",
+    "lru-cache": "7.9.0"
```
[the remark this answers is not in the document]
boneskull: hmm, I was trying to fix the `lru-cache` bug.  I'll rip that out and try again.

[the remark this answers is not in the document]
boneskull: yeah, wasn't needed.

Issue #16916, opened 2022-05-07
NPM 8: ChromeDriver is not installed with Appium


mirao:
## The problem
I would expect that this command installs appium and ChromeDriver. But no "ChromeDriver postinstall" step is performed:
```bash
$ sudo npm install -g appium@1.15 --chromedriver_version="98.0.4758.102"
added 588 packages, and audited 589 packages in 27s

37 vulnerabilities (6 low, 21 moderate, 8 high, 2 critical)

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
```
It only installs appium and doesn't install ChromeDriver (missing folder `chromedriver`):
```
$ ll /usr/lib/node_modules/appium/node_modules/appium-chromedriver
total 64
drwxr-xr-x   4 root root  4096 May  7 16:05 ./
drwxr-xr-x 446 root root 20480 May  7 16:05 ../
drwxr-xr-x   3 root root  4096 May  7 16:05 build/
-rw-r--r--   1 root root   192 May  7 16:05 index.js
-rwxr-xr-x   1 root root  2952 May  7 16:05 install-npm.js*
drwxr-xr-x   2 root root  4096 May  7 16:05 lib/
-rw-r--r--   1 root root 11383 May  7 16:05 LICENSE
-rw-r--r--   1 root root  2075 May  7 16:05 package.json
-rw-r--r--   1 root root  6157 May  7 16:05 README.md
```
Also tested with the options `--unsafe-perm=true --allow-root`, but with the same result, still no success.

If I try the same command with NPM 6.14.17, it works well (it requires `--unsafe-perm=true --allow-root`, see https://github.com/appium/appium/issues/10020):

```bash
$ sudo npm install -g appium@1.15 --chromedriver_version="98.0.4758.102" --unsafe-perm=true --allow-root
/usr/bin/appium -> /usr/lib/node_modules/appium/build/lib/main.js
/usr/bin/authorize-ios -> /usr/lib/node_modules/appium/node_modules/.bin/authorize-ios

> appium-windows-driver@1.6.0 install /usr/lib/node_modules/appium/node_modules/appium-windows-driver
> node install-npm.js

Not installing WinAppDriver since did not detect a Windows system

> pre-commit@1.2.2 install /usr/lib/node_modules/appium/node_modules/pre-commit
> node install.js


> wd@1.11.4 install /usr/lib/node_modules/appium/node_modules/wd
[...]

PR #16928, opened 2022-05-11
chore(appium,support): cache result of check for local appium

boneskull:
This is just a perf improvement.  Anecdotally, this shaves 1-to-5 seconds off of an `appium driver` or `appium plugin` command, depending on system resources (lower end being CI VM's).

While poking around looking for a solution to #16916, I noticed that `npm.list()` in `@appium/support` was getting called twice.  Each call is not fast; an `npm` child process must be forked.  It ended up being two separate calls to `hasAppiumDependency()`.  First, we need to resolve the Appium home dir, so we look for `appium` in `npm ls` output in the directory containing the closest `package.json`.  Second, Appium's `Manifest` class needs to know this information, so it also calls `hasAppiumDependency()` which results in another `npm ls`.  However, that is called with the value of Appium home--not necessarily the same directory as the first call.

It could be argued that the call in `Manifest` is actually a bug, since it shouldn't look at Appium home--it should look at the dir of the closest `package.json`. However, if the result of `hasAppiumDependency(appiumHome)` is `true` then this _is_ the same directory.  But semantically, it's incorrect.  Anyway, it was weird.  To avoid all this nonsense, I decided to just determine the answer earlier so we only need to do it _once_, and hand the result to the `Manifest` factory when loading extensions.  If it's called again for whatever reason, the result will now be memoized (as will the call to find the closest `package.json`).

### DIFF
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -2,7 +2,7 @@
 
 import _ from 'lodash';
 import path from 'path';
-import {npm, fs, util, env} from '@appium/support';
+import {npm, fs, util} from '@appium/support';
 import {log, spinWith, RingBuffer} from './utils';
 import {SubProcess} from 'teen_process';
 import {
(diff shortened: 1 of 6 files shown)


boneskull:
this needs more time to bake

boneskull:
* 2.0
  * **PR #16928** <a href="https://app.graphite.dev/github/pr/appium/appium/16928" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16928?utm_source=stack-comment).

jlipps:
ok, let us know when it is baked enough for review

boneskull:
This is too out-of-sync

PR #16953, opened 2022-05-16
chore(appium,base-plugin,types): plugin name is derived from manifest

boneskull:
This PR changes the shape of plugin classes as well as how CLI args get assigned.

- The `pluginName` static property has been removed.  `name` is now assigned via the `BasePlugin` constructor, and is derived from the plugin name in the manifest.  This means that an implementor only needs to provide a `pluginName` in the manifest, and the data is not duplicated.  The value in this map (the plugin name) is used to instantiate the plugin, instead of passing static `pluginName` property into the constructor.
- CLI args are now passed in through the constructor instead of assigned to the plugin instance after the fact.

Implementation details:

`getActivePlugins()` and `getActiveDrivers()` now return a `Map` of an extension _class_ to an extension name.  The key is _not_ an instance, and is effectively a singleton, so we are not at risk of a memory leak.

This change does not otherwise affect drivers.

* * *

AFAICT this is not a breaking change.

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -14,6 +14,7 @@ import AsyncLock from 'async-lock';
 import {parseCapsForInnerDriver, pullSettings} from './utils';
 import {util, node, logger} from '@appium/support';
 import {getDefaultsForExtension} from './schema';
+import {DRIVER_TYPE, PLUGIN_TYPE} from './constants';
 
 /**
  * Invariant set of base constraints
@@ -62,9 +63,9 @@ class AppiumDriver extends DriverCore {
 
   /**
    * List of active plugins
-   * @type {PluginClass[]}
+   * @type {Map<PluginClass,string>}
    */
-  pluginClasses = [];
+  pluginClasses;
 
   /**
    * map of sessions to actual plugin instances per session
@@ -180,10 +181,9 @@ class AppiumDriver extends DriverCore {
    * @template {ExtensionType} ExtType
    * @param {ExtType} extType 'driver' or 'plugin'
    * @param {string} extName the name of the extension
-   * @param {InstanceType<import('appium/types').ExtClass<ExtType>>} extInstance the driver or plugin instance
-   * @returns {void}
+   * @returns {Record<string,any>|undefined}
    */
-  assignCliArgsToExtension(extType, extName, extInstance) {
+  getCliArgsForExtension(extType, extName) {
     const allCliArgsForExt = /** @type {Record<string,unknown>|undefined} */ (
       this.args[extType]?.[extName]
     );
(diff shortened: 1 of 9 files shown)


jlipps (approved):
yeah, makes sense

jlipps:
(once the build passes anyway)

boneskull:
@jlipps 

the failure is regarding the setting of CLI args.  Let's pretend the fix is obvious and I am totally not trying to weasel out of anything.  

in a test, the `cliArgs` prop of a plugin instance is `{}` when it was expected to be `undefined`.  This property is expected to be `undefined` if there are _no non-default arguments_ provided by a user.  I see a benefit in passing `undefined` in that a plugin author can quickly check if _any_ such arguments have been provided (`if (this.cliArgs)`).

_But_ ... the way this works now is that if a plugin defines default arguments, `cliArgs` will not provide them (I wrote this code, but it was just a refactor of what was previously there... I think).    

```js
  getCliArgsForExtension(extType, extName) {
    // this includes whatever the user provided + the default values
    const allCliArgsForExt = /** @type {Record<string,unknown>|undefined} */ (
      this.args[extType]?.[extName]
    );
    if (!_.isEmpty(allCliArgsForExt)) {
      const defaults = getDefaultsForExtension(extType, extName);
      const cliArgs = _.isEmpty(defaults)
        ? allCliArgsForExt
        // omits any values equal to default values. I do not really know why
        : _.omitBy(allCliArgsForExt, (value, key) => _.isEqual(defaults[key], value)); 
      if (!_.isEmpty(cliArgs)) {
        return cliArgs;
      }
    }
  }
```

This seems kind of odd to me, given that a plugin author will have to _duplicate_ the definition of any such default values (like assigning them in the constructor), because they will not be present in `cliArgs`.  This presents an opportunity for them to get out of sync, of course.  Perhaps more confusingly, if a user passes `--foo=1` to the plugin, and there is no default set in the schema for `foo`, then `cliArgs.foo === 1`.  But if the default of `foo` is set to `1`, then `cliArgs === undefined`.

For reasons of consistency, I think it's probably better to:

- Always provide an object for `cliArgs`, even if it's empty.
- Always provide default values in the `cliArgs` object (unless a user overrides them on the CLI)

Such a change would mean that plugin authors would not be able to easily tell which props of `cliArgs` were user-provided and which were defaults, but since the defaults are accessible within the plugin schema (and at runtime via the schema API) if an author really needs this, they are not SOL.

I know this diverges from how the drivers work...but this already diverges from how the drivers work.  It would be nice to get the drivers and plugins in sync with a more consistent API but is probably not worth doing (breaking change!) unless extension authors are becoming confused.

jlipps:
> Such a change would mean that plugin authors would not be able to easily tell which props of cliArgs were user-provided and which were defaults, but since the defaults are accessible within the plugin schema (and at runtime via the schema API) if an author really needs this, they are not SOL.

Yeah I think I'm OK with this.

boneskull:
None of the plugins seem to leverage `cliArgs` currently except `fake-plugin`.

PR #16966, opened 2022-05-19
feat(appium,support): extension check improvements

boneskull:
This PR changes a few things:

- Extension problems are now displayed correctly at `appium server` startup. This was done by changing the default `npmlog` level, then reesetting it to `silent` when `winston` takes over (in `logsink.init()`)
- Extension problems are also displayed at install time.
- Improved performance of `appium <driver|plugin> install` in many cases by actually checking if the thing is installed _before_ installing it.
- `ExtensionCommand` uses a new console logging mechanism (`console` in `@appium/support`) tailored to CLI usage, since it is intended to be run via the CLI. Using this, we handle the result of extension checks both at server runtime and during CLI tool usage in the appropriate manner. The implementation is just a dumb wrapper around `console.log` and should eventually be removed when we revisit the logging architecture.
- Separated extension checks further into "errors" ("Problems") and "warnings". As these demand distinct log levels, `ExtensionConfig` no longer accepts a `logFn` in its constructor (since this restricted to a single level ("error"))
- Generally improved specificity, helpfulness, and reduced extraneous verbosity (like this sentence) of warnings
- Removed `strip-colors`, as this functionality is present in `@colors/colors`. The E2E test helper now automatically strips colors from child processes.
- More tests

---

The "peer dep" PRs will be targeting this PR (after I do that).


boneskull:
* 2.0
  * **PR #16966** <a href="https://app.graphite.dev/github/pr/appium/appium/16966" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16966?utm_source=stack-comment).

packages/appium/lib/cli/extension-command.js
```diff
@@ -56,6 +57,26 @@ class ExtensionCommand {
...
+   * in other words, _calling_ a function which always throws an exception is not enough.
+   * @param {string|Error} msgOrError
+   * @protected
+   * @returns {Error}
+   */
+  _fatalError(msgOrError) {
+    if (_.isString(msgOrError)) {
+      return new Error(this.log.decorate(msgOrError, 'error'));
+    }
+    if (_.isString(msgOrError.message)) {
```
mykola-mokhnach, line 73: would this be enough to assume `msgOrError ` is of Error type or we have to check that explicitly?

packages/appium/lib/cli/extension-command.js
```diff
@@ -56,6 +57,26 @@ class ExtensionCommand {
...
+  /**
+   * Logs a message and returns an {@linkcode Error} to throw.
+   *
+   * For TS to understand that a function throws an exception, it must actually throw an exception;
+   * in other words, _calling_ a function which always throws an exception is not enough.
+   * @param {string|Error} msgOrError
+   * @protected
+   * @returns {Error}
+   */
+  _fatalError(msgOrError) {
```
mykola-mokhnach, line 69: I usually add verbs to method names

packages/appium/lib/cli/extension-command.js
```diff
@@ -245,30 +276,68 @@ class ExtensionCommand {
...
           installType = INSTALL_TYPE_NPM;
         }
       }
+      installOpts = {installSpec, pkgName, pkgVer};
+    }
 
-      extData = await this.installViaNpm({installSpec, pkgName, pkgVer});
+    // fail fast here if we can
+    if (probablyExtName && this.config.isInstalled(probablyExtName)) {
+      throw this._fatalError(
```
mykola-mokhnach, line 293: not sure if this should be an error or just a warning

packages/appium/lib/cli/extension-command.js
```diff
@@ -552,7 +619,7 @@ class ExtensionCommand {
     await this.config.updateExtension(installSpec, extData);
   }
 
-  /**
+  /**Promise
```
mykola-mokhnach, line 622: typo here?

packages/appium/lib/logsink.js
```diff
@@ -8,7 +8,7 @@ logger.patchLogger(npmlog);
 global._global_npmlog = npmlog;
 
 // npmlog is used only for emitting, we use winston for output
```
mykola-mokhnach, line 10: is the comment still relevant?

packages/appium/lib/cli/extension-command.js
boneskull: if the thing is already installed?  it was an error before.  now it's just an error...earlier.

packages/appium/lib/logsink.js
boneskull: it's... mostly still relevant.  it's true when the server is running.

packages/appium/lib/cli/extension-command.js
boneskull: I think it's probably fair to restrict it to `string`, since I'm not actually passing an error elsewhere. We can add that later if needed.

packages/appium/lib/cli/extension-command.js
boneskull: not a typo.  sometimes refactoring in vscode while files are changing in a background process causes mass hysteria.  these were everywhere.  good catch.

packages/appium/lib/cli/extension-command.js
boneskull: I could go for "if already installed, exit with a nice message instead of a scary error"

boneskull:
This PR will eventually contain all of the other peer-deps PRs for plugins (and `fake-driver`) in this repo.  `fake-driver` and `fake-plugin` have been merged into this branch.  I screwed something up so had to manually reapply those PRs

boneskull:
I don't think Lerna is going to issue updates to these `peerDependencies` fields when we `lerna publish`.  I think we need to manage them manually (this shouldn't be a big deal).  

As a rule, peer deps should be as broad as possible.  So, to that end, we'll start with whatever the current version of Appium is, and either allow minor or patch updates.  _If we allow minor updates_, we need to agree that Appium's version number simply reflects semver--not for marketing reasons--or we will quickly find ourselves hamstrung.  I _assume_ we want to follow semver strictly going forward, but I don't want to assume.  😄

boneskull:
As an example, the `appium` peer dep of all of these will be something like `^2.0.0-beta.36`.

boneskull:
Also note:

Any change to peer dependencies in any of these plugins is a breaking change in the plugin itself.  Let's keep it broad for now and narrow later if needed.

boneskull:
@jlipps this is ready.  ignore everything in the plugin dirs, as those are from PRs that you already approved.

jlipps (approved):

packages/appium/lib/cli/extension-command.js
```diff
@@ -245,30 +270,68 @@ class ExtensionCommand {
...
+          `Did you mean to update? Run "appium ${this.type} update". See ` +
+          `installed ${this.type}s with "appium ${this.type} list --installed".`
+      );
     }
 
+    extData = await this.installViaNpm(installOpts);
+
+    // this _should_ be the same as `probablyExtName` as the one derived above unless
+    // install type is local.
     const extName = extData[/** @type {string} */ (`${this.type}Name`)];
```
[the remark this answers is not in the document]
boneskull: IDK:

```
Type '`${ExtType}Name`' cannot be used to index type 'ExtensionFields<ExtType>'. ts(2536)
```

packages/appium/lib/cli/extension-command.js
```diff
@@ -56,6 +57,20 @@ class ExtensionCommand {
...
+   * Logs a message and returns an {@linkcode Error} to throw.
+   *
+   * For TS to understand that a function throws an exception, it must actually throw an exception--
+   * in other words, _calling_ a function which is guaranteed to throw an exception is not enough--
+   * nor is something like `@returns {never}` which does not imply a thrown exception.
+   * @param {string} message
+   * @protected
+   * @returns {Error}
+   */
+  _createFatalError(message) {
```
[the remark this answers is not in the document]
boneskull: we already have method `error()` which does something entirely different.  so that's why it's not `_error()`.

the intent here is twofold:

1. the return value of this method should immediately be thrown
2. when the return value is thrown, the process should quickly exit. i.e. a "fatal" error

so... yeah.  how can I improve it?

boneskull:
Alright, this looks OK.

boneskull:
wait... one more test.

boneskull:
build failure is an npm timeout.

PR #16992, opened 2022-05-26
refactor image-handling code

boneskull:
This PR contains improvements as follows:

- Both `@appium/images-plugin` and `@appium/opencv` are now type-error-free. They aren't "fully typechecked" because not everything that needs annotating is annotated. But what _is_ there passes `tsc` without throwing errors.
- `@appium/support` exposes an `AppiumImage` class which wraps a `Jimp` instance. Given that a) we currently monkeypatch a `Jimp` instance because we want to use Jimp's `Promise`-returning methods _by default_ instead of referring specifically to `*Async` methods (the non-`*Async` versions being callback-based), and b) much of Jimp's API is of the "chainable" variety (it's rather haphazard), it is impossible to guarantee the consistency of the return type (is it a `Jimp` or is it a _monkeypatched_ `Jimp`? What does `getBuffer()` return?) when calling chainable methods without some extra work. This PR represents that extra work.
- `getJimpImage()` in `@appium/support`'s `image-util` is now considered deprecated, as `AppiumImage` is superior. It should be eventually removed.

However...

Various drivers may depend on methods in the `Jimp` instance as returned by `getJimpImage()`. If any are missing (e.g., `Jimp#setPixelColor()`, `Jimp#scan()`, etc.), they will need to be explicitly added to `AppiumImage` if drivers are to move away from `getJimpImage()`. Drivers currently assume they are working with a monkeypatched `Jimp` instace; `AppiumImage` does not wrap `Jimp` _in its entirety_.

Further, `@appium/opencv` may benefit from using `AppiumImage`, but this has not been changed.

Other notes:

- I ended up _not_ changing the return value of `ImageElementFinder#findByImage()` because it was a rabbit hole. I don't think external code interacts with `ImageElementFinder` directly, so this is more of a nice-to-have.
- There are some other methods (e.g., `cropImage()`, `imageToBase64()`) in the `image-util` module which seem to operate directly on `Buffer`s instead of `Jimp` instances. I'm unsure why this is, since presumably `Jimp` can crop images. Anyway, some drivers use this code, but it isn't used anywhere else in this codebase.

### DIFF
--- a/packages/base-driver/lib/protocol/errors.js
+++ b/packages/base-driver/lib/protocol/errors.js
@@ -8,8 +8,17 @@ const w3cLog = logger.getLogger('W3C');
 
 const W3C_UNKNOWN_ERROR = 'unknown error';
 
-// base error class for all of our errors
+/**
+ * Base error class for all of our errors
+ */
 export class ProtocolError extends ES6Error {
+  /**
+   *
+   * @param {string} [msg]
+   * @param {*} [jsonwpCode]
+   * @param {*} [w3cStatus]
+   * @param {*} [error]
+   */
   constructor(msg, jsonwpCode, w3cStatus, error) {
     super(msg);
     this.jsonwpCode = jsonwpCode;
@@ -45,6 +54,7 @@ export class NoSuchDriverError extends ProtocolError {
   static error() {
     return 'invalid session id';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'A session is either terminated or not started',
@@ -65,6 +75,7 @@ export class NoSuchElementError extends ProtocolError {
   static error() {
     return 'no such element';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'An element could not be located on the page using the given ' + 'search parameters.',
@@ -85,6 +96,7 @@ export class NoSuchFrameError extends ProtocolError {
   static w3cStatus() {
     return HTTPStatusCodes.NOT_FOUND;
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -107,6 +119,7 @@ export class UnknownCommandError extends ProtocolError {
   static error() {
     return 'unknown command';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -130,6 +143,9 @@ export class StaleElementReferenceError extends ProtocolError {
   static error() {
     return 'stale element reference';
   }
+  /**
+   * @param {string} [err] - the error message
+   */
   constructor(err) {
     super(
       err ||
@@ -152,6 +168,7 @@ export class ElementNotVisibleError extends ProtocolError {
   static error() {
     return 'element not visible';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -174,6 +191,7 @@ export class InvalidElementStateError extends ProtocolError {
   static error() {
     return 'invalid element state';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -196,9 +214,10 @@ export class UnknownError extends ProtocolError {
   static error() {
     return W3C_UNKNOWN_ERROR;
   }
+  /** @param {string|Error} [errorOrMessage] */
   constructor(errorOrMessage) {
-    const origMessage = _.isString((errorOrMessage || {}).message)
-      ? errorOrMessage.message
+    const origMessage = _.isString(/** @type {Error} */ ((errorOrMessage) || {}).message)
+      ? /** @type {Error} */ (errorOrMessage).message
       : errorOrMessage;
     const message =
       'An unknown server-side error occurred while processing the command.' +
@@ -217,6 +236,7 @@ export class UnknownMethodError extends ProtocolError {
   static error() {
     return 'unknown method';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'The requested command matched a known URL but did not match an method for that URL',
@@ -237,6 +257,7 @@ export class UnsupportedOperationError extends ProtocolError {
   static error() {
     return 'unsupported operation';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'A server-side error occurred. Command cannot be supported.',
@@ -257,6 +278,7 @@ export class ElementIsNotSelectableError extends ProtocolError {
   static w3cStatus() {
     return HTTPStatusCodes.BAD_REQUEST;
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'An attempt was made to select an element that cannot be selected.',
@@ -299,6 +321,7 @@ export class ElementNotInteractableError extends ProtocolError {
   static w3cStatus() {
     return HTTPStatusCodes.BAD_REQUEST;
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -314,6 +337,7 @@ export class InsecureCertificateError extends ProtocolError {
   static error() {
     return 'insecure certificate';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -335,6 +359,7 @@ export class JavaScriptError extends ProtocolError {
   static error() {
     return 'javascript error';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'An error occurred while executing user supplied JavaScript.',
@@ -355,6 +380,7 @@ export class XPathLookupError extends ProtocolError {
   static error() {
     return 'invalid selector';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'An error occurred while searching for an element by XPath.',
@@ -375,6 +401,7 @@ export class TimeoutError extends ProtocolError {
   static error() {
     return 'timeout';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'An operation did not complete before its timeout expired.',
@@ -395,6 +422,7 @@ export class NoSuchWindowError extends ProtocolError {
   static w3cStatus() {
     return HTTPStatusCodes.NOT_FOUND;
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -417,6 +445,7 @@ export class InvalidArgumentError extends ProtocolError {
   static w3cStatus() {
     return HTTPStatusCodes.BAD_REQUEST;
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'The arguments passed to the command are either invalid or malformed',
@@ -437,6 +466,7 @@ export class InvalidCookieDomainError extends ProtocolError {
   static w3cStatus() {
     return HTTPStatusCodes.BAD_REQUEST;
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -459,6 +489,7 @@ export class NoSuchCookieError extends ProtocolError {
   static error() {
     return 'no such cookie';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err ||
@@ -480,6 +511,7 @@ export class UnableToSetCookieError extends ProtocolError {
   static error() {
     return 'unable to set cookie';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || "A request to set a cookie's value could not be satisfied.",
@@ -500,6 +532,7 @@ export class UnexpectedAlertOpenError extends ProtocolError {
   static error() {
     return 'unexpected alert open';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'A modal dialog was open, blocking this operation',
@@ -520,6 +553,7 @@ export class NoAlertOpenError extends ProtocolError {
   static error() {
     return 'no such alert';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'An attempt was made to operate on a modal dialog when one ' + 'was not open.',
@@ -542,6 +576,7 @@ export class ScriptTimeoutError extends ProtocolError {
   static error() {
     return 'script timeout';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'A script did not complete before its timeout expired.',
@@ -562,6 +597,7 @@ export class InvalidElementCoordinatesError extends ProtocolError {
   static error() {
     return 'invalid coordinates';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'The coordinates provided to an interactions operation are invalid.',
@@ -584,6 +620,7 @@ export class IMENotAvailableError extends ProtocolError {
   static error() {
     return 'unsupported operation';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'IME was not available.',
@@ -624,6 +661,7 @@ export class InvalidSelectorError extends ProtocolError {
   static error() {
     return 'invalid selector';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'Argument was an invalid selector (e.g. XPath/CSS).',
@@ -644,6 +682,7 @@ export class SessionNotCreatedError extends ProtocolError {
   static error() {
     return 'session not created';
   }
+  /** @param {string} [details] */
   constructor(details) {
     let message = 'A new session could not be created.';
     if (details) {
@@ -669,6 +708,7 @@ export class MoveTargetOutOfBoundsError extends ProtocolError {
   static error() {
     return 'move target out of bounds';
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'Target provided for a move action is out of bounds.',
@@ -683,6 +723,7 @@ export class NoSuchContextError extends ProtocolError {
   static code() {
     return 35;
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(err || 'No such context found.', NoSuchContextError.code());
   }
@@ -692,6 +733,7 @@ export class InvalidContextError extends ProtocolError {
   static code() {
     return 36;
   }
+  /** @param {string} [err] */
   constructor(err) {
     super(
       err || 'That command could not be executed in the current context.',
@@ -702,11 +744,13 @@ export class InvalidContextError extends ProtocolError {
 
 // These are aliases for UnknownMethodError
 export class NotYetImplementedError extends UnknownMethodError {
+  /** @param {string} [err] */
   constructor(err) {
     super(err || 'Method has not yet been implemented');
   }
 }
 export class NotImplementedError extends UnknownMethodError {
+  /** @param {string} [err] */
   constructor(err) {
     super(err || 'Method is not implemented');
   }
--- a/packages/base-driver/lib/protocol/errors.js
+++ b/packages/base-driver/lib/protocol/errors.js
@@ -15,9 +15,9 @@ export class ProtocolError extends ES6Error {
   /**
    *
    * @param {string} [msg]
-   * @param {*} [jsonwpCode]
-   * @param {*} [w3cStatus]
-   * @param {*} [error]
+   * @param {number} [jsonwpCode]
+   * @param {number} [w3cStatus]
+   * @param {string} [error]
    */
   constructor(msg, jsonwpCode, w3cStatus, error) {
     super(msg);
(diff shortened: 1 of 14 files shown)


boneskull:
* 2.0
  * **PR #16966** <a href="https://app.graphite.dev/github/pr/appium/appium/16966" target="_blank">Graphite</a> 
    * **PR #16992** <a href="https://app.graphite.dev/github/pr/appium/appium/16992" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/16992?utm_source=stack-comment).

mykola-mokhnach:
It looks like the PR title is not a canonical one

packages/base-driver/lib/protocol/errors.js
```diff
@@ -8,8 +8,17 @@ const w3cLog = logger.getLogger('W3C');
...
 
-// base error class for all of our errors
+/**
+ * Base error class for all of our errors
+ */
 export class ProtocolError extends ES6Error {
+  /**
+   *
+   * @param {string} [msg]
+   * @param {*} [jsonwpCode]
```
mykola-mokhnach, line 18: number

packages/base-driver/lib/protocol/errors.js
```diff
@@ -8,8 +8,17 @@ const w3cLog = logger.getLogger('W3C');
...
+   * @param {*} [w3cStatus]
```
mykola-mokhnach, line 19: number

packages/base-driver/lib/protocol/errors.js
```diff
@@ -8,8 +8,17 @@ const w3cLog = logger.getLogger('W3C');
...
+   * @param {*} [error]
```
mykola-mokhnach, line 20: string

packages/images-plugin/lib/compare.js
```diff
@@ -8,76 +8,93 @@ const MATCH_TEMPLATE_MODE = 'matchTemplate';
...
-      throw new errors.InvalidArgumentError(
-        `'${mode}' images comparison mode is unknown. ` +
-          `Only ${JSON.stringify([
-            MATCH_FEATURES_MODE,
-            GET_SIMILARITY_MODE,
-            MATCH_TEMPLATE_MODE,
-          ])} modes are supported.`
-      );
+
+  if (isMatchFeaturesMode(mode)) {
```
mykola-mokhnach, line 67: why switch does not work here?

packages/images-plugin/lib/compare.js
```diff
@@ -8,76 +8,93 @@ const MATCH_TEMPLATE_MODE = 'matchTemplate';
...
+    const opts = /** @type {MatchingOptions} */ (options);
+    /** @type {MatchingResult} */
+    let result;
+    try {
+      result = await getImagesMatches(img1, img2, opts);
```
mykola-mokhnach, line 72: we could return right here

packages/images-plugin/lib/compare.js
```diff
@@ -90,4 +107,33 @@ export {
...
+ * @typedef {import('@appium/opencv').SimilarityOptions} SimilarityOptions
+ * @typedef {import('@appium/opencv').MatchingOptions} MatchingOptions
+ * @typedef {import('@appium/opencv').SimilarityResult} SimilarityResult
+ * @typedef {import('@appium/opencv').MatchingResult} MatchingResult
+ */
+
+/**
+ * @template {boolean} Multiple
+ * @typedef {import('@appium/opencv').OccurrenceOptions<Multiple>} OccurrenceOptions
+
```
mykola-mokhnach, line 119: extra line break

packages/images-plugin/lib/compare.js
```diff
@@ -90,4 +107,33 @@ export {
...
+ * @typedef {typeof GET_SIMILARITY_MODE} GetSimilarityMode
+ * @typedef {typeof MATCH_FEATURES_MODE} MatchFeaturesMode
+ * @typedef {typeof MATCH_TEMPLATE_MODE} MatchTemplateMode
+ * @typedef {GetSimilarityMode|MatchFeaturesMode|MatchTemplateMode} CompareMode
+ */
+
+/**
+ * @template {CompareMode} Mode
+ * @template {boolean} Multiple
+ * @typedef {Mode extends GetSimilarityMode ? SimilarityOptions : Mode extends MatchFeaturesMode ? MatchingOptions : Mode extends MatchTemplateMode ? OccurrenceOptions<Multiple> : never} CompareImagesOptions
```
mykola-mokhnach, line 132: this line is long and hard to read

packages/images-plugin/lib/finder.js
```diff
@@ -72,16 +73,19 @@ const DEFAULT_SETTINGS = {
...
+   */
   driver;
 
   /** @type {LRU<string,ImageElement>} */
   imgElCache;
 
   /**
    *
-   * @param {ExternalDriver} driver
+   * @param {ExternalDriver} [driver]
```
mykola-mokhnach, line 88: why brackets are needed around param names?

packages/images-plugin/lib/finder.js
```diff
@@ -103,34 +107,30 @@ export default class ImageElementFinder {
    */
   registerImageElement(imgEl) {
     this.imgElCache.set(imgEl.id, imgEl);
-    const protoKey = this.driver.isW3CProtocol() ? W3C_ELEMENT_KEY : MJSONWP_ELEMENT_KEY;
-    return imgEl.asElement(protoKey);
+    if (!this.driver) {
+      throw new ReferenceError('No driver set!');
```
mykola-mokhnach, line 111: would it be possible to provide more helpful error description?

mykola-mokhnach: also I personally don't like to put exclamation marks into error messages: too much drama %)

packages/images-plugin/lib/finder.js
```diff
@@ -103,34 +107,30 @@ export default class ImageElementFinder {
...
-   *
-   * @returns {Promise<Element|Element[]|ImageElement>} - WebDriver element with a special id prefix
+   * @param {FindByImageOptions<Multiple,CheckStaleness>} opts - additional options
+   * @returns {Promise<FindByImageResult<Multiple,CheckStaleness>>} - WebDriver element with a special id prefix
    */
   async findByImage(
     b64Template,
-    {shouldCheckStaleness = false, multiple = false, ignoreDefaultImageTemplateScale = false}
+    {
+      shouldCheckStaleness = /** @type {CheckStaleness} */ (false),
```
mykola-mokhnach, line 130: these are simple booleans

packages/images-plugin/lib/finder.js
```diff
@@ -224,8 +220,10 @@ export default class ImageElementFinder {
     }
 
     if (_.isEmpty(results)) {
-      if (multiple) {
-        return [];
+      if (multiple === true) {
```
mykola-mokhnach, line 223: why `if (multiple)` would not work?

packages/images-plugin/lib/finder.js
```diff
@@ -287,10 +287,10 @@ export default class ImageElementFinder {
    * @returns {Promise<Screenshot & {scale?: ScreenshotScale}>} base64-encoded screenshot and ScreenshotScale
    */
   async getScreenshotForImageFind(screenWidth, screenHeight) {
-    if (!this.driver.getScreenshot) {
+    if (!this.driver?.getScreenshot) {
```
mykola-mokhnach, line 290: if driver is null then the error below won't make much sence

packages/images-plugin/lib/finder.js
```diff
@@ -484,7 +483,7 @@ export default class ImageElementFinder {
         ` to ${scaledWidth}x${scaledHeight}`
     );
     log.info(`The ratio is ${xScale} and ${yScale}`);
-    imgTempObj = await imgTempObj.resize(scaledWidth, scaledHeight);
+    imgTempObj = imgTempObj.resize(scaledWidth, scaledHeight);
```
mykola-mokhnach, line 486: is this method synchronous?

packages/images-plugin/lib/plugin.js
```diff
@@ -2,9 +2,9 @@
 
 import _ from 'lodash';
 import {errors} from 'appium/driver';
-import BasePlugin from 'appium/plugin';
+import {BasePlugin} from 'appium/plugin';
```
mykola-mokhnach, line 5: is this a breaking change?

packages/images-plugin/test/unit/finder.spec.js
```diff
@@ -2,8 +2,8 @@ import _ from 'lodash';
 import {imageUtil} from 'appium/support';
 import {BaseDriver} from 'appium/driver';
 import {ImageElementPlugin, IMAGE_STRATEGY} from '../../lib/plugin';
-import ImageElementFinder from '../../lib/finder';
-import ImageElement from '../../lib/image-element';
+import {ImageElementFinder, W3C_ELEMENT_KEY} from '../../lib/finder';
```
mykola-mokhnach, line 5: should W3C_ELEMENT_KEY be imported from `@appium/support`?

packages/opencv/lib/index.js
```diff
@@ -335,7 +282,10 @@ async function getImagesMatches(img1Data, img2Data, options = {}) {
         width: rect2.width,
         height: rect2.height,
       });
-      result.visualization = await jimpImgFromCvMat(visualization).getBufferAsync(Jimp.MIME_PNG);
+      result.visualization = base64Encode(
```
mykola-mokhnach, line 285: Is it a breaking change? I remember we had a call of `convertVisualizationToBase64` and `visualization` itself was always stored as raw buffer

packages/opencv/lib/index.js
```diff
@@ -344,57 +294,42 @@ async function getImagesMatches(img1Data, img2Data, options = {}) {
       img1.delete();
       img2.delete();
       detector.delete();
-      result1.keyPoints.delete();
-      result1.descriptor.delete();
-      result2.keyPoints.delete();
-      result2.descriptor.delete();
+      result1?.keyPoints.delete();
+      result1?.descriptor.delete();
+      result2?.keyPoints.delete();
```
mykola-mokhnach, line 299: I assume the `delete()` would still fail in this case if result is null

packages/opencv/lib/index.js
```diff
@@ -344,57 +294,42 @@ async function getImagesMatches(img1Data, img2Data, options = {}) {
...
- * @typedef SimilarityResult
- * @property {number} score The similarity score as a float number in range [0.0, 1.0].
- * 1.0 is the highest score (means both images are totally equal).
- * @property {?Buffer} visualization The visualization of the matching result
- * represented as PNG image buffer. This image includes both input pictures where
- * difference regions are highlighted with rectangles.
+ * @param {Buffer} buf
+ * @returns {string}
  */
+function base64Encode(buf) {
```
mykola-mokhnach, line 311: not sure we need a dedicated function for that

packages/opencv/lib/index.js
```diff
@@ -682,3 +587,130 @@ export {getImagesMatches, getImagesSimilarity, getImageOccurrence, initOpenCv};
...
+ * - `TM_SQDIFF_NORMED`
+ * Read https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_imgproc/py_template_matching/py_template_matching.html
+ * for more details.
+ */
+
+/**
+ * @typedef SimilarityResult
+ * @property {number} score The similarity score as a float number in range [0.0, 1.0].
+ * 1.0 is the highest score (means both images are totally equal).
+ * @property {Buffer} [visualization] The visualization of the matching result
```
mykola-mokhnach, line 611: it won't be buffer anymore after we call `toString()` on it

packages/support/lib/image-util.js
```diff
@@ -1,13 +1,126 @@
...
+   */
+  static async from(strOrBuf) {
+    return await (_.isString(strOrBuf)
+      ? AppiumImage.fromString(strOrBuf)
+      : AppiumImage.fromBuffer(strOrBuf));
+  }
+
+  /**
+   *
+   * @param {string} str
```
mykola-mokhnach, line 92: I assume it should explicitly say that base64-encoded image payload is expected. Also it would make sense to list supported image formats

packages/support/lib/image-util.js
```diff
@@ -1,13 +1,126 @@
...
+   * @returns {Promise<AppiumImage>}
+   */
+  static async fromString(str) {
+    return await AppiumImage.fromBuffer(Buffer.from(str, 'base64'));
+  }
+
+  /**
+   *
+   * @param {Buffer} buf
```
mykola-mokhnach, line 101: I would expect more details on this variable

packages/support/test/unit/image-util.spec.js
```diff
@@ -0,0 +1,133 @@
+/* eslint-disable promise/prefer-await-to-callbacks */
+/* eslint-disable require-await */
+import {createSandbox} from 'sinon';
+import rewiremock from 'rewiremock/node';
+
+const {expect} = chai;
```
mykola-mokhnach, line 6: cannot we just import it?

packages/support/test/unit/image-util.spec.js
```diff
@@ -0,0 +1,133 @@
...
+    sandbox.restore();
+  });
+
+  describe('constructor', function () {
+    it('should instantiate an AppiumImage', function () {
+      expect(new AppiumImage(mockJimp)).to.be.an.instanceof(AppiumImage);
+    });
+  });
+
+  describe('static method', function () {
```
mykola-mokhnach, line 59: should this contain something?

packages/images-plugin/lib/compare.js
boneskull: it's a typescript thing--a type guard.  they have to be functions which return a type of `{argName is SomeType}` and you can't do that with a switch, afaik.

packages/images-plugin/lib/compare.js
boneskull: we could, but that'd mean two type coercions (`/** @type {CompareImagesResult<Mode,Multiple>} */ (result)`) instead of one.

packages/images-plugin/lib/compare.js
boneskull: short of making new typedefs, there's nothing to be done, because linebreaks break the type definition.

packages/images-plugin/lib/finder.js
boneskull: this means that the params are optional. they can be `undefined`. they cannot be `null`, however; that'd look like `@param {ExternalDriver?} [driver]` which allows both `null` and `undefined`, or `@param {ExternalDriver?} driver` which wants an `ExternalDriver` or `null` but not `undefined`.

packages/images-plugin/lib/finder.js
boneskull: yeah, I blame copilot.  I think saying something like "call setDriver() first` would work

packages/images-plugin/lib/finder.js
boneskull: yes, but we can't use conditional types without them (the things that look like ternary statements).  the type of the return value depends on whether or not these are true or false, and there's no other way to reference the type. [this explains it better than I can](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html)

packages/images-plugin/lib/finder.js
boneskull: it might; I think I changed this trying to fix something else.

packages/images-plugin/lib/finder.js
boneskull: true. I'll change this

packages/images-plugin/lib/finder.js
boneskull: yes.

packages/images-plugin/lib/plugin.js
boneskull: no... in what?

packages/images-plugin/test/unit/finder.spec.js
boneskull: yeah I can do that

packages/support/lib/image-util.js
boneskull: will do

packages/support/test/unit/image-util.spec.js
boneskull: we could. but it's already in the global context (as to avoid the boilerplate of `import chai from 'chai'`) everywhere, so ..

packages/support/test/unit/image-util.spec.js
boneskull: I should write tests, yes

boneskull:
@mykola-mokhnach 

> It looks like the PR title is not a canonical one

what do you mean?  conventional commits are for commit messages... the only reason they end up in PR titles is they get autofilled if there's only a single commit in a PR.

packages/images-plugin/lib/compare.js
mykola-mokhnach: maybe worth a comment in the code

packages/images-plugin/lib/finder.js
mykola-mokhnach: oh, that's getting too complicated IMHO. Should  we really care about differentiating undefined and null except of some (rare) special cases?

packages/images-plugin/lib/plugin.js
mykola-mokhnach: I mean that BasePlugin is not a default import anymore. Then this line is going to fail in all other clients which were importing the class like that

packages/support/test/unit/image-util.spec.js
mykola-mokhnach: my point is that explicit is better than implicit. We don't know what's in the global context by just reading the source here. I'd rather prefer having one extra line

packages/images-plugin/lib/finder.js
boneskull: well, we kind of have to, because they are different types.  I'm not sure there's any way for TS to treat them as the same via config.  

rant time!

There's no consensus around use of `undefined` vs `null` in JS. _Mostly_, I've seen `null` avoided entirely (I generally subscribe to this philosophy), since (again, _mostly_) using one or the other is simply a matter of convention (aside: the only place where `null` _must_ be used is `Object.create(null)`).   In other words, there's no inherent meaning in the use of `null` other than what the developer ascribes to it.

For example, if the type of `sessionId` in `BaseDriver` was `string|undefined` instead of `string|null`, it'd be functionality equivalent (though `sessionId === null` comparisons will need to be replaced with `sessionId === undefined`).  Anything (within reason) you can do with a `sessionId` of `null` you can do with a `sessionId` of `undefined`.

What's problematic: when there really _isn't_ a convention, and sometimes things are `null` and it's unclear why they aren't just `undefined`, or vice-versa.  😜

packages/images-plugin/lib/plugin.js
boneskull: right, but if so, they would fail anyway, because they would still have to change to `appium/plugin` from `@appium/base-plugin`.  

either way: the default export is still `BasePlugin`. I just changed it here because I like named exports.

packages/opencv/lib/index.js
boneskull: it is if you consider this to be a public API, but the only place this code is called is in this module.  the code which calls this function encoded the result; I just moved the encoding to in here since it seemed appropriate.

packages/opencv/lib/index.js
boneskull: if `result1` is `null`, the method will not be called.

packages/opencv/lib/index.js
boneskull: we use it in a couple places, but it makes sense organizationally as well

packages/opencv/lib/index.js
boneskull: ah, you're right.  there's no strict type checking going on here.

packages/support/test/unit/image-util.spec.js
boneskull: sorry... if we start adding `import whatever from 'chai'` to test files, the convention becomes inconsistent, and for consistency we might as well avoid all of the globals.  I value the reduction of boilerplate more than any FUD around the assignment and referencing of global variables; that's a conscious decision.

packages/opencv/lib/index.js
mykola-mokhnach: yep, my point is we need to be quite careful with such breaking changes to not do too much damage to the depending code. I would rather keep it as is then have to do with all the hassle about finding and fixing all the dependencies that may potentially be broken. Especially when we are talking about more stylistic change like this one rather than some mandatory bugfix or update

packages/opencv/lib/index.js
mykola-mokhnach: I'm still not quite sure about that. Just tried 

```
let users = {
  'u1': { foo: () => 'yolo' },
};

result = users?.u4.foo()
```
in the console and got `users.u4 is undefined` error

jlipps: I think @boneskull is saying that if `result1` exists, we can assume that it will have a `keyPoints` object with a `delete` method on it, because that's guaranteed by the nature of what `result1` is

PR #17033, opened 2022-06-02
fix(base-driver): don't assign log to jwproxy from opts

jlipps:
This should hopefully resolve the #JWProxy issue that's been popping up.

cc @boneskull

### DIFF
--- a/packages/base-driver/lib/jsonwp-proxy/proxy.js
+++ b/packages/base-driver/lib/jsonwp-proxy/proxy.js
@@ -22,6 +22,18 @@ const COMPACT_ERROR_PATTERNS = [/\bECONNREFUSED\b/, /socket hang up/];
 
 const {MJSONWP, W3C} = PROTOCOLS;
 
+const ALLOWED_OPTS = [
+  'scheme',
+  'server',
+  'port',
+  'base',
+  'reqBasePath',
+  'sessionId',
+  'timeout',
+  'log',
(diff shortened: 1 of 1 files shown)


packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -39,7 +39,7 @@ class JWProxy {
   timeout;
 
   constructor(opts = {}) {
-    const options = _.defaults(opts, {
+    const options = _.defaults(_.omit(opts, 'log'), {
```
mykola-mokhnach, line 42: this will break the existing stuff from xcuitest driver. Basically there we share the same logger instance to the jwproxy as we use in the driver, which allows to transparently connect all logs to the same driver instance.

mykola-mokhnach: below we have ` this._log = opts.log;`, so it would probably make sense to have allowed properties list rather than exclude some. It would be much safer from the perspective that we assign options to `this` and we never know what `this` is going to contain

jlipps: ok @mykola-mokhnach i made a list of allowed options so we know what we're dealing with

jlipps:
CI failure looks like a github/npm network flake

mykola-mokhnach (approved):

PR #17038, opened 2022-06-03
fix(support): log-symbols is a prod dep

jlipps:
cc @boneskull


jlipps:
i suppose we need some pre- or post-deploy tests that run a sandboxed prod-deps-only install of appium and test something out.

mykola-mokhnach (approved):

jlipps:
ok this fix was published to appium support, and all depending packages including new appium 2 beta 40

PR #17045, opened 2022-06-08
chore(ci): run smoke test in all packages

boneskull:
This uses the [boneskull/nodejs-production-test-action](https://github.com/boneskull/nodejs-production-test-action) GH action to run a smoke test across all packages.

The action does the following:

1. Runs `npm pack` on each package, writing the tarball to a temp dir. This tarball contains the package as a user would `npm install` it.
2. Changes into the temp dir and runs `npm install <tarball>` on each
3. Runs a specified script (in our case, `test:smoke`) in each installed package's dir in `<tmp_dir>/node_modules`

Because `devDependencies` will not be present, this should avoid a class of issues where production deps are miscategorized as dev deps.

The action should probably use a separate temp dir for each--in the case of a dependency being present but only because some other package depends on it--but it's not smart enough to figure out peer dependencies yet.

### DIFF
--- a/packages/fake-driver/package.json
+++ b/packages/fake-driver/package.json
@@ -43,6 +43,7 @@
     "prepare": "npm run build",
     "test": "npm run test:unit",
     "test:e2e": "mocha --timeout 20s --slow 10s \"./test/e2e/**/*.spec.js\"",
+    "test:smoke": "node ./build/lib/index.js",
     "test:unit": "mocha \"./test/unit/**/*.spec.js\""
   },
   "dependencies": {
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,7 @@ jobs:
         with:
           script: test:smoke
           workspaces: true
+          quiet: true
       - name: Run unit & E2E tests
         run: npm run test:ci
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ permissions:
 
 jobs:
   test:
-    name: Test
+    name: Tests
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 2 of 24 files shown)


boneskull:
* 2.0
  * **PR #17045** <a href="https://app.graphite.dev/github/pr/appium/appium/17045" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/17045?utm_source=stack-comment).

boneskull:
Punting back into draft until I can figure out why the `workspaces` property in `package.json` causes E2E tests to fail.

boneskull:
@jlipps [Here's some output](https://github.com/appium/appium/runs/6837842703?check_suite_focus=true#step:6:1) from a smoke test run.  The first thing we see is a huge JSON blob of info from `npm pack`, but [here](https://github.com/appium/appium/runs/6837842703?check_suite_focus=true#step:6:3857) we can see it's running the `test:smoke` script in each package.  These scripts don't really _do_ anything other than try to load the entry point.  We have to be careful to avoid doing unnecessary work, so `doctor` and `fake-driver` do not load the entry point (as those are CLI executables).  `appium` _does_ load its entry point (`main.js`), but only to run `--show-build` and exit.

Unless we're doing lazy loading, this should be sufficient to detect many cases of missing dependencies and publish misconfigurations.

packages/fake-driver/package.json
```diff
@@ -43,6 +43,7 @@
     "prepare": "npm run build",
     "test": "npm run test:unit",
     "test:e2e": "mocha --timeout 20s --slow 10s \"./test/e2e/**/*.spec.js\"",
+    "test:smoke": "node ./build/lib/index.js",
```
jlipps, line 46: any reason for all of these that we're specifying a specific js file rather than say `node .` and relying on resolution through the `main` field (which is theoretically what users are also doing?)

jlipps (approved):

packages/fake-driver/package.json
boneskull: yeah, we could use `node .` in _most_ places, but some of them (like `fake-driver`) actually start a server or do some extra work we don't need.  I think I mentioned this...somewhere. maybe it was in my head

boneskull: > We have to be careful to avoid doing unnecessary work, so doctor and fake-driver do not load the entry point (as those are CLI executables). appium does load its entry point (main.js), but only to run --show-build and exit.

PR #17050, opened 2022-06-09
docs(appium): add building drivers doc (resolves #16788)

### LINKED ISSUE #16788: appium 2 docs: write guide on building drivers

jlipps:
Here is my stab at a guide for building drivers. It's pretty long. I tried to include everything driver authors might reasonably want to do, and organize it appropriately. Feedback appreciated!

### DIFF
--- a/packages/appium/docs/en/ecosystem/build-drivers.md
+++ b/packages/appium/docs/en/ecosystem/build-drivers.md
@@ -2,4 +2,683 @@
 title: Building Appium Drivers
 ---
 
-TODO
+Appium wants to make it easy for anyone to develop their own automation drivers as part of the
+Appium ecosystem. This guide will explain what's involved and how you can accomplish various driver
+development tasks using the tools Appium provides. This guide assumes you (1) are a competent user of
+Appium, (2) are a competent Node.js developer, and (3) that you have read and understood the
+[Driver Intro](../intro/drivers.md).
+
+If that describes you, great! This guide will get you started.
+
+## Before you create your driver
+
+Before you get to work implementing your driver, it's important to have a few things sorted out.
+For example, you need to know what your driver will do. Which platform is it trying to expose
+WebDriver automation for?
+
+Appium doesn't magically give you the power to automate any platform. All it does is give you a set
+of convenient tools for implementing the WebDriver Protocol. So if you want to create, for example,
+a driver for a new app platform, you'll need to know how to automate apps on that platform *without Appium*.
+
+This usually means that you need to be very familiar with app development for a given platform. And
+it usually means that you will rely on tools or SDKs provided by the platform vendor.
+
+Basically, if you can't answer the question **"how would I launch, remotely trigger behaviours, and
+read state from an app on this platform?" then you're not quite ready to write an Appium driver**.
+Make sure you do the research to feel comfortable that there *is* a path forward. Once there is,
+coding it up and making it available as an Appium driver should be the easy part!
+
+## Other drivers to reference
+
+One of the greatest things about building an Appium driver is that there are already a number of
+open source Appium drivers which you can look at for reference. There is
+a [fake-driver](https://github.com/appium/appium/tree/2.0/packages/fake-driver) sample driver which
+does basically nothing other than showcase some of the things described in this guide.
+
+And of course, all of Appium's official drivers are open source and available in repositories at
+the project's GitHub organization. So if you ever find yourself asking, "how does a driver do X?",
+read the code for these drivers! Also don't be afraid to ask questions of the Appium developers if
+you get stuck; we're always happy to help make sure the driver development experience is a good
+one!
+
+## Basic requirements for Appium drivers
+
+These are the things your driver *must* do (or be), if you want it to be a valid Appium driver.
+
+### Node.js package with Appium extension metadata
+
+All Appium drivers are fundamentally Node.js packages, and therefore must have a valid
+`package.json`. Your driver is not _limited_ to Node.js, but it must provide an adapter written in Node.js so it can be loaded by Appium.
+
+Your `package.json` must include `appium` as a `peerDependency`. The requirements for the
+dependency versions should be as loose as possible (unless you happen to know your driver will only
+work with certain versions of Appium). For Appium 2.0, for example, this would look something like
+`^2.0.0`, declaring that your driver works with any version of Appium that starts with 2.x.
+
+Your `package.json` must contain an `appium` field, like this (we call this the 'Appium extension
+metadata'):
+
+    ```json
+    {
+      ...,
+      "appium": {
+        "driverName": "fake",
+        "automationName": "Fake",
+        "platformNames": [
+          "Fake"
+        ],
+        "mainClass": "FakeDriver"
+      },
+      ...
+    }
+    ```
+
+The required subfields are:
+
+* `driverName`: this should be a short name for your driver.
+* `automationName`: this should be the string users will use for their `appium:automationName`
+  capability to tell Appium to use *your* driver.
+* `platformNames`: this is an array of one or more platform names considered valid for your driver.
+  When a user sends in the `platformName` capability to start a session, it must be included in
+  this list for your driver to handle the session. Known platform name strings include: `iOS`,
+  `tvOS`, `macOS`, `Windows`, `Android`.
+* `mainClass`: this is a named export (in CommonJS style) from your `main` field. It must be a
+  class which extends Appium's `BaseDriver` (see below).
+
+### Extend Appium's `BaseDriver` class
+
+Ultimately, your driver is much easier to write because most of the hard work of implementing the
+WebDriver protocol and handling certain common logic is taken care of already by Appium. This is
+all encoded up as a class which Appium exports for you to use, called `BaseDriver`. It is exported
+from `appium/driver`, so you can use one of these styles to import it and create your *own* class
+that extends it:
+
+```js
+import {BaseDriver} from 'appium/driver';
+// or: const {BaseDriver} = require('appium/driver');
+
+export class MyDriver extends BaseDriver {
+}
+```
+
+### Make your driver available
+
+That's basically it! With a Node.js package exporting a driver class and with correct Appium
+extension metadata, you've got yourself an Appium driver! Now it doesn't *do* anything, but you can
+load it up in Appium, start and stop sessions with it, etc...
+
+To make it available to users, you could publish it via NPM. When you do so, your driver will be
+installable via the Appium CLI:
+
+```
+appium driver install --source=npm <driver-package-on-npm>
+```
+
+It's a good idea to test your driver first, of course. One way to see how it works within Appium is
+to install it locally first:
+
+```
+appium driver install --source=local /path/to/your/driver
+```
+
+### Developing your driver
+
+How you develop your driver is up to you. It is convenient, however, to run it from within Appium
+without having to do lots of publishing and installing. The most straightforward way to do this is
+to include the most recent version of Appium as a `devDependency`, and then also your own driver,
+like this:
+
+```json
+{
+    "devDependencies": {
+        ...,
+        "appium": "^2.0.0",
+        "your-driver": "file:.",
+        ...
+    }
+}
+```
+
+Now, you can run Appium locally (`npm exec appium` or `npx appium`), and because your driver is
+listed as a dependency alongside it, it will be automatically "installed" and available. You can
+design your e2e tests this way, or if you're writing them in Node.js, you can simply import
+Appium's start server methods to handle starting and stopping the Appium server in Node. (TODO:
+reference an implementation of this in one of the open source drivers when ready).
+
+Another way to do local development with an existing Appium server install is to simply install
+your driver locally:
+
+```
+appium driver install --source=local /path/to/your/driver/dev/dir
+```
+
+## Standard driver implementation ideas
+
+These are things you will probably find yourself wanting to do when creating a driver.
+
+### Set up state in a constructor
+
+If you define your own constructor, you'll need to call `super` to make sure all the standard state
+is set up correctly:
+
+```js
+constructor(args) {
+    super(args);
+    // now do your own thing
+}
+```
+
+The `args` parameter here is the object containing all the CLI args used to start the Appium
+server.
+
+### Define and validate accepted capabilities
+
+You can define your own capabilities and basic validation for them. Users will always be able to
+send in capabilities that you don't define, but if they send in capabilities you have explicitly
+defined, then Appium will validate that they are of the correct type (and will check for the
+presence of required capabilities).
+
+If you want to turn capability validation off entirely, set `this.shouldValidateCaps` to `false` in
+your constructor.
+
+To give Appium your validation constraints, set `this.desiredCapConstraints` to a validation object
+in your constructor. Validation objects can be somewhat complex. Here's an example from the
+UiAutomator2 driver:
+
+```js
+{
+  app: {
+    presence: true,
+    isString: true
+  },
+  automationName: {
+    isString: true
+  },
+  browserName: {
+    isString: true
+  },
+  launchTimeout: {
+    isNumber: true
+  },
+}
+```
+
+### Start a session and read capabilities
+
+Appium's `BaseDriver` already implements the `createSession` command, so you don't have to. However
+it's very common to need to perform your own startup actions (launching an app, running some
+platform code, or doing different things based on capabilities you have defined for your driver).
+So you'll probably end up overriding `createSession`. You can do so by defining the method in your
+driver:
+
+```js
+async createSession(jwpCaps, reqCaps, w3cCaps, otherDriverData) {
+    const [sessionId, caps] = super.createSession(w3cCaps);
+    // do your own stuff here
+    return [sessionId, caps];
+}
+```
+
+For legacy reasons, your function will receive old-style JSON Wire Protocol desired and required
+caps as the first two arguments. Given that the old protocol isn't supported anymore and clients
+have all been updated, you can instead only rely on the `w3cCaps` parameter. (For a discussion
+about what `otherDriverData` is about, see the section below on concurrent drivers).
+
+You'll want to make sure to call `super.createSession` in order to get the session ID as well as
+the processed capabilities (note that capabilities are also set on `this.caps`; modifying `caps`
+locally here would have no effect other than changing what the user sees in the create session
+response).
+
+So that's it! You can fill out the middle section with whatever startup logic your driver requires.
+
+### End a session
+
+If your driver requires any cleanup or shutdown logic, it's best to do it as part of overriding the
+implementation of `deleteSession`:
+
+```js
+async deleteSession() {
+    // do your own cleanup here
+    // don't forget to call super!
+    await super.deleteSession();
+}
+```
+
+It's very important not to throw any errors here if possible so that all parts of session cleanup
+can succeed!
+
+### Access capabilities and CLI args
+
+You'll often want to read parameters the user has set for the session, whether as CLI args or as
+capabilities. The easiest way to do this is to access `this.opts`, which is a merge of all options,
+from the CLI or from capabilities. So for example to access the `appium:app` capability, you could
+simply get the value of `this.opts.app`.
+
+If you care about knowing whether something was sent in as a CLI arg *or* a capability, you can
+access the `this.cliArgs` and `this.caps` objects explicitly.
+
+In all cases, the `appium:` capability prefix will have been stripped away by the time you are
+accessing values here, for convenience.
+
+### Implement WebDriver commands
+
+You handle WebDriver commands by implementing functions in your driver class. Each member of the
+WebDriver Protocol, plus the various Appium extensions, has a corresponding function that you
+implement if you want to support that command in your driver. The best way to see which commands
+Appium supports and which method you need to implement for each command is to look at Appium's
+[routes.js](https://github.com/appium/appium/blob/2.0/packages/base-driver/lib/protocol/routes.js).
+Each route object in this file tells you the command name as well as the parameters you'd expect to
+receive for that command.
+
+Let's take this block for example:
+```js
+'/session/:sessionId/url': {
+    GET: {command: 'getUrl'},
+    POST: {command: 'setUrl', payloadParams: {required: ['url']}},
+}
+```
+
+Here we see that the route `/session/:sessionId/url` is mapped to two commands, one for a `GET`
+request and one for a `POST` request. If we want to allow our driver to change the "url" (or
+whatever that might mean for our driver), we can therefore implement the `setUrl` command, knowing
+it will take the `url` parameter:
+
+```js
+async setUrl(url) {
+    // your implementation here
+}
+```
+
+A few notes:
+- all command methods should be `async` functions or otherwise return a `Promise`
+- you don't need to worry about protocol encoding/decoding. You will get JS objects as params, and
+  can return JSON-serializable objects in response. Appium will take care of wrapping it up in the
+  WebDriver protocol response format, turning it into JSON, etc...
+- all session-based commands receive the `sessionId` parameter as the last parameter
+- all element-based commands receive the `elementId` parameter as the second-to-last parameter
+- if your driver doesn't implement a command, users can still try to access the command, and will
+  get a `501 Not Yet Implemented` response error.
+
+### Implement element finding
+
+Element finding is a special command implementation case. You don't actually want to override
+`findElement` or `findElements`, even though those are what are listed in `routes.js`. Appium does
+a lot of work for you if instead you implement this function:
+
+```js
+async findElOrEls(strategy, selector, mult, context) {
+    // find your element here
+}
+```
+
+Here's what gets passed in:
+
+- `strategy` - a string, the locator strategy being used
+- `selector` - a string, the selector
+- `mult` - boolean, whether the user has requested one element or all elements matching the
+  selector
+- `context` - (optional) if defined, will be a W3C Element (i.e., a JS object with the W3C element
+  identifier as the key and the element ID as the value)
+
+And you need to return one of the following:
+
+- a single W3C element (an object as described above)
+- an array of W3C elements
+
+Note that you can import that W3C web element identifier from `appium/support`:
+
+```js
+import {util} from 'appium/support';
+const { W3C_WEB_ELEMENT_IDENTIFIER } = util;
+```
+
+What you do with elements is up to you! Usually you end up keeping a cache map of IDs to actual
+element "objects" or whatever the equivalent is for your platform.
+
+### Define valid locator strategies
+
+Your driver might only support a subset of the standard WebDriver locator strategies, or it might
+add its own custom locator strategies. To tell Appium which strategies are considered valid for
+your driver, create an array of strategies and assign it to `this.locatorStrategies`:
+
+```js
+this.locatorStrategies = ['xpath', 'custom-strategy'];
+```
+
+Appium will throw an error if the user attempts to use any strategies other than the allowed ones,
+which enables you to keep your element finding code clean and deal with only the strategies you
+know about.
+
+By default, the list of valid strategies is empty, so if your driver isn't simply proxying to
+another WebDriver endpoint, you'll need to define some. The protocol-standard locator strategies
+are defined [here](https://www.w3.org/TR/webdriver/#locator-strategies).
+
+### Throw WebDriver-specific errors
+
+The WebDriver spec defines a [set of error
+codes](https://github.com/jlipps/simple-wd-spec#error-codes) to accompany command responses if an
+error occurred. Appium has created error classes for each of these codes, so you can throw the
+appropriate error from inside a command, and it will do the right thing in terms of the protocol
+response to the user. To get access to these error classes, import them from `appium/driver`:
+
+```
+import {errors} from 'appium/driver';
+
+throw new errors.NoSuchElementError();
+```
+
+### Log messages to the Appium log
+
+You can always use `console.log`, of course, but Appium provides a nice logger for you as
+`this.log` (it has `.info`, `.debug`, `.log`, `.warn`, `.error` methods on it for differing log
+levels). If you want to create an Appium logger outside of a driver context (say in a script or
+helper file), you can always construct your own too:
+
+```js
+import {logging} from 'appium/support';
+const log = logging.getLogger('MyDriver');
+```
+
+## Further possibilities for Appium drivers
+
+These are things your driver *can* do to take advantage of extra driver features or do its job more
+conveniently.
+
+### Add a schema for custom command line arguments
+
+You can add custom CLI args if you want your driver to receive data from the command line when the
+Appium server is started (for example, ports that a server administrator should set that should not
+be passed in as capabilities.
+
+To do this you define a `schema` field within the `appium` field in `package.json`, pointing to
+a JS or JSON file defining an argument schema. (See the schema in `fake-driver`, for example).
+Arguments you define there will be available as a `--driver-<name>-<arg-name>` flag on the command
+line. If you define `my-arg` in the schema, for example, and if your driver's name is `awesome`,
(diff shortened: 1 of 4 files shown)


packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+## Before you create your driver
+
+Before you get to work implementing your driver, it's important to have a few things sorted out.
+For example, you need to know what your driver will do. Which platform is it trying to expose
+WebDriver automation for?
+
+Appium doesn't magically give you the power to automate any platform. All it does is give you a set
+of convenient tools to make implementing the WebDriver Protocol easy. So if you want to create
+a driver for some new app platform, let's say, you'll need to already know how you would automate
+apps on that platform *without Appium*.
```
boneskull, line 22:
```suggestion
of convenient tools for implementing the WebDriver Protocol. So if you want to create, for example, 
a driver for a new app platform, you'll need to know how to automate apps on that platform *without Appium*.
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+This usually means that you need to be very familiar with app development for a given platform. And
+it usually means that you will rely on tools or SDKs provided by the platform vendor.
+
+Basically, if you can't answer the question "how would I launch, remotely trigger behaviours, and
+read state from an app on this platform?" then you're not quite ready to write an Appium driver.
```
boneskull, line 28:
```suggestion
Basically, if you can't answer the question **"how would I launch, remotely trigger behaviours, and
read state from an app on this platform?" then you're not quite ready to write an Appium driver**.
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+## Basic requirements for Appium drivers
+
+These are the things your driver *must* do (or be), if you want it to be a valid Appium driver.
+
+### Node.js package with an Appium manifest
+
+All Appium drivers are fundamentally Node.js packages, and therefore must have a valid
+`package.json`. Your driver is not limited to Node.js, but whatever you do with your driver must
+ultimately be triggered by Node.js.
```
boneskull, line 53:
```suggestion
`package.json`. Your driver is not _limited_ to Node.js, but it must provide an adapter written in Node.js so it can be loaded by Appium.
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+Your `package.json` must include `appium` as a `peerDependency`. The requirements for the
+dependency versions should be as loose as possible (unless you happen to know your driver will only
+work with certain versions of Appium). For Appium 2.0, for example, this would look something like
+`^2.0.0`, declaring that your driver works with any version of Appium that starts with 2.x.
+
+Your `package.json` must contain an `appium` field, like this (we call this the 'Appium manifest'):
```
boneskull, line 60: > (we call this the 'Appium manifest')

We do? 😄 

I think we have some convolution of terminology going on (and this is my fault too).  Internally, I've been calling  `extensions.yaml` the "manifest" (Manifest).  The data in the Manifest _which is derived from an extension's `package.json`_ is the "extension metadata".  A Manifest contains records of this extension metadata _and_ internal (computed) metadata for each extension.

If we're going to settle on terminology, no time like the present!

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+        "appium": {
+            "driverName": "fake",
+            "automationName": "Fake",
+            "platformNames": [
+              "Fake"
+            ],
+            "mainClass": "FakeDriver"
+        },
+        ...
+    }
```
boneskull, line 74:
```suggestion
    {
      ...,
      "appium": {
        "driverName": "fake",
        "automationName": "Fake",
        "platformNames": [
          "Fake"
        ],
        "mainClass": "FakeDriver"
      },
      ...
    }
```

:trollface:

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+    ```
+
+The required subfields are:
+
+* `driverName`: this should be a short name for your driver.
+* `automationName`: this should be the string users will use for their `appium:automationName`
+capability to tell Appium to use *your* driver.
+* `platformNames`: this is an array of platform names considered valid for your driver. When a user
```
boneskull, line 82:
```suggestion
* `platformNames`: this is an array of _one or more_ platform names considered valid for your driver. When a user
```

also maybe refer to a list of known platform names? I know a driver author may want to create a new one (or multiple).

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+sends in the `platformName` capability to start a session, it must be included in this list for
+your driver to handle the session.
+* `mainClass`: this is a named export from your `main` field. It must be a class which extends
+Appium's `BaseDriver` (see below).
```
boneskull, line 86:
```suggestion
* `mainClass`: this is a named export (in CommonJS style) from your `main` field. It must be a class which extends
Appium's `BaseDriver` (see below).
```

...just to avoid confusion with ESM named exports.  _Or_ we need to be explicit that Appium expects you to compile your ESM to CommonJS.  

We can support native ESM in the future, but that's the way things are now

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+```
+
+### Make your driver available
+
+That's basically it! With a Node.js package exporting a driver class and with a correct Appium
+manifest, you've got yourself an Appium driver! Now it doesn't *do* anything, but you can load it
+up in Appium, start and stop sessions with it, etc...
+
+To make it available to users, you could publish it via NPM. When you do so, your driver will be
+isntallable via the Appium CLI:
```
boneskull, line 111: Maybe you should have the author `appium driver install --source=local /path/to/your/package` first.

Also typo on `installable`

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+    "devDependencies": {
+        ...,
+        "appium": "^2.0.0",
+        "your-driver": "file:.",
+        ...
+    }
+}
+```
+
+Now, you can run Appium locally (`npx appium`), and because your driver is listed as a dependency
```
boneskull, line 135:
```suggestion
Now, you can run Appium locally (`npm exec appium` or `npx appium`), and because your driver is listed as a dependency
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+alongside it, it will be automatically "installed" and available. You can design your e2e tests
+this way, or if you're writing them in Node.js, you can simply import Appium's start server methods
+to handle starting and stopping the Appium server in Node.
```
boneskull, line 138: We will probably want to reference the basedriver E2E suite function as well.  This is being moved into `appium/test` in #16907

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+These are things you will probably find yourself wanting to do when creating a driver.
+
+### Set up state in a constructor
+
+If you define your own constructor, you'll need to call `super` to make sure all the standard state
+is set up correctly:
+
+```js
+constructor(opts) {
```
boneskull, line 150: we may want to mention what `opts` is

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+You can define your own capabilities and basic validation for them. Users will always be able to
+send in capabilities that you don't define, but if they send in capabilities you have explicitly
+defined, then Appium will validate that they are of the correct type (and will check for the
+presence of required capabilities).
+
+If you want to turn capability validation off entirely, set `this.shouldValidateCaps` to `false` in
+your constructor.
+
+To give Appium your validation constraints, set `this.desiredCapConstraints` to a validation object
```
boneskull, line 166: ugh, need schemas.

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+```js
+async createSession(jwpCaps, reqCaps, w3cCaps, otherDriverData) {
+    const [sessionId, caps] = super.createSession(w3cCaps);
+    // do your own stuff here
+    return [sessionId, caps];
+}
+```
+
+For legacy reasons, your function will receive old-style JSON Wire Protocol desired and required
```
boneskull, line 204: I'm not sure this is true?  IIRC those first two args are always `undefined`.  We really should sort this out.

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+it will take the `url` parameter:
+
+```
+async setUrl(url) {
+    // your implementation here
+}
+```
+
+A few notes:
+- all commands should be `async` functions
```
boneskull, line 272:
```suggestion
- all "command" methods must `async` functions or otherwise return a `Promise`
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+- you don't need to worry about protocol encoding/decoding. You will get JS objects as params, and
+  can return JS objects in response. Appium will take care of wrapping it up in the WebDriver
```
boneskull, line 274:
```suggestion
  can return JSON-serializable JS objects in response. Appium will take care of wrapping it up in the WebDriver
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+```
+async findElOrEls(strategy, selector, mult, context) {
+    // find your element here
+}
+```
+
+Here's what gets passed in:
+
+- `strategy` - a string, the locator strategy being used
```
boneskull, line 295: May want to list the default valid strategies

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+```
+getProxyAvoidList() {
+    return [
+        ['POST', new RegExp('^/session/[^/]+/appium')]
+    ];
+}
+```
+
+The proxy avoidance list should be an array of arrays, where each inner array has an HTTP method as
```
boneskull, line 408: seems like this ought to support a different data structure someday

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+its first member, and a regular expression as its second. If the regular expression matches the
+route, then the route will not be proxied and instead will be handled by your driver. In this
+example, we are avoiding proxying all `POST` routes that have the `appium` prefix.
+
+Next, we have to set up the proxying itself. The way to do this is to use a special class from
+Appium called `JWProxy`. You'll want to create a `JWProxy` object using the details required to
```
boneskull, line 414: What does JW stand for?

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+    port: 1234,
+    base: '/',
+});
+
+this.proxyReqRes = proxy.proxyReqRes.bind(proxy);
+this.proxyCommand = proxy.command.bind(proxy);
+```
+
+Here we are creating a proxy object and assigning some of its methods to `this` under the names
+`proxyReqRes` and `proxyCommand`. This is required for Appium to use the proxy, so don't forget
```
boneskull, line 431: We could make this more ergonomic

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+remote endpoint you're proxying to. Appium will take care of all the proxying for you.
+
+### Extend the existing protocol with new commands
+
+You may find that the existing commands don't cut it for your driver. If you want to expose
+behaviours that don't map to any of the existing commands, you can create new commands in one of
+two ways:
+
+1. Extending the WebDriver protocol and creating client-side plugins to access the extensions.
+1. Overloading the Execute Script command
```
boneskull, line 455:
```suggestion
1. Overloading the "Execute Script" command
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+With those pieces in play, you won't have to reimplement anything that's already implemented by the
+remote endpoint you're proxying to. Appium will take care of all the proxying for you.
+
+### Extend the existing protocol with new commands
+
+You may find that the existing commands don't cut it for your driver. If you want to expose
+behaviours that don't map to any of the existing commands, you can create new commands in one of
+two ways:
+
+1. Extending the WebDriver protocol and creating client-side plugins to access the extensions.
```
boneskull, line 454:
```suggestion
1. Extending the WebDriver protocol and creating client-side plugins to access the extensions
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+1. Overloading the Execute Script command
+
+If you want to follow the first path, you can direct Appium to recognize new methods and add them
+to its set of allowed HTTP routes and command names. You do this by assigning the `newMethodMap`
+static variable in your driver class to an object of the same form as Appium's `routes.js` object.
+For example, here is the `newMethodMap` for the FakeDriver example driver:
```
boneskull, line 460:
```suggestion
For example, here is the `newMethodMap` for the `FakeDriver` example driver:
```

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+}
+```
+
+### Implement handling of Appium settings
+
+Appium users can send parameters to your driver via CLI args as well as via capabilities. But these
+cannot change during the course of a test, and sometimes users want to adjust parameters mid-test.
+Appium has a [Settings](../guides/settings.md) API for this purpose.
+
+To support settings in your own driver, first of all define `this.settings` to be an instance of
```
boneskull, line 519: Why don't we do this by default?

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+If you want to assign some default settings, or run some code on your end whenever settings are
+updated, you can do both of these things as well.
+
+```js
+constructor() {
+  const defaults = {setting1: 'value1'};
+  this.settings = new DeviceSettings(defaults, this.onSettingsUpdate.bind(this));
+}
+
+async onSettingsUpdate(key, value) {
```
boneskull, line 540: I wonder if this should work more like a standard EE

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+Let's say your driver uses up some system resources, like ports. There are a few ways to make sure
+that multiple simultaneous sessions don't use the same resources:
+
+1. Have your users specify resource IDs via capabilities (`appium:driverPort` etc)
+1. Just always use free resources (find a new random port for each session)
+1. Have each driver express what resources it is using, then examine currently-used resources from
+   other drivers when a new session begins.
+
+To support this third strategy, you can implement `get driverData` in your driver to return what
```
boneskull, line 555: I wonder why this is a computed property and not a function, like many of the others

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+Now, when a new session is started on your driver, the `driverData` response from any other
+simultaneously running drivers will also be included, as the last parameter of the `createSession`
+method:
+
+```
+async createSession(jwpCaps, reqCaps, w3cCaps, driverData)
+```
+
+You can dig into this `driverData` array to see what resources other drivers are using to help
+determine which ones you want to use for this particular session.
```
boneskull, line 573: I think I was a little confused about this, and I'm not sure what would be an accurate suggestion, but this seems to mean "instances of other drivers", and naturally those are per-session, and may an instance of your driver or some other driver

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+### Log events to the Appium event timeline
+
+Appium has an [Event Timing API](../guides/event-timing.md) which allows users to get timestamps for
```
boneskull, line 577: expanding a little on the use case here would be helpful

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+Appium has a feature-flag based [security model](../guides/security.md) that allows driver authors
+to hide certain features behind security flags. What this means is that if you have a feature you
+deem insecure and want to require server admins to opt in to it, you can require that they enable
+the feature by adding it to the `--allow-insecure` list or turning off server security entirely.
+
+To support the check within your own driver, you can call `this.isFeatureEnabled(featureName)` to
+determine whether a feature of the given name has been enabled. Or, if you want to simply
+short-circuit and throw an error if the feature isn't enabled, you can call
+`this.ensureFeatureEnabled(featureName)`.
```
boneskull, line 598: we should rename this to `assertFeatureEnabled()` :wink:

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+### Use a temp dir for files
+
+If you want to use a temporary directory for files your driver creates that are not important to
+keep around between computer or server restarts, you can simply read from `this.opts.tmpDir`.
```
boneskull, line 603: Is this a string? does it reference a temp dir that already exists? how can the user control when it's cleaned up? what is an example use-case?

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+might detect that some external service has crashed and nothing will work anymore. In this case, it
+can call `this.startUnexpectedShutdown(err)` with an error object including any details, and Appium
+will attempt to gracefully handle any remaining requests before shutting down the session.
+
+If you want to perform some of your own cleanup logic when you encounter this condition, you can
+either do so immediately before calling `this.startUnexpectedShutdown`, or you can attach a handler
+to the unexpected shutdown event and run your cleanup logic "out of band" so to speak:
+
+```js
+this.onUnexpectedShutdown(handler)
```
boneskull, line 617: what does `handler` receive?

boneskull (approved):
I learned a lot, so that means this is good.

jlipps:
@boneskull thanks, I'll go through all of this. I realized while writing it, as you did while reading it, that there is a lot of room for improvement and standardization here. We should probably note a lot of this stuff for Appium 3 (which I don't mind if we release much sooner), if we want to make breaking changes to the driver API. we could also do that now for Appium 2 but i don't know how much more time we want to spend fiddling with all the drivers vs shipping stuff. plus we could maybe take a step back after 2 is released and wait to see what people build in terms of extensions. no doubt other breaking changes will need to be made at some point soon.

KazuCocoa (approved):

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+up in Appium, start and stop sessions with it, etc...
+
+To make it available to users, you could publish it via NPM. When you do so, your driver will be
+isntallable via the Appium CLI:
+
+```
+appium driver install --source=npm <driver-package-on-npm>
+```
+
+### Developing your driver
```
mykola-mokhnach, line 117: How about installing it from `local` source instead (which is `npm link` under the hood)? This is what I usually do

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+### Start a session and read capabilities
+
+Appium's `BaseDriver` already implements the `createSession` command, so you don't have to. However
+it's very common to need to perform your own startup actions (launching an app, running some
+platform code, or doing different things based on capabilities you have defined for your driver).
+So you'll probably end up overriding `createSession`. You can do so by defining the method in your
+driver:
+
+```js
+async createSession(jwpCaps, reqCaps, w3cCaps, otherDriverData) {
```
mykola-mokhnach, line 197: this is now different and all the first 3 params are assumed to be be `w3cCaps`

packages/appium/docs/en/ecosystem/build-drivers.md
mykola-mokhnach: I'm sure this is not true anymore. We simply cut off lagacy caps

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+You'll want to make sure to call `super.createSession` in order to get the session ID as well as
+the processed capabilities (note that capabilities are also set on `this.caps`; modifying `caps`
+locally here would have no effect other than changing what the user sees in the create session
+response).
+
+So that's it! You can fill out the middle section with whatever startup logic your driver requires.
+
+### End a session
+
+If your driver requires any cleanup or shutdown logic, it's best to do it as part of overriding the
```
mykola-mokhnach, line 218: it is probably important to mention one has to be careful about throwing exceptions in this method call as we usually want all session deletion steps to be executed before existing the method. So it must be designed in the way that each api call there is error-proof.

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+- an array of W3C elements
+
+Note that you can import that W3C web element identifier from `appium/support`:
+
+```js
+import {util} from 'appium/support';
+const { W3C_WEB_ELEMENT_IDENTIFIER } = util;
+```
+
+What you do with elements is up to you! Usually you end up keeping a cache map of IDs to actual
```
mykola-mokhnach, line 314: should we also mention the set of standard errors we have to map w3c error codes?

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+### Log messages to the Appium log
+
+You can always use `console.log`, of course, but Appium provides a nice logger for you as
+`this.log` (it has `.info`, `.debug`, `.log`, `.warn`, `.error` methods on it for differing log
+levels). If you want to create an Appium logger outside of a driver context (say in a script or
+helper file), you can always construct your own too:
+
+```js
+import {logging} from 'appium/support';
+const log = logging.getLogger('MyDriver');
```
mykola-mokhnach, line 340: this has been changed. Each driver now receives its own `log` property, which also automatically contains unique driver fingerpint and the session identifier, which greatly helps to parse server logs coming from multiple drivers.
This log property could also be passed to JWProxy

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+
+Next, we have to set up the proxying itself. The way to do this is to use a special class from
+Appium called `JWProxy`. You'll want to create a `JWProxy` object using the details required to
+connect to the remote server:
+
+```js
+// import {JWProxy} from 'appium/driver';
+
+const proxy = new JWProxy({
+    server: 'remote.server',
```
mykola-mokhnach, line 421: I would describe all possible options

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+`execute` override, and do the appropriate thing:
+
+```js
+async execute(script, args) {
+  if (script === 'soundz: playSong') {
+    const {song, artist} = args[0];
+    // play the song based on song and artist details
+    return;
+  }
+  // otherwise do the normal thing, or throw a not yet implemented
```
mykola-mokhnach, line 509: I'd add a more concrete example on what exception to throw and where to import it

packages/appium/docs/en/ecosystem/build-drivers.md
KazuCocoa: Sounds good idea

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -2,4 +2,617 @@
...
+```
+
+### Make itself aware of resources other concurrent drivers are using
+
+Let's say your driver uses up some system resources, like ports. There are a few ways to make sure
+that multiple simultaneous sessions don't use the same resources:
+
+1. Have your users specify resource IDs via capabilities (`appium:driverPort` etc)
+1. Just always use free resources (find a new random port for each session)
+1. Have each driver express what resources it is using, then examine currently-used resources from
```
mykola-mokhnach, line 552: this strategy is only going to work if drivers are triggered in scope of a single server. If a common system resource is being used by drivers running on the same machine but under different server processes then armageddon is guaranteed. In such case the usage of inter-process locks (for example file-based ones) could only be helpful.

packages/appium/docs/en/ecosystem/build-drivers.md
mykola-mokhnach: the idea behind this is that one could provide a path to this dir from a server command line arg and then clean it safely up from the client code after the server process gets killed.

mykola-mokhnach: this is also the main reason to prefer using tmpDir module from `@appium/support` instead of directly using ones provided by Node.

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: > We do?

aspirationally? i dunno. what should we call this? most specifically, it is "the `appium` field in the extension's package.json`".

I think I'm fine with "Appium extension metadata"?

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: yeah, will add

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: yeah, i mention that just above. these lines are in case you are outside of a driver context and have no access to `this.log`

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: ideally this would be generated by API documentation and linked to here. this document is supposed to be a guide, not exhaustive API docs

jlipps:
thanks for comments everyone, I made the changes

packages/appium/docs/en/ecosystem/build-drivers.md
mykola-mokhnach: maybe not everything was cut off. Basically the code there just checks which of these three arguments is a valid w3c caps one (starting from the last one) and accepts it or throws an exception if none is found. My point was we probably don't want to do a breaking change to all drivers and also don't wan't to expose legacy jwp support.

jlipps:
> thanks for comments everyone, I made the changes

nevermind, there were so many comments many of them were hidden. still working on it.

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: yeah, want to make another schema system lol?

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: we should test this. i was using a client (webdriverio) that i think sends both styles of caps, and I got a value for the first parameter. i needed to use the 3rd param for it to work.

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: 👍

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: it's legacy. I'll make a note. we could also export an alias of it as `WDProxy` maybe?

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: there is a mix of when we use properties and when we use functions. it's not pretty.

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: actually it's just drivers of this driver's type. i'll add a note.

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: yeah. i think all the info is still there from the client, but now it doesn't matter which one comes in which position as long as a w3c object is in there. but it does mean driver authors have to be careful when they are calling `super`, for example, not to send in 3 copies of the jwp style one

jlipps:
ok, *now* it's done. note there's a code change here as well. accidentally squashed it in so there's not really an easy way to give it its own commit.

packages/appium/docs/en/ecosystem/build-drivers.md
boneskull: wfm

packages/appium/docs/en/ecosystem/build-drivers.md
boneskull: luckily I can reuse a ton of it.  

should get done, but less of a high priority and more of a "fun" thing for me

packages/appium/docs/en/ecosystem/build-drivers.md
boneskull: we should probably just rip off the bandaid here and accept one w3c caps arg.

packages/appium/docs/en/ecosystem/build-drivers.md
boneskull: I'd be happy with `WebDriverProxy`

packages/appium/docs/en/ecosystem/build-drivers.md
boneskull: ah, okay. 

in that case it seems like it'd make sense--in a utopian future where the API is perfect--to just have a function or property which defines the defaults, and the thing that instantiates `DeviceSettings` could just use it. (similarly to other things here)

boneskull (approved):
lgtm

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: agreed.

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: https://github.com/appium/appium/issues/17061

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: ok, renamed it to that.

boneskull:
There seems to be some overlap between this doc and `intro/drivers.md`

jlipps:
Yep.

PR #17078, opened 2022-06-19
fix: Update the way build info is fetched from GitHub

mykola-mokhnach:
## Proposed changes

We have changed the way our monorepo is tagged and thus it's necessary to update the build info retrieval logic that uses GitHub REST API. Previously the /status call was returning some random stuff

## Types of changes

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -108,10 +108,13 @@ class AppiumDriver extends DriverCore {
 
     // allow this to happen in the background, so no `await`
     // catch this to avoid an unhandled rejection
-    // eslint-disable-next-line promise/prefer-await-to-then,promise/prefer-await-to-callbacks
-    updateBuildInfo().catch((err) => {
-      this.log.debug(err);
-    });
+    (async () => {
+      try {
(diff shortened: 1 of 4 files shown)


packages/appium/lib/appium.js
```diff
@@ -108,10 +108,13 @@ class AppiumDriver extends DriverCore {
 
     // allow this to happen in the background, so no `await`
     // catch this to avoid an unhandled rejection
```
jlipps, line 110: `make sure we catch any possible errors to avoid unhandled rejections`

jlipps (approved):

packages/appium/lib/appium.js
mykola-mokhnach: not sure what the issue is?

jlipps: we are not catching "this" anymore. now we are using a catch inside the IIFE async function. i was giving a suggestion for rewording the comment to apply more accurately.

mykola-mokhnach: nvm, updated the comment

Issue #17073, opened 2022-06-17
Appium2 - Installing drivers "may be incompatible with the current version of Appium v2.0.0-beta.40 due to an invalid or missing peer dependency on Appium


KimEklund13:
## The problem

I installed the latest Appium [per the documentation](https://appiumpro.com/editions/122-installing-appium-20-and-the-driver-and-plugins-cli)
`npm install -g appium@next`

When I try and install any of the drivers (espresso, uiautomator2...) I get warnings of incompatibility:
Ex: When I run `appium driver install uiautomator2` 
```
✔ Installing 'uiautomator2' using NPM install spec 'appium-uiautomator2-driver'
⚠ Driver "uiautomator2" has 1 potential problem:
  - Driver "uiautomator2" (package `appium-uiautomator2-driver`) may be incompatible with the current version of Appium (v2.0.0-beta.40) due to an invalid or missing peer dependency on Appium. Please ask the developer of `appium-uiautomator2-driver` to add a peer dependency on `^appium@2.0.0-beta.40`.
Driver uiautomator2@2.2.0 successfully installed
```

When I run `appium install driver xcuitest`, it can't even be found.
`appium: error: unrecognized arguments: install driver xcuitest`

## Environment

* Appium version (or git revision) that exhibits the issue: 2.0.0-beta.40

* Driver version that exhibits the issue (only for Appium2):
uiautomator2@2.2.0
espresso@2.5.0
xcuitest: N/A, could not download

* Last driver version that did not exhibit the issue (if applicable, only for Appium2): N/A, first time installing and setting up Appium 2 with drivers.

* Desktop OS/version used to run Appium: MacOS 12.4 Monterey

* Node.js version (unless using Appium.app|exe): v18.2.0

* Package manager name and version (we only officially support NPM): NPM version 8.12.1

* Mobile platform/version under test: N/A
* Real device or emulator/simulator: N/A
* Appium CLI or Appium.app|exe: N/A (though I am about to use Appium server GUI (latest) with the separate inspector tool 
Inspector tool specs: 2022.5.4 / Electron 13.6.9 / Node.js 14.16.0.
The Appium desktop GUI gives me `showAppInfo` so I can't check the GUI version. Although, it says there's no updates available.

mykola-mokhnach:
The warning above is expected for now. We've made the server module to a peer dependency and it takes a while too update all drivers to support this approach. It does not affect drivers functionality though.

[...]

PR #17083, opened 2022-06-21
fix(appium,support): fix installation problems

boneskull:
Ref: #17073

this change uses `--legacy-peer-deps` for newer versions of npm to
avoid installation of appium when installing extensions in `~/.appium`.  this does not work if the extension contains a shrinkwrap with appium in it, though.
in that case, an extra copy of appium will always be installed.

it creates a symlink, e.g., `~/.appium/node_modules/appium` to the currently-running appium, wherever that is.  could not abuse `NODE_PATH` to do this, as changing it on-the-fly has no effect (it must be set at start of `node` process)

this is rather naive and does not check that the symlink is actually the same as the running appium if it already exists. that might cause some weird problems with multiple installs of appium on the same machine, but if it does, we can fix them as they arise.

### DIFF
--- a/packages/appium/lib/main.js
+++ b/packages/appium/lib/main.js
@@ -4,7 +4,7 @@ import {init as logsinkInit} from './logsink'; // this import needs to come firs
 import logger from './logger'; // logger needs to remain second
 // @ts-ignore
 import {routeConfiguringFunction as makeRouter, server as baseServer} from '@appium/base-driver';
-import {logger as logFactory, util, env} from '@appium/support';
+import {logger as logFactory, util, env, fs} from '@appium/support';
 import {asyncify} from 'asyncbox';
 import _ from 'lodash';
 import {AppiumDriver} from './appium';
@@ -15,6 +15,7 @@ import {
   checkNodeOk,
   getGitRev,
   getNonDefaultServerArgs,
+  rootDir,
   showConfig,
   showBuildInfo,
   validateTmpDir,
@@ -26,8 +27,9 @@ import {DRIVER_TYPE, PLUGIN_TYPE, SERVER_SUBCOMMAND} from './constants';
 import registerNode from './grid-register';
 import {getDefaultsForSchema, validate} from './schema/schema';
 import {inspect} from './utils';
+import path from 'path';
 
-const {resolveAppiumHome} = env;
+const {resolveAppiumHome, hasAppiumDependency} = env;
 
 /**
  *
@@ -169,6 +171,19 @@ function areServerCommandArgs(args) {
  */
 async function init(args) {
   const appiumHome = args?.appiumHome ?? (await resolveAppiumHome());
+
+  // this is here so extensions installed in e.g., `~/.appium` can find the peer dependency
+  // of `appium`, which lives wherever it lives (see `rootDir`).
+  if (!(await hasAppiumDependency(appiumHome))) {
+    try {
+      await fs.mkdirp(path.join(appiumHome, 'node_modules'));
+      await fs.symlink(rootDir, path.join(appiumHome, 'node_modules', 'appium'), 'junction');
+    } catch (err) {
+      if (err.code !== 'EEXIST') {
+        throw new Error(`Unable to create symlink to appium: ${err.message}`);
+      }
+    }
+  }
 
   const {driverConfig, pluginConfig} = await loadExtensions(appiumHome);
 
--- a/packages/appium/package.json
+++ b/packages/appium/package.json
@@ -56,7 +56,7 @@
     "prepare": "npm run build",
     "publish:docs": "APPIUM_DOCS_PUBLISH=1 npm run build:docs",
     "test": "npm run test:unit",
-    "test:e2e": "mocha --timeout 75s --slow 30s \"./test/e2e/**/*.spec.js\"",
+    "test:e2e": "mocha --timeout 1m --slow 30s \"./test/e2e/**/*.spec.js\"",
     "test:smoke": "node ./index.js --show-build-info",
     "test:unit": "mocha \"./test/unit/**/*.spec.js\""
   },
--- a/packages/support/lib/env.js
+++ b/packages/support/lib/env.js
@@ -40,7 +40,12 @@ export const MANIFEST_RELATIVE_PATH = path.join(
  * Note that we are _not_ performing the check "is the currently running Appium the same one as found in the list of dependencies", because I'm not sure it actually matters (assuming the versions are compatible).
  * Regardless we may want to make this more robust in the future, as we hit edge cases.
  *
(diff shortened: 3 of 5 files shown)


jlipps (approved):
ok let's give it a shot!

jlipps:
hmm seems like build is unhappy with this

packages/appium/lib/main.js
```diff
@@ -170,6 +172,18 @@ function areServerCommandArgs(args) {
 async function init(args) {
   const appiumHome = args?.appiumHome ?? (await resolveAppiumHome());
 
+  // this is here so extensions installed in e.g., `~/.appium` can find the peer dependency
+  // of `appium`, which lives wherever it lives (see `rootDir`).
+  if (!(await hasAppiumDependency(appiumHome))) {
+    try {
+      await fs.symlink(rootDir, path.join(appiumHome, 'node_modules', 'appium'));
```
mykola-mokhnach, line 179: would this work on Windows?

mykola-mokhnach: I assume we need to use junction there

packages/appium/lib/main.js
```diff
@@ -170,6 +172,18 @@ function areServerCommandArgs(args) {
...
+    } catch (err) {
```
mykola-mokhnach, line 180: is rootDir pointing to the root of the global Appium installation?
Maybe then it would make sense to put a symlink to `$HOME/.node_modules` according to https://nodejs.org/api/modules.html#loading-from-the-global-folders

packages/appium/lib/main.js
mykola-mokhnach: I assume we need to call mkdirp first to create `node_modules` root in `appiumHome`

boneskull:
> I assume we need to call mkdirp first to create `node_modules` root in `appiumHome` 

right, I think so.

I am not sure about windows.  I think we need to expand our CI matrix.

boneskull:
> Maybe it would make sense to put a symlink to `$HOME/.node_modules` according to https://nodejs.org/api/modules.html#loading-from-the-global-folders 

hmm, yes, that may be a way forward. I'll investigate

packages/appium/lib/main.js
boneskull: are you sure we need `junction`?  I thought that was a hard link--e.g., if `~/.appium/node_modules/appium` is deleted then e.g., `$NPM_PREFIX/node_modules/appium` also gets deleted?

mykola-mokhnach: on Windows - yes. it cannot create links for folders. Check https://github.com/nodejs/node/issues/18518, but you could try it yourself. Btw I know a company, which provides a whole VM cloud, where Windows-related stuff could also be tested ;)

mykola-mokhnach (approved):
LGTM, however I'd still check this functionality on Windows before merging just to be on the safe side

jlipps:
Am trying to check on Windows, however it seems there are some other problems on Windows potentially so it's not straightforward. Really need to get Windows in our CI I guess!

boneskull:
Ok, I tested this on Windows 10.  This is what I did:

1. Install [nvm-windows](/coreybutler/nvm-windows)
2. Open PowerShell as admin
3. `nvm install latest` (18.x)
4. `nvm use latest`
5. clone this repo
6. `npm install` (build appium with Node v18.x and npm v8.x) **this updates `package-lock.json` -- it needs to be added to this PR
7. `nvm install 14` (comes w/ npm v6.x)
8. `nvm use 14.19.3`
9. create some new tmp dir and navigate to it.  I did this just to be safe, but it's unclear if skipping this step would cause a problem
10. `node \path\to\packages\appium\build\lib\main.js --show-config` works and creates a link to `appium` working copy (subpath `./packages/appium`) in `~/.appium/node_modules`
11. `node \path\to\packages\appium\build\lib\main.js driver install xcuitest` succeeds, but I see a warning about a `test` driver. This is a bug and needs further investigation (but likely only affects dev envs)
13. Running `node \path\to\packages\appium\build\lib\main.js` shows another error about the "test" driver and refuses to load it but otherwise does not complain about xcuitest and loads it successfully
14. Uninstallation of `xcuitest` works.  However, it seems to blow away the symlink.   This may be avoidable by adding an `appium` dependency in `~/.appium/package.json` pointing back to the install path, but may not be an issue at all--the symlink is restored upon subsequent runs. I want to double-check this doesn't just happen at installation of an extension.
15. `nvm use latest` and repeating the `xcuitest` install and running of the server is identical to the output when using Node v14.x/npm v6.x.

So I guess I'm not entirely happy with it, but it seems to be good enough.  I ran into problems at some point due to the `package-lock.json` mismatch which caused really weird errors (@jlipps also experienced this, it sounds like).  

Windows needs to be in our build matrix and that should be high priority.

jlipps:
Nice! I'm fine to merge this then when the requisite action items are complete

boneskull:
oh i also saw that warning about fake-driver being a dep **and** a devdep. I'll see if that's real or not.

packages/appium/lib/main.js
```diff
@@ -170,6 +172,19 @@ function areServerCommandArgs(args) {
 async function init(args) {
   const appiumHome = args?.appiumHome ?? (await resolveAppiumHome());
 
+  // this is here so extensions installed in e.g., `~/.appium` can find the peer dependency
+  // of `appium`, which lives wherever it lives (see `rootDir`).
+  if (!(await hasAppiumDependency(appiumHome))) {
+    try {
+      await fs.mkdirp(path.join(appiumHome, 'node_modules'));
+      await fs.symlink(rootDir, path.join(appiumHome, 'node_modules', 'appium'), 'junction');
```
jlipps, line 180: woah what is this `junction` business?

jlipps: oh sorry i missed the `)` and misread `junction` as the last argument of `path.join`, not an arg of `fs.symlink`

boneskull:
got sidetracked, looking at this test driver thing now

boneskull:
The "test driver" thing was a bug. I added a fix/test.

boneskull:
I should probably explain:

The symlink to `appium` was tricking `@appium/support` into thinking that--since `node_modules/appium` lives under `$APPIUM_HOME`--Appium was installed as a dependency. It checks for this by doing an `npm list appium` in `$APPIUM_HOME`, and sure enough, there it is.  However, since it's not in `$APPIUM_HOME/package.json`, it's considered "extraneous" by npm. The fix was return `false` if that flag is present in the output of the `npm` command.

jlipps (approved):

boneskull:
[This is cute](https://github.com/appium/appium/runs/7256786600?check_suite_focus=true#step:7:13179)

boneskull:
The `--legacy-peer-deps` flag broke the install of `uiautomator2` with npm v7+ because our `appium` symlink gets blown away during the `npm install uiautomator2`. Some portion of this behavior depends on the contents of the shrinkwrap file, which has recently changed.

PR #17116, opened 2022-06-30
chore(renovate): add supports-color to major version exclusion list

boneskull:
newer versions of `supports-color` are ESM-only.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -19,6 +19,7 @@
         "log-symbols",
         "read-pkg",
         "strip-ansi",
+        "supports-color",
         "term-size",
         "wrap-ansi",
         "write-pkg"


jlipps:
1. is `chore(renovate)` a valid thing? i guess we decided it doesn't really matter?
2. is there really no interop for ESM even now? seems like this is something babel would have figured out?

jlipps (approved):

boneskull:
1. I'm not sure, but `chore(renovate)` seems to be what renovate itself uses. What looks invalid about it?
2. No; that'd require Babel to transpile stuff in `node_modules`, which it doesn't want to do.  We _could_ use `await import()` to use these, but that opens a can of worms (is it OK to import async in this context? does typescript/babel understand? etc).

jlipps:
I have tried `await import` in the past and it doesn't work. Generates the same error as regular `import`. Just seems bonkers that the whole ecosystem decided to fork.

PR #17143, opened 2022-07-06
fix(appium,types): include @appium/types in appium

boneskull:
This makes a small handful of fixes related to type consumption:

- `@appium/types` is now a direct dep of `appium`.  Without this, types leveraging `@appium/types`--e.g., `BaseDriver`, which is based on a `Driver` interface defined in `@appium/types`--fail to be understood fully by a consumer (such as an extension).  Extension authors would need to install `@appium/types` as their own dev dependency.  This change makes addition of `@appium/types` unneeded when consuming types exposed by `appium`, `appium/driver`, `appium/support`, etc.
- At this time, consumers will still need to add `@appium/types` if they wish to directly reference types defined in here (such as `W3CCapabilities`).
- In manual testing of the exported types, I found two issues in `appium/lib/schema/arg-spec.js` and `appium/lib/schema/schema.js` which were relative type imports and only worked in the context of the monorepo.  I fixed these.
- `@appium/types` previously re-exported types defined within `axios` (`HTTPMethod`); I just copied and pasted what we needed to avoid the production dependency on `axios`.
- `@types/express`, `@types/npmlog`, `type-fest` and `@wdio/types` were added as production deps to `@appium/types`, as these are needed.  However, these are packages _only_ containing types, unlike `axios`. There's a lesson to be learned here, but I don't know what it is.

**UPDATE**:

This PR also moves a bunch of types from the root monorepo into production deps of various packages that need them.  This is how you have to do it, apparently.


jlipps (approved):
tyyyyypes

boneskull:
Ugh, it's worse than this.  If strict typing is enabled for extensions, then TS compilation fails because various external type declarations do not exist.  For example, we refer to types for the `uuid` package, but `@types/uuid` is not installed -- it's a devdep in the monorepo root.  If strict typing is _not_ enabled for extensions, the types referencing these packages just become `any`, which is undesirable.

There are like 8-10 of these such packages.

(Other packages such as `axios`--and `appium`!--supply their own types, so e.g., a `@types/axios` is unneeded, and their own type declarations are published.  If you think of it that way, it makes sense.  I don't love it, but what can you do.)

boneskull:
Bonus: I got a change into `@types/teen_process`, so I've updated that to the newest version.

boneskull:
Yeah! It was fairly recent--about 9 mos ago.

boneskull:
OK, _finally_, an extension using strict type checks can use Appium.  This is a good thing.

boneskull:
stupid "degraded performance"

PR #17146, opened 2022-07-07
ci: Add a script that verifies Appium deployment on different OSs

mykola-mokhnach:


KazuCocoa (approved):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+        # node-version: [14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    steps:
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install Appium
+        run: npm i -g appium@next
```
jlipps, line 28: what if this script runs on a git tag `appium@2.3.1`? the npm `next` tag might still point to something else? do we not want to run `npm i -g appium@${{ tag }}` assuming we can get the tag as a variable?

jlipps (approved):
very cool, so this is like a postpublish script that we run whenever we make a `next` release?

.github/workflows/deploy.yml
mykola-mokhnach: `tag` is only defined when the pipeline is triggered by a new tag. It is going to be unset if the pipeline is triggered manually though

mykola-mokhnach: changed to use the current tag

Issue #17167, opened 2022-07-08
fix npm-related e2e test flake


boneskull:
## The problem

Our E2E tests which install extensions from npm fail _frequently_ due to timeouts.  This may be an npm network thing, it may be a GHA infrastructure thing, it may be rate limiting--who knows.  I've increased the timeout on these tests maybe 8-10 times already, and I don't want to do it anymore.

There's no guarantee that disabling the test timeouts entirely would work, either (but it might).  It may be that the `npm install` command would _never_ resolve.  Who knows.

One way that _might_ work to fix this is caching.  

The good news: because we now have a `package-lock.json` under VCS, we can cache our modules in CI.  In addition to the usual caching of `node_modules` and/or `~/.npm` (which should speed up our builds anyway), we may need to add a custom cache of the modules retrieved from npm during these E2E tests.  We can either a) add them to the `~/.npm` cache, b) cache the entire installed extension (wherever it's installed), or c) both.  A caveat with b) is that `APPIUM_HOME` is usually a random temp dir, so this behavior would need to be changed to be something still temporary, but static.

A second approach would be to eschew the use of the registry entirely, and just leverage npm to install a local module (e.g., `@appium/fake-driver`).  I think this is probably OK, and I'm 99% confident it'd work.  I am less confident in the previously-mentioned approach.

> Aside: in fact, `npm install /path/to/some/module` is the same thing as using `--source=local` which performs an `npm link`.  To that end, I'm fairly certain `--source=local` can just be an alias for `--source=npm`, and we shouldn't need to call `npm link` at all.  

@jlipps Thoughts?

boneskull:
If we were going to do local installs instead, I'm not sure of the best way to handle the "upgrade available" test(s)...

jlipps:
What if NPM changes the way that install works and introduces some difference between local and registry. I'd still like to have confidence that installing from the registry works. Maybe we can get away with just 1 test that does this. Or maybe we push this to our post deploy tests, where we are installing actual drivers rather than fake driver anyway.

PR #17168, opened 2022-07-08
chore(ci): attempt to cache deps

boneskull:
This change adds [bahmutov/npm-install](https://github.com/bahmutov/npm-install) to help cache dependencies, and enables caching against global modules via [setup-node](https://github.com/actions/setup-node).

Related: #17167


boneskull:
I estimate that on a cache hit, this is going to shave at least 30s off of the runtime per npm-installing job. The happy path has 9 such jobs, so it may save around 4m 30s per build.

jlipps (approved):

PR #17204, opened 2022-07-14
docs: delete old 1.x docs in favour of a ref to the 1.x branch

jlipps:
Best not to have conflicting docs in the same repo. We still need to pull a bunch of content from the 1.x docs but this can be done from the other branch.

This may break docs auto-generation for appium.io. We should probably turn that off at this point anyway.


KazuCocoa:
Maybe we can suspend or delete:

https://github.com/appium/appium/blob/master/ci-jobs/update-appium-io.yml

https://dev.azure.com/AppiumCI/Appium%20CI/_build?definitionId=37
https://dev.azure.com/AppiumCI/Appium%20CI/_build?definitionId=38

KazuCocoa:
have paused these tasks.

KazuCocoa (approved):

PR #17240, opened 2022-07-19
typecheck the tests

boneskull:
Currently, we don't typecheck anything in `packages/*/test/**` (though your editor might).  This PR causes (more accurately: _will cause_) those checks to happen.  These changes ensure that when writing tests, we're not making silly type errors causing false positives or whatever.

This change can happen piecemeal (per package). I'll break it into separate PRs one I sort out the dependency stack.

One side-effect (which isn't really feasible to avoid) is that type declaration files are generated from test source files--e.g., `somewhere/build/test` will be full of declarations (not JS compiled by Babel, though).  

We don't actually need them _at all_ AFAICT, as the way the build works is that packages can depend on the declarations from _other_ packages--and as of recently, no package depends on the `test` folder of any other package.  Anyway--it's difficult to avoid emitting these files without a completely separate composite project, so I don't want to take on the complexity of trying to manage that.  Declaration files for tests get generated--we'll just have to deal with it.  it _does_ mean we'll probably want to avoid publishing them since they are effectively useless.

### DIFF
--- a/config/tsconfig.base.json
+++ b/config/tsconfig.base.json
@@ -7,12 +7,13 @@
     "declaration": true,
     "declarationMap": true,
     "emitDeclarationOnly": true,
-    "module": "es2015",
+    "module": "esnext",
+    "target": "es2015",
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "strictNullChecks": true,
     "removeComments": false,
-    "target": "es2015",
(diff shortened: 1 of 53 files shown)


jlipps (commented):
I'm probably not going to be able to review this in detail. I say merge it when it passes the build and when you're satisfied it's correct. And yeah, let's not publish test types.

boneskull:
too old, will try again

PR #17251, opened 2022-07-20
feat(appium): pass unknown args to extension scripts

### LINKED ISSUE #17250: driver scripts should accept command-line flags
Drivers (or at least one of them) don't seem to accept flags/params/args on the command-line.  I don't know if this is a regression (@jlipps?).

boneskull:
Resolves #17250

This change passes any extra unknown arguments to extension scripts _if and only if_ `appium <extension> <name> run` is executed.  In all other cases, unknown arguments result in an error and the `appium` CLI will abort as before.

Example:

```bash
appium driver fake run fake-success --foo --bar
```

In the `fake-success` script of the `fake` driver, `process.argv.slice(2)` will be `['--foo', '--bar']`.  Extra arguments can appear anywhere after the `appium` executable in the command and do not need to appear after the script name.

- In addition, added some missing arguments for extension subcommands to the types (`json`, `suppressOutput`).  Extra arguments are stored in property `extraArgs`.
- Fixed a type problem surfaced by the previous item
- Changed help text for the `run` subcommand to hopefully read more clearly
- Reorganized E2E tests for the `run` subcommand
- Slightly changed error output when unknown argument is provided (given that we now need to DIY this)

### DIFF
--- a/packages/appium/lib/cli/parser.js
+++ b/packages/appium/lib/cli/parser.js
@@ -7,6 +7,8 @@ import {finalizeSchema, getArgSpec, hasArgSpec} from '../schema';
 import {rootDir} from '../config';
 import {getExtensionArgs, getServerArgs} from './args';
 
+export const EXTRA_ARGS = 'extraArgs';
+
 /**
  * If the parsed args do not contain any of these values, then we
  * will automatially inject the `server` subcommand.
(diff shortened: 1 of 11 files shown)


packages/appium/lib/cli/parser.js
```diff
@@ -97,8 +99,17 @@ class ArgParser {
     }
 
     try {
-      const parsed = this.parser.parse_args(args);
-      return ArgParser._transformParsedArgs(parsed);
+      const parsed = this.parser.parse_known_args(args);
+      const [knownArgs, unknownArgs] = parsed;
+      if (
+        (unknownArgs?.length && knownArgs.driverCommand === 'run') ||
+        knownArgs.pluginCommand === 'run'
```
jlipps, line 106: is this boolean check correct? seems weird for us not to care about `unknownArgs.length` if we have a plugin command?

boneskull: I think the parens are in the wrong place.

PR #17252, opened 2022-07-20
chore(): attempt to run e2e tests in parallel

### LINKED ISSUE #16125: CI/test performance improvements
There are some obvious ways we should be able to speed up our tests and CI:

- [ ] Run tests (especially E2E) in parallel.  `npm run e2e-test` actually just iterates over all of the package dirs in series, running `gulp e2e-test` in each.  Ideally, we would be able to just run all of it in parallel, since tests in package _a_ are unlikely to break tests in package _b_--but it's also likely that some tests would begin to timeout since they will now have fewer cores to work with.  Hosted GHA VM's have 2 cores--this change would see the most significant speedup on developer machines.  The first thing I'd try is running tests via `lerna run --parallel`
- [x] CI: Maintain the npm cache between runs.  May be a substantial speedup.
- ~~[ ] CI: Cache `APPIUM_HOME`. Probably not so impactful~~ **UPDATE:** Caching `node_modules` (which would be what this is) is [not recommended](https://github.com/actions/cache/blob/main/examples.md#node---npm).  I think we're fine here.

Any other ideas?

boneskull:
Resolves #16125


boneskull:
this is _likely_ failing because our E2E tests do not clean up after themselves properly; e.g. server still listening, socket still connected, etc.

Issue #17287, opened 2022-07-29
driver installation in appium 2 fails because of appium@2.0.0-beta.41 as peerDependencies


KazuCocoa:
## The problem

CI in ruby_linb_core started failing to install drivers since yesterday?:
My local installation also started failing with the same error. Something changed? After installing `npm -g install @babel/core @babel/cli` additionally. but still babel error occurred. 

```
STDERR:
npm ERR! code 127
npm ERR! path /Users/runner/hostedtoolcache/node/16.16.0/x64/lib/node_modules/appium
npm ERR! command failed
npm ERR! command sh -c npm run build
npm ERR! > appium@2.0.0-beta.41 build
npm ERR! > babel lib --root-mode=upward --out-dir=build/lib
npm ERR! sh: babel: command not found
{
  "error": {
    "code": 127,
    "summary": "command failed",
    "detail": "sh -c npm run build\n> appium@2.0.0-beta.41 build\n> babel lib --root-mode=upward --out-dir=build/lib\nsh: babel: command not found"
  }
}
```

https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=22725&view=logs&j=4fb3a4db-6fd8-54a2-5c9f-0d6cc4b8481f&t=fa8e92e4-0ef8-556b-09ec-7d22643adb66

https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=22725&view=results is the root CI tasl.
https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=22711&view=results was ok. This was on Wed 27th 5pm PDT.

I wondered 2.0.0-beta 41 had something initially, but the same error continued with `appium@2.0.0-beta.40`

part of https://github.com/appium/ruby_lib_core/blob/master/ci-jobs/functional/run_appium.yml:
```
steps:
- task: NodeTool@0
  inputs:
    versionSpec: 16.x
  displayName: Install Node 16.x to get NPM v8
- script: npm install -g appium@${APPIUM_VERSION}
  displayName: Install appium beta
- script: npm install -g mjpeg-consumer
  displayName: Install MJPEG Consumer
- script: npm list --depth 2 -g || echo 'ok'
  displayName: Installed node dependencies
- script: appium driver install ${APPIUM_DRIVER}
  displayName: Install appium driver ${APPIUM_DRIVER}
  env:
    APPIUM_HOME: $(Agent.HomeDirectory)/appium
```

[...]

PR #17288, opened 2022-07-30
fix(support): use legacy peer deps for npm v7+

boneskull:
This partially addresses issue #17287

npm v6 still seems confused, but I think I just need to do further testing on it.  we don't support npm v6 for a dev env so I'm not sure if I should worry about it, but need to determine if the weirdness is more widespread.

### DIFF
--- a/packages/support/lib/env.js
+++ b/packages/support/lib/env.js
@@ -63,7 +63,10 @@ export async function hasAppiumDependency(cwd) {
   try {
     listResult = await npm.list(cwd, 'appium');
 
+    // this is behavior for npm v7
+
     // short-circuit if `appium` is hanging around but the user hasn't added it to `package.json`.
+    // npm v7+
     if (listResult?.dependencies?.appium?.extraneous) {
       return false;
     }
@@ -72,6 +75,11 @@ export async function hasAppiumDependency(cwd) {
     resolved = listResult?.dependencies?.appium?.resolved ?? '';
     // ...however, it cannot do so unless `version` is nonempty.
     version = listResult?.dependencies?.appium?.version ?? '';
+
+    // npm v6 is dumb and doesn't have "extraneous", so
+    if ('appium-monorepo' === listResult?.name) {
+      return false;
+    }
   } catch {
     try {
       const pkg = await readPackageInDir(cwd);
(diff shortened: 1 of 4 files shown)


boneskull:
To avoid this, part of the smoke test script for `appium` should attempt to install a driver.

boneskull:
yeah this doesn't work either.

boneskull:
solution: fix the tests

boneskull:
(or, alternatively, move the symlinking code elsewhere)

KazuCocoa (approved):
i don't know if i tested this correctly.... but at least a couple of packages' `appium driver install` worked with this branch by replacing a symlink in APPIUM_HOME/node_modules from beta 41 to this branch,

KazuCocoa:
lol, right https://github.com/appium/appium/pull/17288#issuecomment-1200057253
mm

packages/support/lib/env.js
```diff
@@ -161,7 +168,8 @@ export const resolveManifestPath = _.memoize(
   async function _resolveManifestPath(appiumHome) {
     // can you "await" in a default parameter? is that a good idea?
     appiumHome = appiumHome ?? (await resolveAppiumHome());
-    return path.join(appiumHome, MANIFEST_RELATIVE_PATH);
+    const retval = path.join(appiumHome, MANIFEST_RELATIVE_PATH);
```
mykola-mokhnach, line 171: do we need this intermediate variable?

mykola-mokhnach (approved):

mykola-mokhnach:
It looks like tests are not quite happy though

boneskull:
OK, I think I just want to revert 2a6a056187ce925d5776b7acc4954b10ecf9221b until we can find another solution to avoid extra installations of `appium`.  This means that it will work like beta 39 and earlier; `appium driver install foo`--if `foo` has a `peerDependency` on `appium` (which it should)--will cause `appium` to be installed within `APPIUM_HOME` with newer versions of `npm`.

I think this may have been broken with `npm` pre-v7, so we can choose to drop that or possibly do the symlink (which 2a6a056187ce925d5776b7acc4954b10ecf9221b adds) if the detected `npm` is pre-v7.

KazuCocoa:
I see, sounds reasonable to me.

PR #17289, opened 2022-07-30
docs: address npm version as npm v7+

KazuCocoa:
## Proposed changes

Recently we had a couple of issues like https://github.com/appium/appium/pull/17288 and a few reported issues related to npm version in package installation...
My understanding about main this reason was we use `peerDependencies` to resolve dependencies especially main packages in `@appium` like base driver for others.

So, I wonder we should address supported npm version as well.

I know of nodejs14 bundles npm v6, so users who use nodejs14 need to update the npm version. Others like nodejs16 bundles npm v8, so this won't be an issue for them.

what do you think...?
@mykola-mokhnach @jlipps @boneskull 
cc @saikrishna321 @SrinivasanTarget @wswebcreation

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ to run tests on. Each of the drivers above documents the requirements for their
 particular brand of automation. At a minimum, you will need to be able to run
 the recent [LTS](https://nodejs.org/en/about/releases/) Node.js version.
 
-NPM version is v7+.
+NPM version needs to be v7+.
 
 ### Get Started
 
--- a/README.md
+++ b/README.md
@@ -201,6 +201,8 @@ to run tests on. Each of the drivers above documents the requirements for their
 particular brand of automation. At a minimum, you will need to be able to run
 the recent [LTS](https://nodejs.org/en/about/releases/) Node.js version.
 
+NPM version is v7+.
+
 ### Get Started
 
 Check out our [Getting Started](/docs/en/about-appium/getting-started.md) guide


saikrishna321 (approved):

mykola-mokhnach:
Currently package.json of appium packages contains the limitation of `node >=14, npm >= 6`, also node 14 is still being supported till spring 2023. The fact peer depencencies did not work is related to the issue package lookup issue, which might be resolved by linking `appium` to the folder where drivers are installed. 

The main different of npm7 from npm6 is that the latter forces download of peer dependencies when these are not found, which is not that we want. This unnecessary increases space usage and creates duplicates, although helps to workaround "module not found" exception while the fix above is being merged.

mykola-mokhnach:
@boneskull to confirm

mykola-mokhnach (changes requested):
see my comments above

KazuCocoa:
Ah.. mainly space issue. I haven't deeply caught with this so I thought something dependency resolution stuff. I understand.

My concern was only in case we release potential non-workaround on user side issues like https://github.com/appium/appium/issues/17287 which prevent installing drivers, for example, to keep compatibility for npm v6 in our future work. 
(Actually, I haven't tried but the 17287 could avoid by adding peerDependency as beta.40 in the project root's peerDeps...?)

KazuCocoa:
I succeeded in installing npm version v6 with NodeJS 16, so the 17287 actually had a workaround. https://github.com/appium/appium/issues/17287#issuecomment-1200111788 (I did not have an idea to downgrade the npm version at that time.)
Current npm version issue has workarounds to update/downgrade npm, so probably we can keep nodejs's bundled version.

Let me close this issue now.

PR #17297, opened 2022-08-01
fix(appium): partially revert 2a6a056187ce925d5776b7acc4954b10ecf9221b

boneskull:
This removes the symlink we were creating to avoid an extra install of `appium` in `APPIUM_HOME`.  Using a symlink obviously breaks stuff, because (for whatever reason) `npm` wants to run lthe `prepare` lifecycle script in the symlinked `appium`, which will not work outside of a dev environment due to missing dev depenencies.

Because npm v7+'s behavior is to install peer dependencies automatically, this will cause `appium` to be installed alongside any extensions in `APPIUM_HOME`. Each extension will have its own `appium`. This may just be the price we have to pay for using `APPIUM_HOME`, or there may be another solution to avoid the extra install (I'm not sure what it is).  I think it _also_ exposes us to future bugs where the running Appium differs from whatever the peer dependency is set to... which sort of defeats the purpose of peer dependencies.  Either way, the user will get a warning if the currently-running Appium is too new for an extension.

### DIFF
--- a/packages/appium/lib/main.js
+++ b/packages/appium/lib/main.js
@@ -4,7 +4,7 @@ import {init as logsinkInit} from './logsink'; // this import needs to come firs
 import logger from './logger'; // logger needs to remain second
 // @ts-ignore
 import {routeConfiguringFunction as makeRouter, server as baseServer} from '@appium/base-driver';
-import {logger as logFactory, util, env, fs} from '@appium/support';
+import {logger as logFactory, util, env} from '@appium/support';
 import {asyncify} from 'asyncbox';
 import _ from 'lodash';
 import {AppiumDriver} from './appium';
@@ -15,7 +15,6 @@ import {
   checkNodeOk,
   getGitRev,
   getNonDefaultServerArgs,
-  rootDir,
   showConfig,
   showBuildInfo,
   validateTmpDir,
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
We must sent the minimum NPM version to 7 in order for that to work

mykola-mokhnach:
> Using a symlink obviously breaks stuff, because (for whatever reason) npm wants to run lthe prepare lifecycle script in the symlinked appium, which will not work outside of a dev environment due to missing dev depenencies.

Also, it would probably make sense to report it as an NPM issue. Perhaps, there is a better way to handle such situation we don't know about

boneskull:
IDK if we've come to an agreement about the minimum npm version?

But yes, this PR needs to address that in some way.

boneskull:
On the one hand I'd like to drop npm v6 due to the behavior differences, but on the other hand Node.js v14 won't EOL until April of next year. Really just depends how many users will be held back from upgrading Appium due to this...

boneskull:
as per @jlipps it sounds like he's fine dropping v6.  I have another idea I'd like to try.

boneskull:
okay, giving up on my other idea.  let's go with this.

mykola-mokhnach:
I don't quite support this revert. Installing peer dependencies into each driver's node_modules does not make it different from non-peer dependencies strategy and does not save the disk space. Moreover there might be issues with server updates afterwards. 

One of the idea I had could be to add the path to `appium` into `NODE_PATH` while performing postinstall, so it could still be resolved from a temp folder

PR #17299, opened 2022-08-01
fix(*): update engines

boneskull:
See #17297 

- also add a missing dep in `@appium/schema`

PR #17301, opened 2022-08-01
fix(appium): add npm version check

boneskull:
This adds a warning for a too-old `npm` version and elimiates engine-related errors from being output twice.

### DIFF
--- a/packages/appium/lib/config.js
+++ b/packages/appium/lib/config.js
@@ -3,7 +3,6 @@ import _ from 'lodash';
 import {system, fs} from '@appium/support';
 import axios from 'axios';
 import {exec} from 'teen_process';
-import logger from './logger';
 import semver from 'semver';
 import findUp from 'find-up';
 import {getDefaultsForSchema, getAllArgSpecs} from './schema/schema';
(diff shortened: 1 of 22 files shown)


boneskull:
* master
  * **PR #17299** <a href="https://app.graphite.dev/github/pr/appium/appium/17299" target="_blank">Graphite</a> 
    * **PR #17301** <a href="https://app.graphite.dev/github/pr/appium/appium/17301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/17301?utm_source=stack-comment).

KazuCocoa (approved):

packages/appium/lib/config.js
```diff
@@ -28,6 +28,15 @@ function getNodeVersion() {
   return /** @type {import('semver').SemVer} */ (semver.coerce(process.version));
 }
 
+/**
+ * Returns version of `npm`
+ * @returns {Promise<string>}
+ */
+async function getNpmVersion() {
+  const {stdout} = await exec('npm', ['--version']);
```
jlipps, line 36: we shouldn't do anything clever to find the version of NPM installed in this node's global node modules? i can't remember whether "npm in the path" is already a requirement.

jlipps (approved):

packages/appium/lib/config.js
boneskull: we need `npm` to be in the `PATH` in order to install extensions, yes?

anyway this will break on windows I think. should be `npm.cmd`

boneskull: (imo finding `npm` if it _isn't_ in the `PATH` is kind of a fool's errand)

boneskull: fixed the windows business

PR #17306, opened 2022-08-02
docs(appium): migrate the migration docs and fix urls

jlipps:


packages/appium/docs/en/guides/migrating-1-to-2.md
```diff
@@ -2,4 +2,413 @@
...
+
+Appium 2.0 is the most major new release of Appium in over 5 years. The changes in Appium 2.0 are _not_ primarily related to changes in automation behaviors for specific platforms. Instead, Appium 2.0 reenvisions Appium as a _platform_ where "drivers" (code projects that introduce support for automation of a given platform) and "plugins" (code projects that allow for overriding, altering, extending, or adding behaviors to Appium) can be easily created and shared.
+
+At the same time, the Appium project is taking the opportunity to remove many old and deprecated bits of functionality.
+
+Together these do introduce a few breaking changes to how Appium is installed, how drivers and various features are managed, and protocol support. These are detailed below.
+
+## Breaking Changes
+
+Have a look at the [Appium 2.0 release notes](https://github.com/appium/appium/releases/tag/v2.0.0-beta) for a more comprehensive list of changes. Here we call out the breaking changes and what you need to do do account for them.
```
KazuCocoa, line 17: Will this link change?

packages/appium/docs/en/guides/migrating-1-to-2.md
```diff
@@ -2,4 +2,413 @@
...
+```
+appium driver install xcuitest
+appium driver install uiautomator2
+```
+
+At this point, your drivers are installed and ready. There's a lot more you can do with this CLI so be sure to check out the docs on it.
+If you're running in a CI environment or want to install Appium along with some drivers all in one step, you can do so using some special flags during install, for example:
+
+```
+npm install -g appium --drivers=xcuitest,uiautomator2
```
KazuCocoa, line 32:
```suggestion
npm install -g appium --use-drivers=xcuitest,uiautomator2
```

KazuCocoa (commented):
(still reading)

packages/appium/docs/en/guides/migrating-1-to-2.md
jlipps: yeah, good point. i don't think there's a single link that we know will work now so i'll make it more general.

PR #17321, opened 2022-08-04
feat(base-driver,fake-driver,appium): add convenience methods for defining execute script overloads

jlipps:
This is such a common pattern in pretty much every driver, I thought I'd take a stab at conventionalizing it. Open to feedback! Once we agree, I can add some unit tests in base-driver.

### DIFF
--- a/packages/appium/docs/en/ecosystem/build-drivers.md
+++ b/packages/appium/docs/en/ecosystem/build-drivers.md
@@ -582,7 +582,7 @@ driver.executeScript('soundz: playSong', [{song: 'Stairway to Heaven', artist: '
 ```
 
 Then in your driver code you could define the static property `executeMethodMap` as
-a mapping of script names to member functions on your driver. It has the same basic form as
+a mapping of script names to methods on your driver. It has the same basic form as
 `newMethodMap`. All you need to do at that point is implement the `execute` the command to take
 advantage of the built-in `executeMethod` method available on all drivers:
 
--- a/packages/appium/docs/en/ecosystem/build-drivers.md
+++ b/packages/appium/docs/en/ecosystem/build-drivers.md
@@ -581,7 +581,7 @@ call something like:
 driver.executeScript('soundz: playSong', [{song: 'Stairway to Heaven', artist: 'Led Zeppelin'}]);
 ```
 
-Then in your driver code you could simply define the magic static field `executeMethodMap` as
+Then in your driver code you could define the static property `executeMethodMap` as
 a mapping of script names to member functions on your driver. It has the same basic form as
 `newMethodMap`. All you need to do at that point is implement the `execute` the command to take
 advantage of the built-in `executeMethod` method available on all drivers:
--- a/packages/appium/docs/en/ecosystem/build-drivers.md
+++ b/packages/appium/docs/en/ecosystem/build-drivers.md
@@ -537,7 +537,7 @@ behaviours that don't map to any of the existing commands, you can create new co
 two ways:
 
 1. Extending the WebDriver protocol and creating client-side plugins to access the extensions
-1. Overloading the Execute Script command
+1. Overloading the Execute Script command by defining "Execute Methods"
 
 If you want to follow the first path, you can direct Appium to recognize new methods and add them
 to its set of allowed HTTP routes and command names. You do this by assigning the `newMethodMap`
@@ -566,11 +566,11 @@ create and release client-side plugins for each language you want to support (di
 examples can be found at the relevant client docs).
 
 An alternative to this way of doing things is to overload a command which all WebDriver clients
-have access to already: Execute Script. A common driver pattern is to create a magic prefix which
-the driver looks for at the beginning of any Execute Script invocation, and if it's present, to
-treat the rest of the script string as the name of a custom command to execute. This works in large
-part because for most platforms beyond web browsers, executing arbitrary JavaScript isn't a thing
-that makes sense.
+have access to already: Execute Script. Appium provides some convenience methods for making this
+easy. The convention is to create a sort of magic prefix for your users, which the driver looks for
+at the beginning of any Execute Script invocation, and if it's present, to treat the rest of the
+script string as the name of a custom command to execute. This works in large part because for most
+platforms beyond web browsers, executing arbitrary JavaScript isn't a thing that makes sense.
 
 Let's say you are building a driver for stereo system called `soundz`, and you wanted to create
 a command for playing a song by name. You could expose this to your users in such a way that they
@@ -581,21 +581,37 @@ call something like:
 driver.executeScript('soundz: playSong', [{song: 'Stairway to Heaven', artist: 'Led Zeppelin'}]);
 ```
 
-Then in your driver code you could simply look for `soundz:` at the beginning of your
-`execute` override, and do the appropriate thing:
+Then in your driver code you could simply define the magic static field `executeMethodMap` as
+a mapping of script names to member functions on your driver. It has the same basic form as
+`newMethodMap`. All you need to do at that point is implement the `execute` the command to take
+advantage of the built-in `executeMethod` method available on all drivers:
 
 ```js
-async execute(script, args) {
-  if (script === 'soundz: playSong') {
-    const {song, artist} = args[0];
-    // play the song based on song and artist details
-    return;
-  }
+static executeMethodMap = {
+  command: 'soundz: playSong',
+  params: {required: ['song', 'artist'], optional: []},
+}
 
-  // otherwise do the normal thing, or throw errors.NotYetImplementedError if you have nothing to
-  // do (see the section on WebDriver errors in this guide)
+async soundzPlaySong(song, artist) {
+  // play the song based on song and artist details
+}
+
+async execute(script, args) {
+  return await this.executeMethod(script, args);
 }
 ```
+
+A couple notes about this system:
+1. Arguments sent via `executeScript` will be unwrapped, the first argument selected as containing
+   parameters, validated, and then applied to your overload method in the order specified in
+   `executeMethodMap`. I.e., this framework assumes only a single actual argument sent in via
+   `executeScript` (and this argument should be an object with keys/values representing the
+   parameters your Execute Method expects).
+1. Your overload will always be bound to the driver instance.
+1. Appium does not automatically implement `execute` for you. You may wish, for example, to only
+   call the `executeMethod` helper function when you're not in proxy mode!
+1. The `executeMethod` helper will throw an error if a script name doesn't match one of the
+   script names defined as a command in `executeMethodMap`, or if there are missing parameters.
 
 ### Implement handling of Appium settings
 
--- a/packages/appium/docs/en/ecosystem/build-drivers.md
+++ b/packages/appium/docs/en/ecosystem/build-drivers.md
@@ -609,7 +609,7 @@ A couple notes about this system:
 1. Your overload will always be bound to the driver instance.
 1. Appium does not automatically implement `execute` for you. You may wish, for example, to only
    call the `executeMethod` helper function when you're not in proxy mode!
-1. The `executeMethod` helper will throw an error if a script name doesn't match one of the
+1. The `executeMethod` helper will reject with an `Error` if a script name doesn't match one of the
    script names defined as a command in `executeMethodMap`, or if there are missing parameters.
 
 ### Implement handling of Appium settings
--- a/packages/appium/docs/en/ecosystem/build-drivers.md
+++ b/packages/appium/docs/en/ecosystem/build-drivers.md
@@ -583,8 +583,7 @@ driver.executeScript('soundz: playSong', [{song: 'Stairway to Heaven', artist: '
 
 Then in your driver code you could define the static property `executeMethodMap` as
 a mapping of script names to methods on your driver. It has the same basic form as
-`newMethodMap`. All you need to do at that point is implement the `execute` the command to take
-advantage of the built-in `executeMethod` method available on all drivers:
+`newMethodMap`. Once `executeMethodMap` is defined, you'll need to implement the `execute` command:
 
 ```js
 static executeMethodMap = {
--- a/packages/appium/docs/en/ecosystem/build-drivers.md
+++ b/packages/appium/docs/en/ecosystem/build-drivers.md
@@ -566,29 +566,31 @@ create and release client-side plugins for each language you want to support (di
 examples can be found at the relevant client docs).
 
 An alternative to this way of doing things is to overload a command which all WebDriver clients
-have access to already: Execute Script. Appium provides some convenience methods for making this
-easy. The convention is to create a sort of magic prefix for your users, which the driver looks for
-at the beginning of any Execute Script invocation, and if it's present, to treat the rest of the
-script string as the name of a custom command to execute. This works in large part because for most
-platforms beyond web browsers, executing arbitrary JavaScript isn't a thing that makes sense.
-
-Let's say you are building a driver for stereo system called `soundz`, and you wanted to create
-a command for playing a song by name. You could expose this to your users in such a way that they
-call something like:
+have access to already: Execute Script. Appium provides some a convenient tool for making this
+easy. Let's say you are building a driver for stereo system called `soundz`, and you wanted to
+create a command for playing a song by name. You could expose this to your users in such a way that
+they call something like:
 
 ```js
-// webdriverio example
+// webdriverio example. Calling webdriverio's `executeScript` command is what trigger's Appium's
+// Execute Script command handler
 driver.executeScript('soundz: playSong', [{song: 'Stairway to Heaven', artist: 'Led Zeppelin'}]);
 ```
 
-Then in your driver code you could define the static property `executeMethodMap` as
-a mapping of script names to methods on your driver. It has the same basic form as
-`newMethodMap`. Once `executeMethodMap` is defined, you'll need to implement the `execute` command:
+Then in your driver code you can define the static property `executeMethodMap` as a mapping of
+script names to methods on your driver. It has the same basic form as `newMethodMap`, described
+above. Once `executeMethodMap` is defined, you'll also need to implement the Execute Script command
+handler, which according to Appium's routes mapping is called `execute`. The implementation can
+call a single helper function, `this.executeMethod`, which takes care of looking at the script and
+arguments the user sent in and routing it to the correct custom handler you've defined. Here's an
+example:
 
 ```js
 static executeMethodMap = {
-  command: 'soundz: playSong',
-  params: {required: ['song', 'artist'], optional: []},
+  'soundz: playSong', {
+    command: 'soundzPlaySong',
+    params: {required: ['song', 'artist'], optional: []},
+  }
 }
 
 async soundzPlaySong(song, artist) {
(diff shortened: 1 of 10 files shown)


packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -581,22 +581,30 @@ call something like:
...
-async execute(script, args) {
-  if (script === 'soundz: playSong') {
-    const {song, artist} = args[0];
-    // play the song based on song and artist details
-    return;
-  }
+executeOverloads = {'soundz: playSong': this.soundzPlaySong}
+
+async soundzPlaySong({song, artist}) {
+  const {song, artist} = args[0];
```
mykola-mokhnach, line 593: where args is defined here?

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -566,11 +566,11 @@ create and release client-side plugins for each language you want to support (di
 examples can be found at the relevant client docs).
 
 An alternative to this way of doing things is to overload a command which all WebDriver clients
```
mykola-mokhnach, line 568: I call these "driver extensions" in README

packages/fake-driver/lib/driver.js
```diff
@@ -5,6 +5,12 @@ import {FakeApp} from './fake-app';
 import commands from './commands';
 
 class FakeDriver extends BaseDriver {
+  executeOverloads = {
```
mykola-mokhnach, line 8: would this require breaking changes in existing drivers?

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: copy/paste error.

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: since we use the word "extension" to mean something different (a driver or plugin), maybe we could come up with something else. also, drivers can add new commands via the `extraMethodMap` strategy as well.

packages/fake-driver/lib/driver.js
jlipps: no, existing drivers already implement something like this on their own and could continue to do so. but we could refactor them to take advantage of this feature and remove some code.

jlipps:
hmm, i think i may rewrite this to enable an interface more like the one we already have for new method map, something like:

```js
static executeOverloads = {
    'mobile: swipe': {
        command: 'executeMobile_swipe',
        params: {required: ['direction'], optional: ['element']}
    },
    'mobile: installApp': {
        command: 'executeMobile_installApp',
        params: {required: ['path']}
    }
}
```

i have a feeling this would enable a better dev experience and better documentability (easier docs auto generation; could use all the same tools for generating standard command docs.

packages/appium/docs/en/ecosystem/build-drivers.md
mykola-mokhnach: it still there. Where the changes not pushed?

packages/base-driver/lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,37 @@
...
+   */
+  class ExecuteCommands extends Base {
+    /**
+     * @this {Driver}
+     * @param {string} script
+     * @param {any[]} args
+     */
+    async executeOverload(script, args) {
+      const availableScripts = _.keys(this.executeOverloads);
+      if (!availableScripts.includes(script)) {
```
boneskull, line 20: seems like this could just be 

```js
if (!this.executeScripts[script]) {
```

yeah?

boneskull (commented):
will review more tomorrow

jlipps:
I changed the way I was doing this to bring it up to parity with `newMethodMap`. @mykola-mokhnach @boneskull what do you think of this? I think it will provide us a convenient way to document these additional methods as well

packages/base-driver/lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+export function ExecuteMixin(Base) {
+  /**
+   * @implements {IExecuteCommands}
+   */
+  class ExecuteCommands extends Base {
+    /**
+     * @this {Driver}
+     * @param {string} script
+     * @param {Array<Record<string, any>>} protoArgs
```
boneskull, line 17: Given the comment and code below,

```suggestion
     * @param {[Record<string, any>]|[]} protoArgs
```

...is more appropriate.  This is a "tuple" (a JS array having a specific number and order of items) with one member of type `Record<string,any>` _or_ an empty array.  It cannot be `[Record<string,any>|undefined]` because that would be an array having a single member `undefined`. It cannot be `[Record<string,any>?]` because that's another way of writing `[Record<string,any>|null]`, then the previous sentence applies.

packages/base-driver/lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,58 @@
...
+      if (!_.isPlainObject(args)) {
+        throw new Error(`Did not receive an appropriate executeOverload parameters object`);
+      }
+
+      const Driver = /** @type {DriverClass} */ (this.constructor);
+      const availableScripts = _.keys(Driver.executeOverloadMap);
+      if (!availableScripts.includes(script)) {
+        throw new Error(
+          `Unsupported overloaded execute script '${script}'. Available scripts ` +
+            `are: ${JSON.stringify(availableScripts)}`
```
boneskull, line 36: would probably `availableScripts.join(', ')` but that's me

packages/base-driver/lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,58 @@
...
+        );
+      }
+      let argsToApply = [];
+      const commandMetadata = Driver.executeOverloadMap[script];
+      if (!commandMetadata.params) {
+        commandMetadata.params = {required: [], optional: []};
+      }
+      checkParams(commandMetadata.params, args, null);
```
boneskull, line 44: if `!commandMetadata.params` is `true`, what is the use of calling `checkParams` the "empty" one we created?

packages/fake-driver/test/e2e/general-tests.js
```diff
@@ -106,6 +106,39 @@ function generalTests() {
...
+      await driver.executeScript('fake: addition', [{num1: 2, num2: 3}]).should.eventually.eql(5);
+      await driver
+        .executeScript('fake: addition', [{num1: 2, num2: 3, num3: 4}])
+        .should.eventually.eql(9);
+    });
+
+    it('should throw not implemented if an execute overload isnt supported', async function () {
+      await driver
+        .executeScript('fake: blarg', [])
+        .should.eventually.be.rejectedWith(/Unsupported overload/);
```
boneskull, line 128:
```suggestion
        .should.be.rejectedWith(/Unsupported overload/);
```

😄

packages/fake-driver/test/e2e/general-tests.js
```diff
@@ -106,6 +106,39 @@ function generalTests() {
...
+    });
+
+    it('should throw an error if a required overload param is missing', async function () {
+      await driver
+        .executeScript('fake: addition', [{num3: 4}])
+        .should.eventually.be.rejectedWith(/Parameters were incorrect/);
```
boneskull, line 134:
```suggestion
        .should.be.rejectedWith(/Parameters were incorrect/);
```

packages/fake-driver/test/e2e/general-tests.js
```diff
@@ -106,6 +106,39 @@ function generalTests() {
...
+    });
+
+    it('should throw an error if sending in wrong types of params', async function () {
+      await driver
+        .executeScript('fake: addition', [4, 5])
+        .should.eventually.be.rejectedWith(/appropriate.+parameters/);
```
boneskull, line 140:
```suggestion
        .should.be.rejectedWith(/appropriate.+parameters/);
```

packages/types/lib/driver.ts
```diff
@@ -44,6 +44,19 @@ export interface SessionCommands {
   getSession(): Promise<SingularSessionData>;
 }
 
+export interface ExecuteCommands {
+  executeOverload(script: string, args: any[]): Promise<any>;
```
boneskull, line 48: This signature should match the one [here](https://github.com/appium/appium/pull/17321/files#diff-68bc5501503dcc37a3bc53d4dbd921a1f7d02f7e015f330aa0e9b14b3ed1e880R19), yes?

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -581,22 +581,36 @@ call something like:
 driver.executeScript('soundz: playSong', [{song: 'Stairway to Heaven', artist: 'Led Zeppelin'}]);
 ```
 
-Then in your driver code you could simply look for `soundz:` at the beginning of your
-`execute` override, and do the appropriate thing:
+Then in your driver code you could simply define the magic static field `executeOverloadMap` as
```
boneskull, line 584: I'm not sure I dig `executeOverload`.  While I appreciate the ultraviolence it evokes, maybe `executeMap` would work just as well?

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -581,22 +581,36 @@ call something like:
...
 ```js
-async execute(script, args) {
-  if (script === 'soundz: playSong') {
-    const {song, artist} = args[0];
-    // play the song based on song and artist details
-    return;
-  }
+static executeOverloadMap = {
+  command: 'soundz: playSong',
+  parameters: {required: ['song', 'artist'], optional: []},
```
boneskull, line 592: This example code is wrong (`executeOverloadMap` is of the wrong shape)

boneskull: I'll buy you a sandwich if you change the API to accept a JSON schema instead of `{required, optional}`.  I can hook up the validation later.  I'd rather just not keep adding more of these DSLs for describing the shape of parameters

boneskull: that'd look like, for the above example (using the presumably correct object shape):

```js
static executeOverloadMap = {
  'soundz: playSong': {
    properties: {
      song: {type: 'string'},
      artist: {type: 'string'}
    },
    required: ['song', 'artist']
  }
}
```

We may be able to add a shorthand syntax, like:

```js
static executeOverloadMap = {
  'soundz: playSong': {
    'appium:stringProps': ['song', 'artist'],
    required: ['song', 'artist']
  }
}

```

either way, I think using JSON schema semantics here, while more verbose than what you originally had, is negligible when considering the utility the new API offers.

packages/base-driver/lib/basedriver/commands/execute.js
jlipps: there is no use

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: i think the main reason not to do that now is it is the way we have defined all appium commands. i think changing all of that is a bigger task. i am merely trying to make this new feature idea map better to existing ways of defining commands.

packages/base-driver/lib/basedriver/commands/execute.js
jlipps: ah, good thinking!

packages/fake-driver/test/e2e/general-tests.js
jlipps: oh right haha

packages/types/lib/driver.ts
jlipps: yeah, fixing

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: maybe `executeMethodMap` on parity with `newMethodMap`. we could call these `Execute Methods`

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: so yeah, i'm going to keep it for now since it's the same as everything we have in `routes.js` (except of course we don't have an HTTP method, and we call it `params` instead of `payloadParams` (I am open to renaming `params` in this PR to `payloadParams` for greater consistency).

but I agree with you, migrating our overall command definition to use json schema might be a good thing!

packages/base-driver/lib/basedriver/commands/execute.js
jlipps: yeah

packages/base-driver/lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * @this {Driver}
+     * @param {string} script
+     * @param {[Record<string, any>]|[]} protoArgs
+     */
+    async executeMethod(script, protoArgs) {
+      // the w3c protocol will give us an array of arguments to apply to a javascript function.
+      // that's not what we're doing. we're going to look for a JS object as the first arg, so we
+      // can perform validation on it. we'll ignore everything else.
+      let args = protoArgs[0];
```
mykola-mokhnach, line 23: would this works if protoArgs is undefined?

mykola-mokhnach: maybe `protoArgs?.[0]` ?

packages/base-driver/lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+      if (!_.isPlainObject(args)) {
+        throw new Error(`Did not receive an appropriate executeMethod parameters object`);
+      }
+
+      const Driver = /** @type {DriverClass} */ (this.constructor);
+      const availableScripts = _.keys(Driver.executeMethodMap);
+      const commandMetadata = Driver.executeMethodMap[script];
+      if (!commandMetadata) {
+        throw new Error(
```
mykola-mokhnach, line 35: maybe we can throw a more specific W3C error?

packages/base-driver/lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,61 @@
...
+          `Unsupported overloaded execute script '${script}'. Available scripts ` +
+            `are: ${availableScripts.join(', ')}`
+        );
+      }
+      let argsToApply = [];
+      if (!commandMetadata.params) {
+        commandMetadata.params = {required: [], optional: []};
+      } else {
+        commandMetadata.params.required ??= [];
```
mykola-mokhnach, line 44: oh I did not know we have such operator

mykola-mokhnach (approved):

packages/base-driver/lib/basedriver/commands/execute.js
jlipps: yes, good idea. i added some more validation including this case.

packages/base-driver/lib/basedriver/commands/execute.js
jlipps: yep, found some more specific protocol errors.

packages/base-driver/lib/basedriver/commands/execute.js
jlipps: it transpiles anyway!

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -581,22 +581,38 @@ call something like:
...
+advantage of the built-in `executeMethod` method available on all drivers:
 
 ```js
-async execute(script, args) {
-  if (script === 'soundz: playSong') {
-    const {song, artist} = args[0];
-    // play the song based on song and artist details
-    return;
-  }
+static executeMethodMap = {
```
boneskull, line 590: this is still wrong, I think

packages/types/lib/driver.ts
```diff
@@ -44,6 +44,19 @@ export interface SessionCommands {
...
 }
 
+export interface ExecuteCommands {
+  executeMethod(script: string, args: [Record<string, any>]|[]): Promise<any>;
+}
+
+export interface ExecuteOverloadDef {
+  command: string,
+  params?: {
+    required?: string[],
```
boneskull, line 54: we should be able to use this to define the type (sort of) of e.g. `soundzPlaySong(song, artist)`.  I can take a look later, but at minimum we will be able to determine that one or both are optional.  are these always `string` values?

boneskull (approved):
just some grumbling about documentation, otherwise lgtm

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -566,11 +566,11 @@ create and release client-side plugins for each language you want to support (di
 examples can be found at the relevant client docs).
 
 An alternative to this way of doing things is to overload a command which all WebDriver clients
-have access to already: Execute Script. A common driver pattern is to create a magic prefix which
-the driver looks for at the beginning of any Execute Script invocation, and if it's present, to
-treat the rest of the script string as the name of a custom command to execute. This works in large
-part because for most platforms beyond web browsers, executing arbitrary JavaScript isn't a thing
-that makes sense.
+have access to already: Execute Script. Appium provides some convenience methods for making this
```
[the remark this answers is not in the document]
jlipps: fair enough.

packages/appium/docs/en/ecosystem/build-drivers.md
```diff
@@ -581,22 +581,38 @@ call something like:
...
+  return await this.executeMethod(script, args);
 }
 ```
 
+A couple notes about this system:
+1. Arguments sent via `executeScript` will be unwrapped, the first argument selected as containing
+   parameters, validated, and then applied to your overload method in the order specified in
+   `executeMethodMap`. I.e., this framework assumes only a single actual argument sent in via
+   `executeScript` (and this argument should be an object with keys/values representing the
+   parameters your Execute Method expects).
```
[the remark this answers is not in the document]
jlipps: you're right. i clarified a bunch.

packages/base-driver/lib/basedriver/commands/execute.js
```diff
@@ -0,0 +1,67 @@
...
+ * @param {SessionBase} Base
+ * @returns {ExecuteBase}
+ */
+export function ExecuteMixin(Base) {
+  /**
+   * @implements {IExecuteCommands}
+   */
+  class ExecuteCommands extends Base {
+    /**
+     * @this {Driver}
```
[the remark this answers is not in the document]
jlipps: turns out it's not. i copied this template from another file and it was there. i removed it.

packages/appium/docs/en/ecosystem/build-drivers.md
jlipps: you're right. i fixed it.

packages/types/lib/driver.ts
jlipps: `required` and `optional` are lists of parameter *names*. so yes the names are always strings. their values could be anything.

so yeah if we want to update this we'll need to migrate to a schema system like the one you proposed.

packages/fake-driver/test/e2e/general-tests.js
```diff
@@ -106,6 +106,45 @@ function generalTests() {
       res[0].type.should.eql('pointer');
       res[0].actions.should.have.length(2);
     });
+
+    it('should get and set a fake thing via execute overloads', async function () {
+      let thing = await driver.executeScript('fake: getThing', []);
+      should.not.exist(thing);
+      await driver.executeScript('fake: setThing', [{thing: 1234}]);
```
KazuCocoa, line 113: Just a question.
Do we allow `fake:setThing` like current mobile commands?, or will this strictly check the format? I saw some users did not set spaces in past, so I wondered spaces. If we make the format strict, we may need to address as a breaking change in a migration guide (or do something in drivers' update)

jlipps: Yes, this change does not do anything special with spaces. A way to make it a non-breaking change in the driver would be to define the method metadata and then include both ways of accessing, for example:

```js
const SET_THING_CMD = {
  command: 'setThing',
  params: {required: ['thing'], optional: []},
};

static executeMethodMap = {
  'fake: setThing': SET_THING_CMD,
  'fake:setThing': SET_THING_CMD,
}
```

which is slightly more verbose but would keep it from being a breaking change if we want.

or, we could add custom logic into `executeMethods` that looks for both versions if it finds a colon in the command. that seemed too unnecessary for this PR but we could do it.

KazuCocoa: I see, it makes sense as this base driver 🆗  Drivers may need to consider the space thing.

PR #17323, opened 2022-08-05
feat(many): static updateServer method receives 'cliArgs' param

### LINKED ISSUE #17304: plugins don't have access to cliArgs in updateServer
Plugins should reasonably have access to cli args passed in via the plugin cli arg interface.

Currently they have access via the `this.cliArgs` property on a plugin instance.

But plugins can also implement `static async updateServer`, a method which should also reasonably have access to the plugin's CLI args. Obviously, `this.cliArgs` is not a thing here.

Options:
1. make `updateServer` an instance method
2. figure out how to pass the cli args into `updateServer` so they're accessible

boneskull:
This PR passes a `cliArgs` parameter to the static method `updateServer` of drivers and plugins so they can take action on the server configuration based on the content of the arguments.

Note that `cliArgs` is a misnomer, because these args can also come from configuration files.

Closes #17304

### DIFF
--- a/packages/appium/lib/main.js
+++ b/packages/appium/lib/main.js
@@ -304,6 +304,7 @@ async function main(args) {
   const serverUpdaters = getServerUpdaters(driverClasses, pluginClasses);
   const extraMethodMap = getExtraMethodMap(driverClasses, pluginClasses);
 
+  /** @type {import('@appium/base-driver').ServerOpts} */
   const serverOpts = {
     routeConfiguringFunction,
     port: parsedArgs.port,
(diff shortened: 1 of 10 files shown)


saikrishna321:
@boneskull Does this PR also address accessing cliArgs from plug-in constructor ?

boneskull:
the plugin constructor already received cliArgs

saikrishna321:
Can you show an example on the usage?

We tried the below and didn't work.

constructor(name, cliArgs = {}) {
    super();
    this.name = name;
    this.cliArgs = cliArgs;
    console.log(`hello: ${JSON.stringify(this.cliArgs, null, 2)}`);
  }

boneskull:
See #17327

jlipps (approved):

PR #17336, opened 2022-08-08
chore(eslint-config-appium): upgrade to ESLint v8

boneskull:
BREAKING CHANGE:

This change upgrades to ESLint v8 and removes `@babel/eslint-parser`, which is only necessary if we're using mid-stage TC39 proposals (which we aren't).  I think `@babel/core` was a peer dep of `@babel/eslint-parser`, so I removed that too.  And removed cruft from the main configuration.

ESLint rules _very likely_ had breaking changes, but I didn't experience any on our codebase.  However, this version of ESLint seems to be incompatible with `gulp-eslint`, so `@appium/gulp-plugins` should be held back from upgrading.

In addition:

- Updated `README.md`
- Updated some fields in `package.json`
- Loosened `peerDependencies`, as they're supposed to be loose.


mykola-mokhnach:
@boneskull would you be able to also bump driver dependencies after this is merged? I'm confused about what needs to be bumped and what's not

boneskull:
packages still using gulp cannot upgrade.  so mostly nothing to do

mykola-mokhnach (approved):
I hope this should not be an issue, since we anyway import dev deps directly in drivers. These are not declared as peer dependencies.

jlipps (approved):

PR #17339, opened 2022-08-08
fix(support): path to APPIUM_HOME must always be absolute

### LINKED ISSUE #17338: extension resolution potentially broken when APPIUM_HOME is a relative path
I noticed while testing a plugin that when `APPIUM_HOME=../some/relative/path`, resolution of the extension breaks, because the module resolution begins within the `appium` install itself (`extension-config.js`).  

`APPIUM_HOME` must be converted to an absolute path as soon as it's determined.  That would happen in `env.js` of `@appium/support`.

boneskull:
Closes #17338

### DIFF
--- a/packages/support/lib/env.js
+++ b/packages/support/lib/env.js
@@ -112,6 +112,8 @@ export const readPackageInDir = _.memoize(
  * - If `APPIUM_HOME` is set in the environment, use that
  * - If we have an `extensions.yaml` in {@linkcode DEFAULT_APPIUM_HOME}, then use that.
  * - If we find a `package.json` in or above `cwd` and {@linkcode shouldUseCwdForAppiumHome} returns `true`, then use the directory containing the `package.json`.
+ *
+ * All returned paths will be absolute.
  */
 export const resolveAppiumHome = _.memoize(
   /**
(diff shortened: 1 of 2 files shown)


packages/support/lib/env.js
```diff
@@ -112,21 +112,23 @@ export const readPackageInDir = _.memoize(
...
   async function _resolveAppiumHome(cwd = process.cwd()) {
-    if (process.env.APPIUM_HOME) {
-      return process.env.APPIUM_HOME;
-    }
-
     if (!path.isAbsolute(cwd)) {
       throw new TypeError('`cwd` parameter must be an absolute path');
     }
 
+    if (process.env.APPIUM_HOME) {
```
mykola-mokhnach, line 128: should we also validate if `process.env.APPIUM_HOME` is a relative path?

boneskull: If it's an absolute path, `path.resolve(someAbsolutePath, process.env.APPIUM_HOME)` will just be `process.env.APPIUM_HOME`.  If it's relative, then `path.resolve()` will make it absolute.  I don't see any need to check if it's relative before performing `path.resolve()`.

mykola-mokhnach (approved):

packages/support/lib/env.js
mykola-mokhnach: this was not obvious to me

jlipps (approved):

PR #17350, opened 2022-08-10
feat(appium,support): enable detection of local extensions for improved DX

### LINKED ISSUE #17340: third-party extension working copies should be discoverable
## The problem

If `appium` (the package) is a dependency of the project in the current working directory, `appium` (the executable) will automatically look into `node_modules` for installed extensions.  _However_, it should _also_ look at `$CWD/package.json`.

This should prevent the workaround to symlink `my-extension/node_modules/my-extension` to `my-extension`

boneskull:
This change makes Appium recognize the existence of an extension in the current (or parent) dir, and automatically adds it to the extension manifest.

Resolution of Appium (e.g., _how_ Appium was installed) is no longer taken into account when determining if we're "in" a package that depends upon Appium.  So you can easily `npm link appium` in an extension working copy if you wish, and this will not impact Appium's behavior.

To support this change, `hasAppiumDependency()` in `@appium/support`'s `env` has changed to be a wrapper around the new `findAppiumDependency()`.  The strategy here is to use `npm list --long` to get detailed information about the current directory.  The result will have an object with a `path` property, and that property points to the current workspace root (or current project root).  We can then read the `package.json` in the `path` property--or just the CWD--to make the determination about an `appium` dependency (or lack thereof).  For extension working copies, this will return `true` because `appium` will be present in `devDependencies`, `peerDependencies`, or both.  For any other project, it will usually just be in `devDependencies`.  If there's no `package.json` in the CWD and we're not in a workspace, Appium continues to use the default `APPIUM_HOME`.

One side effect of this is that running `appium` from the monorepo itself treats the dir hierarchy as a package which has an `appium` dependency (because it does!).  To that end, the default `APPIUM_HOME` will _not_ be used when running `appium` from _anywhere in the monorepo_ unless you explicitly set `APPIUM_HOME` in an environment variable.  Generally, our tests prescribe their own temporary `APPIUM_HOME` directories, so this should have no effect on them (in the general case).  It _will_ have an effect on running `appium` manually in a development environment, as the workspace root will be treated as `APPIUM_HOME`; running `node ./packages/appium/index.js driver install uiautomator2` (or `npm start -- driver install uiautomator2`) will install that driver as a dev dependency (and save it).  I could add a special case _just for the monorepo_ if we don't want this to happen, but it may introduce complexity we/I don't want to maintain, so I left it out.  Furthermore, `npm start -- driver list` will show the `fake` driver, and `npm start -- plugin list` will show all of the plugins in `packages/`.

Regardless, I view this as a solid win for extension developers.

Bonus:

While I don't have benchmarks, this should considerably improve performance when detecting locally-installed (`node_modules/some-extension`) or extension working copies.  Instead of using `klaw` to walk the filesystem (in serial!) we can just use `glob` to look for relevant `package.json` files.  Instead of waiting until `glob` is completely finished searching, we can listen for its `match` evens and begin processing `package.json` files as soon as they are found. Because `@appium/support` only exposes a promisifed `glob()`--with no access to the `EventEmitter` it returns--`appium` depends on `glob` directly.

* * *

Closes #17340


boneskull:
For example, I've `npm link`'d my `appium` into my `appium-uiautomator2-driver` working copy, then reset my environment.  We can see `uiautomator2` is automatically detected, and points to the current dir:

```
appium-uiautomator2-driver on  master is 📦 v2.4.6 via  v18.4.0 took 4s
❯ export APPIUM_HOME=

appium-uiautomator2-driver on  master is 📦 v2.4.6 via  v18.4.0
❯ rm -rf ~/.appium

appium-uiautomator2-driver on  master is 📦 v2.4.6 via  v18.4.0
❯ appium driver list
✔ Listing available drivers
- uiautomator2@2.4.6 [installed (NPM)]
- xcuitest [not installed]
- youiengine [not installed]
- windows [not installed]
- mac [not installed]
- mac2 [not installed]
- espresso [not installed]
- tizen [not installed]
- flutter [not installed]
- safari [not installed]
- gecko [not installed]

appium-uiautomator2-driver on  master is 📦 v2.4.6 via  v18.4.0
❯ ls ~/.appium
ls: /Users/boneskull/.appium: No such file or directory

appium-uiautomator2-driver on  master is 📦 v2.4.6 via  v18.4.0
❯ bat node_modules/.cache/appium/extensions.yaml
  File: node_modules/.cache/appium/extensions.yaml
  drivers:
    uiautomator2:
      automationName: UiAutomator2
      platformNames:
        - Android
      mainClass: AndroidUiautomator2Driver
      pkgName: appium-uiautomator2-driver
      version: 2.4.6
      appiumVersion: ^2.0.0-beta.40
      installType: npm
      installSpec: appium-uiautomator2-driver@2.4.6
  plugins: {}
  schemaRev: 2
```

packages/appium/lib/extension/manifest.js
```diff
@@ -175,26 +160,58 @@ export class Manifest {
...
+     * @returns {Promise<void>}
+     */
+    const onMatch = async (filepath) => {
+      try {
+        const pkg = JSON.parse(await fs.readFile(filepath, 'utf8'));
+        if (isDriver(pkg) || isPlugin(pkg)) {
+          const changed = this.addExtensionFromPackage(pkg, filepath);
+          didChange = didChange || changed;
+        }
+      } catch {}
```
mykola-mokhnach, line 181: when node started to support this syntax?

packages/support/test/unit/env.spec.js
```diff
@@ -36,6 +36,10 @@ describe('env', function () {
     delete process.env.APPIUM_HOME;
 
     env = rewiremock.proxy(() => require('../../lib/env'), overrides);
+
+    env.findAppiumDependencyPackage.cache = new Map();
```
mykola-mokhnach, line 40: should we use LRUCache instead?

packages/appium/lib/extension/manifest.js
boneskull: I don't recall, but it does.

packages/support/test/unit/env.spec.js
boneskull: no, this is just `_.memoize()`, which adds a `cache` prop of type `Map<args, return value>` to the function.  this just ensures the memoization is reset after every test, for defensive purposes

jlipps (approved):
lgtm when ci passes

boneskull:
@mykola-mokhnach @KazuCocoa Is there anything happening in CI that you know of that would be effected by this?  e.g., E2E tests leveraging `appium` that assume `APPIUM_HOME` is `~/.appium`?

KazuCocoa:
Searched APPIUM_HOME in our repos. For now, only ruby_lib_core tests have `$(Agent.HomeDirectory)/appium` as APPIUM_HOME, but others do not specify it. So mostly tests will use the default APPIUM_HOME

jlipps:
Pretty sure a lot of the plugin tests I wrote before (unless they were changed), set a `local_appium_home` directory for their tests, and used the appiumHome config when starting the appium server.

boneskull:
@jlipps Yeah, this does not seem to impact the tests in the monorepo at all, since they do not assume anything about `APPIUM_HOME`

jlipps:
OK, sounds like the plugin tests were adjusted then when brought into the monorepo.

boneskull:
the potential problem is repos that _do_ assume `APPIUM_HOME` is somewhere.  It's a little difficult to run all the E2E tests of everything in a local machine, but I guess I could try.

boneskull:
@KazuCocoa I don't think anything besides extensions would be impacted, so no clients, if that's what you're talking about

boneskull:
I see a problem here (or maybe even without this PR) where the extension update check throws an exception if the extension is not published to npm

boneskull:
Created #17357 for that

boneskull:
I think this is OK to merge.  Let's give it another run thru CI after rebasing.

jlipps:
seems like we're good with CI now?

boneskull:
@KazuCocoa @mykola-mokhnach @jlipps fyi, this is going in.  remember, when running `appium` from within the monorepo, _the monorepo root_ is considered `APPIUM_HOME` _unless you provide an env var_.

PR #17351, opened 2022-08-10
fix(types): add missing @types/ws dep

boneskull:
Cannot consume these types without this dep...

```
node_modules/@appium/types/build/index.d.ts:3:41 - error TS2307: Cannot find module 'ws' or its corresponding type declarations.

3 import type { Server as WSServer } from 'ws';
                                          ~~~~
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7424,8 +7424,8 @@
     },
     "node_modules/@types/ws": {
       "version": "8.5.3",
-      "dev": true,
-      "license": "MIT",
+      "resolved": "https://registry.npmjs.org/@types/ws/-/ws-8.5.3.tgz",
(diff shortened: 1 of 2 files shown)


jlipps (approved):

jlipps:
is there a way for us to add this kind of check into CI?

boneskull:
@jlipps of course!  but I'm not sure the best way to go about it.  we _might_ be able to get away with just writing a test with a dummy `tsconfig.json` and trying to `tsc` it, _or_ there may be a tool that checks this sort of thing, _or_ it might need to be a smoke test _and if it does need to be a smoke test_ then we'd need to publish some sort of test fixture for it (which might be fine) _but if we don't want to do that_ then I'll need to figure out how e.g., `midnight-smoker` can best accept arbitrary fixtures or otherwise hack around the fact that we don't want some test fixture published in the tarball.

jlipps:
fair enough :-) doesn't break anything for our user users as it is, anyway

PR #17352, opened 2022-08-10
fix(appium): do not throw ENAMETOOLONG on long cli arg

### LINKED ISSUE #17308: Appium2 cli can't handle long start server commands
Appium2 doesn't seem to be able to handle long server start commands. For both beta 40 and 41

command in question:

appium -p 11000 -pa /wd/hub --session-override  --local-timezone --relaxed-security --log-timestamp  -bp 11100 --default-capabilities '{"appium:systemPort": 11300, "appium:webviewDevtoolsPort": 11600, "appium:udid": "19271JEG502669", "appium:platformName": "ANDROID", "appium:platformVersion": "11", "appium:chomedriverPorts": [[11400, 11499]], "appium:deviceName": "19271JEG502669_Pixel_5a"}' --debug

Log: 
WARN Appium Appium encountered 1 warning while validating drivers found in manifest /local_home/exec/.appium/node_modules/.cache/appium/extensions.yaml
WARN Appium Driver "uiautomator2" has 1 potential problem: 
WARN Appium   - Driver "uiautomator2" (package `appium-uiautomator2-driver`) may be incompatible with the current version of Appium (v2.0.0-beta.41) due to an invalid or missing peer dependency on Appium. A newer version of `appium-uiautomator2-driver` is available; please attempt to upgrade "uiautomator2" to v2.4.6 or newer.

ENAMETOOLONG: name too long, open '{"appium:systemPort": 11300, "appium:webviewDevtoolsPort": 11600, "appium:udid": "19271JEG502669", "appium:platformName": "ANDROID", "appium:platformVersion": "11", "appium:chomedriverPorts": [[11400, 11499]], "appium:deviceName": "19271JEG502669_Pixel_5a"}'

Tested the exact command against 1.20.2 and it doesn't have any issues. 

Edit: For now I can probably use the relaxed-cap plugin to reduce the command length, but this seems like an arbitrary limitation and probably not intended.

boneskull:
Some CLI args can be interpreted as filenames or JSON/CSV blobs.  Appium first assumes the value is a file and attempts to open it.  But if a blob is given and it is of some _problematic length_, the OS will throw an `ENAMETOOLONG` exception trying to open a filepath with a huge blob.  If the JSON is reasonably small, an `ENOENT` will be thrown instead.  Previous to this change, we were only looking for `ENOENT`; now we treat `ENAMETOOLONG` errors meaning "we have a JSON blob".  If the problem is that the filename _is_ actually too long, the user will get an "invalid JSON" error, which will probably be confusing, but is less likely to happen than the more-common error this addresses.

If someone complains about it in the future, we can take another look, but it's going to take more effort than it's worth atm to fix it (imo).

Closes #17308

### DIFF
--- a/packages/appium/test/e2e/cli.e2e.spec.js
+++ b/packages/appium/test/e2e/cli.e2e.spec.js
@@ -1,5 +1,7 @@
 // @ts-check
 
+import getPort from 'get-port';
+import {exec} from 'teen_process';
 import {npm, env, fs, tempDir, util} from '@appium/support';
 import B from 'bluebird';
 import path from 'path';
@@ -15,13 +17,14 @@ import {
   KNOWN_DRIVERS,
   PLUGIN_TYPE,
 } from '../../lib/constants';
-import {FAKE_DRIVER_DIR, resolveFixture} from '../helpers';
+import {FAKE_DRIVER_DIR, PACKAGE_ROOT, resolveFixture} from '../helpers';
 import {
   installLocalExtension,
   runAppiumJson,
   runAppiumRaw,
   readAppiumArgErrorFixture,
   formatAppiumArgErrorOutput,
+  EXECUTABLE,
 } from './e2e-helpers';
 
 const {MANIFEST_RELATIVE_PATH} = env;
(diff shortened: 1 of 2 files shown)


boneskull:
FWIW: any property defined in the config schema of type `object` can be passed to the command-line as a filepath to a JSON file (in theory). e.g., `--default-capabilities=my-default-caps.json` works.

packages/appium/test/e2e/cli.e2e.spec.js
```diff
@@ -523,7 +526,41 @@ describe('CLI behavior', function () {
...
+        it('should not throw an ENAMETOOLONG exception', async function () {
+          this.timeout(10000);
+          const capsArg = JSON.stringify({
+            'appium:platformName': 'ANDROID',
+            'appium:platformVersion': '11',
+            'appium:deviceName':
+              'Spicy jalapeno bacon ipsum dolor amet deserunt tempor pork belly aliqua drumstick, occaecat dolor venison et labore. Rump meatball pork chop tail. Consequat adipisicing kielbasa occaecat laborum pig. Qui pork chop chicken nostrud boudin fugiat. Proident ut culpa, chuck nulla sunt pastrami ut tri-tip. Buffalo dolore adipisicing, labore venison elit beef fatback kevin burgdoggen tail pancetta filet mignon. Dolor turducken rump, anim kevin sunt exercitation ham filet mignon beef ribs ad officia eiusmod id cillum.',
+          });
+
+          await expect(
```
boneskull, line 541: It occurs to me that this kind of sucks, but maybe is just the best way to launch `appium server` in a child process and make assertions about its behavior.  We aren't actually doing that elsewhere, which is sort of surprising.

boneskull: ..._kevin_...?

Jiurong-Yang:
> FWIW: any property defined in the config schema of type `object` can be passed to the command-line as a filepath to a JSON file (in theory). e.g., `--default-capabilities=my-default-caps.json` works for a valid `my-default-caps.json`.

This is a good alternative and all. But the new config conventions is not documented anywhere as far as I can tell. If certain limitations is too much trouble than it's worth to patch then it should be documented. Possibly even in the cli help message

jlipps:
@Jiurong-Yang they'll be documented in the appium 2 docs, currently living [here](https://appium.github.io/appium/docs/en/2.0/guides/config/) until we figure out a permanent home.

packages/appium/test/e2e/cli.e2e.spec.js
jlipps: super sketchy lorem ipsum

jlipps (approved):

boneskull:
@Jiurong-Yang As @jlipps wrote, this is beta software and our documentation is in-flux.  To be clear, the limitation I was writing about was this:  if a user provides a long `.json` filepath (exceeding the OS/FS limit) as an `appium` CLI argument, the resulting error seen by the user would be confusing/misdirecting.  I am content to worry about it when/if someone actually hits this error.

PR #17377, opened 2022-08-17
feat(): experimental support for typedoc generation

boneskull:
Run `npm run typedoc` to generate docs in `typedoc-docs` dir.

I thought I'd take a few mins and set this up.  It was _mostly_ straightforward, except:

1. Figuring out the "entry point" for each package was difficult until I realized `.js` entry points are assumed to be compiled, and thus require a `sourceMappingUrl` reference. IMO this is probably a bug in TypeDoc
2. The `sourceMappingUrl` must reference an external sourcemap file; not an inline source map.  If it's an inline source map, the system tries to `fs.open` a large data-url.
3. The babel config does not control whether external sourcemaps are created; `@babel/cli` does. So I needed to change those.
4. Practically speaking, I don't see any drawback to external sourcemaps _yet_
5. TypeDoc has poor support for cross-package type references in monorepos, and I had to add a plugin which implements a workaround
6. Per some suggestions in the docs, I modified the order of references in the root `tsconfig.json` to start with "the biggest one first"

I didn't modify any types or docstrings anywhere.

The intent here is to show a proof-of-concept.


boneskull:
See https://github.com/TypeStrong/typedoc/issues/2037

PR #17399, opened 2022-08-22
relocate driver/plugin-specific test utils into their own modules

### LINKED ISSUE #17398: @appium/test-support should not have a peer dep of appium
Recently I had moved some test harnesses from `@appium/base-driver` and `@appium/base-plugin` into `@appium/test-support`.  In retrospect, this was probably a bad idea.

Because of this change, `@appium/test-support` requires a peer dep of `appium`--otherwise we wind up with a dependency cycle where `@appium/test-support` depends on `@appium/base-driver` which depends on `@appium/test-support`.

I think the mistake here was the assumption that `@appium/test-support` was intended for extensions (drivers/plugins).  If this was the case, the peer dep of `appium` would be reasonable, because that's expected of extensions.  However, non-extensions get use from `@appium/test-support` (e.g. `appium-adb`), and if a package is not an extension, it should not necessarily need to install `appium`.

To that end I'd like to rip out the driver-and-plugin specific harnesses in `@appium/test-support` and put them back in their original locations.  However, they will be accessible via the public API instead of buried deep within the `test` directory.

Example:

```js
// old
import {driverE2ETestSuite} from 'appium/test'; // or '@appium/test-support'

// new
import {driverE2ETestSuite} from 'appium/driver/test';
```

This will necessitate `driver` and `plugin` dirs at the root of the `appium` package and associated modules in `@appium/base-driver` and `@appium/base-plugin` roots.

Alternatively, we could export from `appium/driver` itself, but that feels kind of strange to me.

boneskull:
This closes #17398.

The main idea:

- Removed driver and plugin test helpers from `@appium/test-support` (**BREAKING**)
- Put them in new packages `@appium/driver-test-support` and `@appium/plugin-test-support` respectively
- `appium` is no longer a peer dependency of `@appium/test-support`
- Removed `test` subpath export from `appium`; `import * from 'appium/test'` no longer allowed. (**BREAKING**)

I'm not sure about that last item on the above list.  We could keep it or not.  If we _do_ keep it, we probably want to re-export everything from the two new packages, since that's the main value the export provides.


boneskull:
I can split this into separate PRs if it seems overwhelming.  Most of the changesets are small changes to `package.json` files.

mykola-mokhnach (approved):

mykola-mokhnach:
Are there any non-monorepo modules under appium org that are going to be affected by this change? Should be bump the major module version after this PR is merged?

boneskull:
@mykola-mokhnach `@appium/test-support` will have its major bumped.  I don't see any usages of the code that's moving  outside of the monorepo though.  It's _possible_ someone else is using it... but that's what the major bump is for.

config/tsconfig.plugin.json
```diff
@@ -1,10 +1,17 @@
 {
   "compilerOptions": {
     "paths": {
-      "appium": ["../packages/appium"]
+      "appium": ["../packages/appium"],
+      "appium/plugin": ["../packages/appium/plugin"],
+      "@appium/plugin-test-support": ["../packages/plugin-test-support"]
```
jlipps, line 6: whyj do we need plugin-test-support stuff here, but not driver-test-support stuff?

jlipps: maybe i don't understand what this file is

packages/images-plugin/tsconfig.json
```diff
@@ -2,18 +2,14 @@
   "compilerOptions": {
     "outDir": "build",
     "paths": {
-      "@appium/opencv": ["../opencv"],
-      "appium": ["../appium"]
+      "@appium/opencv": ["../opencv"]
     }
   },
-  "extends": "../../config/tsconfig.base.json",
+  "extends": "../../config/tsconfig.plugin.json",
```
jlipps, line 8: oh. i get it now.

packages/plugin-test-support/README.md
```diff
@@ -0,0 +1,49 @@
+# @appium/pluginb-test-support
```
jlipps, line 1: `pluginb` -> `plugin`

jlipps (approved):

config/tsconfig.plugin.json
boneskull: `tsconfig.plugin.json` is reusable ts config _for plugins_.  therefore we don't care about the driver support module

PR #17406, opened 2022-08-23
fix FOSSA workflow

boneskull:
We'll see if this works...


boneskull:
OK.  Since the job works against the `master` branch, we won't see the effects until this is merged.

The nightly build might fail due to license/security reasons; I've tried to address these but I don't think it'll "stick" until this is merged.

jlipps (approved):

PR #17407, opened 2022-08-23
chore(renovate): add pkg-dir to ignored packages

boneskull:
See #17374

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -20,6 +20,7 @@
         "globby",
         "inquirer",
         "log-symbols",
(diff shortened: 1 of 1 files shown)


jlipps (approved):

jlipps:
so like is this just a time bomb in the nodejs community? i hate that we can't keep updating so many packages. why doesn't esm interop work with babel? is it because we'd have to transpile dependencies and we have no idea to do that in a guaranteed way?

PR #17408, opened 2022-08-23
fix(appium): eat update info errors

### LINKED ISSUE #17357: exception thrown if extension does not exist in npm registry
When developing a new extension, `appium` can throw an exception _if_ the extension has not (yet) been published to the npm registry _when_ `appium` attempts to check for updates.

To fix, we can probably safely eat any exceptions thrown by spawning `npm` during a check for updates.

boneskull:
If a user runs e.g., "appium driver list --updates" and one of the installed drivers is _not_ published to npm, an error will no longer be emitted.

Closes #17357

### DIFF
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -576,13 +576,14 @@ class ExtensionCommand {
     // this is a helper method, 'ext' is assumed to already be installed here, and of the npm
     // install type
     const {version, pkgName} = this.config.installedExtensions[ext];
+    /** @type {string?} */
     let unsafeUpdate = await npm.getLatestVersion(this.config.appiumHome, pkgName);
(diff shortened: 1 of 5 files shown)


jlipps (approved):
i'm just a bit concerned that if we have bugs in any of our library code here they will not be surfaced and this will just not work. can we not just eat specific known errors?

boneskull:
I'll see what I can do.

boneskull:
@jlipps OK, I changed the strategy here.  `@appium/support`'s `npm` module now examines the resulting error for the `E404` string, which happens if the package is not registered.  If it's found, the error is eaten.  unfortunately, even with the `--json` flag being sent to `npm`, it still dumps a bunch of non-JSON stuff to the console, so there's really no better way to do it unless we want to write some heuristics to parse mixed output 😜

jlipps:
thanks, yeah this is better :-) just have a healthy disregard for the perfection of code i've written

PR #17417, opened 2022-08-25
wip: automatic docs generation

jlipps:
Figured this PR could be a basis for collaboration on automatic docs generation for drivers and plugins.

In its initial state, it shows examples of minimal updates to docstrings that could be used in conjunction with a custom docs generation plugin (for e.g. typedoc or jsdoc).

To take one example:

```js
/**
 * Get the text of an alert
 *
 * @appiumCommand
 * @returns {Promise<string>}
 */
commands.getAlertText = async function getAlertText() {
  this.assertAlert();
  return this.appModel.alertText();
};
```

All we have added to the docstring here is the "tag" `@appiumCommand`. This is what I see happening as a result in the plugin:

0. The plugin determines the function name for all functions tagged with `@appiumCommand`
1. The plugin looks up information about this command in 3 places:
    - Appium's route map
    - The extension's `newMethodMap`
    - The extension's `executeMethodMap`
2. Using the information found, it constructs appropriate documentation output  (e.g., command name, command parameters and types, response values, and W3C protocol information like HTTP method and route. Response values should be adjusted from e.g., `Promise<null>` to `null`, since while `Promise<null>` is the accurate type in our code, it probably doesn't map to anything for an end user, who may be using Java or Python or Ruby.
3. Documentation should be organized so the user knows how to use these various commands. E.g. separated into sections like "Standard appium commands", "Custom appium commands", and "Execute method commands", since the way of accessing these command types might differ.

Note there is a 4th category of commands we want to document, which have no examples in the fake-driver project. These are proxied commands. For example, the Android driver, when in proxy mode, supports the `getUrl` command. But this function is not represented anywhere in the driver codebase, because it is only proxied to Chromedriver. We might still want to allow driver authors to note that these commands are supported (rather than a hand-wavy "we support whatever chromedriver supports when in proxy mode"). So we could allow "empty" docstrings that look like this, maybe at the end of a file:

```js
/**
 * @appiumProxyCommand {Chromedriver} getUrl
 */
```

(TBD whether these should note param/return info; theoretically we can back these out of an appium routes lookup for any proxying to a standard webdriver endpoint)

In terms of the overall architecture of this system, I take it the following should also be requirements:
- docs generation system abstracted and made part of generic tooling that 3rd parties can run in their own extensions. (e.g., a 3rd party plugin author can `npm i @appium/docs && appium-docs gen-commands` and see some pretty output somewhere in HTML or Markdown (preferably markdown for easy copy/paste into READMEs)). probably should support a variety of configuration options that will become clear
- we use this system to generate docs for our own drivers/plugins
- some kind of integration should be possible with our own broader docs generation system. if we are using mkdocs for our drivers and plugins, there should be a way for us to pull in and link these command docs

Still a lot that can be discussed and figured out, for example do we have one doc page per command? Or put them all in a big table? Or different modes?


boneskull:
does it make more sense for the driver author to specify what _doesn't_ proxy to chromedriver?

jlipps:
> does it make more sense for the driver author to specify what doesn't proxy to chromedriver?

I think one option for driver authors is to just have something in their docs that says "when in a webview context, commands are generally proxied to chromedriver. see chromedriver / w3c docs for more information about what commands are available."

the option I'm proposing here is for driver authors that want to have a more self-contained set of docs for their users.

at this point specifying IN vs OUT is kinda arbitrary, but I'd say that specifying OUT would be a moving target. IMO it'd be better to specify IN since theoretically proxied drivers like chromedriver are going to be a smaller set of commands?

jlipps:
i'm also fine if we decide that `@appiumProxyCommand` is not on the table for this iteration of docs preparedness for appium 2. we can have the proxy mode copout message on all our docs.

PR #17452, opened 2022-09-06
feat(support): add fs.isExecutable()

boneskull:
This is handy for drivers which need to find executables in the filesystem.

### DIFF
--- a/packages/support/lib/fs.js
+++ b/packages/support/lib/fs.js
@@ -25,6 +25,7 @@ import sanitize from 'sanitize-filename';
 import which from 'which';
 import log from './logger';
 import Timer from './timing';
+import {isWindows} from './system';
 import {pluralize} from './util';
 
 const ncpAsync =
@@ -36,13 +37,38 @@ const findRootCached = _.memoize(pkgDir.sync);
 const fs = {
   /**
    * Resolves `true` if `path` is _readable_, which differs from Node.js' default behavior of "can we see it?"
-   * @param {import('fs').PathLike} path
+   *
+   * On Windows, ACLs are not supported, so this becomes a simple check for existence.
+   *
+   * This function will never reject.
+   * @param {PathLike} path
    * @returns {Promise<boolean>}
    */
   async hasAccess(path) {
     try {
       await fsPromises.access(path, constants.R_OK);
-    } catch (err) {
+    } catch {
+      return false;
+    }
+    return true;
+  },
+
+  /**
+   * Resolves `true` if `path` is executable; `false` otherwise.
+   *
+   * On Windows, this function delegates to {@linkcode fs.hasAccess}.
+   *
+   * This function will never reject.
+   * @param {PathLike} path
+   * @returns {Promise<boolean>}
+   */
+  async isExecutable(path) {
+    try {
+      if (isWindows()) {
+        return await fs.hasAccess(path);
+      }
+      await fsPromises.access(path, constants.R_OK | constants.X_OK);
+    } catch {
       return false;
     }
     return true;
@@ -50,7 +76,7 @@ const fs = {
 
   /**
    * Alias for {@linkcode fs.hasAccess}
-   * @param {import('fs').PathLike} path
+   * @param {PathLike} path
    */
   async exists(path) {
     return await fs.hasAccess(path);
@@ -104,7 +130,7 @@ const fs = {
 
   /**
    * Create an MD5 hash of a file.
-   * @param {import('fs').PathLike} filePath
+   * @param {PathLike} filePath
    * @returns {Promise<string>}
    */
   async md5(filePath) {
@@ -136,7 +162,7 @@ const fs = {
 
   /**
    * Create a hex digest of some file at `filePath`
-   * @param {import('fs').PathLike} filePath
+   * @param {PathLike} filePath
    * @param {string} [algorithm]
    * @returns {Promise<string>}
    */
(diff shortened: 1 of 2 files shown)


packages/support/lib/fs.js
```diff
@@ -48,9 +48,23 @@ const fs = {
...
   },
 
+  /**
+   * Returns `true` if `path` is executable
+   * @param {PathLike} path
+   * @returns {Promise<boolean>}
+   */
+  async isExecutable(path) {
+    try {
+      await fsPromises.access(path, constants.R_OK | constants.X_OK);
```
mykola-mokhnach, line 58: is just X_OK not enough?

packages/support/lib/fs.js
```diff
@@ -48,9 +48,23 @@ const fs = {
     return true;
   },
 
+  /**
+   * Returns `true` if `path` is executable
+   * @param {PathLike} path
+   * @returns {Promise<boolean>}
+   */
+  async isExecutable(path) {
```
mykola-mokhnach, line 56: should this method eventually verify if the given path exists?

jlipps (commented):
LGTM, pending answers to @mykola-mokhnach's questions

packages/support/lib/fs.js
boneskull: if it doesn't, `fs.access` will reject

packages/support/lib/fs.js
boneskull: afaik, no.  some executables must be read, e.g., shell scripts.  we could add an option to the method to _only_ check for the `x` bit in the future if necessary, but it _probably_ won't come up.

packages/support/lib/fs.js
mykola-mokhnach: I mean if this makes sense for the client code, which is going to call this API. At least we should mention this in the docstrings that `false` is returned for non-existing files. Also, it probably does not make sense on Windows (same as X_OK constant in Node.js

boneskull: OK, I can note that the function will never reject.

re: Windows, I see https://github.com/nodejs/node/issues/19192 which makes me think we just need to check for file existence on that platform.  I'll modify the function accordingly.

boneskull: oh, it's even worse than that.  we will need to actually do an `fs.stat()` on windows since `fs.access()` is seemingly a no-op (and we don't want to use `fs.existsSync` _and_ `fs.exists` is deprecated)

packages/support/lib/fs.js
```diff
@@ -36,21 +37,48 @@ const findRootCached = _.memoize(pkgDir.sync);
...
+   *
+   * On Windows, ACLs are not supported; resolves `true` if `path` exists.
+   *
+   * This function will _not_ reject if the file does not exist.
+   * @param {PathLike} path
+   * @returns {Promise<boolean>}
+   */
+  async isExecutable(path) {
+    if (isWindows()) {
+      return fs.hasAccess(path);
```
mykola-mokhnach, line 69: await ?

boneskull: hmm.. eslint didn't complain.   i'll fix it

packages/support/lib/fs.js
```diff
@@ -36,21 +37,48 @@ const findRootCached = _.memoize(pkgDir.sync);
...
+   * On Windows, ACLs are not supported, so we only check for file existence.
+   * @param {PathLike} path
    * @returns {Promise<boolean>}
    */
   async hasAccess(path) {
     try {
-      await fsPromises.access(path, constants.R_OK);
-    } catch (err) {
+      if (isWindows()) {
+        await fsPromises.stat(path);
```
mykola-mokhnach, line 48: Do you have a link that confirms fs.access does not work on Windows? This method was working just fine before?

packages/support/lib/fs.js
```diff
@@ -36,21 +37,48 @@ const findRootCached = _.memoize(pkgDir.sync);
...
+   * On Windows, ACLs are not supported; resolves `true` if `path` exists.
+   *
+   * This function will _not_ reject if the file does not exist.
+   * @param {PathLike} path
+   * @returns {Promise<boolean>}
+   */
+  async isExecutable(path) {
+    try {
+      if (isWindows()) {
+        return await fs.hasAccess(path);
```
mykola-mokhnach, line 70: so on Windows it is just enough to verify the file exists to assume it is executable?

packages/support/lib/fs.js
boneskull: https://nodejs.org/api/fs.html

> On Windows, access-control policies (ACLs) on a directory may limit access to
a file or directory. The `fs.access()` function, however, does not check the
ACL and therefore may report that a path is accessible even if the ACL restricts
the user from reading or writing to it.

packages/support/lib/fs.js
boneskull: I don't know of any way to check this with node. it's possible there's some 3p module that has better windows support, but I don't know of one

packages/support/lib/fs.js
mykola-mokhnach: this just means it does not properly work with NTFS ACL, although it should be just ok to verify if a file exists. Lets keep this method as is for now to avoid possible regressions as it is called in multiple places across Appium

packages/support/lib/fs.js
mykola-mokhnach: I think it would be simpler to explicitly mention in the docs that this method only makes sense for *nix and on Windows it would always return the same value.

boneskull: I'm confused; it doesn't always return the same value. It just defers to `hasAccess()`.  I'll change my wording anyway

mykola-mokhnach (approved):

PR #17457, opened 2022-09-07
refactor(appium): Prefer condition-based flow over exception-based

mykola-mokhnach:
## Proposed changes

IMHO exception-controlled flows might not always be reliable: https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why

## Types of changes

### DIFF
--- a/packages/appium/lib/schema/cli-transformers.js
+++ b/packages/appium/lib/schema/cli-transformers.js
@@ -1,5 +1,5 @@
 import {ArgumentTypeError} from 'argparse';
-import {readFileSync, accessSync, constants} from 'fs';
+import {readFileSync, existsSync} from 'fs';
 import _ from 'lodash';
 
 /**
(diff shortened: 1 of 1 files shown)


packages/appium/lib/schema/cli-transformers.js
```diff
@@ -85,20 +90,18 @@ export const transformers = {
...
-      // sessions, so holding up the event loop does not incur the usual
-      // drawbacks.
-      json = readFileSync(jsonOrPath, 'utf8');
-      loadedFromFile = true;
-    } catch (err) {
-      // unreadable files don't count.
-      // also `ENAMETOOLONG` can happen if we try to open a file that's a huge JSON string.
-      if (err.code !== 'ENOENT' && err.code !== 'ENAMETOOLONG') {
-        throw err;
+    if (existsSync(jsonOrPath)) {
```
jlipps, line 93: will existsSync throw if the path is super long? or will this get around the nametoolong error?

mykola-mokhnach: I don't thinks Node.js devs would allow such issue. But lets try:

```
Welcome to Node.js v18.6.0.
Type ".help" for more information.
> const fs = require('fs');
undefined
> fs.existsSync('x'.repeat(10000))
false
> 
```

jlipps (approved):

PR #17462, opened 2022-09-08
docs(appium): create a cli args docs generation system

jlipps:
- generate markdown tables and insert into args.md
- remove 1.x args generation docs
- write cli args docs file preamble

Screen Shot 2022-09-08 at 12 53 20


jlipps:
cc @appium/documentors

packages/appium/docs/en/cli/args.md
```diff
@@ -2,4 +2,54 @@
...
-TODO
+To start the Appium server, you may either run `appium` or `appium server`. The `server` subcommand
+is considered to be the default, so if you omit it, Appium will interpret this as your request to
+run the Appium server. (You could run different subcommands, like `appium driver` or `appium
+plugin`: see the [Extensions CLI](./extensions.md) for more info).
+
+The invocation of `appium` (or `appium server`) takes a number of arguments, which are detailed
+below.
+
+!!! note
```
mykola-mokhnach, line 13: how about https://twitter.com/diegohaz/status/1527642881384759297?lang=en ?

packages/appium/docs/scripts/gen-cli-args-docs.js
```diff
@@ -0,0 +1,95 @@
...
+    let def = serverProps[arg].default;
+    let enums = serverProps[arg].enum;
+    if (type === 'array' || type === 'object' || (type === 'string' && def === '')) {
+      def = JSON.stringify(def);
+    }
+    if (type === 'array' && items.type) {
+      type = `array<${items.type}>`;
+    }
+    if (type === 'integer' && typeof minimum !== 'undefined' && typeof maximum !== 'undefined') {
+      description = `${description} (Value must be between \`${minimum}\` and \`${maximum}\`)`;
```
mykola-mokhnach, line 60: can it be that only minimum or maximum is defined?

jlipps: dunno, I just made sure we could handle everything I found in the current schema with a degree of niceness. i didn't dig into all the schema possibilities. basically if there are other things we could do, it shouldn't break this code. but it could be an opportunity for more niceness in a future update.

packages/appium/docs/en/cli/args.md
jlipps: that's awesome! this text won't primarily be consumed on github, though, so for now we still need to use the mkdocs-material supported syntax until they support the github syntax

packages/appium/docs/scripts/gen-cli-args-docs.js
jlipps: @mykola-mokhnach how does that sound to you?

mykola-mokhnach: seems reasonable

PR #17466, opened 2022-09-09
fix(base-driver,types): driver constructors accept DriverOpts, not ServerArgs

boneskull:
`ServerArgs` are command-line flags; `DriverOpts` are `ServerArgs` _and_ capabilities


boneskull:
yeah this is basically wrong

PR #17473, opened 2022-09-12
Add colorized console logs to Appium (1.x)

mikesalvia:
## Proposed changes

Add colorized console logs to Appium (1.x)

## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [X] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [X] I have signed the [CLA](https://cla.js.foundation/appium/appium)
- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)
- [X] Any dependent changes have been merged and published in downstream modules

## Screenshot
Screen Shot 2022-09-12 at 3 45 01 PM

PR #17481, opened 2022-09-13
Add colorized console logs to Appium (2.x)

mikesalvia:
## Proposed changes

Adds colorization to log prefixes using a color wheel

## Types of changes

## Checklist

- [X] I have signed the [CLA](https://cla.js.foundation/appium/appium)
- [X] Any dependent changes have been merged and published in downstream modules

## Screenshot
Screen Shot 2022-09-12 at 3 45 01 PM


mikesalvia:
@jlipps This is targeted for 2.0 I believe.

package.json
```diff
@@ -100,6 +100,7 @@
     "@appium/types": "file:packages/types",
     "@appium/universal-xml-plugin": "file:packages/universal-xml-plugin",
     "appium": "file:packages/appium",
+    "chalk": "4.1.2",
```
boneskull, line 103: `@appium/support` uses `@colors/colors` instead of `chalk`, so it'd probably be best to stick to that one unless there's a compelling reason to use `chalk` instead?

boneskull (commented):
I have no problems with the intent of this PR and don't see any issue with the source changes.

My reservations are that a) we would be pulling in a second "ANSI color support" lib (`chalk`) which duplicates basically the entire functionality of one we already use (`@colors/colors`), and b) the current logging system doesn't have a clear separation between human-readable and structured logs, so is already too complicated, and adding an enhancement such as this makes things potentially worse in terms of maintenance burden.

If we want to merge this as-is, then perhaps the `@colors/colors` dep in `@appium/support`  should be replaced by `chalk` as a separate PR.  

Given the above, I'm indifferent about merging this into Appium v2.

I would prefer to _not_ merge #17473 into v1.x given the lack of maintenance resources.

boneskull:
n/m what I wrote about #17473; I just closed it.

jlipps:
I think these are good concerns, and gets to a worry about this whole logging system in general. I'm not sure how to relate that to this PR in particular, which I think is a nice thing to have for users. But at the very least, maybe not duplicating colour libraries is a good idea. @mikesalvia would it be super hard to explore using the existing `@colors/colors` dep instead?

boneskull:
It may be that `@colors/colors` has no equivalent of `chalk.hex()`, idk

mikesalvia:
@boneskull That is correct. There is no hex way to set the color. I could just create the color list out of what `@colors/colors` has available. @jlipps How important is it not to add another dependency?

package.json
mikesalvia: There is no `.hex` in `@colors/colors` I am not against changing to use of the colors available in `@colors/colors` but I don't believe there are 24 evenly spaced colors of the rainbow to create a complete color wheel.

The colors are generated in hex from running `rainbow-colors-array` in a node script and copying the output, it seems like `@colors/colors` will only allow predefined colors.

mikesalvia:
> I think these are good concerns, and gets to a worry about this whole logging system in general. I'm not sure how to relate that to this PR in particular, which I think is a nice thing to have for users. But at the very least, maybe not duplicating colour libraries is a good idea. @mikesalvia would it be super hard to explore using the existing `@colors/colors` dep instead?

@jlipps Explored and there is no hex converter in the existing library. In fact, there is no way to use self-defined colors from what I can see, and there aren't enough colors to cover 24 (which we currently provide) as a constant.

mikesalvia:
Wanted to sync on this. Is there a preference to strip out chalk? The color set will be limited in @colors/colors.

jlipps:
i'm personally fine with a smaller cycle of colours if it helps us not include multiple overlapping deps

mikesalvia:
@jlipps Based on this comment I can very likely make 256 colors without adding the `chalk` dependency.

https://gist.github.com/abritinthebay/d80eb99b2726c83feb0d97eab95206c4?permalink_comment_id=3991965#gistcomment-3991965

packages/appium/lib/logsink.js
```diff
@@ -186,8 +188,29 @@ async function createTransports(args) {
...
+    nextColorIndex++;
+    if (colorNumber > 256) {
+      return '';
+    }
+    const ansiColorCode = '\x1b[38;5;' + (colorNumber) + 'm';
+    return ansiColorCode;
+  } catch (err) {
+    return '';
+  }
+}
```
mikesalvia, line 209: @jlipps @boneskull I can probably clean up this code a little if the strategy of using the ANSI color code is acceptable.

mikesalvia:
@jlipps I am having the strangest problem. I am not seeing my latest changes show up. I cleaned up the logsink.js file but it is not shown in GitHub.

packages/appium/lib/logsink.js
```diff
@@ -186,8 +186,26 @@ async function createTransports(args) {
   return transports;
 }
 
+let encounteredPrefixes = [];
```
jlipps, line 189: let's use `const`

packages/appium/lib/logsink.js
```diff
@@ -186,8 +186,26 @@ async function createTransports(args) {
...
+
+function getColorizedPrefix(prefix) {
+  let prefixId = prefix.split('@')[0].trim();
+  prefixId = prefixId.split(' (')[0].trim();
+  if (encounteredPrefixes.indexOf(prefixId) < 0) {
+    encounteredPrefixes.push(prefixId);
+  }
+  // using a multiple of 16 should cause 16 colors to be created
+  const colorNumber = encounteredPrefixes.indexOf(prefixId) * 16;
```
jlipps, line 198: so once we pass 16 prefixes, we stop colourizing? what about just cycling around? i think i'd rather see two prefixes share a colour than certain prefixes not have a colour?

jlipps (commented):
ok, just a couple more comments. also, maybe undo your package-lock changes so there's no conflicts?

packages/appium/lib/logsink.js
mikesalvia: Great idea! We can modulus 256, and always have a good color. Will do this!

packages/appium/lib/logsink.js
mikesalvia: Will that still let me append?

jlipps: (i.e., `const` doesn't mean "make this thing immutable", it just means "don't let me reassign this name")

mikesalvia: Ahhh! I see!

packages/appium/lib/logsink.js
mikesalvia: This has been resolved

packages/appium/lib/logsink.js
mikesalvia: Changed to const and moved to top

packages/appium/lib/logsink.js
```diff
@@ -186,8 +188,21 @@ async function createTransports(args) {
...
+  // use the modulus to cycle around color wheel
+  return `\x1b[38;5;${colorNumber % 256}m${prefix}\x1b[0m`;
```
mikesalvia, line 200: @jlipps this will insure we always color the prefix.

jlipps (approved):

jlipps:
nice! everything's good, just needs to be rebased for conflicts! then i will merge.

mikesalvia:
@jlipps anything else I need to do?

mikesalvia:
@jlipps looks like it has merge conflicts again.

mikesalvia:
@jlipps Hope you had a good week off. 🌴 I just fixed the conflicts.

Cheers! 🍻

jlipps:
hmm, github is still showing it's conflicted for me, though it doesn't say with what. what does it show for you @mikesalvia ?

mikesalvia:
Screen Shot 2022-10-24 at 4 46 14 PM

jlipps:
weird, some kind of github lag for me i guess. anyway, it's done. thanks @mikesalvia !

mikesalvia:
You bet! The same thing happened to me a week or 2 ago in this exact PR! Thanks! Look forward to seeing it in production! :-)

PR #17496, opened 2022-09-15
feat(types): adds types to derive capabilities from constraints

boneskull:
Driver authors and end-users may find these types useful.  I still haven't fleshed out how to best provide these to end-users, but we can use `DriverCaps<MyDriver>` which is an object containing Appium's caps _in addition to_ those defined by the `desiredCapConstraints` of `MyDriver`, e.g.:

```js

class MyClass {
  desiredCapConstraints = {foo: {isString: true, inclusion: ['bar', 'baz']}, quux: {isNumber: true}}
}

/** @typedef {DriverCaps<MyClass>} MyClassCaps */
// looks like {platformName: string, ..., foo: 'bar' | baz', quux: number}
```

DRY4eva

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,9 @@
     "*.js": [
       "eslint --fix",
       "prettier --write"
+    ],
+    "*.ts*": [
+      "prettier --write"
     ]
   },
   "prettier": {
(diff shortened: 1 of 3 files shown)


boneskull:
The next step here would be to modify the `DriverOpts` type (`this.opts` in a Driver) to use an appropriate generic based on this type.  So we will know that `this.opts.quux` is an actual thing, because we said so in the constraints.

boneskull:
(note that _all_ capabilities _as per the type system_ are still considered "optional", even though they may cause `validateCaps()` to error if missing or incorrect)

jlipps (commented):
wow i don't understand any of this typescript nonsense

PR #17559, opened 2022-10-04
chore(appium,base-driver,driver-test-support,types): overhaul types for constraints & caps

### LINKED ISSUE #17518: desiredCapConstraints makes me sad
## El Problemo

`desiredCapConstraints` is a property of a `Driver` object describing the allowed shape of incoming capabilities from WD clients.  The type of `desiredCapConstraints` is nominally `Readonly<Record<string,Constraint>>` where:

```typescript
// this is not quite what it looks like, but rather what it ought to be
interface Constraint {
  readonly presence?: boolean | Readonly<{allowEmpty: boolean}>;
  readonly isString?: boolean;
  readonly isNumber?: boolean;
  readonly isBoolean?: boolean;
  readonly isObject?: boolean;
  readonly isArray?: boolean;
  readonly deprecated?: boolean;
  readonly inclusion?: ReadonlyArray<string>;
  readonly inclusionCaseInsensitive?: ReadonlyArray<string>;
}
```

Given that they are not theoretically unique to instances, `desiredCapConstraints` should be immutable.  There's an argument here that they should actually be `static` properties, but that's a separate issue.

> Note that they aren't _entirely_ immutable otherwise; there's at least one place in the logic where we're deleting a property from a `Constraints` object--which isn't 100% necessary.

The problem is that while the `Driver` interface wants a `desiredCapConstraints` prop, the `BaseDriverCore` class (the thing that `BaseDriver` is before mixins) defines it with _accessors_.  The "real" `desiredCapConstraints`--which is the "built-in" set of `Constraint`s--is stored in a `_desiredCapConstraints` property.

The intent of the accessors is to allow subclasses to define their own constraints via variable assignment (e.g., `this.desiredCapConstraints = {stuff}`).  Magically, the setter merges the "built-in"'s from `BaseDriver` with whatever the subclass defines.  The getter returns the merged set.  Beyond this, the intent is unclear to me, and may simply be for convenience in `BaseDriver`, since (I'm assuming) the `Driver` author probably won't need to interact directly with constraints once they are defined.

So, due to the presence of a setter, we no longer have an immutable object. Because we don't have an immutable object, we cannot use the constraints to derive the shape of any incoming `Capabilities` (which we can do otherwise; see the `DriverCaps` type).  The end result is that calling e.g., `super.createSession()` from the subclass returns a (tuple containing a) `Capabilities` object that is constrainted _only_ by the shape of the "built-in" `desiredCapConstraints`.  Everything else is just of type `any`.

Furthermore, the accessor means that implementors cannot use things such as public fields to define their constraints; they _must_ use assignment to `this.desiredCapConstraints` (is this documented anywhere?) or Appium will claim that the constraints are unrecognized, and driver implementors will be clueless as to why.  For example, this will cause Appium to emit warnings:

```js
class MyDriver extends BaseDriver {
  desiredCapConstraints = {
    foo: {isString: true, presence: true},
    bar: {isBoolean: true, presence: true},
  };
}
```

Instead, this is what you have to do:

```js
class MyDriver extends BaseDriver {
  constructor(...args) {
    super(...args);
    this.desiredCapConstraints = {
      foo: {isString: true, presence: true},
      bar: {isBoolean: true, presence: true},
    };
  }
}
```

The thing that prompted this was that it _also_ creates problems for us when attempting to automatically generate documentation for drivers.  Unless a driver author explicitly adds types via TS or JSDoc, we cannot know what the shape of the `desiredCapConstraints` is, and thus cannot automatically generate the relevant documentation.

And finally, the shape of the built-in capabilities exist in two places: 1. in `@appium/types` via `@wdio/types`, and 2. as a `Constraints` object in `@appium/base-driver`.  Obviously, we don't need two separate places; we can just derive caps from the constraints.

## Loco Solution

1. Remove the `desiredCapConstraints`/`_desiredCapConstraints` from `BaseDriverCore` _entirely_. 
2. Move it into `@appium/types` as a constant object.  This is because `Capabilities` will now consume the object via `DriverCaps`.  It cannot live in `@appium/base-driver` because that would create a circular dependency.  Note that this may require a tweak to the TS configuration to generate the appropriate `.js` file (or we need to use Babel like the other packages).
3. Update `Capabilities` (and `W3CCapabilities`) to use this object instead of the stuff from wdio.
4. Import these constraints into `BaseDriver` and use it to validate caps, as before.
5. Wherever we are referencing `this.desiredCapConstraints` in `BaseDriver` (so, including the mixins), ensure that we've merged whatever `this.desiredCapConstraints` is (which will be defined by the downstream drivers) with the built-in constraints.
6. Probably re-export these built-in constraints if driver authors need them (unclear).

_Assuming downstream drivers aren't futzing with `this.desiredCapConstraints` directly_, this should (in theory):

1. Not break people
2. Make the built-in `Constraints` object the single-source-of-truth for what capabilities can contain
3. Allow subclasses of `BaseDriver` to define their own constraints via whatever means they prefer (assignment, public field, prototype assignment, etc)
4. Improve typing of capabilities within downstream drivers

## Yes, But

`desiredCapConstraints` should really just be a JSON schema. 😀

boneskull:
The big idea here is:

1. `desiredCapConstraints` should always contain the constraints _for that driver_; it no longer contains the "base" constraints. This is the only thing that really required any logic changes (though they are somewhat spread out).  The "merged" constraints object is available via `_desiredCapConstraints`.
2. `Capabilities` and W3C-style `W3CCapabilities` are now _derived from_ constraints.  This absolves us (and extension authors) from needing to create an interface describing their capabilities.
3. Changes to caps and constraints now reflect in `DriverOpts` (`this.opts`) in drivers
4. `caps` and `originalCaps` declarations have moved from `Core` to `Driver`
5. TypeScript compiles the JS output from `@appium/types` (which is just a static "constraints" object)

Also, from what I can tell, the shape of any given "capabilities" object should be fully typed; the only caveat is that the constraints object must be declared as a "constant" (this is what `/** @type {const} */` is doing, if you see it anywhere) so that the contents are recursively read-only and any primitives are interpreted as literals.  There is still a disconnect between what is optional per the type (everything) vs. what is actually optional per Appium (e.g., `platformName` is not).  It's trivial to disable "everything is optional", but the result would require a lot of work to wipe up--most of which is just appeasing the compiler.

Resolves #17518

### DIFF
--- a/packages/appium/lib/utils.js
+++ b/packages/appium/lib/utils.js
@@ -166,13 +166,10 @@ function parseCapsForInnerDriver(
  */
 function insertAppiumPrefixes(caps) {
   return /** @type {NSCapabilities<C>} */ (
-    /** @type {unknown} */ (
-      _.mapKeys(caps, (value, key) => {
-        if (STANDARD_CAPS_LOWERCASE.has(key.toLowerCase()) || key.includes(':')) {
-          return key;
-        }
-        return `${W3C_APPIUM_PREFIX}:${key}`;
-      })
+    _.mapKeys(caps, (_, key) =>
+      STANDARD_CAPS_LOWERCASE.has(key.toLowerCase()) || key.includes(':')
+        ? key
+        : `${W3C_APPIUM_PREFIX}:${key}`
     )
   );
 }
--- a/packages/appium/lib/utils.js
+++ b/packages/appium/lib/utils.js
@@ -1,11 +1,12 @@
 import _ from 'lodash';
 import logger from './logger';
-import {processCapabilities, PROTOCOLS} from '@appium/base-driver';
+import {processCapabilities, PROTOCOLS, STANDARD_CAPS} from '@appium/base-driver';
 import {inspect as dump} from 'util';
 import {node} from '@appium/support';
 import path from 'path';
 
 const W3C_APPIUM_PREFIX = 'appium';
+const STANDARD_CAPS_LOWERCASE = new Set([...STANDARD_CAPS].map((cap) => cap.toLowerCase()));
 
 /**
  *
@@ -35,16 +36,18 @@ const inspect = _.flow(
  * Takes the caps that were provided in the request and translates them
  * into caps that can be used by the inner drivers.
  *
- * @param {any} jsonwpCapabilities
- * @param {W3CCapabilities} w3cCapabilities
- * @param {import('@appium/types').Constraints} constraints
- * @param {import('@appium/types').DefaultCapabilitiesConfig} [defaultCapabilities]
- * @returns {ParsedDriverCaps|InvalidCaps}
+ * @template {Constraints} C
+ * @template [J=any]
+ * @param {J} jsonwpCapabilities
+ * @param {W3CCapabilities<C>} w3cCapabilities
+ * @param {C} constraints
+ * @param {NSCapabilities<C>} [defaultCapabilities]
+ * @returns {ParsedDriverCaps<C,J>|InvalidCaps<C,J>}
  */
 function parseCapsForInnerDriver(
   jsonwpCapabilities,
   w3cCapabilities,
-  constraints = {},
+  constraints = /** @type {C} */ ({}),
   defaultCapabilities = {}
 ) {
   // Check if the caller sent JSONWP caps, W3C caps, or both
@@ -52,14 +55,14 @@ function parseCapsForInnerDriver(
     _.isPlainObject(w3cCapabilities) &&
     (_.has(w3cCapabilities, 'alwaysMatch') || _.has(w3cCapabilities, 'firstMatch'));
   const hasJSONWPCaps = _.isPlainObject(jsonwpCapabilities);
-  let desiredCaps = /** @type {ParsedDriverCaps['desiredCaps']} */ ({});
-  /** @type {ParsedDriverCaps['processedW3CCapabilities']} */
+  let desiredCaps = /** @type {ParsedDriverCaps<C>['desiredCaps']} */ ({});
+  /** @type {ParsedDriverCaps<C>['processedW3CCapabilities']} */
   let processedW3CCapabilities;
-  /** @type {ParsedDriverCaps['processedJsonwpCapabilities']} */
+  /** @type {ParsedDriverCaps<C>['processedJsonwpCapabilities']} */
   let processedJsonwpCapabilities;
 
   if (!hasW3CCaps) {
-    return /** @type {InvalidCaps} */ ({
+    return /** @type {InvalidCaps<C>} */ ({
       protocol: PROTOCOLS.W3C,
       error: new Error('W3C capabilities should be provided'),
     });
@@ -78,7 +81,7 @@ function parseCapsForInnerDriver(
       for (const [defaultCapKey, defaultCapValue] of _.toPairs(defaultCapabilities)) {
         let isCapAlreadySet = false;
         // Check if the key is already present in firstMatch entries
-        for (const firstMatchEntry of w3cCapabilities.firstMatch || []) {
+        for (const firstMatchEntry of w3cCapabilities.firstMatch ?? []) {
           if (
             _.isPlainObject(firstMatchEntry) &&
             _.has(removeAppiumPrefixes(firstMatchEntry), removeAppiumPrefix(defaultCapKey))
@@ -102,7 +105,9 @@ function parseCapsForInnerDriver(
 
         // Only add the default capability if it is not overridden
         if (_.isEmpty(w3cCapabilities.firstMatch)) {
-          w3cCapabilities.firstMatch = [{[defaultCapKey]: defaultCapValue}];
+          w3cCapabilities.firstMatch = /** @type {W3CCapabilities<C>['firstMatch']} */ ([
+            {[defaultCapKey]: defaultCapValue},
+          ]);
         } else {
           w3cCapabilities.firstMatch[0][defaultCapKey] = defaultCapValue;
         }
@@ -129,7 +134,7 @@ function parseCapsForInnerDriver(
       desiredCaps = processCapabilities(w3cCapabilities, constraints, true);
     } catch (error) {
       logger.info(`Could not parse W3C capabilities: ${error.message}`);
-      return /** @type {InvalidCaps} */ ({
+      return /** @type {InvalidCaps<C,J>} */ ({
         desiredCaps,
         processedJsonwpCapabilities,
         processedW3CCapabilities,
@@ -145,7 +150,7 @@ function parseCapsForInnerDriver(
     };
   }
 
-  return /** @type {ParsedDriverCaps} */ ({
+  return /** @type {ParsedDriverCaps<C,J>} */ ({
     desiredCaps,
     processedJsonwpCapabilities,
     processedW3CCapabilities,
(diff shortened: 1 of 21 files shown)


packages/appium/lib/utils.js
```diff
@@ -155,57 +160,46 @@ function parseCapsForInnerDriver(
...
-      prefixedCaps[name] = value;
-    } else {
-      prefixedCaps[`${W3C_APPIUM_PREFIX}:${name}`] = value;
-    }
-  }
-  return prefixedCaps;
+  return /** @type {NSCapabilities<C>} */ (
+    /** @type {unknown} */ (
+      _.mapKeys(caps, (value, key) => {
+        if (STANDARD_CAPS_LOWERCASE.has(key.toLowerCase()) || key.includes(':')) {
```
mykola-mokhnach, line 171: this could also be a ternary operator

packages/appium/lib/utils.js
```diff
@@ -1,11 +1,12 @@
 import _ from 'lodash';
 import logger from './logger';
-import {processCapabilities, PROTOCOLS} from '@appium/base-driver';
+import {processCapabilities, PROTOCOLS, STANDARD_CAPS} from '@appium/base-driver';
 import {inspect as dump} from 'util';
 import {node} from '@appium/support';
 import path from 'path';
 
 const W3C_APPIUM_PREFIX = 'appium';
+const STANDARD_CAPS_LOWERCASE = new Set([...STANDARD_CAPS].map((cap) => cap.toLowerCase()));
```
mykola-mokhnach, line 9: perhaps Array.of would be more readable than the spread operator in this case?

mykola-mokhnach: also we can provide a mapping function as the second arg to `from`

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -29,37 +33,43 @@ function mergeCaps(primary = {}, secondary = {}) {
...
-  }
+  // Remove the 'presence' constraint if we're not checking for it
+  constraints = /** @type {C} */ (
+    _.mapValues(
+      constraints,
+      skipPresenceConstraint
+        ? /** @param {Constraint} constraint */
+          (constraint) => _.omit(constraint, 'presence')
+        : /** @param {Constraint} constraint */
+          (constraint) =>
```
mykola-mokhnach, line 65: nested ternary operator are very hard to read

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -77,54 +87,78 @@ function validateCaps(caps, constraints = {}, opts = {}) {
...
+      'pageLoadStrategy',
+      'proxy',
+      'setWindowRect',
+      'timeouts',
+      'unhandledPromptBehavior',
+    ])
+  )
+);
+
+const STANDARD_CAPS_LOWER = new Set([...STANDARD_CAPS].map((cap) => cap.toLowerCase()));
```
mykola-mokhnach, line 110: I saw this constant being duplicated somewhere above

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -77,54 +87,78 @@ function validateCaps(caps, constraints = {}, opts = {}) {
...
 function isStandardCap(cap) {
-  return !!_.find(
-    STANDARD_CAPS,
-    (standardCap) => standardCap.toLowerCase() === `${cap}`.toLowerCase()
-  );
+  return STANDARD_CAPS_LOWER.has(cap.toLowerCase());
 }
 
-// If the 'appium:' prefix was provided and it's a valid capability, strip out the prefix (see https://www.w3.org/TR/webdriver/#dfn-extension-capabilities)
-// (NOTE: Method is destructive and mutates contents of caps)
```
mykola-mokhnach, line 101: I like that the method does not mutate stuff anymore

packages/types/lib/capabilities.ts
```diff
@@ -106,30 +81,111 @@ export type AnyCase<T extends string> = string extends T
...
+  Extra extends StringRecord | void = void
+> = Capabilities<BaseDriverCapConstraints & C, Extra>;
+
+/**
+ * W3C-style capabilities for drivers extending `BaseDriver`.
+ *
+ * Includes {@linkcode BaseCapabilities}.
+ *
+ * @example
+ * ```js
```
jlipps, line 170: should this be `ts`?

jlipps (approved):

packages/appium/lib/utils.js
boneskull: you mean `Array.from`, but idk, I don't see either as inherently more "readable".  in other words stuff like this is not a clear win and I'd just as soon ignore it 🤷

packages/appium/lib/utils.js
```diff
@@ -155,57 +160,46 @@ function parseCapsForInnerDriver(
...
-    return caps;
-  }
-
-  /** @type {Capabilities} */
-  const fixedCaps = {};
-  for (let [name, value] of _.toPairs(caps)) {
-    fixedCaps[removeAppiumPrefix(name)] = value;
-  }
-  return fixedCaps;
+  return /** @type {Capabilities<C>} */ (_.mapKeys(caps, (_, key) => removeAppiumPrefix(key)));
```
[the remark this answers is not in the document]
boneskull: if caps was an array (of what?), no, it wouldn't. but our code assumes it isn't, so I'm not sure what you're getting at?

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -10,17 +10,21 @@ const APPIUM_VENDOR_PREFIX = 'appium:';
...
+ * @template {Constraints} [T={}]
+ * @template {Constraints} [U={}]
+ * @param {Capabilities<T>} [primary]
+ * @param {Capabilities<U>} [secondary]
+ * @returns {import('type-fest').Merge<Capabilities<T>, Capabilities<U>>}
+ * @see https://www.w3.org/TR/webdriver/#dfn-merging-capabilities)
  */
 function mergeCaps(primary = {}, secondary = {}) {
-  let result = Object.assign({}, primary);
+  let result = /** @type {import('type-fest').Merge<Capabilities<T>, Capabilities<U>>} */ ({
```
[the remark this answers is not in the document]
boneskull: so, I thought about this, but then I'm wondering if we should.

capabilities (the values thereof), afaict, cannot be plain objects.  they _can_ be arrays.  so what are we doing when we're merging arrays here?  the code (before I changed it) didn't handle this case, and simply overwrites the `primary` array with the `secondary` array.

we're not mutating the arrays to my knowledge (at least not since I changed that fn not to mutate), so I _think_ we should be in the clear, but it begs the question: are we doing the right thing w/r/t merging caps?  @jlipps

packages/base-driver/lib/basedriver/capabilities.js
boneskull: it's not a nested ternary 😄 

the second branch of the ternary is a function which itself contains a ternary.

packages/base-driver/lib/basedriver/capabilities.js
boneskull: yeah, so it lives in both `appium` and `@appium/base-driver`, and it seemed a little questionable to export it as part of the public API from the latter given its limited use-case.  I could go either way here.

packages/types/lib/capabilities.ts
```diff
@@ -87,10 +60,12 @@ type ConstraintToCapKind<C extends Constraint> = C['isString'] extends true
 
 /**
  * Given {@linkcode Constraint} `C`, determine if it is required or optional.
- * 
+ *
  * In practice, _all_ capabilities are considered optional per types, but various errors might be thrown if some are not present.
  */
-type ConstraintToCap<C extends Constraint> = C['presence'] extends true | {allowEmpty: true}
+export type ConstraintToCap<C extends Constraint> = C['presence'] extends
+  | true
```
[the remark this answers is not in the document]
boneskull: it's similar to what it does with `?` and `:` in ternaries.  seems OK to me, I guess.

packages/base-driver/lib/basedriver/capabilities.js
mykola-mokhnach: still hard to read

packages/types/lib/capabilities.ts
boneskull: yes

packages/base-driver/lib/basedriver/core.js
```diff
@@ -45,16 +47,6 @@ class DriverCore {
    */
   initialOpts;
 
-  /**
-   * @type {Capabilities}
-   */
-  caps;
```
mykola-mokhnach, line 51: is this a breaking change?

boneskull: AFAIK neither `DriverCore` nor `AppiumDriver` (the only other thing besides `BaseDriver` which extends it) is part of our public API, so no...

packages/types/lib/constraints.ts
```diff
@@ -0,0 +1,53 @@
+export const BASE_DESIRED_CAP_CONSTRAINTS = {
```
mykola-mokhnach, line 1: Object.freeze?

boneskull:
Usage example from [appium-tizen-tv-driver](/headspinio/appium-tizen-tv-driver):

Given capabilities object (e.g., "desired caps") and an associated type for the driver `TizenTVDriver`:

```js
export const desiredCapConstraints = /** @type {const} */({
  platformName: {
    isString: true,
    inclusionCaseInsensitive: ['TizenTV'],
    presence: true,
  },
  deviceName: {
    isString: true,
    presence: true,
  },
  deviceAddress: {
    isString: true,
  },
  app: {
    isString: true,
  },
  appPackage: {
    isString: true,
  },
  udid: {
    isString: true,
  },
  chromedriverExecutable: {
    isString: true,
    presence: true, // for now require a custom chromedriver to work with tv version
  },
  isDeviceApiSsl: {
    isBoolean: true,
  },
  useOpenDebugPort: {
    isNumber: true,
  },
  powerCyclePostUrl: {
    isString: true,
  },
  rcToken: {
    isString: true,
  },
  resetRcToken: {
    isBoolean: true,
  },
  sendKeysStrategy: {
    isString: true,
  },
  rcMode: {
    isString: true,
    inclusionCaseInsensitive: ['remote', 'js']
  },
  rcDebugLog: {
    isBoolean: true
  },
  rcKeypressCooldown: {
    isNumber: true
  },
  appLaunchCooldown: {
    isNumber: true
  }
});

/**
 * @typedef {typeof desiredCapConstraints} TizenTVDriverCapConstraints
 */
```

the capabilities object(s) accepted by e.g., `TizenTVDriver.createSession(caps)` are of type

```js
/** 
 * @typedef {import('@appium/types').W3CDriverCaps<TizenTVDriverCapConstraints>} W3CTizenTVDriverCaps
 */
```

e.g.:

```js
  /**
   * @param {W3CTizenTVDriverCaps} w3cCapabilities1
   * @param {W3CTizenTVDriverCaps} [w3cCapabilities2]
   * @param {W3CTizenTVDriverCaps} [w3cCapabilities3]
   * @param {DriverData[]} [driverData]
   */
  async createSession(w3cCapabilities1, w3cCapabilities2, w3cCapabilities3, driverData) {
    // ...
  }
```

This provides intellisense and typechecks for all properties of those `w3cCapabilities*` parameters (e.g., it now knows that the `rcMode` cap can only have a value of `js` or `remote` (case-insensitive!!), _in addition to_ all the "base" caps defined in the basedriver.  Previous to this change, we were not able to get that level of detail.

boneskull:
Anyway, this will help us auto-generate command docs.

packages/base-driver/lib/basedriver/capabilities.js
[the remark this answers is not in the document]
boneskull: hmm... capabilities can be objects, but I don't see any way to declare that in constraints.

mykola-mokhnach (approved):

PR #17564, opened 2022-10-05
chore(): remove babel-runtime

### LINKED ISSUE #17536: evaluate need for @babel/runtime
The long and short of it is that minimum Node.js v14 may mean we don't need to ship `@babel/runtime` with anything. That also means removing the plugin that adds the associated `require` statement.

boneskull:
Closes #17536

Given our target platform (Node.js v14) the dep on `@babel/runtime` doesn't seem to buy us much of anything, because there's very little polyfilling going on.   From what I can tell, it's just a collection of functions which help sort out syntax unsupported by the target platform.

In my testing, the resulting tarballs for each package were a few bytes smaller after this change.  This suggests to me that we aren't using any of those provided functions--they'd be strewn throughout the resulting distfiles if we were!

In other words, the only thing we really a compiler for translating from ESM-ish sources to CJS.

_But_ we drop a prod dep for each of them, which seems like a clear win, if not a minor one.

* * *

At some point in the future, I'd like to see if we can replace it with something more lightweight or more performant.  TypeScript's compiler is also something we could use, since we're already using it for other stuff.


boneskull:
I think I'm going to rebase this and create a separate commit for `@appium/gulp-plugins`, because I think it's a breaking change there.

boneskull:
...actually, on second thought, no--it's not a breaking change, because it does not impact what is polyfilled--just _how_ it's polyfilled.  external users of gulp-plugins--_if_ they see a negative impact (which is a big "if")--would only see resulting bundle sizes increase.

mykola-mokhnach (approved):
I recently tried removed gulp plugins from known appium modules, at least these that we want to support further.
The intention is to get rid of gulp usage in appium completely and deprecate the corresponding module for good

jlipps (commented):
\o./

PR #17575, opened 2022-10-06
fix: Load configs after server arguments parsing is completed

mykola-mokhnach:
## Proposed changes

We do not want to spend extra time for configs loading if the server is executed with `-v` or `--show-config` args

## Types of changes

### DIFF
--- a/packages/appium/lib/main.js
+++ b/packages/appium/lib/main.js
@@ -170,12 +170,6 @@ function areServerCommandArgs(args) {
  * const schema = getSchema(); // entire config schema including plugins and drivers
  */
 async function init(args) {
-  const appiumHome = args?.appiumHome ?? (await resolveAppiumHome());
-
-  adjustNodePath();
-
-  const {driverConfig, pluginConfig} = await loadExtensions(appiumHome);
-
   const parser = getParser();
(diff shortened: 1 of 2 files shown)


boneskull (changes requested):
This shouldn't work.  The parser needs to know about extensions beforehand, or it will fail to understand the CLI arguments provided by extensions.  I think we're missing an E2E test to show this.

boneskull:
In order to do what you're proposing, I think we'd need to short-circuit it: run the parser w/o extensions, see if one of the options which exit quickly is present, then handle those; otherwise load extensions and parse again.  Could also just manually parse the args for those relevant options--since parsing twice is wasteful--but I doubt it'd save us more than a few ms.

boneskull:
So basically you just need a driver that declares CLI arguments.  The `fake` driver does, and now that I've merged #17350, you should be able to:

1. rebase this onto `master`
2. run `npm install` just in case
3. run `npm start driver list` and you should see `fake` at the top of the list
4. run `npm start --help` and you should see the two CLI args provided by `fake` driver: `--driver-fake-silly-web-server-port` and `--driver-fake-silly-web-server-host`.

If 3 or 4 are false, then this change has broken it

PR #17578, opened 2022-10-07
chore(appium): Remove logging from adjustNodePath helper

mykola-mokhnach:
## Proposed changes

I can observe these logs appear even after running the binary with `-v`, which is ugly. So it would be better to remove them for good

## Types of changes


KazuCocoa (approved):
makes sense for me

PR #17612, opened 2022-10-14
fix(basedriver): ensure "opts" is defined at time of construction

boneskull:
The override of the `opts` field in `BaseDriver` broke this.


jlipps (approved):
i don't think i understand the bug, doesn't `super` assign to `this.opts` already?

boneskull:
> Note: Before the class fields specification was finalized with the [[Define]] semantic, most transpilers, including [Babel](https://babeljs.io/) and [tsc](https://www.typescriptlang.org/), transformed class fields to the DerivedWithConstructor form, which has caused subtle bugs after class fields were standardized.

PR #17631, opened 2022-10-19
chore(doctor): suggest users to use the full path instead of LOCALAPPDATA on Win

KazuCocoa:
## Proposed changes

When I set `%LOCALAPPDATA%` in the environment variable as ANDROID_HOME on Win, I noticed at least `fs.exists` in nodejs cannot resolve the path properly. The path can be replaced with the value of `process.env.LOCALAPPDATA` in appium's code (I guess like appium-support?), but it would be safe to guide users to use the full path instead of `LOCALAPPDATA` in the doctor, I guess.

error:
image

ok:
image

error message to suggest process.env.LOCALAPPDATA:
image

Potentially... some reports related to ANDROID_HOME in past could be this LOCALAPPDATA thing.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments


packages/doctor/lib/env.js
```diff
@@ -15,9 +15,18 @@ class EnvVarAndPathCheck extends DoctorCheck {
...
-      ? ok(`${this.varName} is set to: ${varValue}`)
-      : nok(`${this.varName} is set to '${varValue}' but this is NOT a valid path!`);
+
+    if (await fs.exists(varValue)) {
+      return ok(`${this.varName} is set to: ${varValue}`);
+    } else {
+      let err_msg = `${this.varName} is set to '${varValue}' but this is NOT a valid path!`;
+      // On Windows, when the env ver has %LOCALAPPDATA$, fs.exists cannot resolve the path.
+      // Then, it would be safe to request the user to set the full path instead.
+      if (varValue.includes('%LOCALAPPDATA%')) {
```
mykola-mokhnach, line 25: is this only the case for this var or for all of them? maybe we should just check for a regular exp like `\b%\w+%\b`

KazuCocoa: Probably https://pureinfotech.com/list-environment-variables-windows-10/ may work as environment variables. (I haven't checked Win 11 since I do not have) I haven't checked all of them with `process.env.XXXX` as well.

`LOCALAPPDATA` was Android's default path to install sdk, so as this pr, I just covered it for now

packages/doctor/lib/env.js
```diff
@@ -15,9 +15,18 @@ class EnvVarAndPathCheck extends DoctorCheck {
...
     }
-    return (await fs.exists(varValue))
-      ? ok(`${this.varName} is set to: ${varValue}`)
-      : nok(`${this.varName} is set to '${varValue}' but this is NOT a valid path!`);
+
+    if (await fs.exists(varValue)) {
+      return ok(`${this.varName} is set to: ${varValue}`);
+    } else {
+      let err_msg = `${this.varName} is set to '${varValue}' but this is NOT a valid path!`;
+      // On Windows, when the env ver has %LOCALAPPDATA$, fs.exists cannot resolve the path.
```
boneskull, line 23: couple typos

```suggestion
      // On Windows, when the env var has %LOCALAPPDATA%, fs.exists cannot resolve the path.
```

but... why not?

boneskull (commented):
I don't really know enough about windows to have any meaningful opinion

KazuCocoa:
i'm not so familiar with Win neither. This was just I found..

packages/doctor/lib/env.js
mykola-mokhnach: maybe then we should add a check for Windows OS

KazuCocoa: Added, https://github.com/appium/appium/pull/17631/commits/cfbd5300d0434a04f7a36bc6f2da5abe2c4e9b1d

packages/doctor/lib/env.js
```diff
@@ -15,9 +15,18 @@ class EnvVarAndPathCheck extends DoctorCheck {
...
-      : nok(`${this.varName} is set to '${varValue}' but this is NOT a valid path!`);
+
+    if (await fs.exists(varValue)) {
+      return ok(`${this.varName} is set to: ${varValue}`);
+    } else {
+      let err_msg = `${this.varName} is set to '${varValue}' but this is NOT a valid path!`;
+      // On Windows, when the env var has %LOCALAPPDATA%, fs.exists cannot resolve the path.
+      // Then, it would be safe to request the user to set the full path instead.
+      if (system.isWindows() && varValue.includes('%LOCALAPPDATA%')) {
+        err_msg += ` Please set "${process.env.LOCALAPPDATA}" instead of "%LOCALAPPDATA%" as the environment variable.`;
```
mykola-mokhnach, line 26: We usually use single quotes

packages/doctor/lib/env.js
```diff
@@ -15,9 +15,18 @@ class EnvVarAndPathCheck extends DoctorCheck {
     if (typeof varValue === 'undefined') {
       return nok(`${this.varName} environment variable is NOT set!`);
     }
-    return (await fs.exists(varValue))
-      ? ok(`${this.varName} is set to: ${varValue}`)
-      : nok(`${this.varName} is set to '${varValue}' but this is NOT a valid path!`);
+
+    if (await fs.exists(varValue)) {
+      return ok(`${this.varName} is set to: ${varValue}`);
+    } else {
```
mykola-mokhnach, line 21: else is redundant after return

mykola-mokhnach (approved):

PR #17640, opened 2022-10-21
chore(base-driver,types): implement better types for custom log formats

boneskull:
Previously, the value of any `Driver#supportedLogTypes` prop would ultimately be unaware of what format the log data is actually stored in (and thus, what the `GET /session/:sessionId/log` cmd responds with).  Now this can be declared by using the `LogDef<C extends Constraints, LogEntry = string>` type.

Example:

```js
  supportedLogTypes = {
    debug: /** @type {import('@appium/types').LogDef<RokuDriverCapConstraints, string>} */ ({
      description: 'Roku debug logs',
      /** @param {RokuDriver} driver */
      async getter(driver) {
        return await ['foo'];
      },
    }),
  };
```

This is a log named `debug` which stores strings.  The `getter` function is always expected to return a type of `LogEntry[]` and receives a type of `Driver<C>`. In this case, it's the default log entry type (`string`) and `RokuDriver`, which implements `Driver<RokuDriverCapConstraints>` where `RokuDriverCapConstraints` extends `Constraints`.

It was kind of tricky since `supportedLogTypes` is an instance field and we don't have access to `this`, so cannot derive the type of the `driver` parameter from it, nor can we use `this` in the `ILogCommands` interface.

Renamed these various mixin interfaces to be prefixed with `I`. All of the mixins themselves (the implementations) are now driver-aware via the `C` type parameter and thus can reference driver-specific caps or options, if needed.

### DIFF
--- a/packages/base-driver/lib/basedriver/commands/index.js
+++ b/packages/base-driver/lib/basedriver/commands/index.js
@@ -1,5 +1,5 @@
 // @ts-check
-
+import _ from 'lodash';
 import {EventMixin} from './event';
 import {FindMixin} from './find';
 import {LogMixin} from './log';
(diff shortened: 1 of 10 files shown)


packages/base-driver/lib/basedriver/commands/index.js
```diff
@@ -9,22 +9,27 @@ import {TimeoutMixin} from './timeout';
...
- * Applies all the mixins to the `BaseDriverBase` class.
- * Returns a `BaseDriver` class.
- * @param {BaseDriverBase} Base
+ * Applies all the mixins to the `BaseDriverBase` class; returns a `BaseDriver` class definition.
+ * Each mixin is applied in the order it is listed here, and each type is a union with the previous.
+ *
+ * @template {Constraints} C
+ * @param {BaseDriverBase<C>} Base
+ */
+export const createBaseDriverClass = _.flow(
```
jlipps, line 18: 😻

packages/types/lib/driver.ts
```diff
@@ -101,20 +102,60 @@ export interface FindCommands {
...
   /**
-   * Gets logs
+   * Definition of the available log types
+   */
+  supportedLogTypes: Readonly<LogDefRecord<C>>;
+
+  /**
+   * Get available log types as a list of strings
+   */
+  getLogTypes(): Promise<(keyof ILogCommands<C>['supportedLogTypes'])[]>;
```
jlipps, line 117: gotta love how we reimplement the function logic in types

jlipps (approved):

packages/base-driver/lib/basedriver/commands/index.js
boneskull: what's really scary is when I get to use lodash/fp

packages/types/lib/driver.ts
boneskull: yeah, there's a limit to its type inference.  Unless we told it otherwise, it'd think the return type was just `string[]`, even if all members are `keyof SomeType`.

PR #17652, opened 2022-10-24
feat(base-driver): add wd-proxy as w3c-only alternative to jsonwp-proxy

jlipps:
Sooner or later we will want to jettison the cruft that comes with maintaining support for talking to JSONWP servers. I decided to reimplement jsonwp-proxy as w3c-only. I slightly changed the return interface for several of the functions but otherwise it's the same idea. Just less code. Figured we could start using it in new situations where we need a new proxy, or slowly convert existing usage of JWProxy to WDProxy.

### DIFF
--- a/packages/base-driver/lib/wd-proxy/proxy.js
+++ b/packages/base-driver/lib/wd-proxy/proxy.js
@@ -1,0 +1,417 @@
+import _ from 'lodash';
+import axios from 'axios';
+import {logger, util} from '@appium/support';
+import {
+  errors,
+  isErrorType,
+  errorFromW3CJsonCode,
+  getResponseForW3CError,
+} from '../protocol/errors';
+import {routeToCommandName} from '../protocol';
+import {MAX_LOG_BODY_LENGTH, DEFAULT_BASE_PATH} from '../constants';
+import http from 'http';
+import https from 'https';
+
+const DEFAULT_LOG = logger.getLogger('WD Proxy');
+const DEFAULT_REQUEST_TIMEOUT = 240000;
+const COMPACT_ERROR_PATTERNS = [/\bECONNREFUSED\b/, /socket hang up/];
+
+const ALLOWED_OPTS = [
+  'scheme',
+  'server',
+  'port',
+  'base',
+  'reqBasePath',
+  'sessionId',
+  'timeout',
+  'log',
+  'keepAlive',
+];
+
+export class WDProxy {
+  /** @type {string} */
+  scheme;
+  /** @type {string} */
+  server;
+  /** @type {number} */
+  port;
+  /** @type {string} */
+  base;
+  /** @type {string} */
+  reqBasePath;
+  /** @type {string?} */
+  sessionId;
+  /** @type {number} */
+  timeout;
+
+  constructor(opts = {}) {
+    opts = _.pick(opts, ALLOWED_OPTS);
+
+    // omit 'log' in the defaults assignment here because 'log' is a getter and we are going to set
+    // it to this._log (which lies behind the getter) further down
+    const options = _.defaults(_.omit(opts, 'log'), {
+      scheme: 'http',
+      server: 'localhost',
+      port: 4444,
+      base: DEFAULT_BASE_PATH,
+      reqBasePath: DEFAULT_BASE_PATH,
+      sessionId: null,
+      timeout: DEFAULT_REQUEST_TIMEOUT,
+    });
+    options.scheme = options.scheme.toLowerCase();
+    Object.assign(this, options);
+
+    this._activeRequests = [];
+    const agentOpts = {
+      keepAlive: opts.keepAlive ?? true,
+      maxSockets: 10,
+      maxFreeSockets: 5,
+    };
+    this.httpAgent = new http.Agent(agentOpts);
+    this.httpsAgent = new https.Agent(agentOpts);
+    this._log = opts.log;
+  }
+
+  get log() {
+    return this._log ?? DEFAULT_LOG;
+  }
+
+  /**
+   * Performs requests to the downstream server
+   *
+   * @private - Do not call this method directly,
+   * it uses client-specific arguments and responses!
+   *
+   * @param {import('axios').AxiosRequestConfig} requestConfig
+   * @returns {Promise<import('axios').AxiosResponse<W3CResponse>>}
+   */
+  async request(requestConfig) {
+    const reqPromise = axios(requestConfig);
+    this._activeRequests.push(reqPromise);
+    try {
+      return await reqPromise;
+    } finally {
+      _.pull(this._activeRequests, reqPromise);
+    }
+  }
+
+  getActiveRequestsCount() {
+    return this._activeRequests.length;
+  }
+
+  cancelActiveRequests() {
+    this._activeRequests = [];
+  }
+
+  /**
+   * @param {string} endpoint
+   */
+  endpointRequiresSessionId(endpoint) {
+    return !_.includes(['/session', '/sessions', '/status'], endpoint);
+  }
+
+  /**
+   * @param {string} url
+   */
+  getUrlForProxy(url) {
+    if (url === '') {
+      url = '/';
+    }
+    const proxyBase = `${this.scheme}://${this.server}:${this.port}${this.base}`;
+    const endpointRe = '(/(session|status))';
+    let remainingUrl = '';
+    if (/^http/.test(url)) {
+      const first = new RegExp(`(https?://.+)${endpointRe}`).exec(url);
+      if (!first) {
+        throw new Error('Got a complete url but could not extract a WebDriver endpoint');
+      }
+      remainingUrl = url.replace(first[1], '');
+    } else if (new RegExp('^/').test(url)) {
+      remainingUrl = url;
+    } else {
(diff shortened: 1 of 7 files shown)


packages/base-driver/lib/wd-proxy/proxy.js
```diff
@@ -0,0 +1,404 @@
...
+  'port',
+  'base',
+  'reqBasePath',
+  'sessionId',
+  'timeout',
+  'log',
+  'keepAlive',
+];
+
+export class WDProxy {
```
mykola-mokhnach, line 31: We could potentially move the common stuff into a base class and then inherit both WDProxy and JWProxy from it to avoid unnecessary code duplication

packages/base-driver/lib/wd-proxy/proxy.js
```diff
@@ -0,0 +1,404 @@
...
+      isResponseLogged = true;
+      const isSessionCreationRequest = /\/session$/.test(url) && method === 'POST';
+      if (isSessionCreationRequest && status === 200) {
+        if (this.sessionId) {
+          this.log.warn(
+            `Warning; sessionId was previously set to ${this.sessionId} but ` +
+              `will be changed as a result of new session response`
+          );
+        }
+        this.sessionId = data.value?.sessionId;
```
mykola-mokhnach, line 246: shall we first assert data.value?.sessionId is a proper string?

packages/base-driver/lib/wd-proxy/proxy.js
```diff
@@ -0,0 +1,404 @@
...
+    throw new errors.UnknownError(
+      `Did not know what to do with response code '${status}' ` +
+        `and response body '${_.truncate(JSON.stringify(data), {
+          length: 300,
+        })}'`
+    );
+  }
+
+  /**
+   * @param {string} url
```
mykola-mokhnach, line 346: string?

packages/base-driver/lib/wd-proxy/proxy.js
```diff
@@ -0,0 +1,404 @@
...
+   */
+  getSessionIdFromUrl(url) {
+    const match = url.match(/\/session\/([^/]+)/);
+    return match ? match[1] : null;
+  }
+
+  /**
+   * @param {any} req
```
mykola-mokhnach, line 354: these types are not any. These are Request and Response types by Node Express server

packages/base-driver/lib/wd-proxy/proxy.js
```diff
@@ -0,0 +1,404 @@
...
+    const commandName = this.requestToCommandName(url, method);
+    if (commandName) {
+      this.log.debug(`Matched '${url}' to command name '${commandName}'`);
+    }
+    return await this.proxy(url, method, body);
+  }
+
+  /**
+   * @param {string} url
+   * @param {import('@appium/types').HTTPMethod} method
```
mykola-mokhnach, line 319: missing return

packages/base-driver/lib/wd-proxy/proxy.js
```diff
@@ -0,0 +1,404 @@
...
+    }
+    if (!commandName && _.includes(url, this.reqBasePath)) {
+      commandName = extractCommandName(new RegExp(`${_.escapeRegExp(this.reqBasePath)}(/.+)`));
+    }
+    return commandName;
+  }
+
+  /**
+   * @param {string} url
+   * @param {import('@appium/types').HTTPMethod} method
```
mykola-mokhnach, line 307: missing return

packages/base-driver/lib/wd-proxy/proxy.js
```diff
@@ -0,0 +1,404 @@
...
+    }
+    remainingUrl = remainingUrl.replace(/\/$/, ''); // can't have trailing slashes
+
+    return proxyBase + remainingUrl;
+  }
+
+  /**
+   * @param {string} url
+   * @param {string} method
+   * @returns {Promise<{status: number, headers: any, data: any}>}
```
mykola-mokhnach, line 177: data and headers are plain objects

mykola-mokhnach: could be imported from axios typedefs

packages/base-driver/test/unit/helpers.js
```diff
@@ -0,0 +1,19 @@
+export function buildReqRes(url, method, body) {
```
boneskull, line 1: this may make more sense in `packages/driver-test-support/lib/helpers.js`

packages/base-driver/test/unit/wd-proxy/mock-request.js
```diff
@@ -0,0 +1,38 @@
+function resFixture(url, method) {
```
boneskull, line 1: fwiw, I think `sinon` could do this, but maybe it's not helpful to do so in this context

packages/base-driver/test/unit/wd-proxy/proxy.spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+describe('proxy', function () {
+  let port;
+
+  function mockProxy(opts = {}) {
+    // sets default server/port
+    opts = {server: TEST_HOST, port, ...opts};
+    let proxy = new WDProxy(opts);
+    proxy.request = async function (...args) {
+      return await request(...args);
```
boneskull, line 15: why not this?

```suggestion
    proxy.request = request;
```

packages/base-driver/test/unit/wd-proxy/proxy.spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+  describe('command proxy', function () {
+    it('should successfully proxy command', async function () {
+      let j = mockProxy();
+      let res = await j.command('/status', 'GET');
+      res.should.eql({foo: 'bar'});
+    });
+    it('should pass along request errors', function () {
+      let j = mockProxy({sessionId: '123'});
+      j.command('/badurl', 'GET').should.eventually.be.rejectedWith('Could not proxy');
```
boneskull, line 127:
```suggestion
      j.command('/badurl', 'GET').should.be.rejectedWith('Could not proxy');
```

😆

boneskull: (`eventually` is _only_ needed for assertions which aren't part of `chai-as-promised`.  so you'd use it if, say, you wanted to assert a number _x_ is greater than _y_.)

boneskull (approved):

packages/base-driver/lib/wd-proxy/proxy.js
jlipps: i'm sort of hoping we'll just delete jwproxy at some point, which would then entail undoing the abstraction

mykola-mokhnach: I would only consider doing that as soon as all internal drivers switch to full w3c support. The current situation is that we have no one, for example, to maintain/update appium remote debugger (iOS Safari stuff). Also, I observe some customers enforcing chromedriver to work in jwp mode.

packages/base-driver/test/unit/helpers.js
jlipps: ok. this made me write more types though :-(

packages/base-driver/test/unit/wd-proxy/mock-request.js
jlipps: meh

packages/base-driver/test/unit/wd-proxy/proxy.spec.js
jlipps: yeah, just copy pasting what was there before, probably written by a slightly dumber me

jlipps:
@mykola-mokhnach if you don't think we should put this in appium, that's ok. i want to use this for another project and i could just put the code there instead if you don't think we really need it for appium.

boneskull:
I have no opinion on whether or not it should be in appium

jlipps:
on another note, i don't imagine it would be _too_ hard to convert WDA to use the w3c spec, right? the actual differences aren't massive, it's mostly in the return wrapper format and the specifics of a few commands (the most complex being new session)

mykola-mokhnach:
> on another note, i don't imagine it would be _too_ hard to convert WDA to use the w3c spec, right? the actual differences aren't massive, it's mostly in the return wrapper format and the specifics of a few commands (the most complex being new session)

WDA has no issues with that, it has been in w3c mode for years. I'm talking about the remote debugger: https://github.com/appium/appium-remote-debugger

jlipps:
ah OK, the remote debugger doesn't relate to jsonwp-proxy though, or does it?

mykola-mokhnach:
> ah OK, the remote debugger doesn't relate to jsonwp-proxy though, or does it?

Maybe not directly. But WinAppDriver, for example does. And we cannot do much about it because the server development is stuck in JWP era thanks to Microsoft, along with its private source code 🤷

jlipps:
I know this is a separate topic, but IMO we could consider dropping support for WinAppDriver. There's only so much we can do with ancient technology that is not maintained. If people want to use it they can always target it directly with a client library, outside of Appium.

jlipps:
closing this for now.

PR #17698, opened 2022-11-01
chore(renovate): blacklist more esm-only packages

boneskull:
yep...


jlipps:
the Great Schism is upon us

PR #17708, opened 2022-11-02
re-enable log filters

boneskull:
Server option `log-filters` is now supported via embedding in Appium config files _or_ via CLI using an external JSON file.  This option became broken during the switchover to config files.  Note that a config file cannot reference an external JSON file.

Example fixtures included.

Note that the `Rule` type has been replaced with the `LogFilter` type automatically generated from the schema (see previous commit).

Additionally, to support this change:

1. `$schema` is now an allowed property in a config file, so that a config file author can refer to the Appium config schema for validation.  We should actually host this schema JSON somewhere and version it properly.
2. As a pre-commit hook, any change to the master Appium config schema `.js` file causes a) a `.json` equivalent to be generated and added to the changeset, and b) the config file TS declaration to get updated and added to the changeset.

### DIFF
--- a/packages/support/lib/log-internal.js
+++ b/packages/support/lib/log-internal.js
@@ -3,6 +3,15 @@ import _ from 'lodash';
 
 const DEFAULT_REPLACER = '**SECURE**';
 
+/**
+ * Type guard for log filter type
+ * @param {object} value
+ * @returns {value is import('@appium/types').LogFilterRegex}
+ */
+function isLogFilterRegex(value) {
+  return 'pattern' in value;
+}
+
 class SecureValuesPreprocessor {
   constructor() {
     this._rules = [];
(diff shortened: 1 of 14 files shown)


packages/types/lib/appium-config.ts
```diff
@@ -49,10 +49,23 @@ export type LocalTimezoneConfig = boolean;
  * Also send log output to this file
  */
 export type LogConfig = string;
+/**
+ * Log filtering rule
+ */
+export type LogFilter = {
```
boneskull, line 55: I just gotta say that [json-schema-to-typescript](https://npm.im/json-schema-to-typescript) is pretty awesome.  thanks @bcherny

jlipps (approved):

packages/support/lib/log-internal.js
```diff
@@ -3,6 +3,15 @@ import _ from 'lodash';
 
 const DEFAULT_REPLACER = '**SECURE**';
 
+/**
+ * Type guard for log filter type
+ * @param {any} value
+ * @returns {value is import('@appium/types').LogFilterRegex}
+ */
+function isLogFilterRegex(value) {
+  return 'pattern' in value;
```
mykola-mokhnach, line 12: is `in` call safe for non-object types?

mykola-mokhnach (approved):

mykola-mokhnach:
Tests seem to be not very happy though

packages/support/lib/log-internal.js
boneskull: no; I'll update the type

PR #17709, opened 2022-11-02
docs(appium): add capabilities doc

jlipps:
@KazuCocoa if you can help add some of the examples again that'd be great! maybe @saikrishna321 can do the recommended Java example?

cc @appium/documentors

### DIFF
--- a/packages/appium/docs/en/guides/caps.md
+++ b/packages/appium/docs/en/guides/caps.md
@@ -2,9 +2,105 @@
 title: Capabilities
 ---
 
-TODO
+"Capabilities" is the name given to the set of parameters used to start an Appium session. The
+information in the set is used to describe what sort of "capabilities" you want your session to
+have, for example, a certain mobile operating system or a certain version of a device. When you
+start your Appium session, your Appium client will include the set of capabilities you've defined
+as an object in the JSON-formatted body of the request. Capabilities are represented as key-value
+pairs, with values allowed to be any valid JSON type, including other objects. Appium will then
+examine the capabilities and make sure that it can satisfy them before proceeding to start the
+session and return an ID representing the session to your client library.
 
-## Special Notes for Cloud Providers
+The W3C WebDriver spec's [section on Capabilities](https://w3c.github.io/webdriver/#capabilities)
+identifies a small set of 10 standard capabilities, including the following:
+
+| Capability Name  | Type     | Description                                    |
+|------------------|----------|------------------------------------------------|
+| `browserName`    | `string` | The name of the browser to launch and automate |
+| `browserVersion` | `string` | The specific version of the browser            |
+| `platformName`   | `string` | The type of platform hosting the browser       |
+
+## Appium capabilities
+
+Appium understands these browser-focused capabilities, but introduces a number of additional
+capabilities. According to the WebDriver spec, any
+non-standard "extension capabilities" must include a namespace prefix (signifying the vendor
+introducing the capability), ending in a `:`. Appium's vendor prefix is
+`appium:`, and so any Appium-specific capabilities must include this prefix. Depending on which
+client you are using, the prefix may be added automatically or in conjunction with certain
+interfaces, but it is always a good practice to explicitly include it for clarity.
+
+Here is a list of all the globally-recognized Appium capabilities. Notes:
+
+- Individual drivers and plugins can support other capabilities, so refer to their documentation for lists of extension-specific
+capability names.
+- Some drivers might not implement support for all of these capabilities
+- A few Appium extension capabilities are required to start any Appium session; these are noted below as "Required"
+
+| Capability&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Type      | Required? | Description                                                                                                                                                                                                                                       |
+|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------|-----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `platformName`                                                                                                                                                                     | `string`  | Required  | The type of platform hosting the app or browser                                                                                                                                                                                                   |
+| `appium:automationName`                                                                                                                                                            | `string`  | Required  | The name of the Appium driver to use                                                                                                                                                                                                              |
+| `browserName`                                                                                                                                                                      | `string`  |           | The name of the browser to launch and automate, if the driver supports web browsers as a special case                                                                                                                                             |
+| `appium:app`                                                                                                                                                                       | `string`  |           | The path to an installable application                                                                                                                                                                                                            |
+| `appium:deviceName`                                                                                                                                                                | `string`  |           | The name of a particular device to automate, e.g., `iPhone 14` (currently only actually useful for specifying iOS simulators, since in other situations it's typically recommended to use a specific device id via the `appium:udid` capability). |
+| `appium:platformVersion`                                                                                                                                                           | `string`  |           | The version of a platform, e.g., for iOS, `16.0`                                                                                                                                                                                                  |
+| `appium:newCommandTimeout`                                                                                                                                                         | `number`  |           | The number of seconds the Appium server should wait for clients to send commands before deciding that the client has gone away and the session should shut down                                                                                   |
+| `appium:noReset`                                                                                                                                                                   | `boolean` |           | If true, instruct an Appium driver to avoid its usual reset logic during session start and cleanup (default `false`)                                                                                                                              |
+| `appium:fullReset`                                                                                                                                                                 | `boolean` |           | If true, instruct an Appium driver to augment its usual reset logic with additional steps to ensure maximum environmental reproducibility (default `false`)                                                                                       |
+| `appium:eventTimings`                                                                                                                                                              | `boolean` |           | If true, instruct an Appium driver to collect [Event Timings](./event-timing.md) (default `false`)                                                                                                                                                |
+| `appium:printPageSourceOnFindFailure`                                                                                                                                              | `boolean` |           | If true, collect the page source and print it to the Appium log whenever a request to find an element fails (default `false`)                                                                                                                     |
+
+
+Some drivers place more complex constraints on capabilities as a group. For example, while the
+`appium:app` and `browserName` capabilities are listed above as optional, if you want to launch
+a session with a specific app, the XCUITest driver requires that at least one of `appium:app`,
+`browserName`, or `appium:bundleId` are included in the capabilities (otherwise it will not know
+what app to install and/or launch and will simply open a session on the home screen). Each driver
+will document how it interprets these capabilities and any other platform-specific requirements.
+
+!!! note
+
+    Capabilities are like parameters used when starting a session. After the capabilities are sent
+    and the session is started, they cannot be changed. If a driver supports updating aspects of
+    its behaviour in the course of a session, it will provide a [Setting](./settings.md) for this
+    purpose instead of, or in addition to, a capability.
+
+Each Appium client has its own way of constructing capabilities and starting a session. For
+examples of doing this in each client library, head to the [Ecosystem](../ecosystem/index.md) page
+and click through to the appropriate client documentation.
+
+## Always-match and first-match capabilities
+
+The W3C spec allows clients to give the Appium server some flexibility in the kind of session it
+creates in response to a new session request. This is through the concept of "always-match" and
+"first-match" capabilities:
+
+- Always-match capabilities consist of a single set of capabilities, every member of which must be satisfied by the server in order for the new session request to proceed.
+- First-match capabilities consist of an array of capability sets. Each set is merged with the always-match capabilities, and the first set that the server knows how to handle will be the set that is used to start the session.
+
+!!! note
+
+    Check out the [spec itself](https://w3c.github.io/webdriver/#processing-capabilities) or
+    a [summarized version](https://github.com/jlipps/simple-wd-spec#processing-capabilities) for
+    a more in-depth description of how capabilities are processed.
+
+In practice, use of first-match capabilities is not necessary or recommended for use with Appium.
+Instead, we recommend that you define the explicit set of capabilities you want the Appium
+server to handle. These will be encoded as the always-match capabilities, and the array of
+first-match capabilities will be empty.
+
+That being said, Appium _does_ understand always-match and first-match capabilities as
+defined in the W3C spec, so if you use these features, Appium will work as expected. The process of
+defining always-match and first-match capabilities is unique to each client library, so refer to
+the documentation for your client library to see examples of how it works.
+
+## Special notes for cloud providers
+
+!!! warning
+
+    This section is not intended for end-users of Appium; it is intended for developers building
+    Appium-compatible cloud services.
 
 When managing an Appium cloud, your users may wish to target various independent versions of Appium
 drivers and plugins. It is of course up to each service provider how they wish to implement the


packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
 
-TODO
+"Capabilities" is the name given to the set of parameters used to start an Appium session. The
+information in the set is used to describe what sort of "capabilities" you want your session to
+have, for example, a certain mobile operating system or a certain version of a device. When you
+start your Appium session, your Appium client will include the set of capabilities you've defined,
+as an object in the JSON-formatted body of the request. Capabilities are represented as key-value
+pairs, with values allowed to be any valid JSON type, including other objects. Appium will then
+examine the capabilities and make sure that it can satisfy them before proceeding to start the
+session and return an ID representing the session to your client library.
```
mykola-mokhnach, line 12: It is also important to mention capabilities are immutable, e.g. once defined on session startup their values persist until the session ends

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
 title: Capabilities
 ---
 
-TODO
+"Capabilities" is the name given to the set of parameters used to start an Appium session. The
+information in the set is used to describe what sort of "capabilities" you want your session to
+have, for example, a certain mobile operating system or a certain version of a device. When you
+start your Appium session, your Appium client will include the set of capabilities you've defined,
```
boneskull, line 8:
```suggestion
start your Appium session, your Appium client will include the set of capabilities you've defined
```

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+| Capability Name  | Type     | Description                                    |
+|------------------|----------|------------------------------------------------|
+| `browserName`    | `string` | The name of the browser to launch and automate |
+| `browserVersion` | `string` | The specific version of the browser            |
+| `platformName`   | `string` | The type of platform hosting the browser       |
+
+## Appium capabilities
+
+Appium understands these browser-focused capabilities, but introduces a number of additional
+capabilities you are free to use to start Appium sessions. According to the WebDriver spec, any
```
boneskull, line 26:
```suggestion
capabilities. According to the WebDriver spec, any
```

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+non-standard "extension capabilities" must include a namespace prefix (signifying the vendor
+introducing the capability), separated from the capability name by a `:`. Appium's vendor prefix is
```
boneskull, line 28:
```suggestion
introducing the capability), ending in a `:`. Appium's vendor prefix is
```

...or just change it so we're clear the prefix is `appium`...

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+`appium:`, and so any Appium-specific capabilities must include this prefix. Depending on which
+client you are using, the prefix may be added automatically, or you might need to ensure they are
+included in capability names yourself (prefixes are added, for example, in the Appium Python
```
mykola-mokhnach, line 31: actually, this depends on how these capabilities are passed. If the Options concept is used then the above is true and the prefix is added automatically unless some other prefix is already specific. Hoverer if one passes capabilities directly (as dictionary) then no magic happens

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+client, but not the case in WebdriverIO).
```
boneskull, line 32:
```suggestion
included in capability names yourself (for example, the Appium Python client adds them, but WebdriverIO does not).
```

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+
+Here is a list of all the globally-recognized Appium capabilities. Notes:
+
+- Individual drivers and plugins can support other capabilities, so refer to their documentation for lists of extension-specific
+capability names.
+- Some drivers might not implement support for all of these capabilities
+- A few Appium extension capabilities are required to start any Appium session; these are noted below as "Required"
+
+| Capability&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;                            | Type      | Required? | Description                                                                                                                                                     |
```
boneskull, line 41: weird

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+|---------------------------------------|-----------|-----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `platformName`                        | `string`  | Required  | The type of platform hosting the app or browser                                                                                                                 |
+| `appium:automationName`               | `string`  | Required  | The name of the Appium driver to use                                                                                                                            |
+| `browserName`                         | `string`  | Optional  | The name of the browser to launch and automate, if the driver supports web browsers as a special case                                                           |
```
boneskull, line 45: might be easier to read if we omit 'optional' on all these

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+| `appium:app`                          | `string`  | Optional  | The path to an installable application                                                                                                                          |
+| `appium:deviceName`                   | `string`  | Optional  | The name of a particular device to automate, e.g., `iPhone 14`                                                                                                  |
```
mykola-mokhnach, line 47: this might be confusing. For example UIA2 and XCUITest drivers ignore this capability while selecting a real device and respect the `udid` one.

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+document how it interprets these capabilities and any other platform-specific requirements.
+
+## Constructing capabilities using clients
+
+Each Appium client has its own way of constructing capabilities and starting a session. Here is
+a minimal example of starting an Appium Android session using different client libraries.
+
+=== "JS (WebDriverIO)"
+
+    ```js
```
boneskull, line 69: GFM doesn't render these properly, as four spaces means a code block.  De-indent these examples?

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+
+    ```dotnet
+    /* TODO */
+    ```
+
+## Always-match and first-match capabilities
+
+The W3C spec allows clients to give the Appium server some flexibility in the kind of session it
+creates in response to a new session request. This is through the concept of "always-match" and
+"first-match" capabilities:
```
boneskull, line 128: might be good to link to the spec or some easy explainer here, idk

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+
+- Always-match capabilities consist of a single set of capabilities, every member of which must be satisfied by the server in order for the new session request to proceed.
+- First-match capabilities consist of an array of capability sets. Each set is merged with the always-match capabilities, and the first set that the server knows how to handle will be the set that is used to start the session.
+
+In practice, knowledge or use of always-match or first-match capabilities are not necessary or
```
boneskull, line 133: don't understand what "knowledge" means here

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+| `appium:fullReset`                    | `boolean` | Optional  | If true, instruct an Appium driver to augment its usual reset logic with additional steps to ensure maximum environmental reproducibility (default `false`)     |
+| `appium:eventTimings`                 | `boolean` | Optional  | If true, instruct an Appium driver to collect [Event Timings](./event-timing.md) (default `false`)                                                              |
+| `appium:printPageSourceOnFindFailure` | `boolean` | Optional  | If true, collect the page source and print it to the Appium log whenever a request to find an element fails (default `false`)                                   |
+
+
+Some drivers place more complex constraints on capabilities as a group. For example, while the
+`appium:app` and `browserName` capabilities are listed above as optional, the XCUITest driver
+requires that at least one of `appium:app`, `browserName`, or `appium:bundleId` are included in the
```
mykola-mokhnach, line 58: this is not quite true. It is possible to start a no-app session on device home screen if none of these caps is provided

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+
+The W3C spec allows clients to give the Appium server some flexibility in the kind of session it
+creates in response to a new session request. This is through the concept of "always-match" and
+"first-match" capabilities:
+
+- Always-match capabilities consist of a single set of capabilities, every member of which must be satisfied by the server in order for the new session request to proceed.
+- First-match capabilities consist of an array of capability sets. Each set is merged with the always-match capabilities, and the first set that the server knows how to handle will be the set that is used to start the session.
+
+In practice, knowledge or use of always-match or first-match capabilities are not necessary or
+recommended for use with Appium. Instead, we recommend that you simply define the exact set of
```
boneskull, line 134:
```suggestion
recommended for use with Appium. Instead, we recommend that you define the explicit set of
```

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+capabilities you want the Appium server to handle. These will be encoded as the always-match
+capabilities, and the array of first-match capabilities will be empty.
+
+That being said, the Appium server does process always-match and first-match capabilities as
```
boneskull, line 138:
```suggestion
That being said, Appium _does_ understand always-match and first-match capabilities as
```

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+
+=== "JS (WebDriverIO)"
+
+    ```js
+    const {remote} = require('webdriverio')
+
+    const capabilities = {
+      platformName: 'Android',
+      'appium:automationName': 'UiAutomator2',
+      'appium:deviceName': 'Android',
```
mykola-mokhnach, line 75: this is not respected. UIA2 uses udid to select real devices

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+capabilities, and the array of first-match capabilities will be empty.
+
+That being said, the Appium server does process always-match and first-match capabilities as
+defined in the W3C spec, so if you use these features, Appium will work as expected. The process of
+defining always-match and first-match capabilities is unique to each client library, so refer to
+the documentation for your client library to see examples of how it works.
+
+## Special notes for cloud providers
+
+!!! warning
```
boneskull, line 145: what is this markdown

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+
+    This section is not intended for Appium end users. It is intended for developers building
```
boneskull, line 147:
```suggestion
    This section is not intended for end-users of Appium; it is intended for developers building
```

boneskull (approved):

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+=== "JS (WebDriverIO)"
+
+    ```js
+    const {remote} = require('webdriverio')
+
+    const capabilities = {
+      platformName: 'Android',
+      'appium:automationName': 'UiAutomator2',
+      'appium:deviceName': 'Android',
+      'appium:appPackage': 'com.android.settings',
```
mykola-mokhnach, line 76: consider also setting noReset to true otherwise the driver will try to reset the app and most likely fail

packages/appium/docs/en/guides/caps.md
boneskull: I wonder if it would be best to just recommend people use the spec.

packages/appium/docs/en/guides/caps.md
boneskull: curious; why is that?

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+    capabilities = {
+      'platformName': 'Android',
+      'automationName': 'UiAutomator2',
+      'deviceName': 'Android',
+      'appPackage': 'com.android.settings',
+      'appActivity': '.Settings',
+    }
+
+    driver = webdriver.Remote(command_executor='http://localhost:4723',
+                              desired_capabilities=capabilities)
```
mykola-mokhnach, line 109: desired_capabilities are deprecated. Use Options instead: https://github.com/appium/python-client#usage

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+      capabilities,
+    }
+
+    const driver = await remote(wdOpts)
+    ```
+
+=== "Java"
+
+    ```java
+    /* TODO */
```
mykola-mokhnach, line 92: https://github.com/appium/java-client/blob/master/src/test/java/io/appium/java_client/drivers/options/OptionsBuildingTest.java

packages/appium/docs/en/guides/caps.md
mykola-mokhnach: 🤷 
It was always working like that. Most likely, some historical reasons...
The only situation where deviceName is actually used is iOS simulator selection

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+
+Some drivers place more complex constraints on capabilities as a group. For example, while the
+`appium:app` and `browserName` capabilities are listed above as optional, the XCUITest driver
+requires that at least one of `appium:app`, `browserName`, or `appium:bundleId` are included in the
+capabilities (otherwise it will not know what app to install and/or launch). Each driver will
+document how it interprets these capabilities and any other platform-specific requirements.
+
+## Constructing capabilities using clients
+
+Each Appium client has its own way of constructing capabilities and starting a session. Here is
```
mykola-mokhnach, line 64: Perhaps it would make sense to just provide links to readme for each client

KazuCocoa: Like the idea of putting links. Each client also has links for their documentation, and some are for their functional test links as live examples. They will help to find concrete method usage examples

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,9 +2,150 @@
...
+=== "Ruby"
+
+    ```rb
+    # TODO
+    ```
+
+=== "C#"
+
+    ```dotnet
+    /* TODO */
```
Dor-bl, line 121: I'll be happy to help with C# example

packages/appium/docs/en/guides/caps.md
jlipps: good point, will discuss that and link to the Settings doc

packages/appium/docs/en/guides/caps.md
jlipps: yeah, will just recommend people use it for clarity

packages/appium/docs/en/guides/caps.md
jlipps: yeah, support for `deviceName` could theoretically be broadened e.g. for android, we could check adb for connected devices, and run commands on them to figure out which device is a "Samsung S4" or whatever, and then use that device.

but the driver isn't currently set up to do that. `deviceName` is also useful for cloud providers, when all you care about is getting a certain kind of device, not targeting a specific device by id.

packages/appium/docs/en/guides/caps.md
jlipps: if we're starting a session on an emulator, though, is deviceName not required?

packages/appium/docs/en/guides/caps.md
jlipps: i added the spaces so that the capability names would not all be wrapped in weird places.

packages/appium/docs/en/guides/caps.md
jlipps: they need to be indented for our docs renderer. and they render fine there.

packages/appium/docs/en/guides/caps.md
jlipps: yeah ok, maybe better to just link to the clients so things don't get out of date.

packages/appium/docs/en/guides/caps.md
jlipps: it's not designed for rendering on github. it's part of mkdocs-material

jlipps:
ok all comments/suggestions addressed

jlipps:
i've rebuilt the docs so this PR and the previous one are now live on the site

PR #17718, opened 2022-11-04
chore(ci): update 1.x CI to avoid failures/warnings from GHA

boneskull:
I merged a typo fix to the 1.x `README.md` and found that the build failed because GHA dropped support for Node.js v12.


boneskull:
if CI keeps failing here I'm just gonna forget about it

boneskull:
yeah, would need to update the tests to use `get-port` or something.  forget it

PR #17727, opened 2022-11-04
docs(appium): add settings api doc

jlipps:


KazuCocoa (approved):

packages/appium/docs/en/guides/settings.md
```diff
@@ -2,4 +2,45 @@
...
 
-TODO
+Appium has introduced a set of extension APIs that allow you to adjust parameters for a given
+session during session execution. Called "Settings", these parameters mirror the role of
+[Capabilities](./caps.md), but Capabilities cannot be adjusted once a session has started. Settings
+can be adjusted any number of times during the course of a session.
+
+There are 3 important points to the concept of Settings:
+ - Settings are mutable; they can be changed during a session using the Settings API
+ - Settings are only relevant during the session in which they are set. They are reset for each new session.
```
mykola-mokhnach, line 12: hm, I assume this is not completely true, at least for the current UIA2 and XCUITest implementations, where settings are static

KazuCocoa: aa.., true

jlipps: really, settings aren't stored on the driver instance? that's how it's implemented in basedriver...?

mykola-mokhnach: server part implements it differently
eventually it is probably ok for uia2 because the server gets terminated after each session
although WDA keeps running between sessions

would you prefer to change the current behaviour and reset settings upon new session startup?

jlipps: oh i see, you're talking about the on device servers. hmm. yeah. i would expect settings to be scoped to sessions since after all their api is `/session/:id/setting`, not `/appium/setting` or whatever. We don't really have any concept in appium of 'singleton for driver type', and i don't think we want one.

PR #17745, opened 2022-11-08
docs(appium): add log-filters guide

jlipps:
brought this over with little update from 1.x. we should make sure it's telling the truth about the way log filters work in appium 2 though!

### DIFF
--- a/packages/appium/docs/en/guides/log-filters.md
+++ b/packages/appium/docs/en/guides/log-filters.md
@@ -2,4 +2,82 @@
 title: Log Filtering
 ---
 
-TODO
+Sometimes it might be necessary to hide sensitive information, like passwords, device identifiers,
+hashes, etc..., from the server log. Appium makes it possible to ensure such information is
+redacted in logs via the `--log-filters` command line argument. This argument allows you to provide
+the path to a special config file, containing one or more log obfuscation rules.
+
+## Config Format
+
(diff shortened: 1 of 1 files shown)


mykola-mokhnach (approved):

packages/appium/docs/en/guides/log-filters.md
```diff
@@ -2,4 +2,77 @@
 title: Log Filtering
 ---
 
-TODO
+Sometimes it might be necessary to hide sensitive information, like passwords, device identifiers,
+hashes, etc..., from the server log. Appium makes it possible to ensure such information is
+redacted in logs via the `--log-filters` command line argument. This argument allows you to provide
+the path to a special config file, containing one or more log obfuscation rules.
+
+## Config Format
```
boneskull, line 10: should probably add a note/link that this can be configured in an appium config file (without an external JSON file)

boneskull (approved):

packages/appium/docs/en/guides/log-filters.md
jlipps: ah good idea

jlipps: @boneskull does it look correct now?

KazuCocoa (approved):

PR #17747, opened 2022-11-09
persist installation path of extensions

### LINKED ISSUE #17728: auto-detection of drivers does not always work properly when creating a session
When developing a driver, we can see that running `appium driver list` shows the driver as installed.  However, when it comes time to _resolve_ the driver and instantiate it, failure may happen due to the return value of `ExtensionCommand#getInstallPath()` which makes Assumptions about the location of the installed package.

boneskull:
This uses the new v3 version of the manifest schema to add an `installPath` property to every extension in `extensions.yaml`. This fixes an issue where extensions may not be able to be _instantiated_ within an extension dev environment. Previously, Appium would _assume_ the path of an extension is in `node_modules` (relative to `APPIUM_HOME`), but that is not always true. In that respect it's a bit of a hotfix to some changes published in the last `appium` release.

In addition, I've added a small "framework" to migrate schema versions, including their types. Please see `packages/types/lib/manifest/README.md` for more information on how to add a new version of the schema.


packages/appium/lib/extension/manifest-migrations.js
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * If `installPath` is missing from the `ExtManifest` for any extension, return `true`.
+   *
+   * We cannot easily determine the path of any given extension here, but returning `true`
+   * will cause the `Manifest` to sync with the filesystem, which will populate the missing field.
+   *
+   * @type {Migration}
+   */
+  3: (data) => {
+    let shouldSync = false;
```
mykola-mokhnach, line 23: do we need this intermediate variable at all?

packages/appium/lib/extension/manifest.js
```diff
@@ -419,34 +425,35 @@ export class Manifest {
...
-        await fs.writeFile(this._manifestPath, YAML.stringify(this._data), 'utf8');
+        await fs.writeFile(this.#manifestPath, YAML.stringify(this.#data), 'utf8');
         return true;
       } catch (err) {
         throw new Error(
-          `Appium could not write to manifest at ${this._manifestPath} using APPIUM_HOME ${this._appiumHome}. ` +
-            `Please ensure it is writable. Original error: ${err.message}`
+          `Appium could not write to manifest at ${this.#manifestPath} using APPIUM_HOME ${
+            this.#appiumHome
+          }. ` + `Please ensure it is writable. Original error: ${err.message}`
```
mykola-mokhnach, line 449: extra catenation here

packages/appium/test/fixtures/manifest/v3.yaml
```diff
@@ -0,0 +1,27 @@
...
+plugins:
+  fake:
+    mainClass: FakePlugin
+    scripts:
+      fake-error: ./build/lib/scripts/fake-error.js
+      fake-success: ./build/lib/scripts/fake-success.js
+    pkgName: '@appium/fake-plugin'
+    version: 1.2.2
+    installType: local
+    installSpec: /Users/alice/projects/appium/node_modules/@appium/fake-plugin
```
mykola-mokhnach, line 25: I thought installSpec should only have ext name...

mykola-mokhnach (approved):

packages/appium/lib/extension/manifest-migrations.js
boneskull: not technically, no, but it's just for clarity of intent

packages/appium/lib/extension/manifest-migrations.js
```diff
@@ -0,0 +1,94 @@
...
+const Migrations = {
+  /**
+   * If `installPath` is missing from the `ExtManifest` for any extension, return `true`.
+   *
+   * We cannot easily determine the path of any given extension here, but returning `true`
+   * will cause the `Manifest` to sync with the filesystem, which will populate the missing field.
+   *
+   * @type {Migration}
+   */
+  3: (data) => {
```
mykola-mokhnach, line 22: shall we use the constant instead?

KazuCocoa (approved):
lgtm

packages/appium/types/manifest/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Originally, the manifest file was versioned; it had a `schemaRev` prop.  Eventually, the requirement for this prop was dropped, and became optional.  As of this writing, it is now required again.  
+
+The idea was that if Appium found an `extensions.yaml` with a `schemaRev` lower than its "current" revision, it would "migrate" the file to the current revision.  When we removed the requirement for the `schemaRev` prop, we removed this logic.
+
+As of _v3_ of the schema, we are reinstating versioning and migrations.
+
+## How To Add A New Version
+
+The manifest file format composed of types used elsewhere in our codebase.  This is probably a design flaw, but the fact remains that the manifest file's format is strongly coupled to Appium's internal logic around extensions.
```
jlipps, line 15: `is composed of`

jlipps (approved):

packages/appium/lib/extension/manifest-migrations.js
boneskull: There is no constant for this...

There's `CURRENT_SCHEMA_REV = 3`, but that is supposed to be updated when we increment the revision, so we cannot use it.

However: It's probably a good idea to define one anyway, so I'll do that.

boneskull: OK, this is done and I added a note in the `README` about it

Issue #17753, opened 2022-11-09
bug: Find Element by Image fails randomly 


jorgoncalves:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

**We have a suite of tests for Android that use the find element by image, on static images, which can all be found, but sometimes it fails randomly with the bellow error.**

NoSuchElementException: Message: An element could not be located on the page using the given search parameters.
Stacktrace:
NoSuchElementError: An element could not be located on the page using the given search parameters.
    at ImageElementFinder.findByImage (/Users/username/.appium/node_modules/@appium/images-plugin/lib/finder.js:230:13)
    at ImageElementPlugin._find (/Users/username/.appium/node_modules/@appium/images-plugin/lib/plugin.js:61:12)
    at ImageElementPlugin.findElement (/Users/username/.appium/node_modules/@appium/images-plugin/lib/plugin.js:45:12)
    at /usr/local/lib/node_modules/appium/lib/appium.js:722:18
    at AppiumDriver.executeWrappedCommand (/usr/local/lib/node_modules/appium/lib/appium.js:762:16)
    at AppiumDriver.executeCommand (/usr/local/lib/node_modules/appium/lib/appium.js:684:17)
    at asyncHandler (/usr/local/lib/node_modules/appium/node_modules/@appium/base-driver/lib/protocol/protocol.js:346:19)

-----------------------

**After getting the above error, subsequent runs of the same suite will get the bellow error.**

WebDriverException: Message: An unknown server-side error occurred while processing the command. Original error: Cannot read properties of undefined (reading 'match')
Stacktrace:
UnknownError: An unknown server-side error occurred while processing the command. Original error: Cannot read properties of undefined (reading 'match')
    at getResponseForW3CError (/usr/local/lib/node_modules/appium/node_modules/@appium/base-driver/lib/protocol/errors.js:956:9)
    at asyncHandler (/usr/local/lib/node_modules/appium/node_modules/@appium/base-driver/lib/protocol/protocol.js:435:35)

**This second error will always appear until the server is restarted and the images will be found again until the first error happens again.**

### Expected Behavior

The images are found or not consistently and the server is able to handle the exception without the needing a restart.

### Minimal Reproducible Example

We are providing an png image into the bellow function, which will then be converted to base64.

    def _find_by_image(self, img_path: str):
           driver = self._get_driver_instance()
    
            with open(img_path, 'rb') as i_file:
[...]

PR #17756, opened 2022-11-09
fix(images-plugin): Handle the case when error message is undefined

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/appium/appium/issues/17753

## Types of changes

### DIFF
--- a/packages/images-plugin/lib/finder.js
+++ b/packages/images-plugin/lib/finder.js
@@ -71,6 +71,9 @@ const DEFAULT_SETTINGS = {
   // It is used for debugging purpose.
   getMatchedImageResult: false,
 };
+const NO_OCCURRENCES_PATTERN = /Cannot find any occurrences/;
+const CONDITION_UNMET_PATTERN = /Condition unmet/;
+
 
 export default class ImageElementFinder {
   /** @type {ExternalDriver} */
@@ -202,7 +205,7 @@ export default class ImageElementFinder {
         // retry, so trap that and just return false to trigger the next round of
         // implicitly waiting. For other errors, throw them to get out of the
         // implicit wait loop
-        if (err.message.match(/Cannot find any occurrences/)) {
+        if (NO_OCCURRENCES_PATTERN.test(err.message)) {
           return false;
         }
         throw err;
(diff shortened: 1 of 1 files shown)


KazuCocoa (approved):

boneskull (commented):
Do _we_ throw these exceptions, or are they coming out of third-party code? 

If we throw them, we should probably just subclass `Error` and do an `instanceof` check instead of a regex on the `message`...

mykola-mokhnach:
> Do _we_ throw these exceptions, or are they coming out of third-party code?
> 
> If we throw them, we should probably just subclass `Error` and do an `instanceof` check instead of a regex on the `message`...

These exceptions are thrown by other package (opencv). I don't know how to properly orchestrate versions bump there in order to implement such strategy. Also I remember there were issues while using instanceof for error classes of these are being imported from other packages

boneskull:
@mykola-mokhnach 

> Also I remember there were issues while using instanceof for error classes of these are being imported from other packages

What are you referring to here?

I mean, the type or shape of a given `Error` should be considered part of its public API, while contents the error message itself is not necessarily.  _In general_--and considering the policy/history of any given dependency--to distinguish exceptions, I'd rather leverage basically _any_ other bit of information.  

In other words, checking a `message` to determine what kind of `Error` you have may be more risky (in terms of unexpected breakage) than other strategies.  Again, it kind of depends on the dependency and how it's maintained.  

I'm not asking you to change anything necessarily, but--assuming we have not already done so--we should do our due diligence on the project (`opencv`?) & its API and see if there's anything else we should be using to distinguish these errors.

mykola-mokhnach:
The opencv package is in the same monorepo. 

What I'm referring to was also the reason of having isErrorType function in https://github.com/appium/appium/blob/c43acc23ee80d461747c1912cdcfb136280001a0/packages/base-driver/lib/protocol/errors.js
@jlipps probably has more background there as this code already existed after I joined.

mykola-mokhnach:
Generally I agree that checking the error type makes more sense over checking its message

boneskull:
Better than `instanceof` would be a `code` prop on the error which is some constant string we can just prepare, like node does

jlipps:
The problem is that `instanceof` checks don't work when the error class imported by the checker is from a different version of the npm package than the error class imported by the thrower.

If we're sure that the error class is imported in the same place by checker and thrower, then the `instanceof` check will work. This was not a guarantee in earlier versions of appium. As you can imagine, this was a weird bug to track down when it first popped up.

jlipps (approved):

jlipps:
> Better than instanceof would be a code prop on the error which is some constant string we can just compare, like node does

That could be a good way to reimplement `isErrorType`

PR #17757, opened 2022-11-09
chore(docutils): remove jsdoc & its ilk

boneskull:
`jsdoc` does not seem to be actually used anywhere.  The only export from `@appium/docutils` I see actually used is `Mike` within the `build-docs` script in `appium/docs`.

referenced by the `jsdoc` config, these have also been removed:
- `docdash`,
- `jsdoc-plugin-typescript`

See #17720


mykola-mokhnach (approved):

packages/docutils/package.json
```diff
@@ -48,9 +48,6 @@
   },
   "dependencies": {
     "@appium/support": "^2.61.1",
```
mykola-mokhnach, line 50: shouldn't this package be imported by relative path?

boneskull: That used to be the case, but since adoption of npm workspaces, we don't need to do this (and in fact it screws some things up--I see a couple places that need fixing).

mykola-mokhnach: so how we could do now changes in multiple modules and reference them properly within a single PR?

For example: I change package A, which needs some updates to package B. How could I reference the updated package B in package A in scope of a single PR then?

jlipps (approved):

packages/docutils/package.json
boneskull: If you change package A and package B, and A depends on B, you don't worry about the version number because npm creates a symlink and package A _always_ gets whatever is in package B.  In other words, you _cannot_ update package B without it affecting package A.  In general, this is a good thing.

PR #17777, opened 2022-11-15
prep for adding command documentation

boneskull:
This PR contains a bunch of changesets--mostly involving types--in preparation for creating a TypeDoc plugin which automatically documents driver/plugin/builtin commands (API endpoints).

The implementation of `typedoc-plugin-appium` will be linked below.


boneskull:
I can probably try harder to sort out the changesets.  Some of them contain reverts of previous ones...

boneskull:
This started as a fork of the `jlipps/command-docs` branch

boneskull:
alright, I broke something here, so gotta fix that.  this is why we write tests!

boneskull:
OK, this happened because of the fake-driver-mixin thing.  It looks like a more robust method of deleting sessions is needed in `AppiumDriver`

boneskull:
The problem was that `caps` was `undefined` when the `driverData` getter was called by `deleteSession` (and _why_ this is being accessed here is still anybody's guess).  I've modified the `BaseDriverCore` constructor to set `this.caps = {}`.  Unclear why TS wasn't raising a stink about this property leaving the constructor w/o getting defined; probably has something to do with the mixins obscuring the prototype chain...

jlipps (approved):
not sure why CI is unhappy but all this type groundwork seems ok

boneskull:
yeah trying to reproduce that failure locally.  wouldn't be surprised if it's a race condition

PR #17778, opened 2022-11-15
feat(typedoc-plugin-appium): create appium typedoc plugin

boneskull:
This is a typedoc plugin based on `typedoc-plugin-markdown`.

It outputs everything that plugin outputs _plus_ it outputs information about commands (API endpoints) as defined in drivers, plugins, and the builtins in `@appium/base-driver`

It's written in TS because that was the easiest way, because a) there was some sort of disagreement between Babel and TS re: exports, and b) everything it's built on is written in TS.

Current preview:

image

The idea is that we will:

1. take the raw `.md` output and stuff it into something like `mkdocs`
2. make this reusable by extension authors

I think 2 is satisfied already; the plugin should "just work"


boneskull:
* master
  * **PR #17777** <a href="https://app.graphite.dev/github/pr/appium/appium/17777" target="_blank">Graphite</a> 
    * **PR #17778** <a href="https://app.graphite.dev/github/pr/appium/appium/17778" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/17778?utm_source=stack-comment).

boneskull:
To run, do a `npm run typedoc` in the monorepo root. You'll see a bunch of warnings. When it's done, see the contents of the `typedoc-docs` directory (we can output somewhere else).

packages/typedoc-plugin-appium/lib/model/types.ts
```diff
@@ -0,0 +1,37 @@
+import {Comment, DeclarationReflection, ProjectReflection} from 'typedoc';
+import {CommandInfo} from './command-info';
+import {CommandsReflection} from './reflection';
+
+/**
+ * I don't think we allow other HTTP methods?
```
jlipps, line 6: these are the only 3 that are part of the w3c spec. i don't think we want to open the door to others

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export function registerHelpers() {
+  Handlebars.registerHelper('reflectionPath', function (this: PageEvent<ContainerReflection>) {
+    if (this.model) {
+      if (this.model.kind && this.model.kind !== ReflectionKind.Module) {
+        if (this.model.kind === (AppiumPluginReflectionKind.COMMANDS as any)) {
+          return `${this.model.name} Commands`;
+        }
+        const title: string[] = [];
+        if (this.model.parent && this.model.parent.parent) {
```
jlipps, line 42: dunno if we really want breadcrumb links, or at least they should be optional? when integrated into our mkdocs setup mkdocs will handle all navigation presentation

packages/typedoc-plugin-appium/lib/converter/builder.ts
```diff
@@ -0,0 +1,104 @@
...
+      this.#createCommandsReflection(project, ctx, projectCmdInfo);
+    } else {
+      this.#log.warn(`No commands found in project ${project.name}`);
+    }
+  }
+
+  /**
+   * Creates and adds a child {@linkcode CommandReflection} to this reflection
+   * @param ref Command reference
+   * @param route Route
```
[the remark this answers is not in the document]
boneskull: They are only descriptive (no types); there's no other way to describe params

packages/typedoc-plugin-appium/lib/converter/builder.ts
```diff
@@ -0,0 +1,104 @@
...
+   * @param ctx - Current context
+   * @param commandInfo - Command information for `module`
+   * @returns New {@linkcode CommandsReflection}
+   */
+  #createCommandsReflection(
+    module: ParentReflection,
+    ctx: Context,
+    commandInfo: CommandInfo
+  ): CommandsReflection {
+    const commandsRef = new CommandsReflection(module.name, ctx.project, commandInfo, module);
```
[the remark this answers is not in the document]
boneskull: makes sense

packages/typedoc-plugin-appium/lib/model/command-info.ts
```diff
@@ -0,0 +1,18 @@
+import {ExecuteCommandSet, RouteMap} from './types';
+
+/**
+ * Data structure describing routes and commands for a particular module (or project)
+ */
+export class CommandInfo {
+  constructor(
+    public readonly routes: RouteMap,
+    public readonly executeCommands: ExecuteCommandSet = new Set()
```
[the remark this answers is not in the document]
boneskull: definitely think we should get terminology straight.  don't have a full picture of prior art though.  maybe easier to discuss offline?

packages/typedoc-plugin-appium/lib/model/reflection/kind.ts
```diff
@@ -0,0 +1,27 @@
...
+  NS,
+  'Root Commands',
+  AppiumPluginReflectionKind.ROOT | AppiumPluginReflectionKind.COMMANDS
+);
+addReflectionKind(
+  NS,
+  'Root Menu',
+  AppiumPluginReflectionKind.ROOT | AppiumPluginReflectionKind.MENU
+);
+AppiumPluginReflectionKind as unknown as ReflectionKind;
```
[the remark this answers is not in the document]
boneskull: you're right; it doesn't seem to actually do anything here.  leftover from copy/paste

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+export const addReflectionKind = (ns: string, name: string, value?: number | null) => {
+  const fullName = `${ns}:${name}`;
+
+  const kindAny = ReflectionKind as any;
+  const existingValue = kindAny[fullName];
+  if (existingValue !== null && existingValue !== undefined) {
+    return existingValue;
+  }
+  const defaultedValue = value ?? getHigherBitMask() * 2;
+  kindAny[fullName] = defaultedValue;
```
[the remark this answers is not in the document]
boneskull: this is part of the code that I lifted from elsewhere (as noted at the top of the document)

packages/typedoc-plugin-appium/lib/model/types.ts
boneskull: k

packages/typedoc-plugin-appium/lib/converter/converter.ts
```diff
@@ -0,0 +1,315 @@
...
+   */
+  #convertMethodMap(refl: DeclarationReflection): RouteMap {
+    const routes: RouteMap = new Map();
+
+    let methodMap: MethodMapDeclarationReflection;
+    const child = isBaseDriverDeclarationReflection(refl)
+      ? refl.getChildByName(NAME_METHOD_MAP)
+      : refl.getChildByName(NAME_NEW_METHOD_MAP);
+
+    if (isMethodMapDeclarationReflection(child)) {
```
mykola-mokhnach, line 192: deeply nested conditions are hard to parse. Consider changing them to guard conditions if possible, for example:

```
if (x) {a1()} else {return a2()}
```

to 

```
if (!x) {
  return a2()
};

a1()
```

packages/typedoc-plugin-appium/lib/logger.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+
+import {format} from 'node:util';
+import {Logger, LogLevel} from 'typedoc';
+
+const LogMethods = {
+  [LogLevel.Error]: 'error',
+  [LogLevel.Warn]: 'warn',
+  [LogLevel.Info]: 'info',
+  [LogLevel.Verbose]: 'verbose',
```
mykola-mokhnach, line 19: I would prefer to have debug level as well

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
```diff
@@ -0,0 +1,60 @@
+import _ from 'lodash';
+import fs from 'node:fs';
+import path from 'node:path';
+import Handlebars from 'handlebars';
+import {ContainerReflection, PageEvent, ReflectionKind} from 'typedoc';
+import {AppiumPluginReflectionKind} from '../../model';
+
+const RESOURCES_PATH = path.join(__dirname, '..', '..', '..', 'resources');
```
mykola-mokhnach, line 8: I's rather use path.abosulte

mykola-mokhnach:
it is hard to understand what this all nice stuff is doing without tests.
Also Appium unit tests don't seem to be happy

packages/typedoc-plugin-appium/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "../../config/tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": "lib",
+    "outDir": "build",
+    "strict": true,
+    "emitDeclarationOnly": false,
+    "module": "CommonJS",
+    "sourceMap": true,
+    "jsx": "react",
```
mykola-mokhnach, line 10: does it have anything to do with react?

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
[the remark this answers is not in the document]
jlipps: i think what hurts my brain is why you are assigning 'defaultedValue' to 'fullName' and 'fullName' to 'defaultedValue'. is this not erroneous?

packages/typedoc-plugin-appium/lib/model/command-info.ts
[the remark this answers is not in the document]
jlipps: my vote would be to call them `executeMethods` since that is what we ultimately decided to call them in our docs unless you have a preference/reason otherwise.

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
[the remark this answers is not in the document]
boneskull: it's a copy-paste job!! not my fault!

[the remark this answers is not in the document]
boneskull: (it's assigning the `fullName` prop of `kindAny` to a `number` as determined by `getHigherBitMask()` or whatever you passed for `value`. I don't use `value`)

packages/typedoc-plugin-appium/lib/logger.ts
boneskull: There's no debug level, since these levels come from typedoc.  typedoc only has "verbose"

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
boneskull: what do you mean?

jlipps: isn't `__dirname` guaranteed to be absolute?

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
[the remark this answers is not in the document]
jlipps: i'm still confused. i would expect to see this instead:

```js
kindAny[fullName] = fullName
kindAny[defaultedValue] = defaultedValue
```

anything else seems like a failure of variable naming?

boneskull:
@jlipps Here's some basic integration with mkdocs (see 813793f unless I overwrite it)

image

boneskull:
@jlipps FWIW I determined that we could use mkdocs 1.4.0

boneskull:
The strategy I used here was to pull in another plugin which allows you to just import a file into another file.  Instead of this, we could put typedoc's output somewhere in the `docs/` directory.  But we should be careful about which files live under version control, since typedoc's output is just build artifacts.   mkdocs seems to need a single root "docs" directory in which it finds files (we cannot give it a list of directories)

jlipps:
hey that's pretty cool! i like that. and yeah a way to pull in the generated docs dynamically rather than committing is good. we had to commit generated docs in our last docs set up and i can't tell you how many times PRs were made to the generated docs instead of the source files.

jlipps:
presumably we could put this stuff under like a Developer Reference nav heading and call it BaseDriver Reference or something?

packages/docutils/requirements.txt
```diff
@@ -0,0 +1,4 @@
+mkdocs==1.4.0
```
mykola-mokhnach, line 1: shall we rather switch to Poetry or Pipenv for deps management?

packages/typedoc-plugin-appium/lib/converter/builder.ts
```diff
@@ -0,0 +1,121 @@
...
+    const modules = project.getChildrenByKind(ReflectionKind.Module);
+
+    // the project itself may have commands, as well as any modules within the project
+    const parents = [...modules, project].filter((parent) => commands.get(parent)?.hasCommands);
+    if (parents.length) {
+      for (const parent of parents) {
+        this.#createCommandsReflection(ctx, parent, commands.get(parent)!);
+      }
+    } else {
+      this.#log.warn('No Appium commands found in the entire project');
```
mykola-mokhnach, line 37: shall we also print the project path?

packages/typedoc-plugin-appium/lib/converter/builder.ts
```diff
@@ -0,0 +1,121 @@
...
+ * @param ctx TypeDoc Context
+ * @param log Plugin logger
+ * @param commandInfo Command info from converter
+ */
+export function createReflections(
+  ctx: Context,
+  log: AppiumPluginLogger,
+  commandInfo: ModuleCommands
+): void {
+  new CommandTreeBuilder(log).createReflections(ctx, commandInfo);
```
mykola-mokhnach, line 120: should we return something?

packages/typedoc-plugin-appium/lib/converter/converter.ts
```diff
@@ -0,0 +1,417 @@
...
+   * @param propName Either required or optional params
+   * @param refl Parent reflection (`params` prop of method def)
+   * @returns List of parameter names
+   */
+
+  #convertCommandParams(
+    propName: typeof NAME_OPTIONAL | typeof NAME_REQUIRED,
+    refl?: DeclarationReflectionWithReflectedType
+  ): string[] {
+    if (refl) {
```
mykola-mokhnach, line 146: I would rather flip the condition and make it to a guard to avoid long nested code blocks

packages/typedoc-plugin-appium/lib/converter/converter.ts
```diff
@@ -0,0 +1,417 @@
...
+  /**
+   * Extracts information about `MethodMap` objects
+   * @param refl - Some reflection we want to inspect. Could refer to a module or a class
+   * @returns Lookup of routes to {@linkcode CommandMap} objects
+   */
+  #convertMethodMap(refl: DeclarationReflection): RouteMap {
+    const routes: RouteMap = new Map();
+
+    const methodMap = isBaseDriverDeclarationReflection(refl)
+      ? refl.getChildByName(NAME_METHOD_MAP)
```
mykola-mokhnach, line 234: applying ternary operator to refl.getChildByName argument might save some code lines

packages/typedoc-plugin-appium/lib/converter/converter.ts
```diff
@@ -0,0 +1,417 @@
...
+      .filter((child) =>
+        isReflectionWithReflectedType(child)
+      ) as DeclarationReflectionWithReflectedType[];
+
+    for (const classRefl of classReflections) {
+      this.#log.verbose('Converting class %s', classRefl.name);
+      const newMethodMap = this.#convertMethodMap(classRefl);
+
+      if (newMethodMap.size) {
+        routes = new Map([...routes, ...newMethodMap]);
```
mykola-mokhnach, line 331: why should we reassign routes every time instead of adding new entries to it?

packages/typedoc-plugin-appium/lib/logger.ts
```diff
@@ -0,0 +1,151 @@
...
+} as const;
+
+export class AppiumPluginLogger extends Logger {
```
mykola-mokhnach, line 22: is this logger specific for this plugin or it is supposed to be used by other plugins as well?

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Adapted from `@knodes/typedoc-pluginutils`
+ * @see https://github.com/knodescommunity/typedoc-plugins
+ * Copyright (c) 2022 KnodesCommunity
+ * Licensed MIT
+ * @see https://github.com/KnodesCommunity/typedoc-plugins/blob/05717565fae14357b1c4be8122f3156e1d46d332/LICENSE
+ */
+
+import {ReflectionKind} from 'typedoc';
+
+const getHigherBitMask = () =>
```
mykola-mokhnach, line 11: should the result be memozied?

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+      .map((v) => v.toString(2))
+      .filter((v) => v.match(/^0*10*$/))
+      .map((v) => parseInt(v, 2))
+  );
+export const addReflectionKind = (ns: string, name: string, value?: number | null) => {
+  const fullName = `${ns}:${name}`;
+
+  const kindAny = ReflectionKind as any;
+  const existingValue = kindAny[fullName];
+  if (existingValue !== null && existingValue !== undefined) {
```
mykola-mokhnach, line 24: you could also use lodash's isNil

packages/typedoc-plugin-appium/lib/model/types.ts
```diff
@@ -0,0 +1,96 @@
...
+export type Route = string;
+
+/**
+ * A method name in `BaseDriver` or subclass thereof
+ *
+ * Currently must be a string, but could theoretically be a {@link PropertyKey}
+ *
+ * Aliased for intent
+ */
+export type Command = string;
```
mykola-mokhnach, line 22: does TS support the concept similar to python's [NewType](https://justincaustin.com/blog/python-typing-newtype/) or this is just a transparent type alias?

packages/typedoc-plugin-appium/lib/output/theme/appium.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export const THEME_NAME = 'appium';
+
+/**
+ * Factory for `AppiumTheme` class; needs custom logger otherwise inaccessible
+ * @param log - Custom logger
+ * @returns `AppiumTheme` class
+ */
+export function getTheme(log: AppiumPluginLogger): new (renderer: Renderer) => MarkdownTheme {
+  return class AppiumTheme extends MarkdownTheme {
```
mykola-mokhnach, line 18: I know this is a correct syntax, but `return class` is still confusing to me

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
mykola-mokhnach: I mean that the resulting string would look like "/bla/../../../resources" if you use `join` and something like `/resources` if absolute path is used. My point is that it is much easier to parse absolute paths in logs in comparison to relative

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+
+registerPartials();
+
+export enum Template {
+  Commands = 'commands.hbs',
+}
+
+export const compileTemplate = _.memoize((template: Template) => {
+  const templatePath = path.join(TEMPLATE_PATH, template);
+  return Handlebars.compile(fs.readFileSync(templatePath, 'utf8'));
```
mykola-mokhnach, line 31: cannot we use async calls?

packages/typedoc-plugin-appium/lib/plugin.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export function load(app: Application) {
+  const log = new AppiumPluginLogger(app.logger, 'appium');
+
+  // register our custom theme.  the user still has to choose it
+  app.renderer.defineTheme(THEME_NAME, getTheme(log));
+
+  app.converter.on(Converter.EVENT_RESOLVE_BEGIN, (ctx: Context) => {
+    // we don't want to do this work if we're not using the custom theme!
+    if (app.renderer.themeName === THEME_NAME) {
```
mykola-mokhnach, line 18: can we flip the condition and turn it into a guard?

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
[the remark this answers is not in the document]
boneskull: I mean, `fullName` is the key and `defaultedValue` is the value.  If you pass a `value`, it uses it, otherwise it computes the next bit mask in sequence.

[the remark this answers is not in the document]
boneskull: Maybe a failure of variable naming.  I'll just change it to make it more clear

packages/docutils/requirements.txt
boneskull: My Python knowledge is from like a decade ago, so whatever you think.

packages/typedoc-plugin-appium/lib/converter/builder.ts
boneskull: This thing basically just makes calls into TypeDoc to mutate its state, so there's nothing to return.

`CommandTreeBuilder` does not need to be a class.  I suppose I could un-class it.

packages/typedoc-plugin-appium/lib/converter/converter.ts
boneskull: doesn't really matter, but if we did that, there'd need to be a `for..of` loop here.

packages/typedoc-plugin-appium/lib/logger.ts
boneskull: it's specific to this plugin. why?

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
boneskull: doesn't need to be. it should only be called like twice

packages/docutils/requirements.txt
mykola-mokhnach: I would advice to use pipenv

packages/typedoc-plugin-appium/lib/model/types.ts
boneskull: It's just an alias to know that we're talking about a string representing a command name.  Likewise with `Route`

packages/typedoc-plugin-appium/lib/converter/converter.ts
mykola-mokhnach: for me it is easier to read when minimum things are reassigned. Also, the oneliner via [forEach](https://stackoverflow.com/questions/32000865/simplest-way-to-merge-es6-maps-sets) also looks good

packages/typedoc-plugin-appium/lib/logger.ts
mykola-mokhnach: I thought we could move it to base plugin. Nvm, if it is specific

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
mykola-mokhnach: it's anyway always the same result. I don't see a reason to calculate it more than once

packages/typedoc-plugin-appium/lib/output/theme/appium.ts
boneskull: I suppose I should explain why I did this.

We have this `AppiumPluginLogger` instance, right.  and the instance is created _from an instance of the default `Logger`_ in `typedoc`--it kind of wraps the instance.  There's no easy & good way--short of a global singleton instance of `AppiumPluginLogger`--to provide it to an `AppiumPluginTheme`.  TypeDoc does the instantiation of the theme class, so I'm not able to just pass arbitrary arguments to its constructor.

There's a "component" architecture in TypeDoc that I could maybe leverage, but it is apparently being phased-out (per the developer docs of TypeDoc).  It looked overcomplicated, so I didn't want to touch it.

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+const RESOURCES_PATH = path.join(__dirname, '..', '..', '..', 'resources');
+const TEMPLATE_PATH = path.join(RESOURCES_PATH, 'templates');
+const PARTIALS_PATH = path.join(RESOURCES_PATH, 'partials');
+
+const Partials = {
+  command: 'command.hbs',
+  executeCommand: 'execute-command.hbs',
+} as const;
+
+function registerPartials() {
```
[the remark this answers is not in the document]
boneskull: It's only called once (directly below)

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
boneskull: No... `compileTemplate` is called by `AppiumTheme#getTemplate`, which is called by `AppiumTheme#mappings` which is a computed property.  It can't return a `Promise` because the parent class expects it to be a value.

packages/typedoc-plugin-appium/lib/plugin.ts
boneskull: No point, really.  We're not using `themeName` to distinguish types; we're just saying "if the enabled theme name is _our_ theme name, then do some work".  If `themeName` isn't `appium`, I don't care what happens.

boneskull:
@jlipps 

> hey that's pretty cool! i like that. and yeah a way to pull in the generated docs dynamically rather than committing is good. we had to commit generated docs in our last docs set up and i can't tell you how many times PRs were made to the generated docs instead of the source files.

Oh, well, I wouldn't commit the generated docs--I don't see any reason to.  They'd be caught by a gitignore.  But I just didn't want to co-mingle generated docs and VCS-controlled docs in the same directory.  I don't feel strongly about it though; it's just one tradeoff for another.  Doing this _would_ eliminate the need for this plugin, though.

packages/typedoc-plugin-appium/lib/output/theme/appium.ts
boneskull: I may also have been able to do this via the event system (`this.emit('here-is-a-logger', log)`), but that seems like abuse of the event system.

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
boneskull: this makes ts unhappy.

packages/typedoc-plugin-appium/lib/logger.ts
boneskull: oh, yeah, no, it's not an appium logger. it's typedoc-specific.

packages/typedoc-plugin-appium/lib/converter/converter.ts
boneskull: iirc our style was to avoid `forEach` in lieu of `for..of`.  but anyway: this whole thing is better as a `reduce`.

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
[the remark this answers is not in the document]
jlipps: i'll take your word for it. for some reason my brain can only read it as a mistake :-)

packages/typedoc-plugin-appium/lib/converter/converter.ts
mykola-mokhnach: ```methodMap =  refl.getChildByName(isBaseDriverDeclarationReflection(refl) ? NAME_METHOD_MAP : ...)```

packages/typedoc-plugin-appium/lib/output/theme/appium.ts
mykola-mokhnach: maybe we could just define class separately, not in the same line with return, to make that easier to read

boneskull:
@mykola-mokhnach FWIW I've added a _lot_ of comments to the plugin so hopefully things are more clear. Please see ebe837f

boneskull:
@mykola-mokhnach Also the `CommandBuilder` class is no longer a class.

packages/typedoc-plugin-appium/lib/converter/builder.ts
```diff
@@ -0,0 +1,120 @@
...
+  const modules = project.getChildrenByKind(ReflectionKind.Module);
+
+  // the project itself may have commands, as well as any modules within the project
+  const parents = [...modules, project].filter((parent) => commandInfo.get(parent)?.hasData);
+  if (!parents.length) {
+    log.warn('No Appium commands found in the entire project');
+    // TODO: maybe we should abort processing gracefully here? or throw?
+  }
+  for (const parent of parents) {
+    createCommandsReflection(log, ctx, parent, commandInfo.get(parent)!);
```
mykola-mokhnach, line 118: we could do this in a .map call above

packages/appium/docs/mkdocs-en.yml
```diff
@@ -44,6 +44,7 @@ nav:
     - guides/event-timing.md
     - guides/log-filters.md
     - guides/grid.md
+  - Reference: reference.md
```
jlipps, line 47:
```suggestion
  - Developer Reference:
      - BaseDriver: reference.md
```

packages/docutils/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Prerequisites
+
+- `mkdocs` v1.4.0
+- `mkdocs-material` v8.2.8
+- `mkdocs-material-extensions` v1.0.3
+
+Run `pip install -r requirements.txt` to install MkDocs & its deps.
+
+## Running
```
jlipps, line 13:
```suggestion
## Running for Main Appium Docs
```

packages/docutils/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+This package does not need to be run directly. Instead, execute this command in the `packages/appium` folder of [this monorepo](https://github.com/appium/appium):
+
+```bash
+npm run dev:docs:en
+````
+
+See the `scripts` field of the `package.json` for other languages.
```
jlipps, line 21:
```suggestion
See the `scripts` field of the `package.json` for other languages.

## Using in Appium Extensions

TODO: give an example of how this package is used in an extension outside the appium monorepo, for example when it's made a part of the XCUITest repo. Or link to the part of the Appium docs where this info will live.
```

packages/docutils/base-mkdocs.yml
```diff
@@ -14,6 +14,8 @@ markdown_extensions:
       emoji_index: !!python/name:materialx.emoji.twemoji
       emoji_generator: !!python/name:materialx.emoji.to_svg
   - footnotes
+  - markdown_include.include:
+      base_path: ../../typedoc-docs/commands
```
jlipps, line 18: will this work when used by an extension? we don't want extension authors to have to recreate this directory structure right?

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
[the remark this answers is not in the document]
jlipps: i now get what's going on. sorry.

packages/typedoc-plugin-appium/lib/converter/builder.ts
boneskull: okay, rewrote this

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
boneskull: I still don't understand.  Do you mean `path.resolve()`?  Or `path.relative()`?  What does this have to do with logging?

packages/typedoc-plugin-appium/resources/templates/commands.hbs
```diff
@@ -0,0 +1,39 @@
...
+{{> comment}}
+
+{{/if}}
+
+{{/with}}
+
+{{#with model}}
+
+{{#if hasRoute}}
+## {{pluralize 'Route' routeCount}}
```
jlipps, line 22: maybe `Command` or `Driver Command` or `API Command` or something? dunno if `Routes` is going to do it for most users' understanding of what these things are.

jlipps (approved):

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
mykola-mokhnach: yes, path.resolve should do the job

packages/typedoc-plugin-appium/lib/model/reflection/utils.ts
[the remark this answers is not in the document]
boneskull: Let me try again.

Enums in TS are both constant types and variables.  There is no way to _add_ stuff to the type of an `enum`; types are immutable.  However, you _can_ treat an `enum` like an object and do whatever you want with it, which is what the author does: `const kindAny = ReflectionKind as any`.

The code as written is dependent on a TS implementation detail (which is unlikely to change); the compiled JS results in an object with its values for its keys and vice-versa.

In other words, if it looks incorrect, it's because it's a hack around the limitations of the TypeDoc API.  It's unfortunate, but necessary.  I had a conversation with a TypeDoc maintainer about this, and think I'll bring it up again...

packages/typedoc-plugin-appium/resources/templates/commands.hbs
boneskull: right.  route is technical jargon.

packages/docutils/base-mkdocs.yml
boneskull: not as written, no.  will be separate PR

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
boneskull: well, what @jlipps was saying, it doesn't need to be `path.resolve()` since `__dirname` is always absolute, so `path.join()` will be the same thing.

packages/typedoc-plugin-appium/tsconfig.json
boneskull: ah, good catch. it did at one point!

packages/typedoc-plugin-appium/lib/output/theme/utils.ts
jlipps: I now understand why @mykola-mokhnach prefers path.resolve. It will eliminate `../../..` in the resulting path, which makes looking at the path as a human more sensible (because you don't have to go down three directories and back up). we might look at this path as a human if e.g. it is printed out in logs.

Issue #17779, opened 2022-11-16
Appium2.0: InvalidArgumentError: 'automationName' can't be blank


atzdmyp:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when i run script, i got these error:
Stacktrace:
InvalidArgumentError: 'automationName' can't be blank
    at processCapabilities (/usr/local/lib/node_modules/appium/node_modules/@appium/base-driver/lib/basedriver/capabilities.js:330:13)
    at parseCapsForInnerDriver (/usr/local/lib/node_modules/appium/lib/utils.js:134:21)
    at AppiumDriver.createSession (/usr/local/lib/node_modules/appium/lib/appium.js:241:26)
    at commandExecutor (/usr/local/lib/node_modules/appium/node_modules/@appium/base-driver/lib/basedriver/driver.js:126:9)
    at AppiumDriver.executeCommand (/usr/local/lib/node_modules/appium/node_modules/@appium/base-driver/lib/basedriver/driver.js:143:15)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at defaultBehavior (/usr/local/lib/node_modules/appium/lib/appium.js:668:16)
    at AppiumDriver.executeWrappedCommand (/usr/local/lib/node_modules/appium/lib/appium.js:762:16)
    at AppiumDriver.executeCommand (/usr/local/lib/node_modules/appium/lib/appium.js:684:17)
    at asyncHandler (/usr/local/lib/node_modules/appium/node_modules/@appium/base-driver/lib/protocol/protocol.js:346:19)

but when i use 'appium' prefix for capabilities, will be no error. who can help me.

### Expected Behavior

create session

### Minimal Reproducible Example

it's my code:

from appium import webdriver

desired_cap = {
                "platformName": "iOS",
                "appium:options":{
                    "platformVersion": platform_version,
                    "udid": udid,
                    "deviceName": device_name,
                    "bundleId": bundle_id,
                    "automationName": "XCUITest",
                    "xcodeOrgId": xcode_org_id,
                    "xcodeSigningId": xcode_signing_id,
                    "wdaLocalPort": wda_local_port,
                    "derivedDataPath": derived_data_path,
                    "appPushTimeout": 150000,
                }
            }
driver = webdriver.Remote(
                "/usr/local/bin/appium -a 127.0.0.1 -p 4723" , desired_cap
            )

### Environment

- Operating system: mac os monterey 12.1
- env
[...]

PR #17781, opened 2022-11-16
fix(base-driver): Properly validate capabilities inside appium:options map

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/appium/appium/issues/17779

## Types of changes

### DIFF
--- a/packages/base-driver/lib/basedriver/capabilities.js
+++ b/packages/base-driver/lib/basedriver/capabilities.js
@@ -377,8 +377,8 @@ function promoteAppiumOptions(originalCaps) {
       return capName;
     };
     const preprocessedOptions = _(appiumOptions)
-      .mapKeys(verifyIfAcceptable)
-      .mapKeys((key) => shouldAddVendorPrefix(key) ? `${APPIUM_VENDOR_PREFIX}${key}` : key)
+      .mapKeys((value, key) => verifyIfAcceptable(key))
+      .mapKeys((value, key) => shouldAddVendorPrefix(key) ? `${APPIUM_VENDOR_PREFIX}${key}` : key)
       .value();
     // warn if we are going to overwrite any keys on the base caps object
     const overwrittenKeys = _.intersection(Object.keys(obj), Object.keys(preprocessedOptions));
--- a/packages/base-driver/lib/basedriver/capabilities.js
+++ b/packages/base-driver/lib/basedriver/capabilities.js
@@ -360,16 +360,20 @@ function promoteAppiumOptions(originalCaps) {
     );
 
     const shouldAddVendorPrefix = (/** @type {string} */ capName) => !capName.startsWith(APPIUM_VENDOR_PREFIX);
-    const verifyIfAcceptable = (/** @type {string} */ capName) => {
+    const verifyIfAcceptable = (capName) => {
+      if (!_.isString(capName)) {
+        throw new errors.SessionNotCreatedError(
+          `Capability names in ${PREFIXED_APPIUM_OPTS_CAP} must be strings. '${capName}' is unexpected`
+        );
+      }
       if (isStandardCap(capName)) {
         throw new errors.SessionNotCreatedError(
-          `${PREFIXED_APPIUM_OPTS_CAP} must only contain vendor-specific capabilties. '${capName}' one is unexpected`
+          `${PREFIXED_APPIUM_OPTS_CAP} must only contain vendor-specific capabilties. '${capName}' is unexpected`
         );
       }
       return capName;
     };
     const preprocessedOptions = _.toPairs(_.cloneDeep(appiumOptions))
-      .map(([name, value]) => [String(name), value])
       .map(([name, value]) => [verifyIfAcceptable(name), value])
       .map(([name, value]) => [shouldAddVendorPrefix(name) ? `${APPIUM_VENDOR_PREFIX}${name}` : name, value])
       .reduce((acc, [name, value]) => {acc[name] = value; return acc;}, {});
--- a/packages/base-driver/lib/basedriver/capabilities.js
+++ b/packages/base-driver/lib/basedriver/capabilities.js
@@ -359,7 +359,10 @@ function promoteAppiumOptions(originalCaps) {
       `Found ${PREFIXED_APPIUM_OPTS_CAP} capability present; will promote items inside to caps`
     );
 
-    const shouldAddVendorPrefix = (/** @type {string} */ capName) => !capName.startsWith(APPIUM_VENDOR_PREFIX);
+    /**
+     * @param {string} capName
+     */
+    const shouldAddVendorPrefix = (capName) => !capName.startsWith(APPIUM_VENDOR_PREFIX);
     const verifyIfAcceptable = (capName) => {
       if (!_.isString(capName)) {
         throw new errors.SessionNotCreatedError(
@@ -373,10 +376,10 @@ function promoteAppiumOptions(originalCaps) {
       }
       return capName;
     };
-    const preprocessedOptions = _.toPairs(_.cloneDeep(appiumOptions))
-      .map(([name, value]) => [verifyIfAcceptable(name), value])
-      .map(([name, value]) => [shouldAddVendorPrefix(name) ? `${APPIUM_VENDOR_PREFIX}${name}` : name, value])
-      .reduce((acc, [name, value]) => {acc[name] = value; return acc;}, {});
+    const preprocessedOptions = _(appiumOptions)
+      .mapKeys(verifyIfAcceptable)
+      .mapKeys((key) => shouldAddVendorPrefix(key) ? `${APPIUM_VENDOR_PREFIX}${key}` : key)
+      .value();
     // warn if we are going to overwrite any keys on the base caps object
     const overwrittenKeys = _.intersection(Object.keys(obj), Object.keys(preprocessedOptions));
     if (overwrittenKeys.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -336,43 +335,70 @@ function processCapabilities(
...
     );
-  }
 
-  // now just apply them to the main caps object
-  caps = {...caps, ...strippedAppiumOptions};
+    const shouldAddVendorPrefix = (/** @type {string} */ capName) => !capName.startsWith(APPIUM_VENDOR_PREFIX);
+    const verifyIfAcceptable = (/** @type {string} */ capName) => {
+      if (isStandardCap(capName)) {
+        throw new errors.SessionNotCreatedError(
+          `${PREFIXED_APPIUM_OPTS_CAP} must only contain vendor-specific capabilties. '${capName}' one is unexpected`
```
jlipps, line 366:
```suggestion
          `${PREFIXED_APPIUM_OPTS_CAP} must only contain vendor-specific capabilties. '${capName}' is unexpected`
```

jlipps (approved):

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -336,43 +335,74 @@ function processCapabilities(
...
+        throw new errors.SessionNotCreatedError(
+          `${PREFIXED_APPIUM_OPTS_CAP} must only contain vendor-specific capabilties. '${capName}' is unexpected`
+        );
+      }
+      return capName;
+    };
+    const preprocessedOptions = _.toPairs(_.cloneDeep(appiumOptions))
+      .map(([name, value]) => [verifyIfAcceptable(name), value])
+      .map(([name, value]) => [shouldAddVendorPrefix(name) ? `${APPIUM_VENDOR_PREFIX}${name}` : name, value])
+      .reduce((acc, [name, value]) => {acc[name] = value; return acc;}, {});
```
boneskull, line 379: `appiumOptions` doesn't need to be cloned here because we do not mutate the value.

```suggestion
    const preprocessedOptions = _.mapKeys(
      _.mapKeys(appiumOptions, verifyIfAcceptable), 
      name => shouldAddVendorPrefix(name) ? `${APPIUM_VENDOR_PREFIX}${name}` : name, value])
    );
```

or if you wanna get real fancy

```suggestion
    const preprocessOptions = _.flow(
      _.mapKeys(_, verifyIfAcceptible),
      _.mapKeys(_, name => shouldAddVendorPrefix(name) ? `${APPIUM_VENDOR_PREFIX}${name}` : name, value])
    );
    const preprocessedOptions = preprocessOptions(appiumOptions);
```

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -336,43 +335,74 @@ function processCapabilities(
...
+    // warn if we are going to overwrite any keys on the base caps object
+    const overwrittenKeys = _.intersection(Object.keys(obj), Object.keys(preprocessedOptions));
+    if (overwrittenKeys.length > 0) {
+      log.warn(
+        `Found capabilities inside ${PREFIXED_APPIUM_OPTS_CAP} that will overwrite ` +
+          `capabilities at the top level: ${JSON.stringify(overwrittenKeys)}`
+      );
+    }
+    const res = _.cloneDeep(obj);
+    delete res[PREFIXED_APPIUM_OPTS_CAP];
```
boneskull, line 389:
```suggestion
    const res = _.omit(obj, PREFIXED_APPIUM_OPTS_CAP);
```

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -336,43 +335,74 @@ function processCapabilities(
...
+  };
 
-  // and remove all traces of the options cap
-  delete caps[APPIUM_OPTS_CAP];
-  return caps;
+  const {alwaysMatch, firstMatch} = originalCaps;
+  const result = {};
+  if (_.isPlainObject(alwaysMatch)) {
+    result.alwaysMatch = promoteForObject(alwaysMatch);
+  } else if (_.has(originalCaps, 'alwaysMatch')) {
```
boneskull, line 397: one thing to note is that TS has a tough time with `_.has()`.  a way to do this that it _does_ understand would be

```suggestion
  } else if ('alwaysMatch' in originalCaps) {
```

This is likely a non-issue in this case, but just for future reference--`_.has()` is not very useful nowadays because of nullish coalescing

packages/base-driver/test/unit/basedriver/capabilities.spec.js
```diff
@@ -540,67 +541,79 @@ describe('caps', function () {
...
+    const nonPrefixedAppiumCaps = {
+      platformVersion: '14.4',
+      deviceName: 'iPhone 11',
+      app: '/foo/bar.app.zip',
+      automationName: 'XCUITest',
     };
-    const nonAppiumCaps = {
+    const appiumCaps = _.toPairs(nonPrefixedAppiumCaps)
+      .map(([name, value]) => [`${APPIUM_VENDOR_PREFIX}${name}`, value])
+      .reduce((acc, [name, value]) => {acc[name] = value; return acc;}, {});
```
boneskull, line 552:
```suggestion
    const appiumCaps = _.mapKeys(nonPrefixedAppiumCaps, (name) => `${APPIUM_VENDOR_PREFIX}${name}`);
```

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -336,43 +335,74 @@ function processCapabilities(
...
 function promoteAppiumOptions(originalCaps) {
-  const appiumOptions = originalCaps[APPIUM_OPTS_CAP];
-  if (!appiumOptions) {
-    return originalCaps;
-  }
+  const promoteForObject = (obj) => {
+    const appiumOptions = obj[PREFIXED_APPIUM_OPTS_CAP];
+    if (!appiumOptions) {
+      return obj;
+    }
```
boneskull, line 349:
```suggestion
    if (_.isEmpty(appiumOptions)) {
      return obj;
    }
```

.. then delete L354-356, since it's redundant.

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -336,43 +335,74 @@ function processCapabilities(
...
 
-  let caps = _.cloneDeep(originalCaps);
-  if (!_.isPlainObject(appiumOptions)) {
-    throw new errors.SessionNotCreatedError(`The ${APPIUM_OPTS_CAP} capability must be an object`);
-  }
+    if (!_.isPlainObject(appiumOptions)) {
```
boneskull, line 351: Is it actually possible for, say, an `array` to reach this function?  Would some validator toss it out before this gets called?

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -336,43 +335,74 @@ function processCapabilities(
...
+      `Found ${PREFIXED_APPIUM_OPTS_CAP} capability present; will promote items inside to caps`
     );
-  }
 
-  // now just apply them to the main caps object
-  caps = {...caps, ...strippedAppiumOptions};
+    const shouldAddVendorPrefix = (/** @type {string} */ capName) => !capName.startsWith(APPIUM_VENDOR_PREFIX);
+    const verifyIfAcceptable = (capName) => {
+      if (!_.isString(capName)) {
+        throw new errors.SessionNotCreatedError(
```
boneskull, line 365: If we have a non-array object, then assuming `capName` is a property of that object, it _must_ be a `string`, because a `symbol` cannot be transmitted over the wire.  Right?

boneskull (approved):
had some suggestions but nothing that should block merge.

boneskull:
so, something broke in CI, and I'm not sure what it was.  I don't the failure has anything to do with this PR, since I see it elsewhere.  I'll take a look into it

mykola-mokhnach:
> so, something broke in CI, and I'm not sure what it was. I don't the failure has anything to do with this PR, since I see it elsewhere. I'll take a look into it

Thanks, I have also assumed these failures are unrelated to my changes

packages/base-driver/lib/basedriver/capabilities.js
mykola-mokhnach: appiumOptions might have dictionary values. I need to make sure these are not going to be accidentally mutated in the returned value

packages/base-driver/lib/basedriver/capabilities.js
mykola-mokhnach: isEmpty also returns true for an empty object (`{}`) I don't want it to be the case here

packages/base-driver/lib/basedriver/capabilities.js
mykola-mokhnach: we do not validate options before, because this is just a simple capability value. this is the only function where it gets detected and validated

packages/base-driver/lib/basedriver/capabilities.js
mykola-mokhnach: in general yes. just being a bit paranoidal

packages/base-driver/test/unit/basedriver/capabilities.spec.js
mykola-mokhnach: changed

packages/base-driver/lib/basedriver/capabilities.js
mykola-mokhnach: changed

packages/base-driver/lib/basedriver/capabilities.js
mykola-mokhnach: changed

packages/base-driver/lib/basedriver/capabilities.js
```diff
@@ -336,43 +335,74 @@ function processCapabilities(
...
-      `Found capabilities inside ${PREFIXED_APPIUM_OPTS_CAP} that will overwrite ` +
-        `capabilities at the top level: ${JSON.stringify(overwrittenKeys)}`
+    log.debug(
+      `Found ${PREFIXED_APPIUM_OPTS_CAP} capability present; will promote items inside to caps`
     );
-  }
 
-  // now just apply them to the main caps object
-  caps = {...caps, ...strippedAppiumOptions};
+    const shouldAddVendorPrefix = (/** @type {string} */ capName) => !capName.startsWith(APPIUM_VENDOR_PREFIX);
```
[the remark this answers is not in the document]
mykola-mokhnach: changed (not sure what's the difference though)

PR #17787, opened 2022-11-17
chore(test): update flattened schema fixture to match actual schema

boneskull:
This fixes the test that is making CI fail.

### DIFF
--- a/packages/appium/lib/schema/schema.js
+++ b/packages/appium/lib/schema/schema.js
@@ -318,7 +318,7 @@ class AppiumSchema {
     }
     const normalizedExtName = _.kebabCase(extName);
     if (this.hasRegisteredSchema(extType, normalizedExtName)) {
-      if (this._registeredSchemas[extType].get(normalizedExtName) === schema) {
+      if (_.isEqual(this._registeredSchemas[extType].get(normalizedExtName), schema)) {
         return;
       }
       throw new SchemaNameConflictError(extType, extName);
--- a/packages/appium/test/e2e/plugin.e2e.spec.js
+++ b/packages/appium/test/e2e/plugin.e2e.spec.js
@@ -11,13 +11,14 @@ import {W3C_PREFIXED_CAPS, TEST_HOST, getTestPort, PROJECT_ROOT} from '../helper
 import {runExtensionCommand} from '../../lib/cli/extension';
 import {tempDir, fs} from '@appium/support';
 import {loadExtensions} from '../../lib/extension';
+import {DRIVER_TYPE, PLUGIN_TYPE} from '../../lib/constants';
 
 const FAKE_ARGS = {sillyWebServerPort: 1234, host: 'hey'};
 const FAKE_PLUGIN_ARGS = {fake: FAKE_ARGS};
 
 const should = chai.should();
 
-/** @type {WebdriverIO.RemoteOptions} */
+/** @type {import('webdriverio').RemoteOptions} */
 const wdOpts = {
   hostname: TEST_HOST,
   connectionRetryCount: 0,
@@ -49,6 +50,8 @@ describe('FakePlugin', function () {
       {
         driverCommand: 'list',
         showInstalled: true,
+        subcommand: DRIVER_TYPE,
+        suppressOutput: true,
       },
       driverConfig
     );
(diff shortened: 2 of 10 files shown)


boneskull:
It'd be nice if this fixture got auto-updated when the schema changes, but the only way to do that is to use the code that's under test, which means it'd defeat the point of the test.

packages/appium/lib/schema/schema.js
```diff
@@ -576,17 +579,19 @@ export class SchemaNameConflictError extends Error {
...
    * @param {string} extName
+   * @param {{conflictedSchema?: SchemaObject}} [opts]
    */
-  constructor(extType, extName) {
-    super(`Name for ${extType} schema "${extName}" conflicts with an existing schema`);
-    this.data = {extType, extName};
+  constructor(extType, extName, {conflictedSchema} = {}) {
+    super(`Name for ${extType} schema "${extName}" conflicts with an existing schema.`);
+    this.data = {extType, extName, conflictedSchema};
+    logger.debug(JSON.stringify(this.data, null, 2));
```
jlipps: we want to print something just when the error is created? is this kind of an equivalent to our  old pattern of using `errAndThrow`?

packages/appium/test/fixtures/flattened-schema.js
```diff
@@ -522,7 +522,7 @@ export default [
     },
     schema: {
       appiumCliDescription:
-        'A list of drivers to activate. Can be a comma-delimited string or path to CSV file. By default, all installed drivers will be activated.',
+        'A list of drivers to activate. Can be a comma-delimited string or path to CSV file. By default, all installed drivers will be activated. Windows environments may require wrapping the comma-delimited string with quotes to escape the comma.',
```
jlipps, line 525: i feel like these changes came in erroneously from another PR?

jlipps: or oh wait this is probably generated from that stuff.

jlipps (approved):

packages/appium/lib/schema/schema.js
boneskull: I tossed this anyway

packages/appium/test/fixtures/flattened-schema.js
boneskull: no, neither; this file needed updating manually as per my comment above in the issue

boneskull:
I can't work on this any more today and will pick it up tomorrow.

the root cause has something to do with the recent changes to auto-discovery of extensions.  as I mentioned in slack, I'm unable to reproduce this locally (so it makes fixing it considerably slower).   there's something in the CI environment that is causing fake-driver to be installed twice in a temp dir.  it _seems_ like `appium driver list` does _not_ list `fake`, but upon `appium driver install fake`, it's already installed.  this may have something to do with CI caching, so disabling that is probably the first thing I'd check.  next, running with `_FORCE_LOGS=1` may give further insight into the `npm` commands being issued.  also running a sanity check on `extensions.yaml` and the contents of the temp dir before the code causing the exception runs...

`appium driver list` _should_ cause locally-installed extensions to be auto-detected (which will update `extensions.yaml`), so that's correct, but it's unknown why the driver would be considered locally-installed unless the temp dir's `node_modules` is somehow in an unexpected state.

packages/appium/test/fixtures/flattened-schema.js
jlipps: 👍

packages/appium/test/e2e/plugin.e2e.spec.js
```diff
@@ -67,13 +71,17 @@ describe('FakePlugin', function () {
...
+        json: true,
+        suppressOutput: true,
       },
       pluginConfig
     );
     if (!_.has(pluginList, 'fake')) {
       await runExtensionCommand(
         {
           pluginCommand: 'install',
+          subcommand: 'plugin',
```
mykola-mokhnach, line 84: PLUGIN_TYPE?

boneskull: good catch thanks

boneskull:
OK, so I've confirmed that `extensions.yaml` is clean when we go to install.  This means that maybe the schema registration is happening twice...

boneskull:
when this is green I'll just squash the whole mess

PR #17808, opened 2022-11-21
docs(1.x): Add missing parameter to start-activity.

garg3133:
## Proposed changes

Add missing `appWaitActivity` parameter to start-activity. This parameter is present in appium-base-dir's [routes.js](https://github.com/appium/appium-base-driver/blob/master/lib/protocol/routes.js#L536) file but is missing from the docs.

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the CLA

## Further comments

### DIFF
--- a/commands-yml/commands/device/activity/start-activity.yml
+++ b/commands-yml/commands/device/activity/start-activity.yml
@@ -76,6 +76,9 @@ endpoint:
     - name: appWaitPackage
       type: string
       description: Automation will begin after this package starts
+    - name: appWaitActivity
+      type: string
+      description: Automation will begin after this activity starts
     - name: intentAction
       type: string
       description: "[Intent](https://developer.android.com/reference/android/content/Intent.html) action which will be used to start activity"


boneskull:
@jlipps are we accepting updates to v1.x docs?

jlipps:
i guess it doesn't hurt. it will probably never be published though.

garg3133:
Ohh, I only made this change because I wanted to reference somewhere all the json parameters that the `startActivity` command accepts. But it's okay if it won't be published, I can just add the link to the modified yml file in that case instead of the Appium website until these commands are available on Appium 2.0 docs.

jlipps:
> I can just add the link to the modified yml

The yml files don't exist in the 2.0 branch, so should not be linked to

> until these commands are available on Appium 2.0 docs

The Appium 2.0 docs will not list driver-specific commands at all. Instead, the drivers will have all the docs for their own commands. The best way to ensure that this parameter is mentioned in the uiautomator2 doc moving forward is to send a PR to that repo

boneskull:
Should this be closed?

garg3133:
> The best way to ensure that this parameter is mentioned in the uiautomator2 doc moving forward is to send a PR to that repo

@jlipps I would love to do that but I couldn't find the docs for the commands anywhere in the [appium-uiautomator2-driver](https://github.com/appium/appium-uiautomator2-driver) repo. I couldn't even find the implementation for the `startActivity` method but the [README](https://github.com/appium/appium-uiautomator2-driver#applications-management) says that it's there.

Can you please direct me where I can look for the doc for `startActivity` command or even the implementation of the same? From [here](https://github.com/appium/appium-uiautomator2-driver/blob/783cec6dd8945896dbbfd565d7c0223c0b05166e/lib/driver.js#L77), it seems to me that the command is not being proxied to `appium-uiautomator2-server` and it doesn't make sense to proxy it to `chromedriver`. I also looked into `appium-adb` but that isn't really a server but just a library. So, how does this and similar commands work?

jlipps:
@garg3133 i think probably in this case we should just do nothing. pretty soon the uiautomator2 command docs will be auto generated from the implementation code, so all the parameters will be properly documented.

PR #17843, opened 2022-11-29
chore(renovate): update semantic commit behavior

boneskull:
The current intended behavior (to use the "parent dir") as the scope is often or always broken.  An empty scope within parens is actually an invalid conventional commit message.  So instead of `chore():` it will be `chore:`.

We could instead pick a scope (like `deps` or something) but figured it was not important enough to do.

### DIFF
--- a/renovate/README.md
+++ b/renovate/README.md
@@ -43,7 +43,8 @@ Appium extension authors--or anyone else--may use this config as well.
 - `:automergeDigest` - Automatically merges "digest" updates (assuming they pass CI)
 - `:enableVulnerabilityAlerts` - For "security" purposes
 - `:rebaseStalePrs` - Renovate will automatically rebase its PRs
-- `:semanticPrefixChore` - Renovate's PRs have the `chore()` scope in its semantic commit message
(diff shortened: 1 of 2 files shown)


jlipps (approved):
so we won't get chore messages scoped to packages anymore? won't that break lerna's auto versioning logic? or this will only affect things in case the scope would be empty?

boneskull:
actually...hm. you're right, i see it actually does work sometimes.  

I can't imagine Lerna uses the scope to determine what packages are affected though--the scope could be literally anything.  I assume it uses the filepaths in the changeset

boneskull:
Let's see if this works

boneskull:
I've changed `semanticCommitScope` to be this Handlebars template: `"{{#if parentDir}}{{parentDir}}{{else}}deps{{/if}}"`

Since the validator probably doesn't validate templates, I've applied this change to another one of my projects, so we can see what it does.  When it's clear to me it works, I'll merge this.

jlipps:
sounds good. pretty cool you can use handlebar templates!

boneskull:
I can confirm this works. will fix conflicts and merge

Issue #17807, opened 2022-11-21
we don't really need babel


boneskull:
Did I make this ticket already?

Given we're targeting v14, we don't actually need most of what Babel gives us. All we really need is to convert from ESM to CJS. I'm sure I've mentioned this before somewhere. 

Technically we don't even _need_ a build step, but I wouldn't recommend going all-in on pure ESM (at least not yet); such a move would require a nontrivial amount of refactoring and exploratory work around tests and mocks. 

So if we're just converting module formats, we don't have to use Babel. ideally I'd like to trim down the toolchain weight and increase build performance. Possible candidates to replace it include (but are not limited to):

- TypeScript; handy because it's already in the toolchain
- esbuild - very fast. pretty easy. uses native modules though. does not do polyfills etc. 
- ncc - basically just compresses an entire package into a single file. don't recall if we can get it to _not_ bundle dependencies
- rollup - only advantage over esbuild is that it's probably more mature

I've got varying levels of experience with all of these, and would probably recommend trying TS or esbuild. 

If we do go with esbuild, then how exactly we use it with Mocha is unclear to me, but I'm sure it's possible. whatever we use should allow us to run source files directly and provide reasonable stack traces like what we have now. 

there seem to be plenty of tools like this nowadays. any others we should consider?

low priority.

boneskull:
also maybe something we could use to solve #17746. some sort of all-in-one thing? problem is that most of these tools are very web-centric, e.g. vite

mykola-mokhnach:
could you please mention what are pros and cons of using babel for the current setup?
what would be the transition strategy if we decide to get rid of it. There are plenty of different packages under Appium github org.
We probably need to get rid of these that are not maintained anymore first (move them to boneyard maybe)?

jlipps:
I'd be in favour of jettisoning babel for tsc if we can make it happen without a huge amount of work. having it as the default toolchain makes it easier to incrementally add typescript to various places, which in my mind would be preferable than the ts-in-comments situation we have now (though i wouldn't advocate for a full rewrite to TS since it seems necessary).

boneskull:
> could you please mention what are pros and cons of using babel for the current setup?

I'm not sure how easily I can split this into "pros" and "cons", since it's all tradeoffs, but:

- We can use any stage-4 syntax regardless of our supported Node.js versions.
- Babel lets us (for better or worse) mix and match ESM and CJS in a Node.js context.
[...]

PR #17845, opened 2022-11-29
chore(appium): experimental build via tsc

### LINKED ISSUE #17746: dev script is broken: too many workspaces
So basically if you run `npm run dev` what _should_ happen is that the `dev` script runs in every package.  

There are like 22 packages in there, and a dev env probably doesn't have that many cores available, so only _some non-deterministic subset_ of those run.  This is not an issue for the `build` script since those scripts do not stay running indefinitely.

Need to figure out how to shoehorn this into a single command.  IDK maybe `nx` has a solution for this, since `lerna` now uses it under the hood.

boneskull:
See #17807.

This shows that `appium` can be built via `tsc`.

Here's [a gist which shows the output difference](https://gist.github.com/boneskull/a0d3b3659a32830335ed7b9f87bfe31e)


config/tsconfig.base.json
```diff
@@ -7,12 +7,12 @@
...
     "declarationMap": true,
     "emitDeclarationOnly": true,
-    "module": "es2015",
+    "module": "CommonJS",
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "strictNullChecks": true,
     "removeComments": false,
-    "target": "es2015",
+    "target": "ESNext",
```
jlipps, line 15: don't we want to make sure the target matches our lowest version of node?

boneskull: hmm, yeah, probably.  I'm not sure what that is.  I'll have to figure it out.

boneskull: OK, this is ES2020 for Node.js v14.17.0 as per [the kangax table](https://kangax.github.io/compat-table/es2016plus/)

boneskull:
If merged, this will close #17746

boneskull:
@jlipps @mykola-mokhnach @KazuCocoa 

This looks pretty good to me.  Should I expand this to compile the whole monorepo?

mykola-mokhnach (approved):

KazuCocoa (approved):
oh, looks cool to me, 🚀

jlipps (approved):
awesome! lgtm once CI is happy. and did you manually run scripts that aren't run in CI just to double check they still work?

.npmrc
```diff
@@ -1,4 +1,4 @@
 # Do not disable package-lock. Disabling package-lock also disables npm-shrinkwrap
 package-lock=true
 save-exact=true
-
+foreground-scripts=true
```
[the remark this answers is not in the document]
boneskull: it's new-ish. they added it because people were begging in `postinstall` scripts.

boneskull:
yeah I don't think this will work but it's a nice thought

```bash
for $script in $(cat package.json | '.scripts | keys[]'); do 
  npm run $script
done`
```

boneskull:
@jlipps btw the only reason CI is failing is because stuff that was silently failing no longer does so silently.

boneskull:
@jlipps OK, so I removed the "build documentation" job from CI because it hasn't been working in quite some time and it was failing silently.  Historically, it was trying to run the `parse-yml-commands` script, but that script is looking for files that no longer exist (`commands-yml/`, `template.md`).

PR #17857, opened 2022-12-02
chore: use commitlint

### LINKED ISSUE #17844: use commitlint
Might be worthwhile to adopt [commitlint](https://www.npmjs.com/package/@commitlint/cli).  I have it running on a couple of my projects and works well.

While we're in there, we should swap out `pre-commit` for `husky` given the former hasn't been published in six (6) years.

boneskull:
This adds [commitlint](https://npm.im/@commitlint/cli) to the project which validates commit messages against the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec via a preset.

Removed the [pre-commit](https://npm.im/pre-commit) package as it has not been updated in six (6) years; replaced with [husky](https://typicode.github.io/husky/) which is widely-used and full-featured.

Customized commitlint so that it will allow commit messages lines of arbitrary length; this is set to 100 by default, but is not in the specification.  If someone has a good reason that we _shouldn't_ disable that rule, please say something!

Closes #17844


KazuCocoa (approved):

jlipps (approved):
could we add a github action step that runs commitlint on the message of HEAD?

boneskull:
@jlipps well.. what would we do about it if that check fails?

boneskull:
okay I think by `HEAD` you mean "in a PR" because once it gets to `master`...

boneskull:
* master
  * **PR #17857** <a href="https://app.graphite.dev/github/pr/appium/appium/17857" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/17857?utm_source=stack-comment).

boneskull:
@jlipps See 7373125

boneskull:
hm.

jlipps (approved):
yeah nice! i like it. we can always still choose to squash and rebase if we want.

PR #17864, opened 2022-12-02
chore: improve package-lock.json merging

boneskull:
Yeah I got sick of this.  It won't help conflicts seen by GitHub, but it'll make our lives easier when rebasing PRs.

So what this does is treats `package-lock.json` as a text file _except_ when attempting to merge. The result will be that if `package-lock.json` is in a conflicted state, it will not contain conflict markers within the file itself like a regular text file would; it will only have the "conflicted" flag set by git.  This means that to resolve a conflict in `package-lock.json` going forward, all one needs to do is run `npm install && git add -A package-lock.json`, as `npm` will fix `package-lock.json` itself.  In theory.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 * text=auto eol=lf
+
+# Handle as a text file but merge as binary.
+package-lock.json merge=binary


mykola-mokhnach (approved):
Did you borrow this idea from somewhere or it's some unique solution?

boneskull:
@mykola-mokhnach Yeah I found this on the internet somewhere.  There are other ways to try to get at this problem, but this one seemed to be low-impact.

@jlipps I'm not sure which one will be the active one.  I can report back next time I run into a conflict.

Basically:

_If your workflow was this:_ `rm package-lock.json && npm i && git add package-lock.json`
_It now becomes this:_ `npm i && git add package-lock.json`

boneskull:
@jlipps The `package-lock.json` is the version that's been rebased onto.  I don't know if that's "theirs" or "mine"--that was never clear to me.  If you're rebasing, you have changes on top of whatever `package-lock.json` is now in your target branch--and it's the `package-lock.json` from your target branch.

boneskull:
(Seems to work well, BTW)

PR #17878, opened 2022-12-05
feat(appium): add easy access to a few more drivers and plugins

jlipps:
these drivers and plugins are being used in the wild and actively maintained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,40 +69,10 @@ appium driver update xcuitest
 appium driver uninstall xcuitest
 ```
 
-#### Drivers Maintained By The Appium Team
-
-These drivers are supported by Appium core members. Please report any issues
-or suggestions regarding them to the corresponding GitHub issue tracker.
-Refer to the corresponding driver Readme to know more on how to use it and its requirements.
-
-Name | Description
---- | ---
-[appium-xcuitest-driver](https://github.com/appium/appium-xcuitest-driver) | Allows automation of iOS native and web apps (via hybrid mode). Also supports tvOS. This driver is based on Apple's [XCTest framework](https://developer.apple.com/documentation/xctest) and can only use macOS as the host system.
-[appium-uiautomator2-driver](https://github.com/appium/appium-uiautomator2-driver) | Allows automation of Android native and web apps (via hybrid mode). The driver is based on Google's [UiAutomator](https://developer.android.com/training/testing/other-components/ui-automator) framework.
-[appium-espresso-driver](https://github.com/appium/appium-espresso-driver) | Allows automation of Android native apps. The driver is based on Google's [Espresso](https://developer.android.com/training/testing/espresso) framework.
-[appium-mac2-driver](https://github.com/appium/appium-mac2-driver) | Allows automation of macOS native apps. Also supports tvOS. This driver is based on Apple [XCTest framework](https://developer.apple.com/documentation/xctest) and can only use macOS as the host system.
-[appium-windows-driver](https://github.com/appium/appium-windows-driver) | Allows automation of Windows native and UWP apps. This driver is based on Microsoft's [WinAppDriver](https://github.com/microsoft/WinAppDriver) implementation and can only use Windows 10 as the host system.
-[appium-gecko-driver](https://github.com/appium/appium-geckodriver) | Allows automation of Gecko-engine-based browsers, like Firefox on mobile and desktop platforms. Uses [geckodriver](https://github.com/mozilla/geckodriver/releases) command line tool provided by Mozilla for various platforms.
-[appium-safari-driver](https://github.com/appium/appium-safari-driver) | Allows automation of Safari browser on mobile and desktop platforms . Uses the [safaridriver](https://www.manpagez.com/man/1/safaridriver/) command line tool provided by Apple as part of macOS.
-
-#### Drivers Provided By Third Parties
-
-These drivers are supported by different companies or communities.
-Please report any issues related to their functionality to their
-respective maintainers. If you are a developer yourself and have created a new useful
-Appium2-compatible driver then don't hesitate to create a PR and add a link to
-your driver into the below list.
-
-<details>
-  <summary>Drivers List</summary>
-
-  Name | Description
-  --- | ---
-  [appium-youiengine-driver](https://github.com/YOU-i-Labs/appium-youiengine-driver) | Appium You.i Engine Driver is a test automation tool for devices of various platforms running applications built with [You.i Engine](http://www.youi.tv/youi-engine/)
-  [appium-flutter-driver](https://github.com/appium-userland/appium-flutter-driver) | Allows automation of apps built using [Flutter](https://flutter.dev/) framework
-  [appium-tizen-driver](https://github.com/Samsung/appium-tizen-driver) | Allows automation of apps built for devices running [Tizen OS](https://www.tizen.org/)
-  [appium-linux-driver](https://github.com/fantonglang/appium-linux-driver) | Allows automation of Linux applications using [AtSpi2](https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/) framework
-</details>
+You can find a full list of
+[officially-supported](https://appium.github.io/appium/docs/en/latest/ecosystem/#drivers) and
+[third-party](https://appium.github.io/appium/docs/en/latest/ecosystem/#other-drivers) drivers at
+the current Appium 2.0 documentation.
 
 ### Plugins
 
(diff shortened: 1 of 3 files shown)


packages/appium/docs/en/ecosystem/index.md
```diff
@@ -53,3 +56,15 @@ the Appium client depends on) since that is what you will use as your primary in
...
+
+Appium Plugins offer ways to extend or modify Appium's behaviour. This is the list of plugins that
+are currently recognized by Appium:
+
+|Plugin|Installation Key|Description|Support|
+|---|---|---|---|
+|[Images](https://github.com/appium/appium/tree/master/packages/images-plugin)|`images`|Image matching and comparison features|Appium Team|
+|[Execute Driver](https://github.com/appium/appium/tree/master/packages/execute-driver-plugin)|`execute-driver`|Run entire batches of commands in a single call to the Appium server|Appium Team|
+|[Relaxed Caps](https://github.com/appium/appium/tree/master/packages/relaxed-caps-plugin)|`relaxed-caps`|Relax Appium's requirement for vendor prefixes on capabilities|Appium Team|
+|[AltUnity](https://github.com/headspinio/appium-altunity-plugin)|`altunity`|Target Unity games and apps for automation with a new context, via the AltUnityTester framework|HeadSpin|
```
KazuCocoa, line 70: https://github.com/appium/appium#plugins-maintained-by-the-appium-team
[universal-xml](https://github.com/appium/appium/tree/master/packages/universal-xml-plugin) as well?

KazuCocoa (approved):

KazuCocoa:
Can we also address readers to add their plugins/drivers here?
I mean not only us but also users can create a PR to add their drivers/plugins here

boneskull (approved):

packages/appium/lib/constants.js
```diff
@@ -49,6 +50,9 @@ export const KNOWN_DRIVERS = Object.freeze(
     flutter: 'appium-flutter-driver',
     safari: 'appium-safari-driver',
     gecko: 'appium-geckodriver',
+    tizentv: 'appium-tizen-tv-driver',
```
mykola-mokhnach, line 53: I am not quite sure it would be fair to have these shortcuts as other developers could add their own drivers for the same platform. My proposal is to only keep shortcuts for officially supported plugins/drivers

packages/appium/docs/en/ecosystem/index.md
```diff
@@ -53,3 +56,15 @@ the Appium client depends on) since that is what you will use as your primary in
 
 In general, any W3C WebDriver spec-compatible client will also integrate well with Appium, though
 some Appium-specific commands may not be implemented in other clients.
+
+## Plugins
```
mykola-mokhnach, line 60: not sure if we need to duplicate these lists here as we already have them in readme

jlipps: this is for the documentation. the readme won't be part of the documentation. when we publish the docs, we should link to these pages in the readme instead.

packages/appium/lib/constants.js
mykola-mokhnach: what then if the particular developers stops supporting a driver or an extension? It is easy to add something but hard to remove.

Yes, I agree to remove tizen and youi shortcuts

packages/appium/docs/en/ecosystem/index.md
mykola-mokhnach: my point is we should not have duplicates

packages/appium/lib/constants.js
jlipps: i understand the point about downgrading official recognition as well. @boneskull @KazuCocoa what do you think?

KazuCocoa: I'd like to keep maintained drivers/plugins as the central place here as my primary thought.
Then, maybe we can print a link to the drivers/plugins list as the output of `driver list`/`plugin list` command and keep 3rd party drivers/plugins in the documentation? Something like...:

```
$ appium driver list
✔ Listing available drivers
- xcuitest@4.13.1 [installed (NPM)]
- uiautomator2 [not installed]
- youiengine [not installed]
- windows [not installed]
- mac [not installed]
- mac2 [not installed]
- espresso [not installed]
- tizen [not installed]
- flutter [not installed]
- safari [not installed]
- gecko [not installed]
More drivers are 'url'
```

KazuCocoa: btw, `mac` also can be removed since we mostly can maintain `mac2` for newer macOS (10.15+)

jlipps:
ok @mykola-mokhnach what do you think about this revision? i've merged everything from the readme into the docs, and made sure that shortcut keys are only used for appium-maintained extensions.

README.md
```diff
@@ -69,40 +69,9 @@ appium driver update xcuitest
...
-  <summary>Drivers List</summary>
-
-  Name | Description
-  --- | ---
-  [appium-youiengine-driver](https://github.com/YOU-i-Labs/appium-youiengine-driver) | Appium You.i Engine Driver is a test automation tool for devices of various platforms running applications built with [You.i Engine](http://www.youi.tv/youi-engine/)
-  [appium-flutter-driver](https://github.com/appium-userland/appium-flutter-driver) | Allows automation of apps built using [Flutter](https://flutter.dev/) framework
-  [appium-tizen-driver](https://github.com/Samsung/appium-tizen-driver) | Allows automation of apps built for devices running [Tizen OS](https://www.tizen.org/)
-  [appium-linux-driver](https://github.com/fantonglang/appium-linux-driver) | Allows automation of Linux applications using [AtSpi2](https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/) framework
-</details>
+You can find a [full list of Appium
```
mykola-mokhnach, line 72: I think it would make sense to mention both official and third party drivers separately here

packages/appium/lib/constants.js
```diff
@@ -40,13 +41,8 @@ export const KNOWN_DRIVERS = Object.freeze(
   /** @type {const} */ ({
     uiautomator2: 'appium-uiautomator2-driver',
     xcuitest: 'appium-xcuitest-driver',
-    youiengine: 'appium-youiengine-driver',
-    windows: 'appium-windows-driver',
```
mykola-mokhnach, line 44: we actually officially support the driver itself, just not the server part of it. It's complicated %)

packages/appium/docs/en/ecosystem/index.md
```diff
@@ -13,36 +13,52 @@ learn more about the inspector here: [Appium Inspector](https://github.com/appiu
...
 
 And of course, you can install any other drivers you find out there by using the Appium driver CLI.
 
+### Other drivers
+
+These drivers are not maintained by the Appium team and can be used to target additional platforms.
+
+|Driver|Installation Key|Platform(s)|Mode(s)|Supported By|
+|--|--|--|--|--|
+|[Flutter](https://github.com/appium-userland/appium-flutter-driver)|`--source=npm appium-flutter-driver`|iOS, Android|Native|Community|
```
mykola-mokhnach, line 40: Yeah, I think it would be fair to sort the list alphabetically

packages/appium/lib/constants.js
jlipps: i'm fine to put it back, but how optimistic are we that this support will continue to make sense? the "driver" from a user perspective is not just the node.js part but the whole package. what are we trying to say about that?

packages/appium/docs/en/ecosystem/index.md
jlipps: yep i figured we should start sorting them all by alpha

README.md
```diff
@@ -69,40 +69,9 @@ appium driver update xcuitest
...
-  <summary>Drivers List</summary>
-
-  Name | Description
-  --- | ---
-  [appium-youiengine-driver](https://github.com/YOU-i-Labs/appium-youiengine-driver) | Appium You.i Engine Driver is a test automation tool for devices of various platforms running applications built with [You.i Engine](http://www.youi.tv/youi-engine/)
-  [appium-flutter-driver](https://github.com/appium-userland/appium-flutter-driver) | Allows automation of apps built using [Flutter](https://flutter.dev/) framework
-  [appium-tizen-driver](https://github.com/Samsung/appium-tizen-driver) | Allows automation of apps built for devices running [Tizen OS](https://www.tizen.org/)
-  [appium-linux-driver](https://github.com/fantonglang/appium-linux-driver) | Allows automation of Linux applications using [AtSpi2](https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/) framework
-</details>
+You can find a [full list of officially-supported and third-party
```
mykola-mokhnach, line 72: I mean we could explicitly mention appropriate paragraphs

mykola-mokhnach (approved):

packages/appium/lib/constants.js
mykola-mokhnach: I don't say anything. I assume M$ should ;)
According the current state of things we can remove it

PR #17891, opened 2022-12-07
chore: run e2e tests in parallel

boneskull:
This change makes Lerna execute E2E tests across two cores.  Some of the slower E2E suites (`appium`, `support`, `opencv` and `base-driver`) also run in parallel via Mocha's `--parallel` flag.  Since Mocha will use all available cores, limiting Lerna to 2 means that Mocha can do more concurrent work.

In Appium, the main culprit was `cli.e2e.spec.js`, which I split into four files: `args` (argument handling), `cli-driver` (`driver` command), `cli-plugin` (`plugin` command) and `local` (behavior when Appium is a dependency).  This should run faster (in theory) since Mocha runs tests in parallel _per file_ and will expand to the available cores.

CI has limited cores, so this won't really help there--but it will speed up local development.  When running tests locally, you may see various delays, as Mocha buffers the output of a single test file.

This is a continuation of the work I did in #17252.


jlipps:
not sure that our e2e tests are designed to be run in parallel, like they might conflict with open ports and things like that.

boneskull:
@jlipps I'm _pretty_ sure ports won't be an issue, because we've long since eliminated hard-coded ports.  `APPIUM_HOME` _shouldn't_ be an issue, because we're using temp dirs, but there may be a place where they aren't.

boneskull:
* master
  * **PR #17891** <a href="https://app.graphite.dev/github/pr/appium/appium/17891" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/17891?utm_source=stack-comment).

jlipps (approved):
if it works, ship it!

PR #17897, opened 2022-12-08
docs(appium): update docs and reference integration

jlipps:
I didn't like how our base mkdocs config was referencing certain paths. Better for the Appium docs generator to pull in what it needs than to infect everything that inherits from our base config with knowledge about the monorepo structure. This frees us to have one central typedoc process, but each package in the monorepo can pull from it what it needs.

Made a couple other random fixes in here too.


.husky/commit-msg
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-commitlint --edit ${1}
+npx commitlint --edit ${1}
```
boneskull, line 4: What's this for?

packages/appium/docs/scripts/build-reference.js
```diff
@@ -0,0 +1,33 @@
+const {fs} = require('@appium/support');
+const {log, LANGS, DOCS_DIR} = require('./utils');
+const path = require('path');
+const monorepoRoot = path.resolve(__dirname, '..', '..', '..', '..');
+const typedocOut = path.resolve(monorepoRoot, 'typedoc-docs');
```
boneskull, line 5: probably more robust to grab this from the typedoc config file 

```suggestion
const {out: typedocOut} = JSON.parse(fs.readFileSync(path.join(monorepoRoot, 'typedoc.json')));
```

packages/appium/docs/scripts/build-reference.js
```diff
@@ -0,0 +1,33 @@
...
+      log.info(`Copying ${sourcePath} to ${destPath}`);
+      await fs.copyFile(sourcePath, destPath);
+    }
+  }
+}
+
+if (require.main === module) {
+  main().catch((err) => {
+    throw err;
+  });
```
boneskull, line 30: I'm not sure what the intent here is, since presumably it's catching to avoid an unhandled rejection, but then re-throwing to create...an unhandled rejection

```suggestion
  main().catch((err) => {
    console.error(err);
    process.exitCode = 1;
  });
```

packages/appium/package.json
```diff
@@ -41,9 +41,10 @@
     "types"
   ],
   "scripts": {
-    "build:docs": "node ./scripts/parse-yml-commands.js",
+    "build:docs": "node docs/scripts/build-docs.js",
```
boneskull, line 44: ok, so I should probably close #17909?

packages/appium/docs/scripts/build-reference.js
```diff
@@ -0,0 +1,33 @@
+const {fs} = require('@appium/support');
+const {log, LANGS, DOCS_DIR} = require('./utils');
+const path = require('path');
+const monorepoRoot = path.resolve(__dirname, '..', '..', '..', '..');
+const typedocOut = path.resolve(monorepoRoot, 'typedoc-docs');
+const {exec} = require('teen_process');
+
+const typedocMap = [[['commands', 'appium_base_driver.md'], ['base_driver.md']]];
```
boneskull, line 8: I think this is OK for now, but also we can likely implement more granular control over file names so this is not needed.

...but I'm also unsure why we wouldn't just configure typedoc to put the file where we want it?

boneskull (commented):
you'll want to remove the `markdown_include` plugin from `packages/docutils/requirements.txt` then

.husky/commit-msg
jlipps: dunno. i couldn't commit without it. i don't have commitlint installed globally i guess, nor do i want to. seems like this is the way to get husky to use the commitlint installed in devdeps.

could also mess with PATH in this script but `npx` seemed easier.

packages/appium/docs/scripts/build-reference.js
jlipps: if we can do that, that's preferable! seemed like there was only one `out` option for typedoc, and we are pulling stuff from multiple packages into that for convenience. but if we can put e.g. the fake_driver.md into the fakedriver docs, and base_driver.md into the appium docs, then we should just do that, which would simplify this script a whole lot

packages/appium/package.json
jlipps: correct. the parse-yml-commands thing is obsolete (not sure why it got put back in your earlier commit), but `build:docs` and `publish:docs` are in active use.

packages/appium/docs/scripts/build-reference.js
jlipps: yes i'm not sure what i was attempting to do haha

packages/appium/docs/scripts/build-reference.js
```diff
@@ -0,0 +1,34 @@
...
+
+  for (const lang of LANGS) {
+    const langRefPath = path.resolve(DOCS_DIR, lang, 'reference');
+    await fs.rimraf(langRefPath);
+    await fs.mkdirp(langRefPath);
+    for (const [from, to] of typedocMap) {
+      const sourcePath = path.resolve(typedocOut, ...from);
+      const destPath = path.resolve(langRefPath, ...to);
+      log.info(`Copying ${sourcePath} to ${destPath}`);
+      await fs.copyFile(sourcePath, destPath);
```
mykola-mokhnach, line 22: I assume doing the copy in parallel would be much faster
Also, do we only have files there and not folders? Otherwise we probably need something like https://github.com/appium/appium-ios-simulator/blob/5725bf7c352112e132c227ccd7839e4b2fd56db5/test/utils.js#L4

jlipps: we're copying a small number of specific files, currently just one :-)

mykola-mokhnach (approved):

boneskull:
fwiw the commitlint CI job was broken as-is; it seemed to choke on the `Co-Authored-By:` in the commit body when run the way we were doing it.  trying an action now

boneskull:
OK, idk how to make it that job work properly without more investigation. I've removed it for now.

PR #17909, opened 2022-12-09
chore(appium): remove more doc-related scripts

boneskull:
Removes `build:docs` and `publish:docs` (which called `build:docs`) from `appium` scripts.  I don't know if `build:docs:assets` & `build:docs:cli` also need to be removed or not?  I think we _are_ using `dev:docs:*`.

### DIFF
--- a/packages/appium/package.json
+++ b/packages/appium/package.json
@@ -41,7 +41,6 @@
     "types"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


jlipps:
per conversation elsewhere, we _are_ using these, though one needs restoring (in linked PR)

PR #17936, opened 2022-12-14
fix: add 'webSocketUrl' as standard cap for bidi support

jlipps:
cc @christian-bromann this should be added to the w3c spec as a standard cap once bidi is ratified, yes? i think we'll need to add it already for appium to support bidi well.

### DIFF
--- a/packages/base-driver/lib/basedriver/capabilities.js
+++ b/packages/base-driver/lib/basedriver/capabilities.js
@@ -104,6 +104,7 @@ export const STANDARD_CAPS = Object.freeze(
       'setWindowRect',
       'timeouts',
       'unhandledPromptBehavior',
+      'webSocketUrl',
     ])
   )
 );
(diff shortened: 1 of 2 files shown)


christian-bromann:
I had a conversation recently with @whimboo about `webSocketUrl` being a bit ambiguous: boolean in requested capabilities and string in received capabilities. I couldn't find the resolution to this conversation anymore but this is how it is currently defined in the spec.

boneskull (changes requested):
@jlipps I think `webSocketUrl` needs to be added to `@appium/types/lib/constraints.js`, then LGTM

jlipps:
@christian-bromann agree that it is not as clear a name as something like `bidiEnabled` or whatever

jlipps:
@boneskull updated

boneskull:
...that build failure looks unrelated.  did you pull in the "e2e tests run in parallel" changeset?  if so, we may want to revert it.

boneskull:
(it may also be a network burp)

jlipps:
This is on top of almost latest master, so probably. I'm rerunning the failed jobs.

whimboo:
> I had a conversation recently with @whimboo about `webSocketUrl` being a bit ambiguous: boolean in requested capabilities and string in received capabilities.

Yes, this is right.

jlipps:
@boneskull how come I see extraneous changes in this PR? are these changes not actually in master yet?

boneskull:
@jlipps I clicked the update button, which created a merge commit, unfortunately. Just rebase onto master to get rid of it.

boneskull:
I rebased it.  looks like the "Update & Rebase" is in the dropdown and the default behavior is to merge.

jlipps:
I'm still confused why I see changes that I didn't make. Is it prettier-related? At any rate, I think we can merge this.

PR #17943, opened 2022-12-15
chore: remove @appium/gulp-plugins

### LINKED ISSUE #17907: determine if @appium/gulp-plugins is used by anything
1. Go to https://libraries.io/npm/@appium%2Fgulp-plugins/dependents
2. For each pkg in this list, click through and look at the development dependencies in the right-hand column near the bottom.  
a. If no packages depend on it, remove `@appium/gulp-plugins` from the monorepo.  I'm not sure where to put it, since it's in a monorepo, but it may make sense to just create a new repo with the source and throw it into the boneyard (for posterity) _or_ just delete it wholesale and direct anyone who asks to the Git history.
b. If packages depend on it, then modify those packages to stop depending on it.  Repeat until a) is true

cc @mykola-mokhnach

boneskull:
Closes #17907


boneskull:
I did an `npm deprecate` on `@appium/gulp-plugins` too

jlipps (approved):
sweet! LGTM when ci is happy

PR #17948, opened 2022-12-15
fix broken manifest migration persistence

### LINKED ISSUE #17901: APPIUM_HOME resolution seems weird if not broken

### Current Behavior

Fairly recently, Appium was able to detect the presence of `@appium/fake-driver` when running `appium` from the monorepo root.  I believe this stopped happening when we 

Whether that's considered the "correct" behavior or not (in retrospect, I think this is incorrect) is debatable.  So I don't _think_ this is a bug as long as `appium` still knows it's running in an extension working copy when in e.g., `appium-xcuitest-driver`.

But anyway:  _something_ is fishy, because it's doing the resolution dance twice:

```bash
appium --log-level=debug

dbug Support Determined package/workspace root from /Users/boneskull/projects/appium/appium => /Users/boneskull/projects/appium/appium
dbug Support No package.json having current Appium dep in /Users/boneskull/projects/appium/appium
dbug Support Using default APPIUM_HOME: /Users/boneskull/.appium
dbug Support Determined package/workspace root from /Users/boneskull/.appium => /Users/boneskull/.appium
dbug Support No package.json having current Appium dep in /Users/boneskull/.appium
```

### Expected Behavior

I should only see `Determined package/workspace root from <somewhere> => <somewhere-else>` _once_.  

Run `npm start -- --log-level=debug` in a working copy of `appium` at `master`.

### Environment

yes

### Link to Appium Logs

_No response_

### Futher Information

This is mainly a concern because it negatively impacts first load; it's an expensive operation.  

We should also assert that `appium` detects the extension-in-development when run from an extension working copy.

boneskull:
- Manifest migrations were not always applied to disk.  This fixes that problem and adds a test.
- I confirmed that #17901 was otherwise a non-issue and verified it works as expected within the context of an external driver's working copy.

Closes #17901.

### DIFF
--- a/packages/appium/lib/extension/manifest-migrations.js
+++ b/packages/appium/lib/extension/manifest-migrations.js
@@ -26,7 +26,7 @@ const Migrations = {
    */
   [SCHEMA_REV_3]: (data) => {
     let shouldSync = false;
-    const allExtData = [...Object.values(data.drivers), ...Object.values(data.plugins)];
+    const allExtData = [...Object.values(data.drivers || {}), ...Object.values(data.plugins || {})];
     for (const metadata of allExtData) {
       if (!('installPath' in metadata)) {
         shouldSync = true;
--- a/packages/appium/test/e2e/manifest.e2e.spec.js
+++ b/packages/appium/test/e2e/manifest.e2e.spec.js
@@ -1,0 +1,61 @@
+import {resolveFixture} from '../helpers';
+import YAML from 'yaml';
+import {runAppiumJson} from './e2e-helpers';
+import {fs, tempDir} from '@appium/support';
+import {
+  CACHE_DIR_RELATIVE_PATH,
+  CURRENT_SCHEMA_REV,
+  DRIVER_TYPE,
+  EXT_SUBCOMMAND_LIST as LIST,
+} from '../../lib/constants';
+import path from 'node:path';
+
+const {expect} = chai;
+
+describe('manifest handling', function () {
+  /**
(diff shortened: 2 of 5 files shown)


packages/appium/lib/extension/manifest-migrations.js
```diff
@@ -26,7 +26,7 @@ const Migrations = {
    */
   [SCHEMA_REV_3]: (data) => {
     let shouldSync = false;
-    const allExtData = [...Object.values(data.drivers), ...Object.values(data.plugins)];
+    const allExtData = [...Object.values(data.drivers || {}), ...Object.values(data.plugins || {})];
```
boneskull, line 29: while these shouldn't ever be be `null` or `undefined` in practice, I guess it's possible.

packages/appium/test/e2e/manifest.e2e.spec.js
```diff
@@ -0,0 +1,61 @@
...
+
+  describe('migration', function () {
+    beforeEach(async function () {
+      await resetAppiumHome();
+      await fs.mkdirp(path.dirname(manifestPath));
+      await fs.copyFile(resolveFixture('manifest/v2-empty.yaml'), manifestPath);
+    });
+
+    it('should update the manifest file to the latest schema revision', async function () {
+      await runList();
```
boneskull, line 56: this triggers the migration. could have been any other command I guess

jlipps (approved):
nice, i think this explains why I kept seeing migration messages

PR #17953, opened 2022-12-16
feat(appium): cull invalid manifest data

boneskull:
If, for some reason, an extension doesn't exist on disk yet does exist in the manifest, it probably shouldn't be in the manifest, right? Anyway:

This change causes invalid entries to be culled in the same operation as "sync with installed packages". Those situations are:

- if the manifest file is new
- if the manifest needs a migration
- if `appium` is a local dep and its `package.json` has changed

If none of these are true, synchronization won't happen (extensions won't be automatically discovered), and entries won't be culled.

An argument could be made that we want to sync and/or cull invalid entries regardless of whether we're using a local `appium` or `APPIUM_HOME`, and likewise on any extension subcommand which can modify the manifest (install, uninstall, update). While good housekeeping is in order, it's not free--the system must inspect _every `package.json` in `node_modules`_ to find extensions.

In fact, I'm thinking this should be _opt-in_ behavior if `appium` is a locally-installed dependency; the sync would happen any time `package.json` is modified on-disk.  This might be something like a flag `--sync` flag for `appium driver/plugin list`.  Anyway: that can be a separate PR.

### DIFF
--- a/packages/appium/lib/extension/manifest.js
+++ b/packages/appium/lib/extension/manifest.js
@@ -4,7 +4,7 @@
 
 import B from 'bluebird';
 import glob from 'glob';
-import {env, fs} from '@appium/support';
+import {env, fs, util} from '@appium/support';
 import _ from 'lodash';
 import path from 'path';
 import YAML from 'yaml';
@@ -29,7 +29,7 @@ const CONFIG_DATA_PLUGIN_KEY = `${PLUGIN_TYPE}s`;
 /**
  * @type {Readonly<ManifestData>}
  */
-const INITIAL_MANIFEST_DATA = Object.freeze({
+export const INITIAL_MANIFEST_DATA = Object.freeze({
   [CONFIG_DATA_DRIVER_KEY]: Object.freeze({}),
   [CONFIG_DATA_PLUGIN_KEY]: Object.freeze({}),
   schemaRev: CURRENT_SCHEMA_REV,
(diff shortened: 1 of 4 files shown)


boneskull:
Actually.. I think I want to split this up into bug and feature.  I am more optimistic about the bug fix and less optimistic about the feature, which probably needs more tests to ensure we don't e.g., blow away the manifest and re-create it every time appium runs.

boneskull:
* master
  * **PR #17954** <a href="https://app.graphite.dev/github/pr/appium/appium/17954" target="_blank">Graphite</a> 
    * **PR #17953** <a href="https://app.graphite.dev/github/pr/appium/appium/17953" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/17953?utm_source=stack-comment).

boneskull:
Note: the bugfix has been merged.  This is just a feature.

packages/appium/lib/extension/manifest.js
```diff
@@ -156,51 +156,117 @@ export class Manifest {
...
+            if (isDriver(pkg)) {
+              missingDrivers.delete(pkg.name);
+            } else {
+              missingPlugins.delete(pkg.name);
+            }
+          }
         }
-      } catch {}
+      } catch (err) {
+        // only eat ENOENT errors in which `package.json` is missing.
```
jlipps, line 230: to me this looks like we are eating all errors, but choosing to log non-ENOENT errors. the idea is that we never want this to fail in any situation?

boneskull: right. if there's no package.json we don't really care.  if the error is something else it may be interesting

boneskull: but not interesting enough to abort operation--a warning is better than nothing. I don't consider that "eating"

packages/appium/lib/extension/manifest.js
```diff
@@ -218,16 +284,56 @@ export class Manifest {
...
+   */
+  #removeMissingExtensions(extType, missingExtMap) {
+    if (missingExtMap.size) {
+      log.warn(
+        `Removing ${util.pluralize('reference', missingExtMap.size)} to ${util.pluralize(
+          extType,
+          missingExtMap.size,
+          true
+        )} in the manifest that could not be found on disk: ${_.map(
+          [...missingExtMap.values()],
```
mykola-mokhnach, line 325: you could also use `Array.from` if you find it more readable

packages/appium/lib/extension/manifest.js
```diff
@@ -218,16 +284,56 @@ export class Manifest {
...
+          'name'
+        ).join(', ')}; `
+      );
+      for (const {name} of missingExtMap.values()) {
+        delete this.#data[`${extType}s`][name];
+      }
+      return true;
+    }
+    return false;
```
mykola-mokhnach, line 334: I would rather flip the condition above

packages/appium/lib/extension/manifest.js
jlipps: ok, i guess i did. just want to make sure the comment reflects your intentions

boneskull:
This should wait for Appium v2.1.0.

PR #17966, opened 2022-12-20
fix(appium): inability to find automationName inside appium:options

jlipps:
we fixed this before but only fixed it in basedriver's createSession, which is not actually called by Appium's createSession (though maybe it should be)


jlipps:
also removed the appium:options promotion and test from basedriver since really basedriver should not know about it. it's an appium server feature, not part of normal w3c capabilities handling

boneskull (approved):

PR #17985, opened 2022-12-27
docs: add ExternalDriver docstrings

jlipps:


jlipps:
@boneskull is this the kind of thing you had in mind for me to do?

boneskull (approved):
yes

boneskull:
It occurs to me that the methods return `X` but the API commands return `{"value": X}`.  Is there any command that breaks that assumption?  I can likely account for this.

jlipps:
the command methods should return normal javascript values. a wrapper function converts those into actual protocol responses. no command method should know anything about how its response is wrapped up as a protocol response. it should only return the javascript value representing the result of its operation.

i.e., i don't think we want to annotate anything like `{value: 'foo'}` anywhere in our docs. clients always unwrap that into language-specific objects anyway

mykola-mokhnach (approved):
nicely done

boneskull:
lgtm

PR #18016, opened 2023-01-05
fix(base-driver): add missing dependency @appium/types

### LINKED ISSUE #18006: bug: Appium 2 fails to install drivers

### Current Behavior

When I try to install a driver (any driver), it fails with the following:

```
$ appium driver install xcuitest
✖ Installing 'xcuitest' using NPM install spec 'appium-xcuitest-driver'
Error: ✖ Encountered an error when installing package: npm command 'install --save-dev --omit=peer --save-exact --global-style --no-package-lock appium-xcuitest-driver --json' failed with code null.

STDOUT:


STDERR:
```

### Expected Behavior

It works, or at least provides me with some output.

N/A

### Environment

- Operating system: macos
- If running via `appium` CLI...
  - Appium CLI version (output of `appium --version`): 2.0.0-beta.48
  - Node.js version (output of `node --version`): 16.16
  - `npm` version (output of `npm --version`):  8.11 / Yarn 3
  - Last `appium` version which did _not_ exhibit the problem: N/A
- If running Appium Desktop, its version:
- Appium plugin(s) and their version(s):
- Platform and version under test:
- Real device or emulator/simulator: simulator
 

### Link to Appium Logs

_No response_

### Futher Information

_No response_

boneskull:
Closes #18006

I hope

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16844,6 +16844,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@appium/support": "^3.0.1",
+        "@appium/types": "^0.7.0",
         "@colors/colors": "1.5.0",
         "@types/async-lock": "1.3.0",
         "@types/bluebird": "3.5.38",
--- a/packages/base-driver/package.json
+++ b/packages/base-driver/package.json
@@ -43,6 +43,7 @@
   },
   "dependencies": {
     "@appium/support": "^3.0.1",
+    "@appium/types": "^0.7.0",
     "@colors/colors": "1.5.0",
     "@types/async-lock": "1.3.0",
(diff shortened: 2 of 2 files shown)


jlipps:
interesting. so `@appium/types` is a prod dep of basedriver? *goes and reads the code* ah yes, the base cap constraints are part of the types package. not entirely intuitive but makes sense.

jlipps (approved):

boneskull:
> interesting. so @appium/types is a prod dep of basedriver? goes and reads the code ah yes, the base cap constraints are part of the types package. not entirely intuitive but makes sense.

yeah, I don't really love it, but the types package needs that information, or we get a dependency cycle.

PR #18037, opened 2023-01-10
feat(base-driver): deprecate non-standard routes

jlipps:
and add deprecation warnings in the log when these routes are accessed

### DIFF
--- a/packages/base-driver/lib/protocol/protocol.js
+++ b/packages/base-driver/lib/protocol/protocol.js
@@ -18,6 +18,9 @@ const CREATE_SESSION_COMMAND = 'createSession';
 const DELETE_SESSION_COMMAND = 'deleteSession';
 const GET_STATUS_COMMAND = 'getStatus';
 
+/** type {Set<string>} */
+const deprecatedCommandsLogged = new Set();
+
 function determineProtocol(createSessionArgs) {
   return _.some(createSessionArgs, isW3cCaps) ? PROTOCOLS.W3C : PROTOCOLS.MJSONWP;
 }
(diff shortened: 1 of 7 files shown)


jlipps:
The idea here is that we are deprecating routes in Appium that are old or used for only one or two drivers. The route definitions supported by those drivers should be migrated to the `newMethodMap` of those drivers (for example, the `startActivity` command should be migrated to the android drivers)

packages/base-driver/lib/protocol/protocol.js
```diff
@@ -262,6 +262,15 @@ function buildHandler(app, method, path, spec, driver, isSessCmd) {
     let currentProtocol = extractProtocol(driver, req.params.sessionId);
 
     try {
+      // if the route accessed is deprecated, log a warning
+      if (spec.deprecated) {
+        getLogger(driver, req.params.sessionId).warn(
+          `Command '${spec.command}' has been deprecated and will be removed in a future ` +
```
mykola-mokhnach, line 268: not sure if the message should be shorter. A long message might seriously increase the log size if the particuluar command is called multiple times.

mykola-mokhnach:
I would rather start with updating drivers first, at least officially supported ones, before we merge this server change

jlipps:
I can work on updating the drivers, though it would be nice to have some help. I'll begin with the uiautomator2 driver as i imagine it will be the most complex with the dependency on the android driver package. once it's clear what the migration path is maybe we can all pitch in to update the other drivers?

mykola-mokhnach:
I can for sure assist with that if I have at least one example.

packages/base-driver/lib/protocol/protocol.js
boneskull: Agree.  One warning per route.  Probably just keep a `Set<string>` of routes that have had this warning emitted and check it.

boneskull (approved):
see comment

jlipps:
@mykola-mokhnach OK i submitted PRs for android driver, uia2 driver, espresso driver, and xcui driver. turned out to be pretty easy with a script i wrote to print out exactly what newMethodMap should be.

- https://github.com/appium/appium-android-driver/pull/782
- https://github.com/appium/appium-uiautomator2-driver/pull/574
- https://github.com/appium/appium-espresso-driver/pull/846
- https://github.com/appium/appium-xcuitest-driver/pull/1487

mykola-mokhnach (approved):

jlipps:
will wait to publish this until the driver updates have been released

PR #18042, opened 2023-01-11
docs(appium): note the breaking change in default base path

jlipps:
cc @appium/documentors


boneskull (approved):
Please add that you can configure this via a config file too 😄

Issue #17946, opened 2022-12-15
check JS types in all monorepo workspaces


boneskull:
The following workspaces' `tsconfig.json` files are not (yet) configured to typecheck everything at build time:

- [ ] `packages/doctor`
- [ ] `packages/docutils`
- [ ] `packages/execute-driver-plugin`
- [ ] `packages/fake-plugin`
- [ ] `packages/images-plugin`
- [ ] `packages/opencv`
- [ ] `packages/relaxed-caps-plugin`
- [ ] `packages/universal-xml-plugin`

> Note: Some individual source files may have a `// @ts-check` pragma.  If _all_ JS sources in a workspace have this pragma, it can be removed and `checkJs: true` can be added to the workspace's `tsconfig.json`.

Eventually, I'd like to set `checkJs: true` in all of those configs and get everything typed.  Once all workspace `tsconfig.json` files contain `checkJs: true`, we can add `checkJs: true` to the `./config/tsconfig.base.json` and remove it from all workspace-specific configs.

This doesn't have to be done all at once, but it's good to track progress here.

Also: Don't worry about test files for now.

* * *

To get an up-to-date list, run:

```bash
❯ grep -riL '"checkJs": true' ./packages/*/tsconfig.json
./packages/doctor/tsconfig.json
./packages/docutils/tsconfig.json
./packages/execute-driver-plugin/tsconfig.json
./packages/fake-plugin/tsconfig.json
./packages/images-plugin/tsconfig.json
./packages/opencv/tsconfig.json
./packages/relaxed-caps-plugin/tsconfig.json
./packages/universal-xml-plugin/tsconfig.json
```

pacozaa:
@boneskull Question, I see `@linkcode` is used in the code but I check here at  https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#link looks like it changes to just `@link` or `@see`. Would you like me to change them as well?

boneskull:
@pacozaa `@linkcode` is supported by VSCode but apparently not documented.  

In addition to what TypeScript understands, some tags are purely for documentation, however.  See [the TypeDoc docs](https://typedoc.org/guides/overview/) for what we can use.

pacozaa:
Just a short note, some of the work in this issue are done in 
https://github.com/appium/appium/pull/17962
https://github.com/appium/appium/pull/17922

I am working on the rest of the packages.

zeufack:
Is this issues still open ? If so, I would like to work on it

PR #18059, opened 2023-01-14
refactor: improv js types

pacozaa:
## Proposed changes

I tried to type things that were mentioned in this issue https://github.com/appium/appium/issues/17946

## Types of changes

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/packages/fake-plugin/lib/plugin.js
+++ b/packages/fake-plugin/lib/plugin.js
@@ -4,9 +4,15 @@
 import {BasePlugin} from 'appium/plugin';
 import B from 'bluebird';
 
+/**
+ * An example plugin for Appium that demonstrate the implementations.
+ *
+ * @extends {BasePlugin}
+ */
 class FakePlugin extends BasePlugin {
   fakeThing = 'PLUGIN_FAKE_THING';
 
+  // Map route path to  to HTTP methods and command details
   static newMethodMap = /** @type {const} */ ({
     '/session/:sessionId/fake_data': {
       GET: {command: 'getFakeSessionData', neverProxy: true},
@@ -41,11 +47,21 @@ class FakePlugin extends BasePlugin {
     res.send(JSON.stringify({fake: 'fakeResponse'}));
   }
 
+  /**
+   * A route handler function for /unexpected
+   *
+   * @param {Object} req - The request object
+   * @param {Object} res - The response object
+   */
   static unexpectedData(req, res) {
     res.send(JSON.stringify(FakePlugin._unexpectedData));
     FakePlugin._unexpectedData = null;
   }
 
+  /**
+   *
+   * @type {import('@appium/types').UpdateServerCallback}
+   */
   // eslint-disable-next-line no-unused-vars,require-await
   static async updateServer(expressApp, httpServer, cliArgs) {
     expressApp.all('/fake', FakePlugin.fakeRoute);
@@ -75,6 +91,13 @@ class FakePlugin extends BasePlugin {
     return `<Fake>${JSON.stringify(args)}</Fake>`;
   }
 
+  /**
+   *
+   * @param {Function} next - The function to be executed after this one
+   * @param {import('@appium/types').ExternalDriver} driver - The driver instance
+   * @param {any[]} args - Additional arguments to be passed to the next function
+   * @returns {Promise<Object>}
+   */
   async findElement(next, driver, ...args) {
     this.logger.info(`Before findElement is run with args ${JSON.stringify(args)}`);
     const originalRes = await next();
@@ -83,22 +106,47 @@ class FakePlugin extends BasePlugin {
     return originalRes;
   }
 
+  /**
+   * Gets the fake session data, adding a delay.
+   *
+   * @param {Function} next - The function to be executed after this one
+   * @param {Object} driver - The driver instance. Using Object because fakesession doesn't exist in BaseDriver
+   * @returns {Promise<Object?>}
+   */
   async getFakeSessionData(next, driver) {
     await B.delay(1);
     return driver.fakeSessionData || null;
   }
 
+  /**
+   *
+   * @param {Function} next - The function to be executed after this one
+   * @param {Object} driver - The driver instance. Using Object because fakesession doesn't exist in BaseDriver
+   * @param {any[]} args - Additional arguments to be passed to the next function
+   * @returns {Promise<null>}
+   */
   async setFakeSessionData(next, driver, ...args) {
     await B.delay(1);
     driver.fakeSessionData = args[0];
     return null;
   }
 
+  /**
+   *
+   * @param {Function} next - The function to be executed after this one
+   * @returns {Promise<string>}
+   */
   async getWindowHandle(next) {
     const handle = await next();
     return `<<${handle}>>`;
   }
 
+  /**
+   *
+   * @param {import('@appium/types').ExternalDriver} driver - The driver instance
+   * @param {string} cause - The cause of the shutdown
+   * @returns {Promise<void>}
+   */
   // eslint-disable-next-line require-await
   async onUnexpectedShutdown(driver, cause) {
     FakePlugin._unexpectedData = `Session ended because ${cause}`;
@@ -111,3 +159,7 @@ class FakePlugin extends BasePlugin {
 
 export {FakePlugin};
 export default FakePlugin;
+
+/**
+ * @typedef {Record<string,unknown>} CLIArgs
+ */
--- a/packages/fake-plugin/lib/plugin.js
+++ b/packages/fake-plugin/lib/plugin.js
@@ -7,7 +7,7 @@ import B from 'bluebird';
 /**
  * An example plugin for Appium that demonstrate the implementations.
  *
- * @extends {BasePlugin}
+ * @implements {Plugin}
  */
 class FakePlugin extends BasePlugin {
   fakeThing = 'PLUGIN_FAKE_THING';
@@ -43,6 +43,12 @@ class FakePlugin extends BasePlugin {
     },
   });
 
+  /**
+   * A route handler function for /fake that sends a JSON string representing a fake response.
+   *
+   * @param {Request} req - The request object
+   * @param {Response} res - The response object
+   */
   static fakeRoute(req, res) {
     res.send(JSON.stringify({fake: 'fakeResponse'}));
   }
@@ -50,8 +56,8 @@ class FakePlugin extends BasePlugin {
   /**
    * A route handler function for /unexpected
    *
-   * @param {Object} req - The request object
-   * @param {Object} res - The response object
+   * @param {Request} req - The request object
+   * @param {Response} res - The response object
    */
   static unexpectedData(req, res) {
     res.send(JSON.stringify(FakePlugin._unexpectedData));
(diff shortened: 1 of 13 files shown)


pacozaa:
@boneskull Please review this. Thank you👍

packages/fake-plugin/lib/plugin.js
```diff
@@ -4,12 +4,22 @@
 import {BasePlugin} from 'appium/plugin';
 import B from 'bluebird';
 
+/**
+ * An example plugin for Appium that demonstrate the implementations.
+ *
+ * @extends {BasePlugin}
```
boneskull, line 10: `@extends` is _only_ necessary if the class that you're extending expects a type argument (like `BaseDriver`).

However, we _do_ want to ensure that `FakeDriver` implements `Plugin`, because for some reason we have to be explicit about it.

```suggestion
 * @implements {Plugin}
```

You'll need a `@typedef` for `Plugin` referencing `import('@appium/types').Plugin` somewhere, because we cannot use `import(..)` in an `@implements` tag (probably a bug).

packages/fake-plugin/lib/plugin.js
```diff
@@ -27,15 +37,31 @@ class FakePlugin extends BasePlugin {
   /** @type {string?} */
   static _unexpectedData = null;
 
+  /**
+   * A route handler function for /fake that sends a JSON string representing a fake response.
+   *
+   * @param {Object} req - The request object
```
boneskull, line 43: These are actually `Request` and `Response` from `express` (e.g., `import('express').Request` -- add a typedef)

packages/fake-plugin/lib/plugin.js
```diff
@@ -27,15 +37,31 @@ class FakePlugin extends BasePlugin {
...
+   * @param {Object} res - The response object
+   */
   static unexpectedData(req, res) {
     res.send(JSON.stringify(FakePlugin._unexpectedData));
     FakePlugin._unexpectedData = null;
   }
 
+  /**
+   *
+   * @type {import('@appium/types').UpdateServerCallback}
```
boneskull, line 63: I _think_ if we use `@implements {Plugin}` then we won't need to type this.  After you add the `@implements`, remove the `@type` here and see if intellisense knows what it is.

packages/fake-plugin/lib/plugin.js
```diff
@@ -45,11 +71,25 @@ class FakePlugin extends BasePlugin {
     });
   }
 
+  /**
+   *
+   * @param {Function} next - The function to be executed after this one
```
boneskull, line 76: any method in a `Plugin` meant to override a method in a `Driver` is actually a `PluginCommand` from `@appium/types`, which has the first two parameters defined for you.  `PluginCommand` looks like this:

```ts
export type PluginCommand<
  D extends ExternalDriver = ExternalDriver,
  TArgs extends readonly any[] = any[],
  TReturn = any
> = (next: NextPluginCallback, driver: D, ...args: TArgs) => Promise<TReturn>;
```

You could use `@type {PluginCommand}` here, since a) we don't care about the driver, and b) the args _really are_ `any`.

packages/fake-plugin/lib/plugin.js
```diff
@@ -58,22 +98,47 @@ class FakePlugin extends BasePlugin {
...
     await B.delay(1);
     driver.fakeSessionData = args[0];
     return null;
   }
 
+  /**
+   *
+   * @param {Function} next - The function to be executed after this one
+   * @returns {Promise<string>}
+   */
```
boneskull, line 130: This should fail the check for a `PluginCommand` because `driver` isn't present.  Given it's not absolutely necessary, maybe we should make that parameter optional in the `PluginCommand` type.

packages/images-plugin/lib/plugin.js
```diff
@@ -9,6 +9,11 @@ import {ImageElement, IMAGE_ELEMENT_PREFIX} from './image-element';
 
 const IMAGE_STRATEGY = '-image';
 
+/**
+ *
+ * @param {Array} args - An array of arguments
```
boneskull, line 14: should be `{any[]}`

packages/images-plugin/lib/plugin.js
```diff
@@ -17,13 +22,22 @@ function getImgElFromArgs(args) {
   }
 }
 
+/**
+ * A plugin for Appium that provides support for finding elements by image and comparing images.
+ *
+ * @extends {BasePlugin}
```
boneskull, line 28: again, don't need this, but _do_ need an `@implements`

packages/images-plugin/lib/plugin.js
```diff
@@ -37,21 +51,54 @@ export default class ImageElementPlugin extends BasePlugin {
...
+  /**
+   *
+   * @param {Function} next - The function to be executed after this one
+   * @param {ExtDriver} driver - The driver instance
+   * @param {string} mode
+   * @param {string} firstImage
+   * @param {string} secondImage
+   * @param {Object} [options={}]
+   * @returns {Promise<any>}
+   */
```
boneskull, line 63: since we actually expect arguments, the type of this would be something like:

```ts
PluginCommand<[string, string, string, object|undefined]>
```

I can't recall if we can create "named" tuple types, but if we can:

```ts
PluginCommand<[mode: string, firstImage: string, secondImage: string, options: object|undefined]>
```

also, `options` is probably not just any old object, and should eventually be typed.

packages/opencv/lib/index.js
```diff
@@ -650,7 +650,7 @@ async function cvMatFromImage(img) {
  * @param {Array<Point>} nonZeroMatchResults matrix of image match results
  * @param {number} matchNeighbourThreshold The pixel distance within which we
  * consider an element being a neighbour of an existing match
- * @return {Array<Point>} the filtered array of matched points
+ * @returns {Array<Point>} the filtered array of matched points
```
boneskull, line 653: 99% of the time you see `Array<Foo>` it's better expressed as `Foo[]`.

packages/relaxed-caps-plugin/lib/plugin.js
```diff
@@ -18,7 +18,16 @@ const STANDARD_CAPS = [
...
+/**
+ *
+ * @extends {BasePlugin}
+ */
 export class RelaxedCapsPlugin extends BasePlugin {
+  /**
+   *
+   * @param {Capabilities} caps
+   * @returns {Capabilities} - The transformed capability object.
+   */
```
boneskull, line 30:
```suggestion
  /**
   * @template {Constraints} C
   * @param {Capabilities<C>} caps
   * @returns {NSCapabilities<C>} - The transformed capability object.
   */
```

you'll need to `@typedef` that stuff, but yeah. this function takes _non_-namespaced `Capabilities` and returns _namespaced_ ones (`NSCapabilities`)

packages/relaxed-caps-plugin/lib/plugin.js
```diff
@@ -47,6 +56,15 @@ export class RelaxedCapsPlugin extends BasePlugin {
     return newCaps;
   }
 
+  /**
+   *
+   * @param {Function} next - The function to be executed after this one.
+   * @param {import('@appium/types').ExternalDriver} driver - The driver instance
+   * @param {W3CCapabilities} jwpDesCaps
```
boneskull, line 63: I'm not sure we actually have a type defined for the parameters here.  just use `any` for now.

basically, each one can either be `Capabilities<C>`, `NSCapabilities<C>`, or a `W3CCapabilities<C>`-ish composed of the first two.  

also if you have a `@template {Constraints} C`, then `driver` should be `ExternalDriver<C>`.

packages/universal-xml-plugin/lib/attr-map.js
```diff
@@ -1,4 +1,17 @@
...
+ * @property {Attribute} x
+ * @property {Attribute} y
+ * @property {Attribute} width
+ * @property {Attribute} height
+ * @property {Attribute} enabled
+ * @property {Attribute} axId
+ * @property {Attribute} text
+ * @property {Attribute} visible
+ * @property {Attribute} value
+ */
```
boneskull, line 14: a better way to do this is wrap the object `{..}` in `/** @type {const} */({..})`, then you won't need to define all the properties nor will `Attribute` be necessary. same for below

boneskull (commented):
Thanks again!  If you can't get one of my suggestions to work, just leave it and move on.  I haven't quite gotten thru the whole thing, but after reading my comments you may be able to apply them to all the changes. please ping me if/when you're done

packages/images-plugin/lib/plugin.js
pacozaa: @boneskull I see that `return convertVisualizationToBase64(result);` in `compareImages` function inside compare.js returns `OccurrenceResult`. Should I use `@returns {Promise<OccurrenceResult>}

packages/relaxed-caps-plugin/lib/plugin.js
pacozaa: Ah I can't understand this one at all. The knowledge goes over my head. From non-namespaced to namespaced and NSCapabilities and @template. Sorry about that. Would you mind give me a pointer where to read about it?

packages/images-plugin/lib/plugin.js
boneskull: yes, if the function is `async`, the return value must be `Promise<something>`

packages/relaxed-caps-plugin/lib/plugin.js
pacozaa: Ah it's the generic type thing

packages/fake-plugin/lib/plugin.js
pacozaa: I tried this but the IntelliSense still doesn't know what it is. So, I still keep it.

packages/images-plugin/lib/plugin.js
pacozaa: @boneskull  hmmm I see now. Should I use `@type {PluginCommand}` for function that takes`next` and `driver` as the first two arguments as an argument in plugin.js of those plugins?

pacozaa:
@boneskull  so I try to use this one.

```js
  /**
   * @type {PluginCommand<[string, string, string, object|undefined], Promise<OccurrenceResult>>}
   */
  async compareImages(next, driver, mode, firstImage, secondImage, options = {}) {
    return await compareImages(mode, firstImage, secondImage, options);
  }
```

But my vscode doesn't seem to know much about the other parameters.

```js
(method) ImageElementPlugin.compareImages(next: any, driver: any, mode: any, firstImage: any, secondImage: any, options?: {}): Promise<OccurrenceResult | OccurrenceResult[]>
```

There seems to be something missing to let the vscode knows what `next` and `driver` is

Compare to the function that I just used

```js
 /**
   * @type {PluginCommand}
   */
```

I am going to need you help on this one.

boneskull:
OccurrenceResult is not Appium. make sure you are importing the right plugincommand from @appium/types

boneskull:
wait sorry. yes it is. idk offhand. will look monday

packages/images-plugin/lib/plugin.js
boneskull: you don't _have_ to, as long as `next` and `driver` are the same type as in `PluginCommand`.

packages/relaxed-caps-plugin/lib/plugin.js
boneskull: so yeah `@template` defines a type arg for a generic type. it'll look like:

```js
/**
 * @template X
 * @param {X[]} x
 */
function foo(x) {}
```

The above is equivalent to:

```js
/**
 * @template {any} X
 * @param {X[]} x
 */
function foo(x) {}
```

...and the equivalent in TS:

```ts
function foo<X>(x: X[]) {}
```

or

```ts
function foo<X extends any>(x: X[]) {}
```

So you can see that in the `@template`, `{something}` means "extends".

The other thing to know about is defaults:

```js
/**
 * @template {Bar} [X=Baz]
 * @typedef {X[]} Foo
 */
```

I'm not going to get too deep into it, but we're basically saying "if the type argument `X` is not provided, _assume_ it is a `Baz`.  Because the constraint is type `Bar`, for this to compile, type `Baz` _must_ extend `Bar` so it satisfies the constraint.  e.g., `Baz` is something like:

```ts
interface Baz extends Bar {}
```

For our purposes the important thing to know is that if you see a type like `ExternalDriver` _without type arguments_, then it is using the "default" value for the type argument.  In this case, `ExternalDriver<C>` wants a `Constraints` `C`, but if we do not provide it, it just uses the constraints from the base driver (`BaseDriverCapConstraints` iirc).

Regarding `Capabilities`, `NSCapabilities`, `W3CCapabilities` -- these are basically just three different ways of expressing the same data.  `W3CCapabilities` is what an end user provides to Appium, and its shape is defined by the W3C standards body.   Both `Capabilities` and `NSCapabilities` are used mostly or entirely internally -- the "relaxed caps" plugin is the only place (to my knowledge) that an end-user provides a `Capabilities` object, and it exists because backwards compatibility.

You can think of them like this:

- `W3CCapabilities` contains one or more "instances" of `NSCapabilities` within its data structure.
- `NSCapabilities` is the same as `Capabilities`, except that the key/prop names differ (any Appium-specific props are prefixed with an `appium:` namespace)
- `Capabilities` are essentially per-session options that an Appium driver can accept.  

They all build on one another, and the generic `Constraints` `C` is ultimately used by `Capabilities` to build a proper interface.  "Constraints" is kind of an adhoc DSL created long before we started using types.  It looks like the object in `packages/types/lib/constraints.js`.  `Capabilities` "converts" _that_ object into a typical TS interface with key names and types.

boneskull:
@pacozaa I think you found a bug in `PluginCommand` ;)

pacozaa:
@boneskull Please check out a new push at `images-plugin/plugin.js` and `images-plugin/compare.js` regarding your commend from https://github.com/appium/appium/pull/18059#issuecomment-1401003408

I have to add `ExtDriver` as a first generic to get the vscode to recognize the correct PluginCommand structure.

Let's see If I am on the right direction. 

For the `npm run typedoc` error, I am not sure if it is related to me or how to fix it

Screenshot 2566-01-24 at 11 34 19

pacozaa:
@boneskull I tried to pull rebase and push but the `npm run typedoc` still returns errors. What should I do next? Not really sure how those errors related to my PR. Please help.

boneskull:
rebasing onto `master` to see if that fixes the issue.  unless the entire `tsc` build fails, what you're doing shouldn't break the docs build

boneskull:
@pacozaa Thanks again for this.  If you don't want to get it over the finish line, I can debug and fix whatever's wrong here and push it.  What do you think?

pacozaa:
> @pacozaa Thanks again for this.  If you don't want to get it over the finish line, I can debug and fix whatever's wrong here and push it.  What do you think?

@boneskull Sorry about that. I couldn't finish it right now. Yes, That would be great. Thank you for understanding.

PR #18086, opened 2023-01-19
typedoc outputs code examples

boneskull:
Contains significant changes to comment derivation.

Given a command method in an external driver, these changes cause the plugin to search a) the method itself, b) the method's signature (which may come from `BaseDriver` or even `DriverCore`), the method map, and the `ExternalDriver` interface for comments (in that order).  It picks the first "summary" / "description" found, but otherwise it will _merge_ any and all block tags found, given the same precedence.  So if, for example, a `@see` tag exists in cmd method `ExternalDriver.foo()` but not in `MyDriver.foo()`, the plugin will find that `@see` tag and put it into the data to be rendered by the theme for the command associated with `MyDriver.foo()`.

This also works with `@example`, with a change: the examples side-step the usual rendering (by `typedoc-plugin-markdown`'s theme; this is due to tight coupling and complexing necessary in overriding the default behavior by simply adding an overriding template or partial), and are rendered per-language (since it's likely/expected that each method will have examples in several languages).  The language of the example is detected via the _fenced code block_ as used in the example.  In markdown, we're currently using some headers to split examples up, but we also render a token as an HTML comment which can be used on the `mkdocs` side of things to detect the example and perform further post-processing (e.g. as a set of tabs).

If a command has examples, the examples will be surrounded by (in addition to headers for each language) this token pair:

```html
<!-- BEGIN:EXAMPLES -->
...
<!-- END:EXAMPLES -->
```

Each language's fenced code block is surrounded by the following token pair _within_ the above pair:

```html
<!-- BEGIN:EXAMPLE lang=$lang -->
...
<!-- END:EXAMPLE -->
```

...where `$lang` is the _proper name_ for a language, e.g., `JavaScript` or `Ruby`.  The fenced code blocks themselves remain unchanged and will contain the "identifier" for the language, e.g., `js` or `javascript` or `rb` or `ruby` (i.e. they are _not_ normalized).

Also:

- changed the type of `KnownMethods` to be a `Map<string, CommandMethodDeclarationReflection>`, as the value of the `comment` prop is better derived via `deriveComments()`. This changes `findAsyncMethodsInReflection()` as well
- memoized `AppiumPluginLogger.createChildLogger()` to avoid creating identical loggers
- renamed `ProjectCommands.empty` => `ProjectCommands.isEmpty`
- `CommandReflection`'s `constructor` now expects a `log` parameter
- Hopefully easier-to-understand verbose logging


boneskull:
* master
  * **PR #18086** <a href="https://app.graphite.dev/github/pr/appium/appium/18086" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18086?utm_source=stack-comment).

boneskull:
example output

image

packages/typedoc-plugin-appium/test/e2e/converter/comment.e2e.spec.ts
```diff
@@ -8,21 +8,16 @@ import {
...
+      expect(knownMethods.get('activateApp')).to.exist;
     });
 
-    describe('when not provided a reflection nor `Comment` parameter', function () {
-      it('should derive the comment from KnownMethods', function () {
-        const commentData = deriveComment('activateApp', knownMethods);
-        expect(commentData).to.exist.and.to.have.keys('comment', 'commentSource');
-      });
-    });
+    it('should', function () {});
```
jlipps, line 21: i don't imagine you intend to commit an empty test?

packages/types/lib/driver.ts
```diff
@@ -615,6 +615,37 @@ export interface Core<C extends Constraints = BaseDriverCapConstraints> {
...
+   * Implementations may optionally include additional meta information as part of the body, but the top-level properties ready and message are reserved and must not be overwritten.
+   *
+   * @example
+   * ```js
+   * // webdriver.io example
+   * await driver.status();
+   * ```
+   *
+   * ```python
+   * selenium.webdriver.common.utils.is_url_connectable(port)
```
jlipps, line 634: i don't think this is right. @KazuCocoa what is `GET /status` in the python bindings?

KazuCocoa: Maybe Python client does not have it. `is_url_connectable` actually calls the local https://github.com/SeleniumHQ/selenium/blob/084c3b16bb11d625536141bc441ec95a8ee05bb7/py/selenium/webdriver/common/utils.py#L122 but maybe it is.

KazuCocoa: hm, actually the selenium client does not have https://github.com/SeleniumHQ/selenium/blob/086e9fb1536348a891af9a52853c53021566dd50/py/selenium/webdriver/remote/command.py#L99
https://github.com/SeleniumHQ/selenium/blob/084c3b16bb11d625536141bc441ec95a8ee05bb7/py/selenium/webdriver/remote/remote_connection.py#L182

let me create a pr as `driver.get_status()`

packages/typedoc-plugin-appium/test/e2e/converter/comment.e2e.spec.ts
boneskull: I didn't, no.

packages/types/lib/driver.ts
boneskull: I just copied the example from the website

KazuCocoa: Yea, i saw the page.
A new version will have `driver.getStatus()` https://github.com/appium/python-client/pull/820 (will push a new version later)

jlipps (approved):

packages/types/lib/driver.ts
KazuCocoa, line 634:
```suggestion
   * driver.get_status()
```

KazuCocoa (approved):
published a new Python client ver which has get_status

PR #18091, opened 2023-01-20
chore(ci): add job to disallow merges of PRs

boneskull:
...if it has the "On Hold" label

### DIFF
--- a/.github/workflows/on-hold.yml
+++ b/.github/workflows/on-hold.yml
@@ -1,0 +1,15 @@
+name: Disallow "On Hold" PRs
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


boneskull:
This works

boneskull:
FWIW, the action needs "write" perms because it rewrites the status of its own previous checks. I think this may cause any other jobs that depend on it to re-run?  I'm not entirely sure.  Anyway, this seems safe enough.

jlipps (approved):
something tells me you may be the only one to use this ^ _ ^ but looks good!

boneskull:
@jlipps yeah, I haven't seen anyone else use `On Hold`, but I felt nervous about it.

PR #18092, opened 2023-01-20
chore(types): remove @wdio/types

boneskull:
Due to problems where multiple versions of `@wdio/types` can be installed in the dep tree of external modules, and the fact we are not using most of the types we pull in, it makes sense to remove this module.

The "standard" W3C caps have been pulled out of `@wdio/types`.  We do not need any webdriverio or vendor-specific cap types.  Those only make sense when consuming wdio itself and/or writing client-side code.

This will also make it easier to upgrade `webdriverio`; it too depends on `@wdio/types`.

Note: there were several Appium-specific caps in `@wdio/types` that I did not recognize.  Perhaps these were driver-specific and not accurately labeled as such.  Driver-specific caps belong in drivers.

See #18090

### DIFF
--- a/packages/types/lib/capabilities.ts
+++ b/packages/types/lib/capabilities.ts
@@ -1,8 +1,9 @@
-import type {Capabilities as WdioCaps} from '@wdio/types';
+import {StandardCapabilities} from './standard-caps';
 import {StringRecord, Constraint, Constraints} from '.';
 import {BaseDriverCapConstraints} from './constraints';
 
-export type StandardCapabilities = WdioCaps.Capabilities;
+export {StandardCapabilities};
+
 export type W3C_APPIUM_PREFIX = 'appium';
 
 /**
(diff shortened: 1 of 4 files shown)


packages/types/lib/capabilities.ts
```diff
@@ -15,16 +16,6 @@ export type BaseCapabilities = ConstraintsToCaps<BaseDriverCapConstraints>;
  */
 export type BaseNSCapabilities = CapsToNSCaps<ConstraintsToCaps<BaseDriverCapConstraints>>;
 
-/**
- * These may (or should) be reused by drivers.
- */
-export type AppiumAndroidCapabilities = WdioCaps.AppiumAndroidCapabilities;
```
mykola-mokhnach, line 21: are these not used anywhere else?

mykola-mokhnach (approved):

packages/types/lib/capabilities.ts
boneskull: not that I'm aware of, no.  the intent was originally for a driver to consume it, but since capabilities can now be derived via constraints, there's no real point to keeping it.  I certainly didn't send a PR into e.g., uiautomator2 to use this

jlipps (approved):

PR #18093, opened 2023-01-20
chore(appium): do not attempt automatic install of extensions in dev env

boneskull:
It sucks to run the "autoinstall extensions" script in CI or the monorepo due to the order in which lifecycle scripts are executed: the `postinstall` script of `appium` will always run _before_ the `prepare` script in the monorepo root.  This completely disables the behavior if Appium has not yet been built.

### DIFF
--- a/packages/appium/scripts/autoinstall-extensions.js
+++ b/packages/appium/scripts/autoinstall-extensions.js
@@ -19,18 +19,11 @@
  * `npm install -g appium --drivers=uiautomator2,xcuitest --plugins=images`
  */
 
-const path = require('path');
-const {exec} = require('teen_process');
 const B = require('bluebird');
 
 B.config({
   cancellation: true,
 });
-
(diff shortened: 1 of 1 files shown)


packages/appium/scripts/autoinstall-extensions.js
```diff
@@ -227,10 +181,7 @@ async function checkAndInstallExtension({
 if (require.main === module) {
   main().catch((e) => {
     log(e);
-    if (process.env.CI) {
-      log('Detected CI environment, exiting with code 1');
-      process.exitCode = 1;
-    }
+    process.exitCode = 1;
```
boneskull, line 184: tbh I'm happy about exiting nonzero on a failure, but it's the only way an end-user is going to know, because otherwise the output from this script is squelched by default.

jlipps (approved):

PR #18101, opened 2023-01-23
chore(execute-driver-plugin,support,types): use named tuples

boneskull:
The "old" way of defining tuples is `[T, U]`, but I think TS 4.7 introduced "named" tuples, which at least provides a little bit of context to consumers, e.g., `[foo: T, bar: U]`.  Since a tuple is just a fixed-length array, the names aren't actually used in code, but are surfaced in type information and error messages regardless.

Anyway, name tuples seem objectively better than unnamed tuples.

Also:

- Fixed bad param name in definition of `SessionHandler`
- Used a semantically-correct type for `StopScreenRecordOptions.formFields`

### DIFF
--- a/packages/execute-driver-plugin/lib/plugin.js
+++ b/packages/execute-driver-plugin/lib/plugin.js
@@ -32,7 +32,7 @@ export default class ExecuteDriverPlugin extends BasePlugin {
    *
    * @returns {Promise<any>} - a JSONifiable object representing the return value of
    * the script
-   * @type {import('@appium/types').PluginCommand<[string, string?, number?]>}
+   * @type {import('@appium/types').PluginCommand<[script: string, scriptType: string?, timeoutMs: number?], any>}
    * @throws {Error}
    */
   async executeDriverScript(
--- a/packages/support/test/mocks.js
+++ b/packages/support/test/mocks.js
@@ -87,5 +87,5 @@ export function initMocks(sandbox = createSandbox()) {
  */
 
 /**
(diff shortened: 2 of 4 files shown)


boneskull:
* master
  * **PR #18101** <a href="https://app.graphite.dev/github/pr/appium/appium/18101" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18101?utm_source=stack-comment).

packages/execute-driver-plugin/lib/plugin.js
```diff
@@ -32,7 +32,7 @@ export default class ExecuteDriverPlugin extends BasePlugin {
    *
    * @returns {Promise<any>} - a JSONifiable object representing the return value of
    * the script
-   * @type {import('@appium/types').PluginCommand<[string, string?, number?]>}
+   * @type {import('@appium/types').PluginCommand<[script:string, scriptType:string?, timeoutMs:number?], any>}
```
jlipps, line 35: everywhere else we have spaces after `:`, probably should here for consistency

jlipps (approved):

packages/execute-driver-plugin/lib/plugin.js
boneskull: yeah, that was on purpose due to a tooling problem (I have a thing that automatically rewraps comment paragraphs which I should probably disable).

PR #18109, opened 2023-01-25
docs(appium): include all reference output in docs build

jlipps:


boneskull (approved):
lgtm. we can omit the readme output at the typedoc level too

PR #18175, opened 2023-02-08
feat: granular control over package titles in command doc output

boneskull:
This changes how packages get their titles (the h1 in the commands page, which is used by mkdocs in the nav bar).

A new option, `packageTitles` is available in `typedoc.json`.  This is an array of objects like so:

```json
{
  "packageTitles": [
    {"name": "@appium/fake-driver": "title": "Fake Driver"},
    {"name": "@appium/fake-plugin": "title": "Fake Plugin"}
  ]
}
```

This allows granular control over how each package is titled.  However, if this option is not provided, the plugin will "extract" a title from the `package.json` of each package, assuming that package is an Appium extension.  By default, it will name drivers `Driver: ${appium.driverName}` and plugins `Plugin: ${appium.pluginName}` where `appium` is a prop in `package.json`.

If the package is _not_ an extension but _does_ have commands -- and currently the only package that fulfills this criteria is `@appium/base-driver` -- _and_ is in a multi-package project (e.g. this monorepo), the title will be `${name} Commands` where `name` is from that package's `package.json`.  Otherwise, if we are documenting a project with a _single_ package (e.g., a 3p extension) _and_ we cannot otherwise "extract" a title, we will just use `Commands` for the title.

Also: contains a PR with better logging for docutils.  I guess that should be a separate PR.  lmk if it's bothersome enough and I can easily do so


mykola-mokhnach (approved):

boneskull:
@jlipps you may want to run `build-docs.js` and look at the output

jlipps (approved):

jlipps:
yeah, looks a lot better!

PR #18177, opened 2023-02-09
fix(docutils): fix bad option name and ignore most falsy args

boneskull:
- in type `DeployOpts`, `mkDocsYml` should have been `mkdocsYml`
- arguments were only being omitted from being passed to `mike` if they were `false`, but it needed
  to also consider `undefined` values and empty string values.  `0` is falsy, but we should not
  disallow this.  thus... "most" falsy values get ignored.


boneskull:
* master
  * **PR #18177** <a href="https://app.graphite.dev/github/pr/appium/appium/18177" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18177?utm_source=stack-comment).

boneskull:
need to add another fix here

jlipps (approved):

PR #18181, opened 2023-02-10
typedoc-plugin-appium: fix parameter description rendering

### LINKED ISSUE #18132: typedoc-plugin-markdown: command parameter description output problem

### Current Behavior

We consider the parameter names as defined in a `MethodDef` to be the source of truth. They override the parameter names for the method implementations; the `MethodDef` names are used and any extraneous method impl parameters are dropped.  Likewise, the `MethodDef` param names override the param names specified by interfaces (in e.g., `@appium/types`' `ExternalDriver`).

The above strategy might want to change, but this is the state of things now.

The `MethodDef` (in the `METHOD_MAP` of `@appium/basedriver`'s `lib/basedriver/routes.js` calls the parameters `desiredCapabilities`, `requiredCapabilities`, and `capabilities`.  In `BaseDriver`'s (`BaseDriverCore`) method implementation, these parameters are named `w3cCapabilities1`, `w3cCapabilities2` and `w3cCapabilities3`.  Similarly, in `ExternalDriver.createSession`, these are named `w3cCaps1`, `w3cCaps2` and `w3cCaps3`.

Since a `MethodDef` offers no property to describe a parameter, the parameter descriptions must be derived from either `ExternalDriver` or the method implementation.  

* * *

In the case of `createSession`, the method implementation provides a `@param` tag for each parameter, but the descriptions are empty.  @jlipps has graciously done so in `ExternalDriver.createSession`.  But in the final markdown output, we only see:

```
##### `Param`
```

### Expected Behavior

- The `Param` header should be omitted
- The descriptions of the parameters should live in the "parameter table" shown beneath (e.g. the generated documentation for `ExternalDriver.createSession` does the thing I want)

- execute `npm run typedoc` in the monorepo root
- examine `typedoc-docs/commands/base-driver.md`

### Environment

_No response_

### Link to Appium Logs

_No response_

### Futher Information

I theorize that this is happening via one or more of (at least) of these mechanisms:

1.  Because the `@param`s are present on `BaseDriver.createSession`'s docstring but lacks a description, all of the tags programmatically "look the same" and are collapsed into a single empty value via some uniqueness check happening somewhere.
2. Because we (naïvely) _clone_ and _rename_ the `ParameterReflection` objects, a reference is broken somewhere in TypeDoc's guts.

We use `typedoc-plugin-markdown`'s facilities to render these "parameter tables", so I'm going to start by looking into the implementation to see what's missing.

boneskull:
In #18132, I found that parameter descriptions of commands were being omitted from the output. Those descriptions have to come from a method, because they cannot be expressed in their original location (a method map) via JS or a docstring.

In addition, if an extension's implementation did not describe the parameters (e.g., in `@param` tags), we need to pull those descriptions out of `ExternalDriver` (assuming they exist).

Furthermore, we were executing the conversion _too early_ in TypeDoc's lifecycle, since its internal plugins haven't run yet. One of those plugins actually moves the `@param` comments from the _signature_ into the appropriate parameter. Once those are there, we can find them.

- Added a new list of "comment finders" which are parameter-specific (could probably do the same thing for return statements)
- Moved some functions into different modules and renamed a couple things to be more accurate.
- Update VSCode launch config
- `CommandData` now requires a reflection of a method. This was true before, but not formalized

Closes #18132


boneskull:
OK, so to mitigate the OOM problem, I changed the E2E tests for `typedoc-plugin-appium` to run each suite in its own process, in serial.  Each test runs TypeDoc at least up through conversion one or more times, so I'm not really surprised... but it makes me suspicious that TypeDoc has memory leaks.  Maybe as a fun exercise I can try to track them down, but not now.

jlipps (approved):
can't say i fully understand what's going on here but i didn't notice any obvious problems

boneskull:
this has some kinks that need working out

boneskull:
before:

image


after: 

image

boneskull:
the types aren't quite right though

boneskull:
there we go

image

PR #18182, opened 2023-02-10
chore(deps): upgrade typescript

boneskull:
anything newer than v4.7.4 fails to ignore an optional arg in `@param {T} [foo]` if `foo` has a default assignment in the function signature.  removing the square brackets fixes the issue.

whether or not this is a bug, I am not entirely sure: it is marked as a bug, but it is also not
something the TS team is in a hurry to fix, so whatevs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,7 +64,7 @@
         "typedoc": "0.23.24",
         "typedoc-plugin-markdown": "3.14.0",
         "typedoc-plugin-resolve-crossmodule-references": "0.3.3",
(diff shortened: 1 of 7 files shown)


boneskull:
* master
  * **PR #18182** <a href="https://app.graphite.dev/github/pr/appium/appium/18182" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18182?utm_source=stack-comment).

jlipps (approved):

PR #18202, opened 2023-02-14
feat(typedoc-plugin-appium): extract descriptions of return values from builtins

### LINKED ISSUE #18196: inherit comments for return values of command methods
Similarly to #18132; while the return value description will show if the class method provides it, e.g.,:

```js
class MyDriver extends BaseDriver {
  /**
   * @returns stuff
   */
  createSession() {
    return 'stuff';
  }
}
```

When omitted, it does not inherit a description from the docstring for `ExternalDriver.createSession`, assuming that docstring contains a `@returns` tag.

(Why do I get the uneasy feeling that I'm implementing docstring tag inheritance?)

boneskull:
If a command is not documented with a `@returns` tag, but a builtin method _does_ have this tag (e.g., `ExternalDriver.createSession()`), then the plugin will use the tag from the builtin.

If a command _does_ have a `@returns` tag, but it provides no description, then the plugin will use the original `@returns` tag (which may have a type) and combine it with the content/description from the builtin (if any).

This requires further inspection of `SignatureReflection` data, which needed a new set of "comment finders", which prompted me to refactor the comment finders into their own files (because it was becoming unwieldy).  The call to `deriveComment` happens in the static `BaseCommandData.rewriteSignature()` method (whereas before it was just a straightforward clone operation).

- Fixed the block tag aggregation to prefer contentful over contentless tags in all cases.
- Modified the `@returns` tag of `FakeDriver.createSession()` for testing
- Moved some types around that were used outside of their sibling modules (e.g. `CommentSource` which used to be called `CommentSourceType`)
- pruned dead imports

Closes #18196


boneskull:
* master
  * **PR #18202** <a href="https://app.graphite.dev/github/pr/appium/appium/18202" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18202?utm_source=stack-comment).

boneskull:
Here, the text at the very bottom was extracted from the docstring in `@appium/types`: https://github.com/appium/appium/blob/64554b967aa6d48864505d03b9c2e57b167059a8/packages/types/lib/driver.ts#L404-L426

image

packages/typedoc-plugin-appium/lib/converter/comment/index.ts
```diff
@@ -0,0 +1,182 @@
...
+  const finalBlockTags: CommentTag[] = [];
+  for (const tags of Object.values(tagsByTag)) {
+    if (tags.length === 1) {
+      finalBlockTags.push(_.first(tags)!);
+    } else if (tags.length > 1) {
+      // prefer a tag with content
+      const tag = tags.find((t) => t.content.length) ?? tags[0];
+      finalBlockTags.push(tag);
+    } else {
+      // should not happen
```
jlipps, line 81: do we want to throw an error here so if it _does_ happen we know that it happened?

packages/typedoc-plugin-appium/lib/converter/comment/method.ts
```diff
@@ -0,0 +1,79 @@
...
+      }
+
+      // if `otherRefl` exists, then the comment could live in several places,
+      // which happen to be findable by the _other_ `CommentFinder`s.  we avoid
+      // `CommentSourceType.OtherMethod`, which corresponds to _this_ `CommentFinder`
+      // to avoid recursion (for now).
+      //
+      // after looping thru the finders, if we have a comment in the list of `commentData`
+      // objects, return the first one found.
+      const comment = MethodCommentFinders.filter(
```
jlipps, line 62: probably easier to read if you put `.filter` on the line below so it lines up with `.map` and `.find`

packages/typedoc-plugin-appium/lib/converter/comment/types.ts
```diff
@@ -0,0 +1,31 @@
+import {Comment, Reflection} from 'typedoc';
+import {CommentSource, KnownMethods} from '../types';
+
+/**
+ * A function which queries some object for a {@linkcode Comment}, and if found,
+ * its {@linkcode Comment.summary summary}.
```
jlipps, line 6: should this be `{@linkcode Comment.summary} summary`?

packages/typedoc-plugin-appium/lib/converter/types.ts
```diff
@@ -276,3 +277,87 @@ export type ExecuteMethodCommandReflection = CommandReflection & {
...
+  ts: 'TypeScript',
+  typescript: 'TypeScript',
+  js: 'JavaScript',
+  javascript: 'JavaScript',
+  py: 'Python',
+  python: 'Python',
+  rb: 'Ruby',
+  ruby: 'Ruby',
+  java: 'Java',
+}) satisfies Record<string, string>;
```
jlipps, line 296: 👏

jlipps (approved):

packages/typedoc-plugin-appium/lib/converter/comment/index.ts
boneskull: (I'll log it)

packages/typedoc-plugin-appium/lib/converter/comment/method.ts
boneskull: prettier won't let me

packages/typedoc-plugin-appium/lib/converter/comment/types.ts
boneskull: it does the same thing:

image

boneskull:
@jlipps If CI passes, I'm going to merge this, but please look at https://github.com/appium/appium/pull/18202/commits/be84cfb0aae305a78b8b7bba8ccd4199b994e584 if you get a chance, because it is shameful

PR #18208, opened 2023-02-15
chore: replace usage of public class field assignments

boneskull:
Basically, public class fields don't work the way you'd expect, so I'd like to avoid assignment to them whenever possible.  Compounding the situation is the fact that `tsc` (and Babel, fwiw) both transpile in a way that deviates slightly from the official specification.

Given the following example:

```js
class A {
  value = 'a';
  constructor() {
    console.log(this.value);
  }
}

class B extends A {
  value = 'b';
}
```

What is logged by `new B()`?  If you guessed `b`, you'd be both reasonable and wrong.  What's happening is that the super class does not know anything about public class fields in subclasses.  If `value` was instead a method or getter, this _would_ work as you'd expect.

Probably better to avoid them whenever possible.  It is perfectly safe to _declare_ them for type inference & the like, but assignment should be avoided until it's absolutely the right thing to do.

Note that `tsc` has a flag to change the transpilation behavior to act in-line with the spec, but it does not change anything in the above regard, AFAICT.

### DIFF
--- a/packages/base-driver/lib/basedriver/core.js
+++ b/packages/base-driver/lib/basedriver/core.js
@@ -35,7 +35,7 @@ class DriverCore {
   /**
    * @type {string?}
    */
-  sessionId = null;
+  sessionId;
 
   /**
    * @type {import('@appium/types').DriverOpts<C>}
@@ -47,7 +47,8 @@ class DriverCore {
    */
   initialOpts;
 
-  helpers = helpers;
+  /** @type {typeof helpers} */
+  helpers;
 
   /**
    * basePath is used for several purposes, for example in setting up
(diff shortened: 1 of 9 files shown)


boneskull:
To be clear: `static` fields, private `static` fields, and private class fields do not suffer from this problem

packages/base-driver/lib/basedriver/core.js
```diff
@@ -56,39 +57,43 @@ class DriverCore {
...
-  managedDrivers = [];
+  managedDrivers;
 
   /** @type {NodeJS.Timeout?} */
-  noCommandTimer = null;
+  noCommandTimer;
 
   /** @type {EventHistory} */
-  _eventHistory = {commands: []};
+  _eventHistory;
```
mykola-mokhnach, line 92: should it be also changed to `#eventHistory` ?

packages/fake-driver/lib/driver.js
```diff
@@ -15,20 +15,45 @@ const FAKE_DRIVER_CONSTRAINTS = /** @type {const} */ ({
...
 
   /** @type {string} */
   curContext;
 
-  appModel = new FakeApp();
+  /** @type {FakeApp} */
+  appModel;
+
+  /** @type {boolean} */
+  _proxyActive;
```
mykola-mokhnach, line 41: same as above

packages/universal-xml-plugin/lib/plugin.js
```diff
@@ -1,20 +1,13 @@
...
 
-import BasePlugin from 'appium/plugin';
+import {BasePlugin} from 'appium/plugin';
 import {errors} from 'appium/driver';
 import {transformSourceXml} from './source';
 import {transformQuery} from './xpath';
 import log from './logger';
 
 export default class UniversalXMLPlugin extends BasePlugin {
-  commands = [
```
mykola-mokhnach, line 10: how this assignment should work now?

boneskull: this code was unused afaict.

boneskull:
* master
  * **PR #18208** <a href="https://app.graphite.dev/github/pr/appium/appium/18208" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18208?utm_source=stack-comment).

packages/fake-driver/lib/driver.js
```diff
@@ -205,82 +232,354 @@ export class FakeDriver extends BaseDriver {
   /*********
    * ALERT *
    *********/
-  assertNoAlert = alert.assertNoAlert;
-  assertAlert = alert.assertAlert;
-  getAlertText = alert.getAlertText;
-  setAlertText = alert.setAlertText;
-  postAcceptAlert = alert.postAcceptAlert;
-  postDismissAlert = alert.postDismissAlert;
+  /**
```
jlipps, line 235: ugh i hate that one line is now 6. it makes it much less desirable to use this file-splitting pattern

packages/universal-xml-plugin/lib/plugin.js
jlipps: yeah i think it came from an old version of our base plugin logic which required registration of overrides.

jlipps (approved):

packages/fake-driver/lib/driver.js
boneskull: I kind of hate it too.  there's another way, which is to just fake it.  let me see if I can do that.

boneskull:
I've changed this to exclude changes to `fake-driver`, since they were more involved; see #18237

boneskull:
labeled "enhancement" since `@appium/types` gets a new type

mykola-mokhnach (approved):

jlipps (approved):

PR #18213, opened 2023-02-16
feat(driver-test-support): add startStoppableAppium()

boneskull:
This adds a function `startStoppableAppium()` which starts an Appium server (via its `main()` entry point), but resolves with an object having a `stop() => Promise<boolean>` method. This method force-quits the server, even if there are active connections.

This functionality will be helpful in driver E2E tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2277,6 +2277,14 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/@types/stoppable": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/@types/stoppable/-/stoppable-1.1.1.tgz",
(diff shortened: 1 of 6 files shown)


jlipps (approved):
LGTM, just want to make sure we're not writing code to hide a bigger problem. Is the actual Appium server behaving appropriately already (by not stopping completely when we call `.close` on it?)

boneskull:
The thing is, `Server.close()` does not stop the server "completely"; it will remain open while sockets are still open; it is the socket's responsibility to disconnect.  This is why `server.closeAllConnections()` exists in Node.js v18.  AFAICT, once our minimum version is Node.js v18, [stoppable](https://npm.im/stoppable) is obsolete and this method can be changed to be an alias of Appium's `main()` and deprecated.

boneskull:
fwiw I haven't seen anything that would suggest to me that Appium is doing something wrong here.  That doesn't mean it isn't doing something wrong, of course...

jlipps:
yep, ok makes sense. just makes me wonder why there are still open socket connections after all sessions have been quit.

PR #18278, opened 2023-03-01
chore: Remove obsolete jpeg-js version override

mykola-mokhnach:
The recent version of jimp does not use this module anymore

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,6 @@
     "access": "public"
   },
   "overrides": {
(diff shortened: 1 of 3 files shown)


boneskull:
I have no idea why that CI job is failing.  @mykola-mokhnach Could you like... amend and force-push?

boneskull (approved):
LGTM, but don't understand why that job is failing. Anyway: it likely doesn't have anything to do with these changes

PR #18266, opened 2023-02-28
fix(base-driver): use new mixin strategy

boneskull:
This applies the same kind of changes made to `FakeDriver` to `BaseDriver`.  Unlike `FakeDriver`, we
already have the interfaces defined in `@appium/types`, so we can reuse them.

Note that `BaseDriverCore` is no longer a thing; the class is just called `BaseDriver`. Hopefully
this will be less confusing.

PR #18265, opened 2023-02-28
a miscellaneous grab bag of fixes

boneskull:
Mostly type-related changes which were identified when converting `BaseDriver` to the new mixin
strategy.

PR #18261, opened 2023-02-27
chore(docutils): remove monkeypatching of typedoc

boneskull:
with the upgrade to the latest typedoc we no longer need to monkeypatch it.

Issue #18159, opened 2023-02-06
typescript-eslint: the time has come


boneskull:
Well, since we now have some TS in the repo, we should probably use [typescript-eslint](https://typescript-eslint.io/getting-started/) since ESLint doesn't know how to read TS without it.

PR #18280, opened 2023-03-01
docutils cli: fail-fast validation of paths

boneskull:
This adds up-front checking of user-provided custom paths to avoid doing a bunch of work and then
failing later ("fail fast").

For example, a user may provide a custom path to `mkdocs.yml` on the CLI:

```bash
appium-docs build --mkdocs-yml /some/bad/path/to/mkdocs.yml
```

In the above case, `appium-docs` will check if `/some/bad/path/to/mkdocs.yml` exists _before_ running
TypeDoc and MkDocs.

An alternative solution would be to just ignore it and use the default path.  This was _clearly_
not the user's intent (because they provided a custom path), so it's a poor solution.

### DIFF
--- a/packages/docutils/lib/cli/check.ts
+++ b/packages/docutils/lib/cli/check.ts
@@ -3,10 +3,9 @@
  * @module
  */
 
-import {fs} from '@appium/support';
+import {fs, util} from '@appium/support';
 import _ from 'lodash';
-import pluralize from 'pluralize';
-import {InferredOptionTypes, Options} from 'yargs';
+import type {Options} from 'yargs';
 import logger from '../logger';
 
 const log = logger.withTag('check');
@@ -72,7 +71,7 @@ export async function checkMissingPaths<T extends Record<string, Options>>(
 
   log.debug(
     'Checking for existence of %s: %s',
-    pluralize('path', pathsToCheck.length),
+    util.pluralize('path', pathsToCheck.length),
     _.map(pathsToCheck, 'path')
   );
 
--- a/packages/docutils/lib/cli/check.ts
+++ b/packages/docutils/lib/cli/check.ts
@@ -1,0 +1,88 @@
+/**
+ * Provides functions to validate user-provided options as part of Yargs' post-parsing "check" phase
+ * @module
+ */
+
+import {fs} from '@appium/support';
+import _ from 'lodash';
+import pluralize from 'pluralize';
+import {InferredOptionTypes, Options} from 'yargs';
+import logger from '../logger';
+
+const log = logger.withTag('check');
+
+/**
+ * Given a list of objects with `id` and `path` props, filters out the ones that do not exist
+ * @param paths Filepaths
+ * @returns Missing files
+ */
+async function filterMissing(paths: MissingFileData[]): Promise<MissingFileData[]> {
+  const exists = await Promise.all(
+    paths.map(async ({id, path}) => ({id, path, exists: await fs.exists(path)}))
+  );
+  const results = _.reject(exists, 'exists');
+  return _.map(results, (result) => _.omit(result, 'exists'));
+}
+
+/**
+ * Data structure describing a missing file; returned by {@linkcode filterMissing}
+ */
+interface MissingFileData {
+  /**
+   * Arbitrary identifier; intent was to map to an option name
+   */
+  id: string;
+  /**
+   * Normalized filepath
+   */
+  path: string;
+}
+
+/**
+ * Takes user-provided paths and checks for existence.
+ *
+ * Filters options to consider based on group name only.
+ *
+ * Meant to be used as a "fail-fast" strategy on the CLI, so we don't go all the way through some
+ * expensive behavior before realizing we're missing a path.
+ * @param opts Options object for a yargs command
+ * @param group Group name to filter on
+ * @param argv User-provided args
+ * @returns `true` if all paths exist or otherwise an error message
+ */
+export async function checkMissingPaths<T extends Record<string, Options>>(
+  opts: T,
+  group: string,
+  argv: Record<string, unknown>
+): Promise<true | string> {
+  const argsToCheck = _.keys(
+    _.pickBy(opts, (opt: Options, id: string) => opt?.group === group && id in argv) as Partial<T>
+  );
+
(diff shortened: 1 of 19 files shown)


boneskull:
* master
  * **PR #18261** <a href="https://app.graphite.dev/github/pr/appium/appium/18261" target="_blank">Graphite</a> 
    * **PR #18265** <a href="https://app.graphite.dev/github/pr/appium/appium/18265" target="_blank">Graphite</a> 
      * **PR #18266** <a href="https://app.graphite.dev/github/pr/appium/appium/18266" target="_blank">Graphite</a> 
        * **PR #18280** <a href="https://app.graphite.dev/github/pr/appium/appium/18280" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18280?utm_source=stack-comment).

packages/docutils/lib/cli/check.ts
```diff
@@ -0,0 +1,88 @@
...
+const log = logger.withTag('check');
+
+/**
+ * Given a list of objects with `id` and `path` props, filters out the ones that do not exist
+ * @param paths Filepaths
+ * @returns Missing files
+ */
+async function filterMissing(paths: MissingFileData[]): Promise<MissingFileData[]> {
+  const exists = await Promise.all(
+    paths.map(async ({id, path}) => ({id, path, exists: await fs.exists(path)}))
```
mykola-mokhnach, line 21: I did not know that .map accepts async functions

packages/docutils/lib/cli/check.ts
```diff
@@ -0,0 +1,88 @@
...
+  );
+  const results = _.reject(exists, 'exists');
+  return _.map(results, (result) => _.omit(result, 'exists'));
```
mykola-mokhnach, line 24: I assume bluebird's filter could do this all in one line

packages/docutils/lib/cli/check.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Takes user-provided paths and checks for existence.
+ *
+ * Filters options to consider based on group name only.
+ *
+ * Meant to be used as a "fail-fast" strategy on the CLI, so we don't go all the way through some
+ * expensive behavior before realizing we're missing a path.
+ * @param opts Options object for a yargs command
+ * @param group Group name to filter on
+ * @param argv User-provided args
+ * @returns `true` if all paths exist or otherwise an error message
```
mykola-mokhnach, line 51: would it make more sence to just throw an error instead of returning a string?

packages/docutils/lib/cli/check.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  const pathsToCheck: MissingFileData[] = _.map(argsToCheck, (id) => ({
+    id,
+    path: String(argv[id]), // this must be a string per the above check
+  }));
+
+  log.debug(
+    'Checking for existence of %s: %s',
+    pluralize('path', pathsToCheck.length),
```
mykola-mokhnach, line 75: there is util.pluralize function in the support module

packages/docutils/lib/cli/command/init.ts
```diff
@@ -144,17 +157,21 @@ const opts = Object.freeze({
...
 type InitOptions = InferredOptionTypes<typeof opts>;
 
 export default {
   command: 'init',
   describe: 'Initialize package for doc generation',
-  builder: opts,
+  builder(yargs) {
+    return yargs
+      .options(opts)
+      .check(async (argv) => checkMissingPaths(opts, InitCommandGroup.Paths, argv));
```
mykola-mokhnach, line 173: do we need await in the lambda body?

packages/docutils/lib/cli/command/validate.ts
```diff
@@ -76,25 +84,25 @@ const opts = Object.freeze({
...
 export default {
   command: 'validate',
   describe: 'Validate Environment',
-  builder: opts,
+  builder(yargs) {
+    return yargs.options(opts).check(async (argv) => {
+      if (!argv.python && !argv.typedoc && !argv.typescript && !argv.mkdocs) {
+        return 'No validation targets specified; one or more of --python, --typescript, --typedoc or --mkdocs must be provided';
+      }
+      return checkMissingPaths(opts, ValidateCommandGroup.Paths, argv);
```
mykola-mokhnach, line 102: await?

packages/docutils/lib/cli/check.ts
jlipps: there's nothing magical going on here. async functions are just functions that return a promise when called. so this use of `.map` will produce a list of promises, which are resolved by `await Promise.all`.

packages/docutils/lib/cli/check.ts
boneskull: I could do this on one line with lodash too, but it was too clever

packages/docutils/lib/cli/check.ts
boneskull: we could do that, but I figure that'd produce a stack trace somewhere, and it's unneeded for checking parameters.  the return value of these functions is just what yargs expects.

packages/docutils/lib/cli/check.ts
boneskull: good catch -- I started depending directly on `pluralize` in `@appium/typedoc-plugin-appium` because I didn't need to pull in all of `@appium/support` (though it's ostensibly already installed).  guess I forgot why I was doing what I was doing

packages/docutils/lib/cli/command/init.ts
boneskull: we don't have lint support for typescript installed yet (see #18159) so idk.  I was never a fan of that rule

packages/docutils/lib/cli/command/validate.ts
boneskull: more about this: my rationale for not using it is that it does not matter what `checkMissingPaths` returns because the function consuming it is `async`; adding `await` here without a `try/catch` serves no practical purpose other than causing churn if it did happen to become a sync function (because keeping a useless `await` will be slow in a hot path and could be the source of subtle bugs).  

anyway, I'm going to leave this for now and I can make my case when we add `typescript-eslint` support.

PR #18281, opened 2023-03-01
chore(ci): remove the "on hold" label checker

boneskull:
It is broken and I do not know why. Was fun while it lasted.


boneskull:
This was in response to #18278 ... I'm not sure what's going on, because it doesn't seem to be breaking elsewhere.  Hmm...

Issue #18273, opened 2023-03-01
error running command: Error: Could not find MIME for Buffer


nanatlantica:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

[...]

PR #18284, opened 2023-03-02
fix(support): Improve error description on image load

mykola-mokhnach:
## Proposed changes

Related to https://github.com/appium/appium/issues/18273

## Types of changes

### DIFF
--- a/packages/fake-driver/lib/driver.js
+++ b/packages/fake-driver/lib/driver.js
@@ -1,6 +1,6 @@
 import B from 'bluebird';
 import {BaseDriver, errors} from 'appium/driver';
-import {deprecatedCommandsLogged} from '@appium/base-driver/lib/protocol/protocol';
+import {deprecatedCommandsLogged} from '@appium/base-driver/build/lib/protocol/protocol';
 import {FakeApp} from './fake-app';
 
 const FAKE_DRIVER_CONSTRAINTS = /** @type {const} */ ({
--- a/packages/fake-driver/lib/driver.js
+++ b/packages/fake-driver/lib/driver.js
@@ -1,6 +1,6 @@
 import B from 'bluebird';
 import {BaseDriver, errors} from 'appium/driver';
-import {deprecatedCommandsLogged} from '@appium/base-driver/build/lib/protocol/protocol';
+import {deprecatedCommandsLogged} from '@appium/base-driver/lib/protocol/protocol';
 import {FakeApp} from './fake-app';
 
 const FAKE_DRIVER_CONSTRAINTS = /** @type {const} */ ({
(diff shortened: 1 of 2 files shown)


packages/support/lib/image-util.js
```diff
@@ -22,18 +22,25 @@ async function getJimpImage(data) {
...
-    if (_.isString(data)) {
-      data = Buffer.from(data, 'base64');
-    }
+
+    const imagePayload = _.isBuffer(data)
+      ? data
+      // if data is a string, assume it is a base64-encoded image
+      : Buffer.from(String(data), 'base64');
+    const truncatedData = _.truncate(
+      _.isBuffer(data) ? data.toString('utf8') : data,
```
boneskull, line 31: `_.isBuffer(data)` isn't type-safe; it returns a `boolean` but does not narrow the type of `data` to `Buffer`.  `Buffer.isBuffer` does, however. 

```suggestion
      Buffer.isBuffer(data) ? data.toString('utf8') : data,
```

packages/fake-driver/lib/driver.js
```diff
@@ -1,6 +1,6 @@
 import B from 'bluebird';
 import {BaseDriver, errors} from 'appium/driver';
-import {deprecatedCommandsLogged} from '@appium/base-driver/build/lib/protocol/protocol';
+import {deprecatedCommandsLogged} from '@appium/base-driver/lib/protocol/protocol';
```
boneskull, line 3: If we need `deprecatedCommandsLogged` outside of `@appium/base-driver`, then it should be exported from `@appium/base-driver`.  that will make it so it can be imported from `appium/driver`, as it should be.  this means we are making `deprecatedCommandsLogged` a public API.

please export this function from `@appium/base-driver/lib/index.js`

boneskull: (note: importing from `@appium/base-driver` or `@appium/support` is discouraged in drivers, since we do not explicitly depend on these packages)

packages/support/lib/image-util.js
mykola-mokhnach: fixed

packages/fake-driver/lib/driver.js
mykola-mokhnach: I have reverted the change. this anyway is not related to the current PR, so should not be mixed. 
Up to @jlipps to decide what to so with this import

jlipps:
Given that it's really only imported for test purposes, I don't think we need to make it a public API.

mykola-mokhnach:
Should we maybe move it to base driver's e2e tests instead of keeping it in fake driver?

jlipps:
Fakedriver is a part of our e2e tests (that is why it was invented). This was the best way I could think of to have an e2e test of the deprecation feature, anyway.

jlipps (approved):

PR #18287, opened 2023-03-02
fix(typedoc-plugin-appium): replace void responses with 'null'

### LINKED ISSUE #18269: typedoc-plugin-appium: null return type
The commands as documented by the plugin should display `null` if the return type is `void` or `undefined`, as this is what will be sent over the wire. 

The return types have already been extracted from their `Promise` wrappers, as this is irrelevant for commands.

boneskull:
The main thrust of this was to resolve #18269.

Implementation-wise, we have to modify an `IntrinsicType` of which both `void` and `undefined` become.  To make it `null`, we must create a new `LiteralType` (`null` is not an "intrinsic" type; I don't know why).  I would think that TypeDoc wouldn't go around creating new `LiteralType`s every time it sees `null`, but I don't see it doing anything else. Anyway, if there is a "type cache" somewhere, I'm not sure how to get at it.

Additionally, the new reflections created for parameters and the call signature were not being "registered" properly with TypeDoc.  So this adds a `static create` to both `CommandData` and `ExecMethodData` which wraps the constructor for each, respectively, then performs the registration.  This keeps a `Context` object _out_ of `CommandData`/`ExecMethodData` instances (it would only be used once at time of instantiation anyway).

Un-memoized a couple methods in `BaseCommandData` since not necessary

### DIFF
--- a/packages/typedoc-plugin-appium/lib/guards.ts
+++ b/packages/typedoc-plugin-appium/lib/guards.ts
@@ -296,7 +296,7 @@ export function isCommandMethodDeclarationReflection(
     ? value.type.declaration.getAllSignatures()
     : value.getAllSignatures();
   return Boolean(
-    signatures?.find((sig) => isReferenceType(sig.type) && sig.type.name === 'Promise')
+    signatures?.find((sig) => sig.type instanceof ReferenceType && sig.type.name === 'Promise')
   );
 }
 
@@ -332,14 +332,6 @@ export function isCallSignatureReflectionWithArity(
 }
 
 /**
- * Guard for {@linkcode ReferenceType}
- * @param value any
- */
-export function isReferenceType(value: any): value is ReferenceType {
-  return value instanceof ReferenceType;
-}
-
-/**
  * Guard for {@linkcode ConstructorDeclarationReflection}
  * @param value any
  */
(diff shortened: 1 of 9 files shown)


boneskull:
* master
  * **PR #18287** <a href="https://app.graphite.dev/github/pr/appium/appium/18287" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18287?utm_source=stack-comment).

packages/typedoc-plugin-appium/lib/guards.ts
```diff
@@ -296,7 +296,7 @@ export function isCommandMethodDeclarationReflection(
     ? value.type.declaration.getAllSignatures()
     : value.getAllSignatures();
   return Boolean(
-    signatures?.find((sig) => isReferenceType(sig.type) && sig.type.name === 'Promise')
+    signatures?.find((sig) => sig.type instanceof ReferenceType && sig.type.name === 'Promise')
```
boneskull, line 299: I think I prefer using `x instanceof T` when the type guard is that simple; this is why I removed `isReferenceType()`

jlipps: yeah makes sense

jlipps (approved):

PR #18293, opened 2023-03-03
fix(appium,fake-driver): expose child process when running an extension script

### LINKED ISSUE #18211: bug: appium driver run scripts can't handle stdin

### Current Behavior

Interoperability with prompt managers like inquirer don't work

### Expected Behavior

I would expect a driver author to be able to do pretty much anything in a driver script including ask for CLI input

n/a

### Environment

n/a

### Link to Appium Logs

_No response_

### Futher Information

_No response_

boneskull:
This change makes it so when running an extension script (e.g., `appium driver foo run some-script`), unless the `--json` flag is present, the child process will do no buffering whatsoever and will inherit `process.stdin`, `process.stdout` and `process.stderr` from Appium.

This means that extension scripts can do stuff like prompt the user for input.  It _also_ means they can do stuff like display progress bars and spinners.

If the user wants JSON data, we revert to the old behavior, which is to buffer the output and dump it once the script has completed.  I still don't understand why we need a `RingBuffer` there, since it seems like it could obliterate JSON output.  I didn't want to touch it though; Chesterton's fence & all that.

* * *

Note that given `teen_process`' `Promise`-based architecture--even using `teen_process.SubProcess`--there's still buffering and wrapping happening and I could not figure out how to make `teen_process` work the way I needed it to.  Maybe the lesson I've learned is that `teen_process` has a specific common use-case in mind--and it is not this one.

### DIFF
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -4,14 +4,15 @@ import _ from 'lodash';
 import path from 'path';
 import {npm, util, env, console} from '@appium/support';
 import {spinWith, RingBuffer} from './utils';
-import {SubProcess} from 'teen_process';
 import {
   INSTALL_TYPE_NPM,
   INSTALL_TYPE_GIT,
   INSTALL_TYPE_GITHUB,
   INSTALL_TYPE_LOCAL,
 } from '../extension/extension-config';
+import {SubProcess} from 'teen_process';
 import {packageDidChange} from '../extension/package-changed';
+import {spawn} from 'child_process';
 
 const UPDATE_ALL = 'installed';
 
@@ -660,17 +661,34 @@ class ExtensionCommand {
   }
 
   /**
-   * Runs a script cached inside the "scripts" field under "appium"
-   * inside of the driver/plugins "package.json" file. Will throw
-   * an error if the driver/plugin does not contain a "scripts" field
-   * underneath the "appium" field in its package.json, if the
-   * "scripts" field is not a plain object, or if the scriptName is
-   * not found within "scripts" object.
+   * Just wraps {@linkcode child_process.spawn} with some default options
+   *
+   * @param {string} cwd - CWD
+   * @param {string} script - Path to script
+   * @param {string[]} args - Extra args for script
+   * @param {import('child_process').SpawnOptions} opts - Options
+   * @returns {import('node:child_process').ChildProcess}
+   */
+  _runUnbuffered(cwd, script, args = [], opts = {}) {
+    return spawn(process.execPath, [script, ...args], {
+      cwd,
+      stdio: 'inherit',
+      ...opts,
+    });
+  }
+
+  /**
+   * Runs a script cached inside the `scripts` field under `appium`
+   * inside of the extension's `package.json` file. Will throw
+   * an error if the driver/plugin does not contain a `scripts` field
+   * underneath the `appium` field in its `package.json`, if the
+   * `scripts` field is not a plain object, or if the `scriptName` is
+   * not found within `scripts` object.
    *
    * @param {RunOptions} opts
    * @return {Promise<RunOutput>}
    */
-  async _run({installSpec, scriptName, extraArgs = []}) {
+  async _run({installSpec, scriptName, extraArgs = [], bufferOutput = false}) {
     if (!this.config.isInstalled(installSpec)) {
       throw this._createFatalError(`The ${this.type} "${installSpec}" is not installed`);
     }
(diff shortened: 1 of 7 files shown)


boneskull:
> I still don't understand why we need a RingBuffer there, since it seems like it could obliterate JSON output

or rather, assuming it's used as an extra buffer on top of whatever is stuck in the stream, maybe it won't break JSON, but even then I'm not sure why it's needed.

packages/appium/lib/cli/extension-command.js
```diff
@@ -699,26 +717,54 @@ class ExtensionCommand {
...
+      try {
+        await new B((resolve, reject) => {
+          this._runUnbuffered(
+            this.config.getInstallPath(installSpec),
+            extScripts[scriptName],
+            extraArgs
+          )
+            .on('error', (err) => {
+              // generally this is of the "I can't find the script" variety
+              reject(err);
```
mykola-mokhnach, line 752: sometimes native errors are quite cryptic for users, e.g. ENOENT one.
Shall we add some customization on top of this one?

packages/fake-driver/lib/scripts/fake-stdin.js
```diff
@@ -0,0 +1,16 @@
...
+const rl = readline.createInterface({input: process.stdin, output: process.stderr});
+
+rl.question('Press ENTER to continue: ', () => {
+  rl.close();
+  // eslint-disable-next-line no-console
+  console.error('You did it!');
+});
+
+rl.once('close', () => {
+  // there is some sort of weirdness where Node.js doesn't know when to stop listening on `process.stdin` if
```
mykola-mokhnach, line 12: perhaps we just need to feed `eof` into the stream?
Are you sure `unref`'ing it won't create a memory leak?

mykola-mokhnach:
```
const { spawn } = require('node:child_process');
const ps = spawn('ps', ['ax']);
const grep = spawn('grep', ['ssh']);

ps.stdout.on('data', (data) => {
  grep.stdin.write(data);
});

ps.stderr.on('data', (data) => {
  console.error(`ps stderr: ${data}`);
});

ps.on('close', (code) => {
  if (code !== 0) {
    console.log(`ps process exited with code ${code}`);
  }
  grep.stdin.end();
});

grep.stdout.on('data', (data) => {
  console.log(data.toString());
});

grep.stderr.on('data', (data) => {
  console.error(`grep stderr: ${data}`);
});

grep.on('close', (code) => {
  if (code !== 0) {
    console.log(`grep process exited with code ${code}`);
  }
});
```

An example from https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

boneskull:
* master
  * **PR #18293** <a href="https://app.graphite.dev/github/pr/appium/appium/18293" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18293?utm_source=stack-comment).

packages/appium/lib/cli/extension-command.js
```diff
@@ -699,26 +717,54 @@ class ExtensionCommand {
...
-      return {error: err.message, output: output.getBuff()};
+      try {
+        await runner.join();
+        this.log.ok(`${scriptName} successfully ran`.green);
+        return {output: output.getBuff()};
+      } catch (err) {
+        this.log.error(`Encountered an error when running '${scriptName}': ${err.message}`.red);
+        return {error: err.message, output: output.getBuff()};
+      }
+    } else {
```
jlipps, line 742: given that we return in the block above, we don't really need the `else` and can save a level of indentation

jlipps (approved):
not sure about @mykola-mokhnach's questions about unref, but otherwise LGTM

packages/appium/lib/cli/extension-command.js
boneskull: It's not actionable by the user; it's a developer bug.  This means "the script file that the developer said would be there is not there"--not that the script isn't defined.

packages/fake-driver/lib/scripts/fake-stdin.js
boneskull: Why would `unref` cause a memory leak?  There's a bug around this, which is why I did what I did: https://github.com/nodejs/node/issues/32291

boneskull: Anyway, the `end()` seems to work in lieu of `unref()`, so I'll just use that.

PR #18308, opened 2023-03-06
chore: add typescript-eslint

### LINKED ISSUE #18159: typescript-eslint: the time has come

boneskull:
**UPDATE**: This PR creates a new package, `@appium/eslint-config-appium-ts`, which projects having TS sources may use.  It no longer affects users of `@appium/eslint-config-appium`.

* * *

This change adds `typescript-eslint` in order to lint our `.ts` files.

Due to the nature of TS, it can be a little more bold with fixes (e.g., changing `let` to `const`).  I changed the following rules from the recommended settings:

- `no-empty-function`: Reduced to a warning.  It's likely we want the functions, but perhaps they should not be empty.
- `no-empty-interface`: Disabled; we use these to merge mixin interfaces with driver classes.
- `no-explicit-any`: Disabled; while transitioning from an untyped codebase to a typed one, we still need to use `any` in places.  Eventually these should all be replaced.
- `no-non-null-assertion`: Disabled; This allows us to continue use of the `!` operator, which is handy when working with untyped or loosely-typed code (especially in tests).
- `no-var-requires`: Reduced to a warning.  Occasionally we may need to use `require()` instead of `import`.

Since the configuration _extends_ our shared config, the following rules have been disabled:

- `import/no-unresolved`: The `eslint-plugin-import` module cannot resolve TS modules.
- `no-unused-vars`: This rule conflicts with `typescript-eslint`'s rule, so it must be disabled per the documentation.
- `promise/no-native`: In our TS-based packages, I haven't needed Bluebird, so it doesn't make sense to require it.
- `require-await`: TS is much more aware of the return values of functions; this is not needed to avoid ambiguity (in TS code).

* * *

The second changeset lints the codebase.


boneskull:
* master
  * **PR #18308** <a href="https://app.graphite.dev/github/pr/appium/appium/18308" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18308?utm_source=stack-comment).

boneskull:
This closes #18159

boneskull:
I think this is actually a breaking change since it changes peer deps.  I wonder if it wouldn't make more sense to publish a separate package for this?

mykola-mokhnach (approved):

boneskull:
No longer breaking

jlipps (approved):
yeah i like this better

Issue #16970, opened 2022-05-20
we need to talk about Logging


boneskull:
I started looking at this issue in #16966.  So, the logging facilities are kind of difficult to maintain.  I am not sure how they got to where they are now, but it seems pretty organic.  From what I understand:

- The logging API which Appium calls internally is a wrapper around `npmlog`.
- `npmlog` does not actually output anything; instead, multiple `winston` transports  get configured and invoked when the `npmlog` instance (is it a singleton?) emits events.
- An end-user has some control over which transports are used
- A programmatic user may have further granular control over what happens with log messages

The actual logging API itself (e.g., `log.info('something')`) doesn't seem to deviate greatly between `winston` and `npmlog`, so I'm assuming the intent here was not due to those differences.  

I spoke on [JS Party](https://changelog.com/jsparty) about this just yesterday--most Node.js packages are either servers, CLI tools, or libraries.  But some are more than one of these.  And while there's often _some_ crossover, the use-cases are wildly different. 

For our purposes, we can split an ideal Appium into three buckets:

1. The server (listening on a port, logging requests and commands, etc)
2. The CLI (help text, warnings, spinners, etc.)
3. Crosscutting stuff (library code)

And some examples of how I think logging should be treated for each, respectively:

1. Structured logging, powerful and configurable. example: `winston`, `pino`
2. Human-readable, friendly output--optionally machine-readable. Terminal-dependent (color/emoji support).  examples: `ora`, `console.log()` with `chalk`, `ink`
3. Silent unless environment flag or configuration variable says otherwise; straightforward, human-readable, and a balance between 1 and 2. Example: `debug`.

If we had three (3) different ways to log based on the context, they shouldn't "conflict" with each other.  In practice, that'd be difficult to consume, since a contributor would need to know a) which of these to use in what bit of the codebase, and b) three different APIs.

If "what is the context in which this code runs?" is murky, then perhaps refactoring or throwing more organization at it would help (e.g., split `appium` into more packages, move things out, etc.).  That's probably good.

[...]

PR #18319, opened 2023-03-08
fix(support): remove all logging from env module

### LINKED ISSUE #18311: bug: plugin list showing stderr debug messages when called with --json

### Current Behavior

When calling `plugin list` with  `--json` flag, debug messages are part of output:

i.e:
```
node packages/appium/index.js plugin list --installed --json 
dbug Support No package.json having current Appium dep in /Users/vrunoa/Projects/appium
dbug Support Using default APPIUM_HOME: /Users/vrunoa/.appium
dbug Support No package.json having current Appium dep in /Users/vrunoa/.appium
{
  "Boost": {
     "pkgNam
...
```

These debug messages are going to stderr so im able to avoid them from output by running
```
node packages/appium/index.js plugin list --installed --json 2>/dev/null
```

but it would be good to avoid this workaround and skip any debug message when using `--json` flag

### Expected Behavior

No debug messages when using `--json` flag

```sh
node packages/appium/index.js plugin list --installed --json 
```

### Environment

- Operating system: **macOS Ventura Version 13.2.1**
- If running via `appium` CLI...
  - Appium CLI version (output of `appium --version`): **appium@2.0.0-beta.48**
  - Node.js version (output of `node --version`): **v16.16.0**
  - `npm` version (output of `npm --version`): **8.11.0**
  - Last `appium` version which did _not_ exhibit the problem: 
- If running Appium Desktop, its version:
- Appium plugin(s) and their version(s):
- Platform and version under test:
- Real device or emulator/simulator:

### Link to Appium Logs

_No response_

### Futher Information

_No response_

boneskull:
Currently there is no way to specify a log level when running extension commands (e.g., `appium driver list`).  Further, the log level used by the support module depends on whether or not Appium set it; used outside of an Appium context, there is no way to programmatically set the log level of the support module via an external API. To that end, I just removed the log messages, because a "proper" fix will be quite involved.  See #16970.

Closes #18311

### DIFF
--- a/packages/support/lib/env.js
+++ b/packages/support/lib/env.js
@@ -4,7 +4,6 @@ import {homedir} from 'os';
 import path from 'path';
 import readPkg from 'read-pkg';
 import {npm} from './npm';
-import log from './logger';
 
 /**
  * Path to the default `APPIUM_HOME` dir (`~/.appium`).
@@ -69,11 +68,6 @@ export const findAppiumDependencyPackage = _.memoize(
           pkg?.peerDependencies?.appium;
         pkgPath = version && !OLD_VERSION_REGEX.test(String(version)) ? cwd : undefined;
       } catch {}
-      if (pkgPath) {
-        log.debug(`Found package.json having current Appium dep in ${pkgPath}`);
-      } else {
-        log.debug(`No package.json having current Appium dep in ${cwd}`);
-      }
       return pkgPath;
     };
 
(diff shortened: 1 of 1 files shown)


boneskull:
I am pleased with the auto-labeling

boneskull:
eh this is fine

mykola-mokhnach:
I had an idea to set logging level to `error` in case of json output. Removing logs completely might make it harder to debug possible issues

jlipps:
yeah LGTM. @mykola-mokhnach's idea is good too

PR #18320, opened 2023-03-08
chore(ci): attempt to avoid running trusted code on PRs from forks

boneskull:
The jobs which set labels can't run on PRs from forked repos because they don't have access to the `GITHUB_TOKEN` secret.


boneskull:
let's just see if this works

PR #18321, opened 2023-03-08
fix(docutils): inherit stdio in long-running processes

boneskull:
This change ensures when running `mkdocs` or `mike` with the `--serve` flag, the output is unbuffered and stdio is inherited from the parent process.

This will show helpful things like _the URL at which the development server is running_.

### DIFF
--- a/packages/docutils/lib/util.ts
+++ b/packages/docutils/lib/util.ts
@@ -4,6 +4,7 @@
  */
 
 import _ from 'lodash';
+import {SpawnOptions, spawn} from 'node:child_process';
 import path from 'node:path';
 import type {SubProcess} from 'teen_process';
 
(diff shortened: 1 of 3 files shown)


packages/docutils/lib/util.ts
```diff
@@ -72,8 +73,29 @@ export const argify: (obj: Record<string, string | number | boolean | undefined>
...
+ * background tasks; we use node's `child_process` directly here to pass `stdio` through, since
+ * `teen_process` basically does not respect `{stdio: 'inherit'}`.
  */
-export type TeenProcessSubprocessStartOpts = Partial<
-  TupleToObject<Parameters<SubProcess['start']>, ['startDetector', 'detach', 'timeoutMs']>
->;
+export async function spawnBackgroundProcess(command: string, args: string[], opts: SpawnOptions) {
+  return new Promise<void>((resolve, reject) => {
+    spawn(command, args, {...opts, stdio: 'inherit'})
+      .on('error', (err) => {
```
mykola-mokhnach, line 88: `on('error', reject)`

mykola-mokhnach (approved):

packages/docutils/lib/util.ts
boneskull: tip: use suggestions for stuff like this

image

jlipps (approved):

PR #18323, opened 2023-03-09
chore(ci): attempt to build and preview docs

boneskull:
This also isolates the docs build into a workflow that only runs when changes are likely to materially affect doc gen.

let's see what happens here...

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,40 @@
+name: Build Docs
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
+    paths:
+      - './packages/appium/docs/**'
+      - './packages/docutils/**'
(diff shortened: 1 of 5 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,40 @@
...
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
+    paths:
+      - './packages/appium/docs/**'
+      - './packages/docutils/**'
+      - './packages/typedoc-plugin-appium/**'
```
jlipps, line 13: we might in the future want to add docs dirs for plugins in the monorepo, but i don't think any exist yet

jlipps (approved):

.github/workflows/docs.yml
boneskull: I'm thinking separate workflows so docs build happens on per-package basis, and only if the pkg is modified

PR #18343, opened 2023-03-14
feat(types,base-driver): allow declaration of extra props to "opts" prop of driver

boneskull:
Some drivers (e.g., XCUITest), for better or worse, put arbitrary data in "opts". This adds a type arg which allows a driver to define what that arbitrary data is. However, we cannot guarantee that the data is present at time of instantiation (unless the driver impl makes the assignment in the constructor, which may not be desirable), so even if defined, it will always be considered a "partial" type.

* * *

FWIW I'm not sure I really _agree_ with this--_why_ does a driver need to put data there?--but in the name of retrofitting types to our codebases, it's either this or refactors.  I don't know how prevalent this strategy is.

### DIFF
--- a/packages/base-driver/lib/basedriver/core.js
+++ b/packages/base-driver/lib/basedriver/core.js
@@ -20,6 +20,7 @@ const ON_UNEXPECTED_SHUTDOWN_EVENT = 'onUnexpectedShutdown';
 
 /**
  * @template {Constraints} [C=BaseDriverCapConstraints]
+ * @template {import('@appium/types').StringRecord} [ExtraOpts=import('type-fest').EmptyObject]
  * @implements {Core<C>}
  */
 class DriverCore {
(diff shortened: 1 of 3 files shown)


boneskull:
* master
  * **PR #18343** <a href="https://app.graphite.dev/github/pr/appium/appium/18343" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18343?utm_source=stack-comment).

jlipps (approved):
I'm not entirely sure why drivers *need* to put things on opts, but in a typeless world it seemed as reasonable a thing to do as otherwise. Moving forward with types it probably makes more sense to put these values as simple member fields instead of on `opts`. We could refactor the existing drivers to do this but it's probably more trouble than it's worth and might introduce bugs. So I think this change is reasonable.

boneskull:
@jlipps Given there's some agreement here that _drivers shouldn't do this_, maybe the right thing to do is close this PR, and drivers can `@ts-ignore` their way out of it.

jlipps:
yeah let's do that, better than adding things to base types which will just cause furrowed brows later. ts-ignore with a todo in the drivers makes sense right now.

PR #18346, opened 2023-03-14
feat(appium): improvements to ext commands when running extension in dev mode

### LINKED ISSUE #18277: bug: "appium driver  uninstall" shows false success most of the times

### Current Behavior

When using the command `appium driver uninstall espresso` the resulting output says the espresso driver is uninstalled successfully. However, if checked with the command `appium driver list --installed` the output says it is still installed. Rerunning uninstall command one more time finally uninstalls it.

### Expected Behavior

The driver should be uninstalled with just one uninstall command.

```bash
❯ appium driver install espresso
✔ Installing 'espresso' using NPM install spec 'appium-espresso-driver'
ℹ Driver espresso@2.15.4 successfully installed
- automationName: Espresso
- platformNames: ["Android"]
❯ appium driver list --installed
✔ Listing installed drivers
- espresso@2.15.4 [installed (NPM)]
❯ appium driver uninstall espresso
✔ Successfully uninstalled driver 'espresso'
❯ appium driver list --installed
info Appium Discovered installed driver "espresso"
✔ Listing installed drivers
- espresso@2.15.4 [installed (NPM)]
❯ appium driver uninstall espresso
✔ Successfully uninstalled driver 'espresso'
❯ appium driver list --installed
✔ Listing installed drivers
❯ appium driver uninstall espresso
Error: ✖ Can't uninstall driver 'espresso'; it is not installed
```

### Environment

- Operating system: MacOS 13.2.1
- Appium CLI version (output of `appium --version`): 2.0.0-beta.56
- Node.js version (output of `node --version`): v18.12.1
- `npm` version (output of `npm --version`):  8.19.2

### Link to Appium Logs

_No response_

### Futher Information

Although this happens most of the time for me. It does however uninstalls the driver with just one command sometimes.

boneskull:
This PR does a handful of things:

1. It adds a `--verbose` flag to `appium <driver|plugin> list` which just dumps a big object (via Node's `util.inspect()`), which contains the underlying manifest file data
2. It creates a new manifest schema revision where the "install type" can be `dev`. Added a migration.
3. Displays a warning if trying to uninstall an extension in dev mode
4. Displays `[installed (dev mode)]` instead of `[installed (npm)]` when running `appium <driver|plugin> list` w/o `--verbose` flag
5. Attempted to simplify the return type of `ExtensionCommand.list()`

Closes #18277

### DIFF
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -9,10 +9,12 @@ import {
   INSTALL_TYPE_GIT,
   INSTALL_TYPE_GITHUB,
   INSTALL_TYPE_LOCAL,
+  INSTALL_TYPE_DEV,
 } from '../extension/extension-config';
 import {SubProcess} from 'teen_process';
 import {packageDidChange} from '../extension/package-changed';
 import {spawn} from 'child_process';
+import {inspect} from 'node:util';
 
 const UPDATE_ALL = 'installed';
 
@@ -99,41 +101,40 @@ class ExtensionCommand {
 
   /**
    * List extensions
-   *
+   * @template {ExtensionType} ExtType
    * @param {ListOptions} opts
-   * @return {Promise<ExtensionListData>} map of extension names to extension data
+   * @return {Promise<ExtensionList<ExtType>>} map of extension names to extension data
    */
-  async list({showInstalled, showUpdates}) {
-    const lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} ${this.type}s`;
+  async list({showInstalled, showUpdates, verbose = false}) {
+    let lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} ${this.type}s`;
+    if (verbose) {
+      lsMsg += ' (verbose mode)';
+    }
     const installedNames = Object.keys(this.config.installedExtensions);
     const knownNames = Object.keys(this.knownExtensions);
-    const exts = [...installedNames, ...knownNames].reduce(
-      (acc, name) => {
-        if (!acc[name]) {
-          if (installedNames.includes(name)) {
-            acc[name] = {
-              ...this.config.installedExtensions[name],
-              installed: true,
-            };
-          } else if (!showInstalled) {
-            acc[name] = {pkgName: this.knownExtensions[name], installed: false};
-          }
+    const listData = [...installedNames, ...knownNames].reduce((acc, name) => {
+      if (!acc[name]) {
+        if (installedNames.includes(name)) {
+          acc[name] = {
+            .../** @type {Partial<ExtManifest<ExtType>>} */ (this.config.installedExtensions[name]),
+            installed: true,
+          };
+        } else if (!showInstalled) {
+          acc[name] = /** @type {ExtensionListData<ExtType>} */ ({
+            pkgName: this.knownExtensions[name],
+            installed: false,
+          });
         }
-        return acc;
-      },
-      /**
-       * This accumulator contains either {@linkcode UninstalledExtensionLIstData} _or_
-       * {@linkcode InstalledExtensionListData} without upgrade information (which is added by the below code block)
-       * @type {Record<string,Partial<InstalledExtensionListData>|UninstalledExtensionListData>}
-       */ ({})
-    );
+      }
+      return acc;
+    }, /** @type {ExtensionList<ExtType>} */ ({}));
 
     // if we want to show whether updates are available, put that behind a spinner
     await spinWith(this.isJsonOutput, lsMsg, async () => {
       if (!showUpdates) {
         return;
       }
-      for (const [ext, data] of _.toPairs(exts)) {
+      for (const [ext, data] of _.toPairs(listData)) {
         if (!data.installed || data.installType !== INSTALL_TYPE_NPM) {
           // don't need to check for updates on exts that aren't installed
           // also don't need to check for updates on non-npm exts
@@ -150,7 +151,12 @@ class ExtensionCommand {
       }
     });
 
-    const listData = /** @type {ExtensionListData} */ (exts);
+    /**
+     * Type guard to narrow "installed" extensions, which have more data
+     * @param {any} data
+     * @returns {data is InstalledExtensionListData<ExtType>}
+     */
+    const extIsInstalled = (data) => Boolean(data.installed);
 
     // if we're just getting the data, short circuit return here since we don't need to do any
     // formatting logic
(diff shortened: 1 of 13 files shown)


boneskull:
verbose mode example:

```bash
> appium driver list --verbose --installed

✔ Listing installed drivers (verbose mode)
{
  xcuitest: {
    pkgName: 'appium-xcuitest-driver',
    version: '4.16.12',
    installType: 'npm',
    installSpec: 'xcuitest',
    installPath: '/Users/boneskull/.appium/node_modules/appium-xcuitest-driver',
    appiumVersion: '^2.0.0-beta.40',
    automationName: 'XCUITest',
    platformNames: [ 'iOS', 'tvOS' ],
    mainClass: 'XCUITestDriver',
    scripts: {
      'build-wda': './build/lib/build-wda.js',
      'open-wda': './scripts/open-wda.js'
    },
    schema: {
      '$schema': 'http://json-schema.org/draft-07/schema',
      type: 'object',
      properties: {
        'webdriveragent-port': {
          appiumCliDest: 'wdaLocalPort',
          default: 8100,
          description: 'Local port used for communication with WebDriverAgent',
          maximum: 65535,
          minimum: 1,
          type: 'integer'
        }
      },
      additionalProperties: false,
      title: 'XCUITest Driver Configuration',
      description: 'Appium configuration schema for the XCUITest driver.',
      '$id': 'driver-xcuitest.json'
    },
    installed: true
  },
  fake: {
    pkgName: '@appium/fake-driver',
    version: '5.2.0',
    installType: 'local',
    installSpec: './packages/fake-driver',
    installPath: '/Users/boneskull/projects/appium/appium/packages/fake-driver',
    appiumVersion: '^2.0.0-beta.35',
    automationName: 'Fake',
    platformNames: [ 'Fake' ],
    mainClass: 'FakeDriver',
    schema: './build/lib/fake-driver-schema.js',
    scripts: {
      'fake-error': './build/lib/scripts/fake-error.js',
      'fake-success': './build/lib/scripts/fake-success.js'
    },
    installed: true
  }
}
```

boneskull:
Example of dev mode in listing:

```shell

❯ appium driver list
✔ Listing available drivers
- xcuitest@4.19.1 [installed (dev mode)]
- uiautomator2 [not installed]
- mac2 [not installed]
- espresso [not installed]
- safari [not installed]
- gecko [not installed]
- chromium [not installed]
```

boneskull:
Upon uninstall:

```bash
❯ appium driver uninstall xcuitest
⚠ Cannot uninstall driver "xcuitest" because it is in development!
```

packages/appium/lib/cli/extension-command.js
```diff
@@ -513,7 +526,12 @@ class ExtensionCommand {
         `Can't uninstall ${this.type} '${installSpec}'; it is not installed`
       );
     }
-    const pkgName = this.config.installedExtensions[installSpec].pkgName;
+    const extRecord = this.config.installedExtensions[installSpec];
+    if (extRecord.installType === INSTALL_TYPE_DEV) {
+      this.log.warn(`Cannot uninstall ${this.type} "${installSpec}" because it is in development!`);
```
mykola-mokhnach, line 531: I don't quite understand the scenario here. Let say I am developing a driver. I checked out the sources and then installed them to appium server. Tested it, everything is fine, then published. Now I want to switch the driver to the NPM branch to verify all works as expected. How do I do it if the dev version cannot be uninstalled (I'm not talking about removing sources physically, but rather removing it from the manifest)

packages/appium/lib/cli/extension-command.js
```diff
@@ -150,20 +151,29 @@ class ExtensionCommand {
...
+    const extIsInstalled = (data) => Boolean(data.installed);
 
     // if we're just getting the data, short circuit return here since we don't need to do any
     // formatting logic
     if (this.isJsonOutput) {
       return listData;
     }
 
+    if (verbose) {
+      this.log.log(inspect(listData, {colors: true, depth: null}));
```
mykola-mokhnach, line 168: should verbose mode respect json output?

packages/appium/lib/cli/extension-command.js
jlipps: i'm pretty sure that 'dev' extensions are not actually 'installed'. they do not live in an appium manifest anywhere. they're dynamically detected based on the current directory. the way to 'uninstall' them is to simply move out of the development directory.

jlipps (approved):
This is good, however I'm concerned that `dev mode` is a misleading name. It's a good name for driver authors, but a potentially misleading name for users who are simply using appium in a local package.json. If I'm running a node project and I have appium and the xcuitest driver in my package.json and the driver is listed as `dev mode` I think that would be confusing.

Maybe `auto-detected` or `locally-detected` or something would be better?

packages/appium/lib/cli/extension-command.js
boneskull: You use a different directory, I guess?  "dev mode" only applies if `CWD` is the directory of an extension (or a subdirectory thereof), _and_ that extension depends on `appium` (as it should).  It doesn't make sense to "uninstall" something that way.  If you want to test your _published_ package, you will need to specify a different `APPIUM_HOME`.  Just like you wouldn't `npm install appium-xcuitest-driver` _from the `appium-xcuitest-driver` working copy_.

packages/appium/lib/cli/extension-command.js
boneskull: hmm, yes, afaict `--json` returns the same data as `--verbose`, but it's just formatted differently.   maybe we don't need `--verbose` at all.  @jlipps

boneskull:
@jlipps 

Also

> Maybe auto-detected or locally-detected or something would be better?

This should be _another_ mode if we want to do it.  Right now, such extensions have an `InstallType` of `npm` (which theoretically makes sense, because `npm` or another pkg manager installed the extension, though `appium` did not).  Further, there should be no restriction (there is currently no such restriction) on removing such auto-detected packages.

jlipps:
> Because of 3, this would not apply for the use case of someone who is just consuming extensions and appium all via npm.

Oh ok great, makes sense. We should still call it dev. But in that case I don't think this necessarily solves the issue that was originally raised, since I think that issue was probably in the context of someone who was using Appium in their test project with a driver defined as a dep in package.json. What happens if you run `appium driver uninstall xcuitest` in that context?

boneskull:
* master
  * **PR #18346** <a href="https://app.graphite.dev/github/pr/appium/appium/18346" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18346?utm_source=stack-comment).

boneskull:
> Oh ok great, makes sense. We should still call it dev. But in that case I don't think this necessarily solves the issue that was originally raised, since I think that issue was probably in the context of someone who was using Appium in their test project with a driver defined as a dep in package.json. What happens if you run appium driver uninstall xcuitest in that context?

I don't think this is what the original use-case was.  If you do this in an empty dir:

- `npm init -y; npm install appium@next appium-xcuitest-driver`
- `appium driver uninstall xcuitest`

The driver is uninstalled.  This PR does not change that behavior.

However, in the original ticket, the author _did_ have a working copy, afaict.

jlipps (approved):

PR #18348, opened 2023-03-15
chore(base-driver): Update HTTP caching logic and align it with the standard

mykola-mokhnach:
## Proposed changes

Now HTTP caching logic properly follows https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching, which allows to reduce the amount of requests and the boilerplate code.

## Types of changes

### DIFF
--- a/packages/base-driver/lib/basedriver/helpers.js
+++ b/packages/base-driver/lib/basedriver/helpers.js
@@ -8,6 +8,8 @@ import AsyncLock from 'async-lock';
 import axios from 'axios';
 import B from 'bluebird';
 
+// for compat with running tests transpiled and in-place
+const {version: BASEDRIVER_VER} = fs.readPackageJsonFrom(__dirname);
 const IPA_EXT = '.ipa';
 const ZIP_EXTS = new Set(['.zip', IPA_EXT]);
 const ZIP_MIME_TYPES = ['application/zip', 'application/x-zip-compressed', 'multipart/x-zip'];
@@ -15,8 +17,9 @@ const CACHED_APPS_MAX_AGE = 1000 * 60 * 60 * 24; // ms
 const MAX_CACHED_APPS = 1024;
 const HTTP_STATUS_NOT_MODIFIED = 304;
 const DEFAULT_REQ_HEADERS = Object.freeze({
-  'user-agent': 'Appium',
+  'user-agent': `Appium (BaseDriver v${BASEDRIVER_VER})`,
 });
+const AVG_DOWNLOAD_SPEED_MEASUREMENT_THRESHOLD_SEC = 2;
 const APPLICATIONS_CACHE = new LRU({
   max: MAX_CACHED_APPS,
   ttl: CACHED_APPS_MAX_AGE, // expire after 24 hours
@@ -453,7 +456,7 @@ async function fetchApp (srcStream, dstPath) {
     `has been downloaded to '${dstPath}' in ${secondsElapsed.toFixed(3)}s`
   );
   // it does not make much sense to approximate the speed for short downloads
-  if (secondsElapsed >= 2) {
+  if (secondsElapsed >= AVG_DOWNLOAD_SPEED_MEASUREMENT_THRESHOLD_SEC) {
     const bytesPerSec = Math.floor(size / secondsElapsed);
     logger.debug(`Approximate download speed: ${util.toReadableSizeString(bytesPerSec)}/s`);
   }
@@ -626,7 +629,14 @@ export default {
   parseCapsArray,
   generateDriverLogPrefix,
 };
-export {configureApp, isPackageOrBundle, duplicateKeys, parseCapsArray, generateDriverLogPrefix};
+export {
+  configureApp,
+  isPackageOrBundle,
+  duplicateKeys,
+  parseCapsArray,
+  generateDriverLogPrefix,
+  BASEDRIVER_VER,
+};
 
 /**
  * @typedef RemoteAppProps
--- a/packages/base-driver/lib/basedriver/helpers.js
+++ b/packages/base-driver/lib/basedriver/helpers.js
@@ -9,7 +9,7 @@ import axios from 'axios';
 import B from 'bluebird';
 
 const IPA_EXT = '.ipa';
-const ZIP_EXTS = ['.zip', IPA_EXT];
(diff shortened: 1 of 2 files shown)


packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -2,16 +2,21 @@ import _ from 'lodash';
...
+import B from 'bluebird';
 
 const IPA_EXT = '.ipa';
 const ZIP_EXTS = ['.zip', IPA_EXT];
 const ZIP_MIME_TYPES = ['application/zip', 'application/x-zip-compressed', 'multipart/x-zip'];
 const CACHED_APPS_MAX_AGE = 1000 * 60 * 60 * 24; // ms
 const MAX_CACHED_APPS = 1024;
+const HTTP_STATUS_NOT_MODIFIED = 304;
+const DEFAULT_REQ_HEADERS = Object.freeze({
+  'user-agent': 'Appium',
```
boneskull, line 18: guess it doesn't really matter but might want to stuff a version in there too.

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -250,90 +185,113 @@ async function configureApp(app, options = /** @type {ConfigureAppOptions} */ ({
...
-          logger.info(
-            `The current file extension '${resultingExt}' is not supported. ` +
+        if (headers['content-disposition'] && /^attachment/i.test(headers['content-disposition'])) {
+          logger.debug(`Content-Disposition: ${headers['content-disposition']}`);
+          const match = /filename="([^"]+)/i.exec(headers['content-disposition']);
+          if (match) {
+            fileName = fs.sanitizeName(match[1], {
+              replacement: SANITIZE_REPLACEMENT,
+            });
+            shouldUnzipApp = shouldUnzipApp || ZIP_EXTS.includes(path.extname(fileName));
```
boneskull, line 267: `ZIP_EXTS` should be a `Set`

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -429,16 +388,76 @@ async function configureApp(app, options = /** @type {ConfigureAppOptions} */ ({
...
+    throw new Error(`Cannot fetch the application: ${err.message}`);
+  }
+
+  const secondsElapsed = timer.getDuration().asSeconds;
+  const {size} = await fs.stat(dstPath);
+  logger.debug(
+    `The application (${util.toReadableSizeString(size)}) ` +
+    `has been downloaded to '${dstPath}' in ${secondsElapsed.toFixed(3)}s`
+  );
+  if (secondsElapsed >= 2) {
```
boneskull, line 455: magic number

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -429,16 +388,76 @@ async function configureApp(app, options = /** @type {ConfigureAppOptions} */ ({
...
+ * @returns {Promise<string>} The same dstPath
+ * @throws {Error} If there was a failure while downloading the file
+ */
+async function fetchApp (srcStream, dstPath) {
+  const timer = new timing.Timer().start();
+  try {
+    const writer = fs.createWriteStream(dstPath);
+    srcStream.pipe(writer);
+
+    await new B((resolve, reject) => {
```
boneskull, line 437: consider using `pipeline` here from `node:stream/promises` after dropping node v14

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -613,4 +632,12 @@ export {configureApp, isPackageOrBundle, duplicateKeys, parseCapsArray, generate
  * @property {Date?} lastModified
  * @property {boolean} immutable
  * @property {number?} maxAge
+ * @property {string?} etag
+ */
+
+/**
+ * @typedef RemoteAppData
```
boneskull, line 639: please add a description and descriptions of props

boneskull (approved):
I can't really comment on the intent of this or if it's "correct" to make these changes, so just had some suggestions.

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: it does not make sense for me to measure approximate speed if the download itself took less than two seconds. Would it be better to just add a comment there or you could propose a good constant name for the minimum number of seconds, which is `2`?

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: I did not know about such thing

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: do we have some helper, which allows to fetch the version number?

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: changed

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: np, added descriptions

packages/base-driver/lib/basedriver/helpers.js
jlipps: could just use basedriver's version number so you don't have to look outside this package

```
Appium (BaseDriver vX.Y.Z)
```

packages/base-driver/lib/basedriver/helpers.js
jlipps: `DOWNLOAD_MEASUREMENT_THRESHOLD`?

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: probably an example of getting the version of base driver package?

jlipps: something like `require('../path/to/package.json').version`

mykola-mokhnach: added

packages/base-driver/lib/basedriver/helpers.js
mykola-mokhnach: changed

Issue #18360, opened 2023-03-17
bug: Crashed when I use findElement(MobileBy.IMAGE)


nanatlantica:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

[ImageElementPlugin] Finding image element with match threshold 0.4
[debug] [XCUITestDriver@efef (908745d6)] Proxying [GET /window/size] to [GET http://127.0.0.1:8100/session/3B727833-3E77-43B5-9864-601A3005D321/window/size] with no body
[debug] [XCUITestDriver@efef (908745d6)] Got response with status 200: {"value":{"width":414,"height":896},"sessionId":"3B727833-3E77-43B5-9864-601A3005D321"}
[debug] [XCUITestDriver@efef (908745d6)] Waiting up to 0 ms for condition
[debug] [XCUITestDriver@efef (908745d6)] Taking screenshot with WDA
[debug] [XCUITestDriver@efef (908745d6)] Matched '/screenshot' to command name 'getScreenshot'
[debug] [XCUITestDriver@efef (908745d6)] Proxying [GET /screenshot] to [GET http://127.0.0.1:8100/session/3B727833-3E77-43B5-9864-601A3005D321/screenshot] with no body
[...]

PR #18369, opened 2023-03-20
chore(base-driver): Check if a valid Error instance is thrown

mykola-mokhnach:
## Proposed changes

Related to https://github.com/appium/appium/issues/18360
This should help to debug issue related to error objections creation.

## Types of changes

### DIFF
--- a/packages/base-driver/lib/protocol/protocol.js
+++ b/packages/base-driver/lib/protocol/protocol.js
@@ -454,7 +454,16 @@ function buildHandler(app, method, path, spec, driver, isSessCmd) {
     } catch (err) {
       // if anything goes wrong, figure out what our response should be
       // based on the type of error that we encountered
-      let actualErr = err;
+      let actualErr;
+      if (!(err instanceof Error) || !(_.has(err, 'stack') && _.has(err, 'message'))) {
+        getLogger(driver, req.params.sessionId || newSessionId).warn(
+          'The thrown error object does not seem to be a valid instance of the Error class. This ' +
+          'might be a genuine bug of a driver or a plugin.'
+        );
+        actualErr = new Error(`${err ?? 'unknown'}`);
+      } else {
+        actualErr = err;
+      }
 
       currentProtocol =
         currentProtocol || extractProtocol(driver, req.params.sessionId || newSessionId);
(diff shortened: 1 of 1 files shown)


packages/base-driver/lib/protocol/protocol.js
```diff
@@ -454,7 +454,16 @@ function buildHandler(app, method, path, spec, driver, isSessCmd) {
...
+      let actualErr;
+      if (!(err instanceof Error) || !(_.has(err, 'stack') && _.has(err, 'message'))) {
+        getLogger(driver, req.params.sessionId || newSessionId).warn(
+          'The thrown error object does not seem to be a valid instance of the Error class. This ' +
+          'might be a genuine bug of a driver or a plugin.'
+        );
+        actualErr = new Error(`${err ?? 'unknown'}`);
+      } else {
+        actualErr = err;
+      }
```
boneskull, line 466: If the symptom is just that we cannot use the `in` operator on something thrown, then all we _really_ need to do is check if `err` is an object:

```js
const actualErr = _.isObject(err) ? err : new Error(`${err ?? 'unknown'}`);
```

right?  because the only thing you can't use `in` on (to my knowledge) are primitives and `null`, both of which `_.isObject()` returns `false` for.

I'm also unsure if _this_ is the only place we need to be making these sorts of checks.  If we cannot make guarantees about the behavior of drivers, we _may_ need to make these coercions happen wherever `BaseDriver` (or `AppiumDriver`, which I think calls any `updateServer` methods?) explicitly catches something thrown by an extension.  This goes for `Promise`s as well, since there's no requirement for a `Promise` to reject with anything in particular.

Unless you think you want to take a look at where else this might be a problem, I'd say let's just get this PR merged and deal with the rest later.  There may be a better way to trap anything coming out of extensions via async hooks or something.

boneskull:
ugh, looks like I need to take another looksee at the config for those failed jobs.  I don't want them to run.

packages/base-driver/lib/protocol/protocol.js
mykola-mokhnach: javascript is weird and null is an object (while undefined isn't). 
Also we don't want to accept any object, but an error, where we have a proper stacktrace attached.

jlipps (approved):

PR #18387, opened 2023-03-23
fix(appium): remove longjohn

### LINKED ISSUE #18357: appium: remove longjohn dependency
[longjohn](https://npm.im/longjohn) is a lib which enables "longer stack traces".

I'm not sure why we need longer stack traces in the first place?  This lib mutates Node.js globals and builtins liberally, and contains this code:

```js
  if (process.env.NODE_ENV === 'production') {
    console.warn('NOTICE: Longjohn is known to cause CPU usage due to its extensive data collection during runtime.\nIt generally should not be used in production applications.');
  }
```

While I'm under no illusions that removing this will give us any sort of notable performance boost, it smells and I'd like to remove it.

cc @jlipps @mykola-mokhnach

boneskull:
This removes `longjohn` and replaces it with a statement which increases the stack trace limit from the default (10) to 100.

I believe setting this overrides whatever was passed to the runtime `--stack-trace-limit` option.  If someone complains, we can try a different strategy.

Closes [#18357](https://github.com/appium/appium/issues/18357)

### DIFF
--- a/packages/appium/lib/main.js
+++ b/packages/appium/lib/main.js
@@ -22,7 +22,7 @@ import {
 } from './config';
 import {readConfigFile} from './config-file';
 import {loadExtensions, getActivePlugins, getActiveDrivers} from './extension';
-import {SERVER_SUBCOMMAND} from './constants';
+import {SERVER_SUBCOMMAND, LONG_STACKTRACE_LIMIT} from './constants';
 import registerNode from './grid-register';
 import {getDefaultsForSchema, validate} from './schema/schema';
 import {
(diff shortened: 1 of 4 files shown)


boneskull:
* master
  * **PR #18387** <a href="https://app.graphite.dev/github/pr/appium/appium/18387" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18387?utm_source=stack-comment).

packages/appium/lib/main.js
```diff
@@ -46,8 +46,7 @@ async function preflightChecks(args, throwInsteadOfExit = false) {
     checkNodeOk();
     await checkNpmOk();
     if (args.longStacktrace) {
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
-      require('longjohn').async_trace_limit = -1;
+      Error.stackTraceLimit = LONG_STACKTRACE_LIMIT;
```
mykola-mokhnach, line 49: what is the default value there?

mykola-mokhnach (approved):

packages/appium/lib/main.js
boneskull: `10`

Issue #18379, opened 2023-03-22
base-driver: bad declarations (and the prevention thereof)


boneskull:
So the declarations generated from `@appium/base-driver` have some sort of bug in them. For whatever reason, `BaseDriver` ends up _extending multiple classes_. 

This is probably a bug in TypeScript, but I don't think I have the gumption to get at the root cause.

My hunch is that it is JS-specific. The first step I'd take is to see if we can't convert the `driver.js` source to TypeScript. 

This was discovered by pulling in the declarations from a project where the TS compiler config has `skipLibCheck: false`. 

To avoid this sort of thing in the future, pulling in a small type-testing framework should be sufficient. Whatever the DefinitivelyTyped project uses should be adequate.

PR #18390, opened 2023-03-23
add tsd for testing declaration files

boneskull:
This adds [tsd](https://npm.im/tsd) as a framework to test our declaration files.  I didn't look too hard for another tool to do this, but this is the one that the DefinitelyTyped project uses, so it's battle-tested, at minimum.

The aim of `tsd` is to ensure that the declaration files you write (by hand, typically; more on that later) are actually valid.  Often enough, projects (which consume declaration files published by a dependency) _also_ validate the declaration files of their dependencies when compiling.  The `skipLibCheck = true` compiler option avoids this, and is recommended in most situations.

However, while `skipLibCheck = true` unblocks consumers from invalid declarations, the problem does not go away; invalid declaration files are still published.  `tsd` allows those writing declarations to make assertions that, in fact, their declarations are valid.

While we don't have many (any?) hand-written declaration (`.d.ts`) files, we have many generated by TypeScript.  As of this writing, TS' support for generating declaration files from JS sources seems buggy (see #18379).  Until it's more stable--or we're able to implement a workaround--we should be smoke-testing the declaration files generated from our JS sources.  _To be clear_, I don't want huge test suites around everything (they are "just types" after all), but I think some amount of basic "does this very important type (e.g., `BaseDriver` or `ExternalDriver`) actually do what we think it does?" will have value.  In addition, if we do have any handwritten `.d.ts` files, we should test those too.

One thing to note is that `tsd` expects the test files (`.test-d.ts`) to be siblings of the `.d.ts` files it is testing.  This leads to  strange module resolution if you do not want to do this.  At any rate, `eslint-plugin-import` does not grok it, so I've disabled the resolution check rule for `.test-d.ts` files.

This PR does not contain any tests; it is just the scaffolding.


boneskull:
* master
  * **PR #18390** <a href="https://app.graphite.dev/github/pr/appium/appium/18390" target="_blank">Graphite</a>  👈
    * **PR #18391** <a href="https://app.graphite.dev/github/pr/appium/appium/18391" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18390?utm_source=stack-comment).

jlipps (approved):

PR #18391, opened 2023-03-23
many fixes to emit working declarations

### LINKED ISSUE #18379: base-driver: bad declarations (and the prevention thereof)

boneskull:
This fixes the invalid declarations generated by TypeScript by rewriting `@appium/base-driver`'s
`lib/basedriver/driver.js` (containing the `BaseDriver` class) in TypeScript.

It also addresses a host of issues found when doing so, particularly around the use of the
`*Capabilities` and `*DriverCaps` types.  The question of which of these an external driver should
use is much less murky (e.g., you want to use `createSession(caps: W3CDriverCaps<MyConstraints>)`).
See the detailed commit message for more information.

It builds upon the structure added in #18390 by adding handful of simple checks for `BaseDriver`.

Closes #18379

### DIFF
--- a/packages/base-driver/lib/protocol/routes.js
+++ b/packages/base-driver/lib/protocol/routes.js
@@ -18,6 +18,7 @@ const SET_ALERT_TEXT_PAYLOAD_PARAMS = {
  * define the routes, mapping of HTTP methods to particular driver commands, and
  * any parameters that are expected in a request parameters can be `required` or
  * `optional`
+ * @satisfies {import('@appium/types').MethodMap<import('../basedriver/driver').BaseDriver>}
  */
 const METHOD_MAP = /** @type {const} */ ({
   '/status': {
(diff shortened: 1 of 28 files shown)


boneskull:
* master
  * **PR #18390** <a href="https://app.graphite.dev/github/pr/appium/appium/18390" target="_blank">Graphite</a> 
    * **PR #18391** <a href="https://app.graphite.dev/github/pr/appium/appium/18391" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18391?utm_source=stack-comment).

mykola-mokhnach (approved):

packages/base-driver/lib/protocol/routes.js
```diff
@@ -18,6 +18,7 @@ const SET_ALERT_TEXT_PAYLOAD_PARAMS = {
  * define the routes, mapping of HTTP methods to particular driver commands, and
  * any parameters that are expected in a request parameters can be `required` or
  * `optional`
+ * @satisfies {import('@appium/types').MethodMap<import('../basedriver/driver').BaseDriver>}
```
jlipps, line 21: oh cool, is this new?

jlipps (approved):

packages/base-driver/lib/protocol/routes.js
boneskull: yes!

PR #18417, opened 2023-03-29
chore: attempt to get labeler workflows running against forks

boneskull:
Currently the `labeler` and `on-hold` workflows fail when run from forks.  Trying to get it working here.


boneskull:
* master
  * **PR #18417** <a href="https://app.graphite.dev/github/pr/appium/appium/18417" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18417?utm_source=stack-comment).

boneskull:
Well, they should have ran here, but they didn't, so this isn't quite right.

boneskull:
Oh, I see.  It doesn't run because it runs _only_ if the workflow exists in `master`, which it doesn't yet.  I'm going to merge this and see what it does, I guess.

PR #18425, opened 2023-03-30
chore(ci): filter more workflows

boneskull:
This causes the "Build" workflow to only trigger if the PR modifies files in paths that might affect the build.

Likewise, the "Build Docs" workflow is expanded to rebuild docs if API changes occur (e.g. in sources).

Also restricts perms a bit more on docs build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,42 @@
 name: Appium Build
-on: [push, pull_request]
+on:
+  push:
+    # this list of paths could potentially impact the build
+    paths:
+      - '**/tsconfig*.json'
+      - 'packages/base-driver/static/**/*'
+      - 'packages/*/test/**'
+      - 'packages/*/bin'
+      - 'packages/**/*.ts'
+      - 'packages/**/*.json'
+      - 'packages/**/*.js'
+      - 'package*.json'
+      - 'lerna.json'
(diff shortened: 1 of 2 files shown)


boneskull:
* master
  * **PR #18425** <a href="https://app.graphite.dev/github/pr/appium/appium/18425" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18425?utm_source=stack-comment).

.github/workflows/build.yml
```diff
@@ -1,5 +1,42 @@
...
+      - '.npmrc'
+      - '.mocharc.js'
+      - '.github/workflows/build.yml' # this file
+      - '.eslintrc'
+      - '.eslintignore'
+      - '!**/sample-code/**'
+      - '!packages/*/docs/**'
+  pull_request:
+    paths:
+      - '**/tsconfig*.json'
```
jlipps, line 24: blech config file duplication

jlipps (approved):

.github/workflows/build.yml
boneskull: I tried to use an anchor, but they don't work for sequences in YAML, because reasons.

boneskull:
preview url is wrong. trying again

PR #18432, opened 2023-03-31
feat(docutils): add "full" watch mode

boneskull:
BREAKING CHANGE: This package is now ESM-only.

TypeDoc itself is unable to watch and rebuild a "composite" TS project (I'm not sure why, but the sources say something about "too hard").  Maybe just _too hard to do the right way_.

So I did it manually; gather a list of files to watch and re-run TypeDoc when needed.  Once TypeDoc outputs files, `mkdocs serve` will see the output and rebuild/reload itself accordingly.

Unfortunately--and this is a questionable decision on my part--the file watcher I found looks a lot nicer than the usual `chokidar`--but it is ESM only.  To use it, `@appium/docutils` must also be ESM-only (or rather, I haven't figured out how I can `await import(someEsmOnlyPkg)` in TS; it definitely depends on how a package distributes itself).  I expect eventually the ecosystem will move more towards ESM modules (for better or worse), and we can think of this as an experimental foray into that sort of thing.

Because `@appium/docutils` is mostly intended to be used as a CLI app, I think we can get away with this.  In `appium`'s `docs/scripts/build-docs.js`, we _do_ need to pull in `@appium/docutils` as a library, but that seems to work fine with `await import()`.

Tests don't run, so the build will break.  If we want to go down this rabbit hole further, I'll see
about getting them to work.  All-in-all it wasn't _too_ bad getting TS to do the right thing, but
maybe just using `chokidar` is for the best (though it pulls in native module(s)).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -55,6 +55,12 @@
           }
         ]
       }
+    },
+    {
+      "files": "./packages/docutils/**/*.ts",
+      "parserOptions": {
+        "sourceType": "module"
+      }
     }
   ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -60,6 +60,22 @@
       ],
       "sourceMaps": true,
       "type": "node"
+    },
+    {
+      "console": "integratedTerminal",
+      "cwd": "${workspaceFolder}/packages/appium",
+      "internalConsoleOptions": "openOnSessionStart",
+      "name": "appium-docs build",
+      "program": "${workspaceFolder}/packages/docutils/build/lib/cli/index.js",
+      "args": ["build", "--serve", "--mkdocs-yml", "./docs/mkdocs-en.yml", "-L", "debug"],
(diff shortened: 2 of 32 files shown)


boneskull:
* master
  * **PR #18432** <a href="https://app.graphite.dev/github/pr/appium/appium/18432" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18432?utm_source=stack-comment).

packages/opencv/lib/index.js
boneskull, line 1: I need to pull this change out into its own PR.

jlipps:
I feel a little hesitant about introducing ESM stuff without a clear strategy for how other tools, particularly in our own ecosystem, could import it. Is there a way for us to make this project ESM but ensure that somehow it also exports a CJS version?

PR #18453, opened 2023-04-05
feat(typedoc-plugin-appium): enable comment derivation via reflection types

boneskull:
In the case of `appium-xcuitest-driver`, each command (or execute) method is declared like so:

```ts
import {someMixin} from './some-mixins';

class XCUITestDriver {
  foo = someMixin.foo;
}
```

The docstring exists on `someMixin.foo`, but not on the `foo` prop of an `XCUITestDriver` instance.

This change makes `@appium/typedoc-plugin-appium` dig into such references and derive the comments.

### DIFF
--- a/packages/typedoc-plugin-appium/lib/converter/comment/index.ts
+++ b/packages/typedoc-plugin-appium/lib/converter/comment/index.ts
@@ -75,7 +75,7 @@ export function deriveComment(opts: DeriveCommentOptions = {}): CommentData | un
   const finalBlockTags: CommentTag[] = [];
   for (const tags of Object.values(tagsByTag)) {
     if (tags.length === 1) {
-      finalBlockTags.push(_.first(tags)!);
+      finalBlockTags.push(tags[0]);
     } else if (tags.length > 1) {
       // prefer a tag with content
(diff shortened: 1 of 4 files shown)


boneskull:
* master
  * **PR #18453** <a href="https://app.graphite.dev/github/pr/appium/appium/18453" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18453?utm_source=stack-comment).

packages/typedoc-plugin-appium/lib/converter/comment/index.ts
```diff
@@ -75,7 +75,7 @@ export function deriveComment(opts: DeriveCommentOptions = {}): CommentData | un
   const finalBlockTags: CommentTag[] = [];
   for (const tags of Object.values(tagsByTag)) {
     if (tags.length === 1) {
-      finalBlockTags.push(_.first(tags)!);
+      finalBlockTags.push(tags[0]);
```
boneskull, line 78: avoid warning about `!`

jlipps (approved):

PR #18458, opened 2023-04-06
fix(docutils): properly respects log levels

boneskull:
The resulting solution keeps a bunch of weak refs to `Consola` objects and sets the log level on all of them if it ever changes.

This is--at minimum--easier to understand than the broken `Proxy` implementation.

### DIFF
--- a/packages/docutils/lib/logger.ts
+++ b/packages/docutils/lib/logger.ts
@@ -137,57 +137,58 @@ class DocutilsReporter extends FancyReporter {
 let globalLevel: LogLevel = LogLevelMap[DEFAULT_LOG_LEVEL];
 
 /**
- * The logger from which all loggers are created.  This one uses a unique tag and our custom reporter.
+ * Type guard to see if a string is a recognized log level
+ * @param level any value
  */
-const rootLogger = createLogProxy(
-  consola.create({defaults: {tag: 'docutils'}, reporters: [new DocutilsReporter()]})
-);
-
-/**
- * @summary Creates a log-level-propagating proxy for a {@linkcode Consola} logger.
- * @description
- * Alright.  So when you create a new logger via {@linkcode Consola.create}, it's basically a clone
- * of its parent with a new set of options.
- *
- * If we change the log level of the root logger (which we do: see `cli/index.ts`), we may (almost
- * certainly) have
- * child loggers which: a) have already been created and b) have inherited the old/default log level
- * from the root logger. We don't _want_ that (though this is likely a reasonable use case) for our
- * purposes.
- *
- * The implementation below solves the problem by maintaining its own singleton log level value, and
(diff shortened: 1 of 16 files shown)


boneskull:
* master
  * **PR #18458** <a href="https://app.graphite.dev/github/pr/appium/appium/18458" target="_blank">Graphite</a>  👈
    * **PR #18459** <a href="https://app.graphite.dev/github/pr/appium/appium/18459" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18458?utm_source=stack-comment).

packages/docutils/lib/logger.ts
```diff
@@ -137,57 +137,44 @@ class DocutilsReporter extends FancyReporter {
...
-      return Reflect.set(target, prop, value, receiver);
-    },
-  });
+const rootLogger = consola.create({
+  defaults: {tag: 'docutils'},
+  reporters: [new DocutilsReporter()],
+  level: globalLevel,
+});
+
+const loggers: Map<string, WeakRef<Consola>> = new Map();
```
mykola-mokhnach, line 156: may it be WeakMap instead?

packages/docutils/lib/logger.ts
```diff
@@ -137,57 +137,44 @@ class DocutilsReporter extends FancyReporter {
...
+  loggers.set(tag, new WeakRef(logger));
+  return logger;
 }
 
-/**
- * The proxied root logger
- * @see {createLogProxy}
- */
-export default rootLogger;
+export function initLogger(level: keyof typeof LogLevelMap | LogLevel) {
```
mykola-mokhnach, line 171: not sure if I understand what this method does. Why should it mutate levels of other loggers when a new logger is being initialized?

packages/docutils/lib/logger.ts
boneskull: no, the WeakMap isn't suitable for the use case for two reasons:

1. we're tracking loggers by their tag name, which is a string.  the key of a `WeakMap` _must_ be an object.
2. a `WeakMap` is not iterable, so there is no way to loop over all of them to change the log level.

packages/docutils/lib/logger.ts
boneskull: We want all loggers to have the same log level.  `initLogger` is intended to be called once -- when we determine what the logging level should be. that is after config and parsing of args.  

currently it's called twice, but that's probably wrong.

mykola-mokhnach: should it be memoized?

boneskull: I've updated it to be wrapped in `_.once()` which is like `_.memoize()` except... once. 😄

mykola-mokhnach (approved):

PR #18479, opened 2023-04-07
feat(appium): make "ls" alias of "list"

### LINKED ISSUE #18448: [Feat]: alias for `driver list`
I find myself often wanting to do `appium driver ls`, it should be easy enough to create that as an alias?

boneskull:
This makes `ls` an alias of `list`; e.g., `appium driver ls`

While doing this I noticed a limitation of `argparser`, which is that an option (not a command) can only have a single (1) alias.  I have noted this behavior in a comment.

I also notice that we need to rewrite `ls` to be `list` ourselves, or we will have a bunch of conditionals checking for both values scattered around.

We need to use yargs...

* * *

Closes #18448

### DIFF
--- a/packages/appium/lib/cli/parser.js
+++ b/packages/appium/lib/cli/parser.js
@@ -101,6 +101,13 @@ class ArgParser {
     try {
       const parsed = this.parser.parse_known_args(args);
       const [knownArgs, unknownArgs] = parsed;
+      // XXX: you'd think that argparse, when given an alias for a subcommand,
+      // would set this value to the original subcommand name, but it doesn't.
+      if (knownArgs?.driverCommand === 'ls') {
+        knownArgs.driverCommand = 'list';
+      } else if (knownArgs?.pluginCommand === 'ls') {
+        knownArgs.pluginCommand = 'list';
+      }
       if (
         unknownArgs?.length &&
         (knownArgs.driverCommand === 'run' || knownArgs.pluginCommand === 'run')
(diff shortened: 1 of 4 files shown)


boneskull:
* master
  * **PR #18479** <a href="https://app.graphite.dev/github/pr/appium/appium/18479" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18479?utm_source=stack-comment).

packages/appium/lib/cli/parser.js
```diff
@@ -242,15 +250,18 @@ class ArgParser {
...
+      for (const {command, args, help, aliases} of parserSpecs) {
+        const parser = extSubParsers.add_parser(command, {help, aliases: aliases ?? []});
 
         ArgParser._patchExit(parser);
 
         for (const [flagsOrNames, opts] of args) {
           // add_argument mutates params so make sure to send in copies instead
-          // @ts-ignore
-          parser.add_argument(...flagsOrNames, {...opts});
+          if (flagsOrNames.length === 2) {
```
mykola-mokhnach, line 260: could you please explain this magic?

mykola-mokhnach (approved):

mykola-mokhnach:
hm, lint is unhappy

KazuCocoa (approved):

packages/appium/lib/cli/parser.js
boneskull: If there was type inference available in GH's review tool (maybe this is some addon app that does this), then you probably wouldn't have asked. 😄 

`args` is of type `ArgumentDefinitions`, which is defined [here](https://github.com/appium/appium/pull/18479/files#diff-ee3a38dac1a7e9aeb4ade3c6bd97e6bcabf8690782f2666621dedcab01008a04R297).  This is a tuple, where the type of the first value (`flagsOrNames`) is `[string] | [string, string]`.  That is _also_ a tuple (an array w/ a fixed number of items) with either one or two items.  

I noticed when looking at `argparser`'s implementation that it only allows a single alias for any given arg.  So we can either give it the option name (`[string]`), or the option name and an alias (`[string, string]`).

Unfortunately, that conflicts with our config schema that says a server option or CLI argument can have an arbitrary number of aliases.  I'm punting on it for now, but I've noted the problem in comment(s).

boneskull: Of note, TS treats function parameters as tuples.  For any function `f`, you can get a tuple of its parameters via `Parameters<typeof f>`.  `argparser`'s `add_argument()` function looks like this:

```ts
add_argument(arg: string, options?: ArgumentOptions): void;
add_argument(arg1: string, arg2: string, options?: ArgumentOptions): void;
```

This is why we're only allowed the name and a single alias.

jlipps (approved):
yeah that was kinda ugly :-/

PR #18496, opened 2023-04-11
fix: upgrades for typedoc v0.24.x

boneskull:
This PR is split into two changesets; one for `@appium/docutils` and another for `@appium/typedoc-plugin-appium`. The latter is a breaking change; the former is not.

* * *

TypeDoc v0.24.0 made some breaking changes which we were bit by.  Of note:

### `@appium/docutils`:

1. Plugins are no longer automatically discovered. 😔
2. The `logger` option is no longer recognized.
3. `typedoc-plugin-resolve-crossmodule-references` is no longer needed.

I've modified `@appium/docutils` to smooth this over.

In addition, I've changed the behavior of `appium-docs init` to create a `typedoc.json` which _will_ consider `README.md`, as this is the most likely scenario for extensions.  I've changed the monorepo's config to explicitly _not_ use `README.md`.

### `@appium/typedoc-plugin-appium`

**BREAKING CHANGE: This package now requires peer dependencies `typedoc@^0.24.1` and `typedoc-plugin-markdown@^3.15.1`.**

The necessary changes here include:

- `originalName` is no longer a property on `Reflection`
- The constructor for `TypeParameterReflection` eliminated some parameters
- The `priority` property of `OptionsReader` is now `order`, and it seems to have the opposite meaning.
- The `supportsPackages` property of `OptionReader` is now required. Our plugin _only_ works with `packages`, fwiw.

_This PR supercedes #18486_


mykola-mokhnach (approved):

boneskull:
OK, this is not gonna work out as-is.  I've created https://github.com/TypeStrong/typedoc/issues/2236.

For now, to fix the build problem, we will need to pin the peer dep of `typedoc-plugin-markdown`.

PR #18519, opened 2023-04-14
fix(support): prevent s3 presigned post request failing in http upload

mohanjith:
## Proposed changes

AWS S3 expects the file field to be the last field in the POST request. Currently, it is the first field; that prevents AWS from parsing the other POST fields and the request gives a 400 error. The change makes it possible to provide formFields necessary for http upload to S3 from appium.

Use case:

mobile: stopMediaProjectionRecording can upload screen recordings to S3 for further processing

## Types of changes

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/packages/support/lib/net.js
+++ b/packages/support/lib/net.js
@@ -70,7 +70,6 @@ async function uploadFileToHttp(
   }
   if (fileFieldName) {
     const form = new FormData();
-    form.append(fileFieldName, localFileStream);
     if (formFields) {
       let pairs = [];
       if (_.isArray(formFields)) {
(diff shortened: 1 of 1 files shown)


packages/support/lib/net.js
```diff
@@ -84,6 +83,7 @@ async function uploadFileToHttp(
         }
       }
     }
+    form.append(fileFieldName, localFileStream);
```
mykola-mokhnach, line 86: Please add the explanation comment here, so this sequence won't be accidentally changed in the future

mohanjith: Added comment "AWS S3 POST upload requires this to be the last field"

mykola-mokhnach (approved):

PR #18525, opened 2023-04-17
blast docs PR preview

boneskull:
~~The PR docs preview was pointing to a directory which contains `en`, `ja`, and `zh`, but no `index.html`.~~

~~This changes the `build-docs` script, when run as a PR preview, to copy a `preview-index.html` file into the `site` dir, which will perform the redirect to `en`.~~

~~Unfortunately, there's no way to tell the action to deploy the `x` dir, but actually append e.g., `en/index.html` to the URL.  As it stands, the alternative is deploying `en` only.~~

Update: this disables the PR preview entirely, since there's [no feasible way to run it for forks](https://github.com/appium/appium/pull/18525#issuecomment-1511913195) given our current configuration.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,7 +1,7 @@
 name: Build Docs
 
 on:
-  pull_request_target:
+  pull_request:
     paths:
       - '**/typedoc.json'
       - '**/tsconfig*.json'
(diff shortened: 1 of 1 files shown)


boneskull:
@jlipps Just a heads up-- disabling the PR preview of docs build [because reasons](https://github.com/appium/appium/pull/18525#issuecomment-1511913195).

jlipps (approved):
ah ok yeah that makes sense. if we could restrict the deploy preview to maintainers only or something like that maybe it would still provide some utility. but for now, just pulling a branch and running a local preview is not super onerous.

Issue #18527, opened 2023-04-17
create a redirect from old docs site to new docs site


jlipps:
Our old docs/marketing site (https://appium.io) is defunct. We're moving to a GitHub pages hosted site (https://appium.github.io/appium). Let's create a permanent redirect from anything on appium.io to the new site.

PR #18528, opened 2023-04-17
docs: just use appium.io for github page frontend

jlipps:
I was attempting to work on https://github.com/appium/appium/issues/18527. It struck me that we should probably just replace all the content at appium.io with the new site. That would obviate the need for a redirect, keep SEO working, etc...


jlipps:
Uh I guess I actually already did it... https://appium.io/ though this PR will be necessary to fix all the README links and maybe some base url links

KazuCocoa:
lgtm.
The old one can stay in https://appium.github.io/appium.io/docs/en/about-appium/api/ so we also can guide users to check them if needed when someone asked us about the previous page's some links

PR #18547, opened 2023-04-19
feat(strongbox): allow rimraf of entire container

boneskull:
BREAKING CHANGE: The signature for an `ItemCtor` has changed. The second parameter is now a type extending `Strongbox` instead of a `container` string.  This allows for deeper integration between a custom `Item` implementation and a custom `Strongbox` implementation. Further, a custom `container` now must be an absolute path after it's been slugified.

- The `Strongbox.clearAll()` method now accepts a `force` param which, if `true`, will rimraf the entire container.
- Added two methods to the `Strongbox` class which can be overridden to accept custom options, apply defaults to those options, then validate and/or transform them.
- Added nearly-complete unit test coverage and added E2E coverage of the `BaseItem` implementation


boneskull:
@jlipps with this, I am more convinced that the implementation works.  So our problems are likely in `tizen-remote`

jlipps (approved):

boneskull:
* master
  * **PR #18547** <a href="https://app.graphite.dev/github/pr/appium/appium/18547" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18547?utm_source=stack-comment).

PR #18555, opened 2023-04-20
feat(image-plugin): Make it possible to find elements inside of other elements

mykola-mokhnach:
## Proposed changes

https://github.com/appium/java-client/issues/1891

## Types of changes

### DIFF
--- a/packages/images-plugin/lib/finder.js
+++ b/packages/images-plugin/lib/finder.js
@@ -2,15 +2,14 @@ import _ from 'lodash';
 import LRU from 'lru-cache';
 import {errors} from 'appium/driver';
 import {imageUtil} from 'appium/support';
-import {
-  ImageElement,
-  DEFAULT_TEMPLATE_IMAGE_SCALE,
-  IMAGE_EL_TAP_STRATEGY_W3C,
-} from './image-element';
-import {MATCH_TEMPLATE_MODE, compareImages, DEFAULT_MATCH_THRESHOLD} from './compare';
+import {ImageElement} from './image-element';
+import {compareImages} from './compare';
 import log from './logger';
+import {
+  DEFAULT_SETTINGS, MATCH_TEMPLATE_MODE, DEFAULT_TEMPLATE_IMAGE_SCALE,
+  DEFAULT_FIX_IMAGE_TEMPLATE_SCALE,
+} from './constants';
 
-const DEFAULT_FIX_IMAGE_TEMPLATE_SCALE = 1;
 // Used to compare ratio and screen width
 // Pixel is basically under 1080 for example. 100K is probably enough fo a while.
 const FLOAT_PRECISION = 100000;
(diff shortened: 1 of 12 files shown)


packages/images-plugin/test/unit/image-element.spec.js
```diff
@@ -48,7 +48,7 @@ describe('ImageElement', function () {
   describe('.asElement', function () {
     it('should get the webdriver object representation of the element', function () {
       const el = new ImageElement(defTemplate, defRect);
-      el.asElement('ELEMENT').ELEMENT.should.match(/^appium-image-el/);
+      el.asElement().ELEMENT.should.match(/^appium-image-el/);
```
jlipps, line 51: should we also verify the w3c element identifier key value is correct? i know this wasn't checked before but seems particularly important now that we only support w3c.

jlipps (approved):

packages/images-plugin/test/unit/image-element.spec.js
mykola-mokhnach: changed to use unwrapElement helper

mykola-mokhnach:
@jlipps Could you please check the PR one more time?

I did some more changes. I've also replaced jimp with sharp for opencv module (the first one was failing to decode some pictures with weird errors).

I would like to get rid of jimp completely for this plugin in favour of sharp and it would probably be a good idea to do the same in appium/support. I believe @boneskull will also be happy about that

packages/images-plugin/lib/finder.js
```diff
@@ -1,113 +1,60 @@
...
+   * @param {number} maxSize
    */
-  constructor(driver, maxSize = MAX_CACHE_SIZE_BYTES) {
-    this.driver = driver;
+  constructor(maxSize = MAX_CACHE_SIZE_BYTES) {
     this.imgElCache = new LRU({
       max: MAX_CACHE_ITEMS,
       maxSize,
-      sizeCalculation: (el) => el.template.length,
+      sizeCalculation: (el) => el.template.length + (el.matchedImage?.length ?? 0),
```
jlipps, line 47: ah so here's the source of the memory leak? nice find!

jlipps (approved):

jlipps:
we should probably release this as a major version bump

packages/images-plugin/lib/finder.js
mykola-mokhnach: no, it is not, but this is anyway more precise than before

mykola-mokhnach:
> we should probably release this as a major version bump

why so? I did not do any breaking changes

jlipps:
drivers used with the plugin must now implement `getWindowRect` instead of `getWindowSize`

mykola-mokhnach:
> drivers used with the plugin must now implement `getWindowRect` instead of `getWindowSize`

Ok, let me put the getWindowSize as a fallback

jlipps (approved):
OK great!

PR #18570, opened 2023-04-25
chore: Tune elements caching logic

mykola-mokhnach:
## Proposed changes

This change ensures elements don't stay in the cache for too long (max 24 hours). Also, all obsolete elements now get deleted on session deletion to make sure the cache only contains actual ones.

## Types of changes

### DIFF
--- a/packages/images-plugin/lib/plugin.js
+++ b/packages/images-plugin/lib/plugin.js
@@ -64,11 +64,15 @@ export default class ImageElementPlugin extends BasePlugin {
     // and execute the command on it
     const imgElId = getImgElFromArgs(args);
     if (imgElId) {
-      if (!this.finder.imgElCache.has(imgElId)) {
+      const imgEl = this.finder.getImageElement(imgElId);
+      if (!imgEl) {
         throw new errors.NoSuchElementError();
       }
-      const imgEl = this.finder.imgElCache.get(imgElId);
       return await ImageElement.execute(driver, imgEl, cmdName, ...args);
+    }
+
+    if (cmdName === 'deleteSession') {
+      this.finder.revokeObsoleteImageElements(driver.sessionId);
     }
 
     // otherwise just do the normal thing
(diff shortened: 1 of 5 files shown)


KazuCocoa (approved):

packages/images-plugin/lib/plugin.js
```diff
@@ -64,13 +64,17 @@ export default class ImageElementPlugin extends BasePlugin {
...
+      const imgEl = this.finder.getImageElement(imgElId);
+      if (!imgEl) {
         throw new errors.NoSuchElementError();
       }
-      const imgEl = this.finder.imgElCache.get(imgElId);
       return await ImageElement.execute(driver, imgEl, cmdName, ...args);
     }
 
+    if (cmdName === 'deleteSession') {
+      this.finder.revokeObsoleteImageElements(driver.sessionId);
```
jlipps, line 75: theoretically, each instance of a plugin will live only for the extent of a single session. so theoretically we could delete all image objects. have you noticed a case where image elements from other sessions are in the same cache? do we have a singleton cache or something?

mykola-mokhnach: hm, I thought a single plugin instance is shared accross all drivers

mykola-mokhnach: if you say there is 1:1 relationship between plugins and drivers then the thing could be slightly simplified

jlipps: that's correct. a set of active plugins are created for each session and deleted at the end of each session.

a set of "sessionless" plugins is also created for the ability to handle `createSession`, since there is not any session when `createSession` is called. but these sessionless plugins are migrated to the session plugins and also deleted at the end of the session.

each plugin instance handling a session command should therefore be isolated to that session. there's nothing to stop a session from creating a global cache, but the images plugin seems to have a per instance cache, so we should be safe to simplify.

PR #18572, opened 2023-04-25
refactor(images-plugin): Replace Jimp usages with Sharp

mykola-mokhnach:
## Proposed changes

Also changes the logic to prefer buffers over base64-encoded strings. Buffers are good because they are natively used by Sharp and also are stored  outside of the Node heap, whose size is limited.


packages/images-plugin/lib/image-element.js
```diff
@@ -88,10 +86,17 @@ export default class ImageElement {
...
+  get originalImage() {
+    return this.template.toString('base64');
+  }
+
+  /**
+   * @returns {string|null} - the base64-encoded image which has matched marks
    */
   get matchedImage() {
-    return this.b64MatchedImage;
+    return this.match?.toString('base64') ?? null;
```
jlipps, line 96: doesn't `this.match?.toString('base64')` already return null if `this.match` is null?

jlipps (approved):

packages/images-plugin/lib/image-element.js
jlipps: 👍

PR #18575, opened 2023-04-26
fix(base-driver): Ignore unknown script arguments

mykola-mokhnach:
## Proposed changes

Sometimes we might pass extra arguments to executeScript command and this should not throw exceptions like in https://dev.azure.com/srinivasansekar/java-client/_build/results?buildId=1666&view=logs&j=61693eb6-c6c2-5404-693f-d0c9a515c613&t=2ca172d0-7775-512d-edc4-316d6d6b3014&l=3205

## Types of changes

### DIFF
--- a/packages/base-driver/lib/protocol/protocol.js
+++ b/packages/base-driver/lib/protocol/protocol.js
@@ -238,9 +238,7 @@ function validateExecuteMethodParams(params, paramSpec) {
   } else {
     paramSpec.required ??= [];
     paramSpec.optional ??= [];
-    const supportedParamNames = [...paramSpec.required, ...paramSpec.optional];
-    const argNames = _.keys(args);
-    const unknownNames = _.difference(argNames, supportedParamNames);
+    const unknownNames = _.difference(_.keys(args), paramSpec.required, paramSpec.optional);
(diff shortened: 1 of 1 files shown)


packages/base-driver/lib/protocol/protocol.js
```diff
@@ -237,6 +238,11 @@ function validateExecuteMethodParams(params, paramSpec) {
   } else {
     paramSpec.required ??= [];
     paramSpec.optional ??= [];
+    const unknownNames = _.difference(_.keys(args), paramSpec.required, paramSpec.optional);
+    if (!_.isEmpty(unknownNames)) {
+      log.info(`The following script arguments are not known and will be ignored: ${unknownNames}`);
```
jlipps, line 243: would probably make this a warning as it indicates the user is doing something wrong

jlipps (approved):

packages/base-driver/lib/protocol/protocol.js
mykola-mokhnach: not necessary. We are doing it ourselves in clients to unify the endpoint for various clients

PR #18639, opened 2023-05-12
docs(appium): fix error in docs build script

jlipps:
in the case where the gh-pages branch was out of sync with local it would just fail. now we will attempt to rebase first.

also fix another error I found in mike args generation


KazuCocoa (approved):

boneskull (commented):
awesome.

I'm gonna guess you edited the snapshots manually, though.

jlipps:
yeah I saw that there was a way to update the snapshots through some CLI tool but i didn't spend time figuring that out since it was a single find/replace in this case.

PR #18650, opened 2023-05-18
fix(docutils): try python3 if python not in PATH

boneskull:
I discovered this while using WSL (Ubuntu).  `apt install python` installs `python2.7` (sigh..). To get v3 you need to `apt install python3` (and also `apt install python3-pip`).   The result is an executable named `python3`; not `python`.

This changes the `init` command to use `which` to find `python3`/`python`, as well as the validator.  `python3` is actually _preferred_, because while `python` could be python 2.x, `python3` is most certainly not.

**UPDATE**:

More changes here.

Instead of looking for `pip`, `mike` and `mkdocs` in the `PATH`, I've changed it to use `/path/to/python3 -m <module>` , because not all systems actually put these executables in the `PATH`.  If you're using, say, Homebrew on Mac, these _are_ automatically added to the `PATH`, but on e.g. Ubuntu they are not.

Furthermore, added the ability to provide a custom path to the `typedoc` executable.  This is unneeded in most cases, but can aid development at minimum.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,12 +11,6 @@
       "rules": {"func-names": "off"}
     },
     {
-      "files": ["packages/*/index.js", "packages/*/scripts/**/*.js", "test/*.js"],
-      "parserOptions": {
-        "sourceType": "script"
-      }
-    },
-    {
       "files": [
         "packages/appium/support.js",
         "packages/appium/driver.js",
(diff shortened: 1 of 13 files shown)


boneskull:
* boneskull/support-which-types
  * **PR #18650** <a href="https://app.graphite.dev/github/pr/appium/appium/18650" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18650?utm_source=stack-comment).

mykola-mokhnach:
This is unlikely to happen, but I would still prefer to point the script explicitly to python3 by default and not just python, which could be an alias to python2

mykola-mokhnach:
also macos has only python3 now

boneskull:
@mykola-mokhnach Yes, that's what it does--it looks for `python3` first.  The validator will also explicitly check the version, regardless of the name of the executable.

boneskull:
hmm, yes, but we'd still need to check the output of `python --version`.

mykola-mokhnach:
> hmm, yes, but we'd still need to check the output of `python --version`.

you mean to make sure the minimum version requirement is satisfied?

boneskull:
@mykola-mokhnach Yeah, because on my mac using homebrew, python 3 is `python`, not `python3`. on linux, `python` is python 2.  so we don't know what `python` is.

jlipps (approved):

PR #18659, opened 2023-05-19
fix(base-driver): add missing @types/lodash

boneskull:
the lack of this module was causing base-driver to fail to compile when installed as a production module.  generally, nobody is going to do such a thing--consumers will use the generated `.d.ts` files--UNLESS they happen to be running typedoc against their extension.

will need to figure out a way to test this to make sure it doesn't happen again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4741,7 +4741,8 @@
     },
     "node_modules/@types/lodash": {
       "version": "4.14.194",
-      "license": "MIT"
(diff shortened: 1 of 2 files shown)


boneskull:
fwiw, the generated `.d.ts` files were _not_ broken, because the code that fails to compile is not exported!  to that end, I'm unsure why TypeDoc can't just use `.d.ts` files.  I don't think I had any success with this before, but maybe it'll work.  If we can make it work, then we won't have to publish `tsconfig.json` files for the relevant modules.

PR #18674, opened 2023-05-22
feat(appium): Allow to provide ipv6 addresses and handle broadcast addresses in logs

mykola-mokhnach:
## Proposed changes

I can sometimes observe a situation when (non-experienced) users try to point their client to a broadcast address (like 0.0.0.0), which obviously fails. This PR improves the startup log, so we actually show the list of actual IP addresses our server listener is listening on. It also adds a possibility to provide IPv6 addresses (surprisingly the "hostname" format validator allows ipv4 addresses as well)

## Types of changes


packages/types/lib/appium-config.ts
```diff
@@ -6,9 +6,13 @@
  */
 
 /**
- * IP address to listen on
+ * IPv4/IPv6 address or a hostname to listen on
  */
-export type AddressConfig = string;
+export type AddressConfig = AddressConfig1 & AddressConfig2;
```
mykola-mokhnach, line 11: I assume this should not be happening and it must be a bug in the types generator itself. Although I have no idea how I could influence that.

mykola-mokhnach: maybe @boneskull could propose a solution?

boneskull: Since we're using `oneOf` here, you're going to have _n_ different types for each possible type.

What looks like a problem is the definition below of `AddressConfig1`:

```ts
export type AddressConfig1 = {
  [k: string]: unknown;
};
```

This is ostensibly coming from `format: 'ipv6'` , since we know that `hostname` translates to just `string`.  `ipv6` should _also_ be `string`.

Try changing `oneOf` to `anyOf`, rebuild, and check result.  The README for `json-schema-to-typescript` claims `oneOf` is "inexpressible".

mykola-mokhnach: Unfortunately changing to anyOf did not help.
Now I get this error: 

```
Type 'string | undefined' is not assignable to type '(AddressConfig1 & string) | undefined'.
  Type 'string' is not assignable to type 'AddressConfig1 & string'.
    Type 'string' is not assignable to type 'AddressConfig1'.
```

Is it a dead end?

jlipps (approved):

KazuCocoa (approved):

PR #18676, opened 2023-05-22
feat(docutils): init: scaffold package.json

boneskull:
This makes `appium-docs init` add a needed prop (`typedoc.entryPoint`) to `package.json` of an extension. It will write to `package.json` _by default_, even if it exists--which is different than the default behavior of other scaffolding tasks.

Note that `appium-docs init` will now _fail_ if `--entry-point <some-file>` (or `-e`) is not provided; there's no way for us to guess what it is (this is the _source_ entry point; not necessarily the `main` file). I guess that's technically breaking, but `appium-docs init` should not be part of anyone's workflow.

### DIFF
--- a/packages/docutils/lib/cli/index.ts
+++ b/packages/docutils/lib/cli/index.ts
@@ -8,19 +8,20 @@
 
 import {getLogger} from '../logger';
 
+import {fs} from '@appium/support';
 import _ from 'lodash';
+import {sync as readPkg} from 'read-pkg';
 import {hideBin} from 'yargs/helpers';
 import yargs from 'yargs/yargs';
 import {DEFAULT_LOG_LEVEL, LogLevelMap, NAME_BIN} from '../constants';
 import {DocutilsError} from '../error';
 import {build, init, validate} from './command';
 import {findConfig} from './config';
-import {fs} from '@appium/support';
-import {sync as readPkg} from 'read-pkg';
 
 const pkg = readPkg({cwd: fs.findRoot(__dirname)});
-
 const log = getLogger('cli');
+const IMPLICATIONS_FAILED_REGEX = /implications\s+failed:\n\s*(.+)\s->\s(.+)$/i;
+
 export async function main(argv = hideBin(process.argv)) {
   const config = await findConfig(argv);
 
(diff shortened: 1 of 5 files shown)


packages/docutils/lib/cli/index.ts
```diff
@@ -70,12 +71,34 @@ export async function main(argv = hideBin(process.argv)) {
...
+         * yargs' default output if an "implication" fails (e.g., arg _A_ requires arg _B_) leaves much to be desired.
+         *
+         * @remarks Unfortunately, we do not have access to the parsed arguments object, since it may have failed parsing.
+         * @param msg Implication failure message
+         * @returns Whether the message was an implication failure
+         */
+        const handleImplicationFailure = (msg: string | null): boolean => {
+          let match: RegExpMatchArray | null | undefined;
+          if ((match = msg?.match(IMPLICATIONS_FAILED_REGEX))) {
+            const [, arg, missingArg] = match;
```
mykola-mokhnach, line 84: I would rather flip this condition and make it to a guard

packages/docutils/lib/cli/index.ts
```diff
@@ -8,19 +8,20 @@
...
 import {DocutilsError} from '../error';
 import {build, init, validate} from './command';
 import {findConfig} from './config';
-import {fs} from '@appium/support';
-import {sync as readPkg} from 'read-pkg';
 
 const pkg = readPkg({cwd: fs.findRoot(__dirname)});
-
 const log = getLogger('cli');
+const IMPLICATIONS_FAILED_REGEX = /implications\s+failed:\n\s*(.+)\s->\s(.+)$/i;
```
mykola-mokhnach, line 23: how stable this regex is going to be?

boneskull: I don't know.  this package needs tests.  but the failure mode here is that the message is output verbatim, which is not ideal, but also not breaking.

packages/docutils/lib/cli/index.ts
boneskull: I don't understand what you're suggesting

jlipps: like
```js
if (!match) {
  return false;
}

const [, arg, missingArg] = match
```

avoids some nesting.

jlipps (approved):

mykola-mokhnach (approved):

packages/docutils/lib/cli/index.ts
boneskull: done

PR #18690, opened 2023-05-24
fix(types): fix broken appium config schema types, temporarily

boneskull:
This appears to be a bug in `json-schema-to-typescript`, where the `ipv6` type converts to `{[key: string]: unknown}` instead of `string`.

For now, we need to drop all formatting requirements from the `address` prop until this is fixed.

This did not break CI because the `build` script of `@appium/types` was run either _after_ building the rest of the project, or not at all, and the resulting change to `appium-config.ts` was not committed.

### DIFF
--- a/packages/appium/test/fixtures/flattened-schema.js
+++ b/packages/appium/test/fixtures/flattened-schema.js
@@ -11,14 +11,6 @@ export default [
       ref: 'appium.json#/properties/server/properties/address',
     },
     schema: {
-      anyOf: [
-        {
-          format: 'hostname'
-        },
-        {
-          format: 'ipv6'
-        }
-      ],
       appiumCliAliases: ['a'],
       default: '0.0.0.0',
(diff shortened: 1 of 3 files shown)


boneskull:
This also changes the build scripts to generate `appium-config.ts` _before_ compiling the rest of the project.

jlipps (approved):

packages/appium/test/fixtures/flattened-schema.js
```diff
@@ -11,14 +11,6 @@ export default [
       ref: 'appium.json#/properties/server/properties/address',
     },
     schema: {
-      anyOf: [
```
mykola-mokhnach, line 14: I am not happy about dropping the whole argument validation just because of a bug in typing.
At least the validation should be added elsewhere instead of removing it completely

boneskull: I'm not happy about it either.  I'll create an issue to deal w/ this

PR #18720, opened 2023-06-01
chore(ci): fix invalid workflow

boneskull:

PR #18721, opened 2023-06-02
sundry type fixes

boneskull:
Mainly:

**BREAKING CHANGE: This changes the def of `findElOrEls` and `findElOrElsWithProcessing` in `ExternalDriver` in a breaking manner.**

So conditional types don't work like how I was using them.  They work in capital-T-types (`type`), but they don't work as return values.  Instead, what we should have is an overload.  in this case, the function can be called one of two ways, and it returns a different thing depending on those values.  _unfortunately_ this is a little hinky to implement, given that overloads are unsupported in _object literals_.  they are supported in function statements or expressions, and they are supported in class definitions, but not object literals.  you'll see what needs to happen to make this work in `FakeDriver` and `BaseDriver` in subsequent changesets; essentially the function must be defined outside of the object literal and then stuffed in there.

The other change here changes an `object` type (which does not allow arbitrary properties) to a `StringRecord` type within the type which converts `Constraints` to `Capabilities`.  This affects caps for some drivers such as `chromedriverArgs` in `appium-android-driver`; anywhere where the constraint has `isObject: true`.

### DIFF
--- a/packages/base-driver/lib/basedriver/commands/find.ts
+++ b/packages/base-driver/lib/basedriver/commands/find.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
 import {Constraints, Element, IFindCommands} from '@appium/types';
 import {errors} from '../../protocol';
 import {BaseDriver} from '../driver';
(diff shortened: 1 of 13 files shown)


boneskull:
* master
  * **PR #18720** <a href="https://app.graphite.dev/github/pr/appium/appium/18720" target="_blank">Graphite</a> 
    * **PR #18721** <a href="https://app.graphite.dev/github/pr/appium/appium/18721" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium/18721?utm_source=stack-comment).

jlipps (commented):
is this a breaking change for drivers written in JS? or just those using TS/TS-in-JSDoc?

jlipps (approved):

boneskull:
@jlipps they break types only

jlipps (approved):

PR #18733, opened 2023-06-05
chore(appium): Make sharp to an optional dependency

mykola-mokhnach:
## Proposed changes

sharp is using native code. It is deployed in a form of downloadable precompiled binaries, but users may have issues fetching these because of possible networking or system issues. And since sharp module as a part of Appium server is only used rarely for various support calls it does not make sense to fail the whole server setup if it cannot be deployed.

### DIFF
--- a/packages/support/lib/image-util.js
+++ b/packages/support/lib/image-util.js
@@ -1,14 +1,32 @@
-import sharp from 'sharp';
+let _sharp;
+
+/**
+ * @returns {import('sharp')}
+ */
+function requireSharp() {
+  if (!_sharp) {
+    try {
+      _sharp = require('sharp');
+    } catch (err) {
+      throw new Error(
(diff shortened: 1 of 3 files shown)


packages/support/lib/image-util.js
```diff
@@ -1,14 +1,32 @@
...
+/**
+ * @returns {import('sharp')}
+ */
+function requireSharp() {
+  if (!_sharp) {
+    try {
+      _sharp = require('sharp');
+    } catch (err) {
+      throw new Error(
+        `Cannot load the 'sharp' module needed for images processing. ` +
```
jlipps, line 12: maybe add a line like `Automatic install of sharp failed during Appium npm install`  here?

jlipps (approved):

jlipps (approved):

DonMcNamara:
Thank you for making this change. This sharp dependency is really complicating my builds.

PR #18562, opened 2023-04-23
feat(base-driver): add https server support

devhazem:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

I have implemented this feature as I need appium to run on HTTPS to make communication from an application running on HTTPS posssible.

## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [X] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the CLA
- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #18754, opened 2023-06-12
fix: ensure logServerAddress gets an actual URL

jlipps:
The server was failing to start entirely. How was CI passing before?

### DIFF
--- a/packages/appium/lib/main.js
+++ b/packages/appium/lib/main.js
@@ -167,8 +167,8 @@ async function prepareAppiumHome(name, appiumHome) {
     }
     throw new Error(
       `The path '${appiumHome}' provided in the ${name} must point ` +
-      `to a valid folder writeable for the current user account '${os.userInfo().username}'. ` +
-      `Original error: ${err.message}`
+        `to a valid folder writeable for the current user account '${os.userInfo().username}'. ` +
+        `Original error: ${err.message}`
     );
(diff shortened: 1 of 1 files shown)


boneskull (approved):

packages/appium/lib/main.js
```diff
@@ -442,7 +442,7 @@ async function main(args) {
     });
   }
 
-  logServerAddress(`${parsedArgs.address}:${parsedArgs.port}${parsedArgs.basePath}`);
+  logServerAddress(`http://${parsedArgs.address}:${parsedArgs.port}${parsedArgs.basePath}`);
```
mykola-mokhnach, line 445: not sure if it is a good idea to hardcode `http`. We could use `parsedArgs.scheme` in order to make it more flexible/error prone

jlipps: yeah good idea. i was looking for some arg like this but couldn't find it. will update.

jlipps: actually @mykola-mokhnach i grepped the codebase for `scheme` but don't see it as an appium arg?

mykola-mokhnach: ah, looks like that [PR](https://github.com/appium/appium/pull/18562) has never been merged. Sorry for the confusion

PR #18751, opened 2023-06-09
fix(docutils): fixes some weirdness with default behavior and the "implies" keyword

boneskull:
- fixes some weird broken CLI arg stuff, I think; see https://github.com/appium/appium-xcuitest-driver/issues/1741
- also squashes an errant warning
- fix some diff logic when writing to `mkdocs.yml`
- add missing `@types/source-map-support` dev dep

PR #18770, opened 2023-06-14
fix(docutils): set mike version to major.minor by default

jlipps:
an attempt to resolve https://github.com/appium/appium/pull/18751#issuecomment-1591783397

### DIFF
--- a/packages/docutils/lib/builder/deploy.ts
+++ b/packages/docutils/lib/builder/deploy.ts
@@ -68,7 +68,7 @@ async function findDeployVersion(packageJsonPath?: string, cwd = process.cwd()):
   // return MAJOR.MINOR as the version by default, if that is a thing we can extract, otherwise
   // just return the version as is
   const versionParts = version.split('.');
-  if (versionParts.length < 2) {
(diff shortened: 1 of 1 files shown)


jlipps:
@boneskull do you think this will work?

packages/docutils/lib/builder/deploy.ts
```diff
@@ -68,7 +64,14 @@ async function findDeployVersion(packageJsonPath?: string, cwd = process.cwd()):
       'No "version" field found in package.json; please add one or specify a version to deploy'
     );
   }
-  return version;
+
+  // return MAJOR.MINOR as the version by default, if that is a thing we can extract, otherwise
+  // just return the version as is
+  const versionParts = version.split('.');
+  if (versionParts.length < 2) {
```
boneskull, line 71: It's not possible for the length of `versionParts` to be `0`.

```suggestion
  if (versionParts.length === 1) {
```

boneskull (approved):
I _think_ this is OK.

you know `mike` better than I do, though

PR #18809, opened 2023-06-26
chore(appium): Use the actual server address in the logged URL

mykola-mokhnach:
## Proposed changes

The normalization is applied to it on server creation, so we want to reflect the actual URL used for that server.

## Types of changes


KazuCocoa (approved):
lgtm

PR #18820, opened 2023-06-29
chore(base-driver,driver-test-support): do not run basedriver e2e tests in parallel

boneskull:
also: rewrote event timings tests due to it being unclear which assertion actually failed. one assertion per test.


jlipps (approved):

boneskull:
this should fix the failing master build

mykola-mokhnach (approved):

boneskull:
slow af

PR #18916, opened 2023-07-26
docs(appium): Added correct doc url path for images comparison and settings

bipin-k:
## Proposed changes

* Added correct doc path for images comparison and settings

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments
N/A

### DIFF
--- a/packages/appium/lib/utils.js
+++ b/packages/appium/lib/utils.js
@@ -280,7 +280,7 @@ function adjustNodePath() {
  * The capabilities argument itself gets mutated, so it does not contain parsed
  * settings anymore to avoid further parsing issues.
  * Check
- * https://github.com/appium/appium/blob/master/docs/en/advanced-concepts/settings.md
+ * https://appium.io/docs/en/latest/guides/settings/
  * for more details on the available settings.
  *
  * @param {?Object} caps - Capabilities dictionary. It is mutated if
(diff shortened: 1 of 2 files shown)


packages/appium/lib/utils.js
```diff
@@ -280,7 +280,7 @@ function adjustNodePath() {
  * The capabilities argument itself gets mutated, so it does not contain parsed
  * settings anymore to avoid further parsing issues.
  * Check
- * https://github.com/appium/appium/blob/master/docs/en/advanced-concepts/settings.md
+ * https://appium.io/docs/en/2.0/guides/settings/
```
eglitise, line 283: For future compatibility it may be better to change this to `https://appium.io/docs/en/latest/guides/settings/`

bipin-k: @eglitise The url as mentioned, `https://appium.io/docs/en/latest/guides/settings/` is getting being redirected to `https://appium.io/docs/en/2.0/guides/settings/`. Do you still want me to make the changes as suggested by you?

eglitise: Yes, it's supposed to redirect, I'm just thinking of the possibility when the docs eventually get updated to 2.1, 2.2, etc, in which case this link may become outdated. @jlipps what do you think?

jlipps: yes, ideally all our doc links should refer to `latest` so that the redirect gets to the most current version.

jlipps (approved):

PR #18922, opened 2023-07-29
fix(base-driver): Tune responses caching logic

mykola-mokhnach:
Related to https://github.com/appium/dotnet-client/issues/625

cc @Dor-bl

### DIFF
--- a/packages/base-driver/lib/express/idempotency.js
+++ b/packages/base-driver/lib/express/idempotency.js
@@ -34,17 +34,15 @@ function cacheResponse(key, req, res) {
     responseStateListener,
   });
   const originalSocketWriter = res.socket.write.bind(res.socket);
-  let response = null;
-  const patchedWriter = (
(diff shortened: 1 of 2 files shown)


KazuCocoa (approved):
it makes sense, lgtm

jlipps (approved):
changes LGTM but I can't say I really understand the point of this code to begin with

mykola-mokhnach:
> changes LGTM but I can't say I really understand the point of this code to begin with

The point of it is to make session creation requests idempontent. Possible scenario:
- The client sends a POST /session request with the idempotency key header '1234'
- The client times out because server takes too long to complete the session creation
- The client retries by sending another POST /session  request, but with the same idempotency key header '1234'
- The server knows it is the same session creation request, so instead of starting doing mess with the current session being created it just connects the retry request to the already running one and replies with the same data

Issue #18969, opened 2023-08-09
[Feat]: initialize settings for new session via single combined capability


eglitise:
**Is your feature request related to a problem? Please describe.**
When starting a new session, if I want to initialize some settings from the get-go, the documentation suggests the syntax `appium:settings[<name>]`. This works well, but it looks a little messy when there's many settings to be changed.

**Describe the solution you'd like**
Support a syntax similar to the combined `appium:options` capability, but for use with settings, e.g.
```json
{
  "appium:settings": {
    "enableMultiWindows": true,
    "waitForIdleTimeout": 50
  }
}
```

**Describe alternatives you've considered**
Writing each setting in the current syntax continues to work fine.

**Additional context**
N/A

eglitise:
After updating to Appium 2.1, I can confirm this is working. Thanks!

PR #18970, opened 2023-08-09
feat(base-driver): Make it possible to provide settings as a map in session capabilities

mykola-mokhnach:
## Proposed changes

Created in response to https://github.com/appium/appium/issues/18969
Although this might be a breaking change if any of existing drivers or plugins already defines a capability called `settings`.

## Types of changes

### DIFF
--- a/packages/appium/docs/en/guides/settings.md
+++ b/packages/appium/docs/en/guides/settings.md
@@ -41,6 +41,18 @@ would construct a set of capabilities that includes the following in its JSON re
 }
 ```
 
+Also, since base-driver version 9.4.0, there is a possibility to provide multiple settings
+in a single `appium:settings` capability value:
+
+```json
+{
+    "appium:settings": {
+        "ignoreUnimportantViews": true,
+        "allowInvisibleElements": true
+    }
+}
+```
+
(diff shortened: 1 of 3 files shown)


KazuCocoa:
> Although this might be a breaking change ...

yea, correct. Hm, should we add a warning message such as _XXX is already reserved by Appium base driver_ for drivers/plugins...?

eglitise:
Thanks for this! I would also suggest adding this info to the Settings API docs page (though perhaps it's safer to do so after the feature is released)

jlipps (approved):

jlipps:
i think we might as well include a settings docs update in this PR to make sure we don't forget.

mykola-mokhnach:
@jlipps Which particular document should be updated?

eglitise:
I think this info should go in [guides/settings.md](https://github.com/appium/appium/blob/master/packages/appium/docs/en/guides/settings.md). Some info can be taken from the `appium:options` docs in [guides/caps.md](https://github.com/appium/appium/blob/master/packages/appium/docs/en/guides/caps.md#appiumoptions), for example, the warning about argument precedence (though in this case the first value seems to take precedence).

KazuCocoa (approved):

PR #18972, opened 2023-08-10
fix(appium): Update the GET /status response to be in sync with the standard

mykola-mokhnach:
## Proposed changes

According to https://www.w3.org/TR/webdriver/#dfn-status the status response must contain `ready` and `message` fields

## Types of changes


KazuCocoa (approved):

jlipps (approved):

jlipps:
should we not call this `fix` instead of `chore` since we are correcting the status response?

PR #19098, opened 2023-08-31
chore(base-driver): Tune the log message for non-recognised capabilities

mykola-mokhnach:
## Proposed changes

Capabilities are specific to the driver, it should not mention the whole ecosystem in the warning message


KazuCocoa (approved):

jlipps (commented):
yep good point

PR #19105, opened 2023-09-02
feat(base-driver): Add server support of TLS and SPDY protocols

mykola-mokhnach:
## Proposed changes

It's a continuation of https://github.com/appium/appium/pull/18562, but more reach on features being proposed .
It allows to provide the certificate and the key via server command line or via config options. If these both options are provided and are valid then the server starts in strict TLS mode, which also allows to enable SPDY and HTTP/2 extensions on it. See https://www.npmjs.com/package/spdy for more details.

## Types of changes

### DIFF
--- a/packages/base-driver/package.json
+++ b/packages/base-driver/package.json
@@ -66,6 +66,7 @@
     "method-override": "3.0.0",
     "morgan": "1.10.0",
     "serve-favicon": "2.5.0",
+    "spdy": "4.0.2",
     "source-map-support": "0.5.21",
     "type-fest": "3.13.1",
     "validate.js": "0.13.1"
--- a/packages/base-driver/package.json
+++ b/packages/base-driver/package.json
@@ -66,10 +66,12 @@
     "method-override": "3.0.0",
     "morgan": "1.10.0",
     "serve-favicon": "2.5.0",
-    "spdy": "4.0.2",
     "source-map-support": "0.5.21",
     "type-fest": "3.13.1",
     "validate.js": "0.13.1"
+  },
+  "optionalDependencies": {
+    "spdy": "4.0.2"
   },
   "engines": {
     "node": "^14.17.0 || ^16.13.0 || >=18.0.0",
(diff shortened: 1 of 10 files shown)


mykola-mokhnach:
cc @devhazem

packages/base-driver/package.json
```diff
@@ -66,6 +66,7 @@
     "method-override": "3.0.0",
     "morgan": "1.10.0",
     "serve-favicon": "2.5.0",
+    "spdy": "4.0.2",
```
KazuCocoa, line 69: Could be an optional deps? Perhaps this usage is limited so it should not block installation as possible failures

mykola-mokhnach: yes, I can change that

packages/base-driver/test/e2e/express/server.e2e.spec.js
```diff
@@ -3,8 +3,19 @@ import axios from 'axios';
 import {createSandbox} from 'sinon';
 import B from 'bluebird';
 import _ from 'lodash';
+import { exec } from 'teen_process';
+import https from 'node:https';
 import {TEST_HOST, getTestPort} from '@appium/driver-test-support';
 
+async function generateCertificate (certPath, keyPath) {
```
jlipps, line 10: this test will not run on windows i'm assuming?

packages/schema/lib/appium-config-schema.js
```diff
@@ -288,6 +288,18 @@ export const AppiumConfigJsonSchema = /** @type {const} */ ({
           title: 'webhook config',
           type: 'string',
         },
+        'ssl-certificate-path': {
```
jlipps, line 291: `ssl-cert-path` might be less verbose and equally descriptive

jlipps (approved):

packages/base-driver/test/e2e/express/server.e2e.spec.js
mykola-mokhnach: it will only if openssl is installed and present in PATH

packages/schema/lib/appium-config-schema.js
mykola-mokhnach: changed

DmitryTheMan:
Hey, I would like to use issued SSL certificate so that I got https, how could I do it now?

Issue #19331, opened 2023-10-21
bug: Repeatedly reducing the template image


gointoit:
### Do I have the most recent component updates?

- [X] I use the most recent available driver/plugin and server versions

### Is the component officially supported by the Appium team?

- [X] I have verified the component repository is present under the Appium organization in GitHub

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

**when set implicitlyWait**
 Image matching fail will rerty within this implicitlyWait interval
 But every time try, scaling template image repeatedly which is incorrect,
 It causes template images to become smaller and smaller

**use: findByImage()**
**config:**
`drv.manage().timeouts().implicitlyWait(DEFAULT_APPIUM_IMPLICITLY_WAIT)`

`setting(drv, Setting.IMAGE_MATCH_THRESHOLD.toString(), "0.82");

setting(drv, Setting.FIX_IMAGE_FIND_SCREENSHOT_DIMENSIONS.toString(), true);
        setting(drv, Setting.FIX_IMAGE_TEMPLATE_SIZE.toString(), false);
        setting(drv, Setting.FIX_IMAGE_TEMPLATE_SCALE.toString(), true);
        setting(drv, Setting.CHECK_IMAGE_ELEMENT_STALENESS.toString(), false);
        setting(drv, Setting.UPDATE_IMAGE_ELEMENT_POSITION.toString(), false);`

**Log:**
![image](https://github.com/appium/appium/assets/14820842/745fb8a8-4175-477d-a95d-be0865bf7ce8)

### Expected Behavior

just scaling template image only Once when matching failed and retry

### Minimal Reproducible Example

Just give a picture that doesn't match and  set implicitlyWait = 5s,
Then trigger a matching retry，

### Environment

- Operating system: MAC
- Appium server version (output of `appium --version`): v2.2.1
- Appium driver(s) and their version(s): uiautomator2@2.29.2,xcuitest@4.34.0
- Appium plugin(s) and their version(s):images@2.1.6
- Node.js version (output of `node --version`):v16.18.0
- `npm` version (output of `npm --version`):8.19.2
- Last component(s) version which did _not_ exhibit the problem:
- Platform and version under test: ios 15.2 ,iPhone 12
- Real device or emulator/simulator: both

mykola-mokhnach:
It is a valid bug. Thanks for the detailed description

mykola-mokhnach:
The issue has been fixed in master and will be available in the next server release

PR #19334, opened 2023-10-22
fix(images-plugin): Prevent repeatable template modification

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/appium/appium/issues/19331
Also enabled better typing on the plugin

## Types of changes

### DIFF
--- a/packages/images-plugin/lib/finder.js
+++ b/packages/images-plugin/lib/finder.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import { LRUCache } from 'lru-cache';
+// @ts-ignore It does export errors
 import {errors} from 'appium/driver';
 import {ImageElement} from './image-element';
 import {compareImages} from './compare';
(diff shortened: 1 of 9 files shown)


packages/images-plugin/lib/finder.js
```diff
@@ -136,16 +136,23 @@ export default class ImageElementFinder {
...
-          ...scale,
-        });
+        if (!didFixTemplateImageScale) {
+          template = await this.fixImageTemplateScale(template, {
+            defaultImageTemplateScale,
+            ignoreDefaultImageTemplateScale,
+            fixImageTemplateScale,
+            ...(scale || {}),
+          });
+          // We do not want `template` to be mutated multiple times when the
```
KazuCocoa, line 152: oh, i see. the lambda did multiple times..

KazuCocoa (approved):
lgtm

packages/images-plugin/lib/finder.js
jlipps: 👍 good catch

PR #19403, opened 2023-11-10
chore: run prettier on some files

jlipps:
Working on these files for the bidi implementation And want a clean PR to review without all this junk


jlipps:
If tests pass this should be an auto-merge, it's just reformatting.

PR #19404, opened 2023-11-10
feat: add webdriver bidi support

jlipps:
This is a major new feature for Appium, enabling [WebDriver BiDi](https://w3c.github.io/webdriver-bidi) support for Appium drivers. The main idea:

- Add websocket handlers for routes to the main appium server
- Define a set of bidi commands which drivers can implement the same way they implement webdriver classic commands
- Allow drivers to declare whether they support bidi
- If a driver supports bidi, then clients can make bidi connections, and commands/params will be unwrapped and applied to the methods implemented by drivers. I.e., the main Appium server will handle all aspects of the BiDi protocol, and drivers don't need to worry about it.
- Default implementation for subscription/unsubscription to bidi events
- Allow drivers to emit bidi events which will be pushed to the client if the client is subscribed
- Add bidi support for FakeDriver (a simple bidi command and a simple bidi event) for testing
- Allow bidi proxying. If a driver sets a bidi proxy url for an upstream socket server (e.g., appium-chromium-driver wanting to proxy bidi connections to chromedriver), then Appium will transparently pass all socket communication back and forth instead of handling 

TODO:
- [ ] allow drivers to define custom bidi commands (similar to how they can define custom webdriver commands) (separate PR)
- [ ] add lots more tests (separate PR)
- [x] add documentation for driver authors on what they need to do to implement bidi support in their drivers
- [ ] fill out the bidi command list with all the currently supported commands from the spec (separate PR)
- [ ] on the current implementation, proxying is single-shot; you can turn it on but you can't toggle it on and off. Unclear if this is desirable but it's how our current webdriver proxying works (separate PR)
- [ ] evaluate whether we want "bidi proxy avoid" commands like we have "webdriver proxy avoid" routes (separate PR)
- [ ] allow create/delete session to work from the 'global' / 'sessionless' bidi connection (unclear if desirable, but it's part of the spec, kinda) (separate PR)

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-unused-vars */
 import _ from 'lodash';
+import B from 'bluebird';
 import {getBuildInfo, updateBuildInfo, APPIUM_VER} from './config';
 import {
   BaseDriver,
@@ -9,6 +10,7 @@ import {
   CREATE_SESSION_COMMAND,
   DELETE_SESSION_COMMAND,
   GET_STATUS_COMMAND,
+  MAX_LOG_BODY_LENGTH,
   promoteAppiumOptions,
   promoteAppiumOptionsForObject,
 } from '@appium/base-driver';
@@ -16,7 +18,8 @@ import AsyncLock from 'async-lock';
 import {parseCapsForInnerDriver, pullSettings, makeNonW3cCapsError} from './utils';
 import {util, node, logger} from '@appium/support';
 import {getDefaultsForExtension} from './schema';
-import {DRIVER_TYPE} from './constants';
+import {DRIVER_TYPE, BIDI_BASE_PATH, BIDI_EVENT_NAME} from './constants';
+import WebSocket from 'ws';
 
 const desiredCapabilityConstraints = /** @type {const} */ ({
   automationName: {
@@ -83,6 +86,12 @@ class AppiumDriver extends DriverCore {
   /** @type {AppiumServer} */
   server;
 
+  /** @type {Record<string, import('ws').WebSocket[]>} */
+  bidiSockets;
+
+  /** @type {Record<string, import('ws').WebSocket>} */
+  bidiProxyClients;
+
   /**
    * @type {AppiumDriverConstraints}
    * @readonly
@@ -113,6 +122,8 @@ class AppiumDriver extends DriverCore {
     this.pluginClasses = new Map();
     this.sessionPlugins = {};
     this.sessionlessPlugins = [];
+    this.bidiSockets = {};
+    this.bidiProxyClients = {};
     this.desiredCapConstraints = desiredCapabilityConstraints;
     this._isShuttingDown = false;
 
@@ -239,6 +250,337 @@ class AppiumDriver extends DriverCore {
         return cliArgs;
       }
     }
+  }
+
+  /**
+   * Initialize a new bidi connection and set up handlers
+   * @param {import('ws').WebSocket} ws The websocket connection object
+   * @param {import('http').IncomingMessage} req The connection pathname, which might include the session id
+   */
+  onBidiConnection(ws, req) {
+    // TODO put bidi-related functionality into a mixin/helper class
+    // wrap all of the handler logic with exception handling so if something blows up we can log
+    // and close the websocket
+    try {
+      const {bidiHandlerDriver, proxyClient, send, sendToProxy, logSocketErr} = this.initBidiSocket(
+        ws,
+        req,
+      );
+
+      this.initBidiSocketHandlers(
+        ws,
+        proxyClient,
+        send,
+        sendToProxy,
+        bidiHandlerDriver,
+        logSocketErr,
+      );
+      this.initBidiProxyHandlers(proxyClient, ws, send);
+      this.initBidiEventListeners(ws, bidiHandlerDriver, send);
+    } catch (err) {
+      this.log.error(err);
+      try {
+        ws.close();
+      } catch (ign) {}
+    }
+  }
+
+  /**
+   * Initialize a new bidi connection
+   * @param {import('ws').WebSocket} ws The websocket connection object
+   * @param {import('http').IncomingMessage} req The connection pathname, which might include the session id
+   */
+  initBidiSocket(ws, req) {
+    let outOfBandErrorPrefix = '';
+    const pathname = req.url;
+    if (!pathname) {
+      throw new Error('Invalid connection request: pathname missing from request');
+    }
+    const bidiSessionRe = new RegExp(`${BIDI_BASE_PATH}/([^/]+)$`);
+    const bidiNoSessionRe = new RegExp(`${BIDI_BASE_PATH}/?$`);
+    const sessionMatch = bidiSessionRe.exec(pathname);
+    const noSessionMatch = bidiNoSessionRe.exec(pathname);
+
+    if (!sessionMatch && !noSessionMatch) {
+      throw new Error(
+        `Got websocket connection for path ${pathname} but didn't know what to do with it. ` +
+          `Ignoring and will close the connection`,
+      );
+    }
+
+    // Let's figure out which driver is going to handle this socket connection. It's either going
+    // to be a driver matching a session id appended to the bidi base path, or this umbrella driver
+    // (if no session id is included in the bidi connection request)
+
+    /** @type {import('@appium/types').ExternalDriver | AppiumDriver} */
+    let bidiHandlerDriver;
+
+    /** @type {import('ws').WebSocket | null} */
+    let proxyClient = null;
+
+    if (sessionMatch) {
+      // If we found a session id, see if it matches an active session
+      const sessionId = sessionMatch[1];
+      bidiHandlerDriver = this.sessions[sessionId];
+      if (!bidiHandlerDriver) {
+        // The session ID sent in doesn't match an active session; just ignore this socket
+        // connection in that case
+        throw new Error(
+          `Got bidi connection request for session with id ${sessionId} which is closed ` +
+            `or does not exist. Closing the socket connection.`,
+        );
+      }
+      const driverName = bidiHandlerDriver.constructor.name;
+      outOfBandErrorPrefix = `[session ${sessionId}] `;
+      this.log.info(`Bidi websocket connection made for session ${sessionId}`);
+      // store this socket connection for later removal on session deletion. theoretically there
+      // can be multiple sockets per session
+      if (!this.bidiSockets[sessionId]) {
+        this.bidiSockets[sessionId] = [];
+      }
+      this.bidiSockets[sessionId].push(ws);
+
+      const bidiProxyUrl = bidiHandlerDriver.bidiProxyUrl;
+      if (bidiProxyUrl) {
+        try {
+          new URL(bidiProxyUrl);
+        } catch (ign) {
+          throw new Error(
+            `Got request for ${driverName} to proxy bidi connections to upstream socket with ` +
+              `url ${bidiProxyUrl}, but this was not a valid url`,
+          );
+        }
+        this.log.info(`Bidi connection for ${driverName} will be proxied to ${bidiProxyUrl}`);
+        proxyClient = new WebSocket(bidiProxyUrl);
+        this.bidiProxyClients[sessionId] = proxyClient;
+      }
+    } else {
+      this.log.info('Bidi websocket connection made to main server');
+      // no need to store the socket connection if it's to the main server since it will just
+      // stay open as long as the server itself is and will close when the server closes.
+      bidiHandlerDriver = this; // eslint-disable-line @typescript-eslint/no-this-alias
+    }
+
+    const logSocketErr = (/** @type {Error} */ err) =>
+      this.log.error(`${outOfBandErrorPrefix}${err}`);
+
+    // This is a function which wraps the 'send' method on a web socket for two reasons:
+    // 1. Make it async-await friendly
+    // 2. Do some logging if there's a send error
+    const sendFactory = (/** @type {import('ws').WebSocket} */ socket) => {
+      const socketSend = B.promisify(socket.send, {context: socket});
+      return async (/** @type {string|Buffer} */ data) => {
+        try {
+          await socketSend(data);
+        } catch (err) {
+          logSocketErr(err);
+        }
+      };
+    };
+
+    // Construct our send method for sending messages to the client
+    const send = sendFactory(ws);
+
+    // Construct a conditional send method for proxying messages from the client to an upstream
+    // bidi socket server (e.g. on a browser)
+    const sendToProxy = proxyClient ? sendFactory(proxyClient) : null;
+
+    return {bidiHandlerDriver, proxyClient, send, sendToProxy, logSocketErr};
+  }
+
+  /**
+   * Set up handlers on upstream bidi socket we are proxying to/from
+   *
+   * @param {import('ws').WebSocket | null} proxyClient - the websocket connection to/from the
+   * upstream socket (the one we're proxying to/from)
+   * @param {import('ws').WebSocket} ws - the websocket connection to/from the client
+   * @param {(data: string | Buffer) => Promise<void>} send - a method used to send data to the
+   * client
+   */
+  initBidiProxyHandlers(proxyClient, ws, send) {
+    // Set up handlers for events that might come from the upstream bidi socket connection if
+    // we're in proxy mode
+    if (proxyClient) {
+      // Here we're receiving a message from the upstream socket server. We want to pass it on to
+      // the client
+      proxyClient.on('message', async (/** @type {Buffer|string} */ data) => {
+        const logData = _.truncate(data.toString('utf8'), {length: MAX_LOG_BODY_LENGTH});
+        this.log.debug(
+          `<-- BIDI Received data from proxied bidi socket, sending to client. Data: ${logData}`,
+        );
+        await send(data);
+      });
+
+      // If the upstream socket server closes the connection, should close the connection to the
+      // client as well
+      proxyClient.on('close', (code, reason) => {
+        this.log.debug(
+          `Upstream bidi socket closed connection (code ${code}, reason: '${reason}'). ` +
+            `Closing proxy connection to client`,
+        );
+        ws.close(code, reason);
+      });
+
+      proxyClient.on('error', (err) => {
+        this.log.error(`Got error on upstream bidi socket connection: ${err}`);
+      });
+    }
+  }
+
+  /**
+   * Set up handlers on the bidi socket connection to the client
+   *
+   * @param {import('ws').WebSocket} ws - the websocket connection to/from the client
+   * @param {import('ws').WebSocket | null} proxyClient - the websocket connection to/from the
+   * upstream socket (the one we're proxying to/from, if we're proxying)
+   * @param {(data: string | Buffer) => Promise<void>} send - a method used to send data to the
+   * client
+   * @param {((data: string | Buffer) => Promise<void>) | null} sendToProxy - a method used to send data to the
+   * upstream socket
+   * @param {import('@appium/types').ExternalDriver | AppiumDriver} bidiHandlerDriver - the driver
+   * handling the bidi commands
+   * @param {(err: Error) => void} logSocketErr - a special prefixed logger
+   */
+  initBidiSocketHandlers(ws, proxyClient, send, sendToProxy, bidiHandlerDriver, logSocketErr) {
+    ws.on('error', (err) => {
+      // Can't do much with random errors on the connection other than log them
+      logSocketErr(err);
+    });
+
+    ws.on('open', () => {
+      this.log.info('Bidi websocket connection is now open');
+    });
+
+    // Now set up handlers for the various events that might happen on the websocket connection
+    // coming from the client
+    // First is incoming messages from the client
+    ws.on('message', async (/** @type {Buffer} */ data) => {
+      if (proxyClient) {
+        const logData = _.truncate(data.toString('utf8'), {length: MAX_LOG_BODY_LENGTH});
+        this.log.debug(
+          `--> BIDI Received data from client, sending to upstream bidi socket. Data: ${logData}`,
+        );
+        // if we're meant to proxy to an upstream bidi socket, just do that
+        // @ts-ignore sendToProxy is never null if proxyClient is truthy, but ts doesn't know
+        // that
+        await sendToProxy(data.toString('utf8'));
+      } else {
+        const res = await this.onBidiMessage(data, bidiHandlerDriver);
+        await send(JSON.stringify(res));
+      }
+    });
+
+    // Next consider if the client closes the socket connection on us
+    ws.on('close', (code, reason) => {
+      // Not sure if we need to do anything here if the client closes the websocket connection.
+      // Probably if a session was started via the socket, and the socket closes, we should end the
+      // associated session to free up resources. But otherwise, for sockets attached to existing
+      // sessions, doing nothing is probably right.
+      this.log.debug(`Bidi socket connection closed (code ${code}, reason: '${reason}')`);
+
+      // If we're proxying, might as well close the upstream connection and clean it up
+      if (proxyClient) {
+        this.log.debug('Also closing bidi proxy socket connection');
+        proxyClient.close(code, reason);
+      }
+    });
+  }
+
+  /**
+   * Set up bidi event listeners
+   *
+   * @param {import('ws').WebSocket} ws - the websocket connection to/from the client
+   * @param {import('@appium/types').ExternalDriver | AppiumDriver} bidiHandlerDriver - the driver
+   * handling the bidi commands
+   * @param {(data: string | Buffer) => Promise<void>} send - a method used to send data to the
+   * client
+   */
+  initBidiEventListeners(ws, bidiHandlerDriver, send) {
+    // If the driver emits a bidi event that should maybe get sent to the client, check to make
+    // sure the client is subscribed and then pass it on
+    let eventListener = async ({context, method, params}) => {
+      // if the driver didn't specify a context, use the empty context
+      if (!context) {
+        context = '';
+      }
+      if (!method || !params) {
+        throw new Error(
+          `Driver emitted a bidi event that was malformed. Require method and params keys ` +
+            `(with optional context). But instead received: ${JSON.stringify({
+              context,
+              method,
+              params,
+            })}`,
+        );
+      }
+      if (ws.readyState !== WebSocket.OPEN) {
+        // if the websocket is not still 'open', then we can ignore sending these events
+        if (ws.readyState > WebSocket.OPEN) {
+          // if the websocket is closed or closing, we can remove this listener as well to avoid
+          // leaks
+          bidiHandlerDriver.eventEmitter.removeListener(BIDI_EVENT_NAME, eventListener);
+        }
+        return;
+      }
+
+      if (bidiHandlerDriver.bidiEventSubs[method]?.includes(context)) {
+        this.log.info(
+          `<-- BIDI EVENT ${method} (context: '${context}', params: ${JSON.stringify(params)})`,
+        );
+        // now we can send the event onto the socket
+        const ev = {type: 'event', context, method, params};
+        await send(JSON.stringify(ev));
+      }
+    };
+    bidiHandlerDriver.eventEmitter.on(BIDI_EVENT_NAME, eventListener);
+  }
+
+  /**
+   * @param {Buffer} data
+   * @param {ExternalDriver | AppiumDriver} driver
+   */
+  async onBidiMessage(data, driver) {
+    let resMessage, id, method, params;
+    const dataTruncated = _.truncate(data.toString(), {length: 100});
+    try {
+      try {
+        ({id, method, params} = JSON.parse(data.toString('utf8')));
+      } catch (err) {
+        throw new errors.InvalidArgumentError(
+          `Could not parse Bidi command '${dataTruncated}': ${err.message}`,
+        );
+      }
+      driver.log.info(`--> BIDI message #${id}`);
+      if (!method) {
+        throw new errors.InvalidArgumentError(
+          `Missing method for BiDi operation in '${dataTruncated}'`,
+        );
+      }
+      if (!params) {
+        throw new errors.InvalidArgumentError(
+          `Missing params for BiDi operation in '${dataTruncated}`,
+        );
+      }
+      const result = await driver.executeBidiCommand(method, params);
+      // https://w3c.github.io/webdriver-bidi/#protocol-definition
+      resMessage = {
+        id,
+        type: 'success',
+        result,
+      };
+    } catch (err) {
+      resMessage = err.bidiErrObject(id);
+    }
+    driver.log.info(`<-- BIDI message #${id}`);
+    return resMessage;
+  }
+
+  /**
+   * Log a bidi server error
+   * @param {Error} err
+   */
+  onBidiServerError(err) {
+    this.log.error(`Error from bidi websocket server: ${err}`);
   }
 
   /**
@@ -410,6 +752,21 @@ class AppiumDriver extends DriverCore {
         );
         await driverInstance.updateSettings(jwpSettings);
       }
+
+      // if the user has asked for bidi support, and the driver says that it supports bidi,
+      // send our bidi url back to the user. The inner driver will need to have already saved any
+      // internal bidi urls it might want to proxy to, cause we are going to overwrite that
+      // information here!
+      if (dCaps.webSocketUrl && driverInstance.doesSupportBidi) {
+        const {address, port, basePath} = this.args;
+        const isUsingSsl = this.args.sslCertificatePath && this.args.sslKeyPath;
+        const scheme = isUsingSsl ? 'wss' : 'ws';
+        const bidiUrl = `${scheme}://${address}:${port}${basePath}${BIDI_BASE_PATH}/${innerSessionId}`;
+        // @ts-ignore webSocketUrl gets sent by the client as a boolean, but then it is supposed
+        // to come back from the server as a string. TODO figure out how to express this in our
+        // capability constraint system
+        dCaps.webSocketUrl = bidiUrl;
+      }
     } catch (error) {
       return {
         protocol,
@@ -515,6 +872,9 @@ class AppiumDriver extends DriverCore {
         // be in otherwise
         delete this.sessions[sessionId];
         delete this.sessionPlugins[sessionId];
+
+        this.cleanupBidiSockets(sessionId);
+
         return dstSession;
       });
       // this may not be correct, but if `dstSession` was falsy, the call to `deleteSession()` would
@@ -532,6 +892,32 @@ class AppiumDriver extends DriverCore {
         protocol,
         error: e,
       };
+    }
+  }
+
+  /**
+   * @param {string} sessionId
+   */
+  cleanupBidiSockets(sessionId) {
+    // clean up any bidi sockets associated with session
+    if (this.bidiSockets[sessionId]) {
+      try {
+        this.log.debug(`Closing bidi socket(s) associated with session ${sessionId}`);
+        for (const ws of this.bidiSockets[sessionId]) {
+          // 1001 means server is going away
+          ws.close(1001, 'Appium session is closing');
+        }
+      } catch {}
+      delete this.bidiSockets[sessionId];
+      const proxyClient = this.bidiProxyClients[sessionId];
+      if (proxyClient) {
+        this.log.debug(`Also closing proxy connection to upstream bidi server`);
+        try {
+          // 1000 means normal closure, which seems correct when Appium is acting as the client
+          proxyClient.close(1000);
+        } catch {}
+        delete this.bidiProxyClients[sessionId];
+      }
     }
   }
 
--- a/packages/appium/lib/main.js
+++ b/packages/appium/lib/main.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+import {WebSocketServer} from 'ws';
 import {init as logsinkInit} from './logsink'; // this import needs to come first since it sets up global npmlog
 import logger from './logger'; // logger needs to remain second
 import {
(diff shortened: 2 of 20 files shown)


jlipps:
cc @christian-bromann @titusfortner @AutomatedTester @shs96c @pujagani

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+      // Can't do much with random errors on the connection other than log them
+      this.log.error(err);
+    });
+
+    try {
+      const pathname = req.url;
+      if (!pathname) {
+        throw new Error('Invalid connection request');
+      }
+      const bidiSessionRe = new RegExp(`${BIDI_BASE_PATH}/([^/]+)$`);
```
mykola-mokhnach, line 270: should these patterns be global constants?

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+            `Ignoring and will close the connection`,
+        );
+      }
+
+      // Let's figure out which driver is going to handle this socket connection. It's either going
+      // to be a driver matching a session id appended to the bidi base path, or this umbrella driver
+      // (if no session id is included in the bidi connection request)
+
+      /** @type {import('@appium/types').ExternalDriver | AppiumDriver} */
+      let bidiHandlerDriver;
```
mykola-mokhnach, line 287: this functions contains too much logic. Consider splitting it into smaller chunks

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+
+      /** @type {import('ws').WebSocket | null} */
+      let proxyClient = null;
+
+      if (sessionMatch) {
+        // If we found a session id, see if it matches an active session
+        const sessionId = sessionMatch[1];
+        await sessionsListGuard.acquire(AppiumDriver.name, () => {
```
mykola-mokhnach, line 295: I assume we don't need the guard if the operation is synchronous

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+          }
+        };
+      };
+
+      // Construct our send method for sending messages to the client
+      const send = sendFactory(ws);
+
+      // Construct a conditional send method for proxying messages from the client to an upstream
+      // bidi socket server (e.g. on a browser)
+      let sendToProxy = null;
```
mykola-mokhnach, line 352: const sendToProxy = proxyClient ? sendFactory(proxyClient) : null

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
 
+  /**
+   * Initialize a new bidi connection
+   * @param {import('ws').WebSocket} ws The websocket connection object
+   * @param {import('http').IncomingMessage} req The connection pathname, which might include the session id
+   */
+  async onBidiConnection(ws, req) {
+    ws.on('error', (err) => {
+      // Can't do much with random errors on the connection other than log them
+      this.log.error(err);
```
mykola-mokhnach, line 262: It makes sense to log session identifier if possible, so we can always connect particular errors with corresponding sessions.

mykola-mokhnach: same below

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+      });
+
+      // Set up handlers for events that might come from the upstream bidi socket connection if
+      // we're in proxy mode
+      if (proxyClient) {
+        // Here we're receiving a message from the upstream socket server. We want to pass it on to
+        // the client
+        proxyClient.on('message', async (/** @type {Buffer} */ data) => {
+          this.log.debug(
+            `<-- BIDI Received data from proxied bidi socket, sending to client. Data: ${data.toString(
```
mykola-mokhnach, line 368: We probably don't want to log the whole data chunk, but abbreviate it instead

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+        });
+      }
+
+      // Now set up handlers for the various events that might happen on the websocket connection
+      // coming from the client
+      // First is incoming messages from the client
+      ws.on('message', async (/** @type {Buffer} */ data) => {
+        if (proxyClient) {
+          this.log.debug(
+            `--> BIDI Received data from client, sending to upstream bidi socket. Data: ${data.toString(
```
mykola-mokhnach, line 396: same as below

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+          throw new Error(
+            `Driver emitted a bidi event that was malformed. Require method and params keys ` +
+              `(with optional context). But instead received: ${JSON.stringify({
+                context,
+                method,
+                params,
+              })}`,
+          );
+        }
+        if (ws.readyState !== 1) {
```
mykola-mokhnach, line 440: Can we use state constants instead?

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+            // if the websocket is closed or closing, we can remove this listener as well to avoid
+            // leaks
+            bidiHandlerDriver.eventEmitter.removeListener(BIDI_EVENT_NAME, eventListener);
+          }
+          return;
+        }
+
+        if (
+          bidiHandlerDriver.bidiEventSubs[method] &&
+          bidiHandlerDriver.bidiEventSubs[method].includes(context)
```
mykola-mokhnach, line 452: you can use `bidiHandlerDriver.bidiEventSubs[method]?.includes(context)` format

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+          const ev = {type: 'event', context, method, params};
+          // TODO add conditional logic to send only if user is subscribed
+          await send(JSON.stringify(ev));
+        }
+      };
+      bidiHandlerDriver.eventEmitter.on(BIDI_EVENT_NAME, eventListener);
+    } catch (err) {
+      this.log.error(err);
+      try {
+        ws.close();
```
mykola-mokhnach, line 467: can it ever throw an error?

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+   * @param {Buffer} data
+   * @param {ExternalDriver | AppiumDriver} driver
+   */
+  async onBidiMessage(data, driver) {
+    let resMessage, id, method, params;
+    try {
+      try {
+        ({id, method, params} = JSON.parse(data.toString('utf8')));
+      } catch (err) {
+        throw new errors.InvalidArgumentError('Could not parse Bidi command');
```
mykola-mokhnach, line 482: it makes sense to also print the command and the error itself

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+      }
+      this.log.info(`--> BIDI message #${id}`);
+      if (!method || !params) {
+        throw new errors.InvalidArgumentError(`Found invalid method or params`);
```
mykola-mokhnach, line 486: I would rather show separate messages for both. This way it's easier to debug

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+      }
+      const result = await driver.executeBidiCommand(method, params);
+      resMessage = {
```
mykola-mokhnach, line 489: It would be nice to reference the spec here in comments

packages/appium/lib/appium.js
```diff
@@ -241,6 +251,261 @@ class AppiumDriver extends DriverCore {
...
+        id,
+        type: 'success',
+        result,
+      };
+    } catch (err) {
+      resMessage = err.bidiErrObject(id || '');
```
mykola-mokhnach, line 495: can id be equal to zero (which is evaluated to false)?

packages/appium/lib/appium.js
```diff
@@ -410,6 +675,19 @@ class AppiumDriver extends DriverCore {
...
         await driverInstance.updateSettings(jwpSettings);
       }
+
+      // if the user has asked for bidi support, and the driver says that it supports bidi,
+      // send our bidi url back to the user. The inner driver will need to have already saved any
+      // internal bidi urls it might want to proxy to, cause we are going to overwrite that
+      // information here!
+      if (dCaps.webSocketUrl && driverInstance.doesSupportBidi) {
+        const {address, port, basePath} = this.args;
+        const bidiUrl = `ws://${address}:${port}${basePath}${BIDI_BASE_PATH}/${innerSessionId}`;
```
mykola-mokhnach, line 685: the protocol might not be `ws` if the server uses SSL. Check https://github.com/mykola-mokhnach/appium-devtools-plugin/blob/b8857ba5335645c1064d8cedb32898399b4d19a0/lib/mixins/proxy.js#L392 for more details.

packages/appium/lib/appium.js
```diff
@@ -410,6 +675,19 @@ class AppiumDriver extends DriverCore {
...
+        // @ts-ignore webSocketUrl gets sent by the client as a boolean, but then it is supposed
+        // to come back from the server as a string. TODO figure out how to express this in our
+        // capability constraint system
```
mykola-mokhnach, line 688: we can mark it as `boolean|string|null`

packages/appium/lib/appium.js
```diff
@@ -515,6 +793,25 @@ class AppiumDriver extends DriverCore {
         // be in otherwise
         delete this.sessions[sessionId];
         delete this.sessionPlugins[sessionId];
+
+        // clean up any bidi sockets associated with session
+        if (this.bidiSockets[sessionId]) {
+          try {
+            for (const ws of this.bidiSockets[sessionId]) {
+              this.log.debug(`Closing bidi socket associated with session ${sessionId}`);
```
mykola-mokhnach, line 801: this same message might be printed multiple times, which is confusing

packages/appium/lib/appium.js
```diff
@@ -515,6 +793,25 @@ class AppiumDriver extends DriverCore {
         // be in otherwise
         delete this.sessions[sessionId];
         delete this.sessionPlugins[sessionId];
+
+        // clean up any bidi sockets associated with session
+        if (this.bidiSockets[sessionId]) {
```
mykola-mokhnach, line 798: can we move this chunk in a separate method?

packages/appium/lib/main.js
```diff
@@ -399,8 +400,13 @@ async function main(args) {
     serverOpts.keepAliveTimeout = parsedArgs.keepAliveTimeout * 1000;
   }
   let server;
+  const bidiServer = new WebSocketServer({noServer: true});
```
mykola-mokhnach, line 403: same as above. The web socket is expected to use SSL if the server already uses it for http communication

packages/appium/test/e2e/driver.e2e.spec.js
```diff
@@ -536,6 +536,68 @@ describe('FakeDriver via HTTP', function () {
...
+  describe('Bidi protocol', function () {
+    withServer();
+    const capabilities = {...caps, webSocketUrl: true};
+
+    it('should respond with bidi specific capability when a driver supports it', async function () {
+      let driver = await wdio({...wdOpts, capabilities});
+      try {
+        should.exist(driver.capabilities.webSocketUrl);
+      } finally {
+        await driver.deleteSession();
```
mykola-mokhnach, line 549: this method should be in afterEach callback

packages/appium/test/e2e/driver.e2e.spec.js
```diff
@@ -536,6 +536,68 @@ describe('FakeDriver via HTTP', function () {
       }
     });
   });
+
+  describe('Bidi protocol', function () {
+    withServer();
+    const capabilities = {...caps, webSocketUrl: true};
+
+    it('should respond with bidi specific capability when a driver supports it', async function () {
+      let driver = await wdio({...wdOpts, capabilities});
```
mykola-mokhnach, line 545: this method call should be in beforeEach callback

packages/appium/test/e2e/driver.e2e.spec.js
```diff
@@ -536,6 +536,68 @@ describe('FakeDriver via HTTP', function () {
...
+        driver.on('message', (data) => {
+          const {method, params} = JSON.parse(data.toString('utf8'));
+          if (method === 'clock.currentTime') {
+            collectedEvents.push(params.time);
+          }
+        });
+
+        // wait for some time to be sure clock events have happened, and assert we don't receive
+        // any yet
+        await B.delay(750);
```
mykola-mokhnach, line 583: can we put a timer instead of the hardcoded delay?

packages/appium/test/e2e/driver.e2e.spec.js
mykola-mokhnach: same below

packages/appium/test/e2e/driver.e2e.spec.js
mykola-mokhnach: same below

packages/base-driver/lib/basedriver/commands/bidi.ts
```diff
@@ -0,0 +1,37 @@
...
+declare module '../driver' {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  interface BaseDriver<C extends Constraints> extends IBidiCommands {}
+}
+
+const BidiCommands: IBidiCommands = {
+  async bidiSubscribe<C extends Constraints>(
+    this: BaseDriver<C>,
+    events: string[],
+    contexts: string[] = [''],
```
mykola-mokhnach, line 14: should it be an empty array instead?

packages/base-driver/lib/basedriver/commands/bidi.ts
```diff
@@ -0,0 +1,37 @@
...
+  },
+
+  async bidiUnsubscribe<C extends Constraints>(
+    this: BaseDriver<C>,
+    events: string[],
+    contexts: string[] = [''],
+  ) {
+    for (const event of events) {
+      if (this.bidiEventSubs[event]) {
+        this.bidiEventSubs[event] = this.bidiEventSubs[event].filter((c) => !contexts.includes(c));
```
mykola-mokhnach, line 28: not sure how `includes` reacts to an empty string

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -99,6 +101,10 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
 
   protocol?: Protocol;
 
+  bidiEventSubs: Record<string, string[]>;
```
mykola-mokhnach, line 104: the abbreviation `Subs` might be confusing. How about `Subscriptions`?

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -376,4 +388,57 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
       this.noCommandTimer = null;
     }
   }
+
+  async executeBidiCommand(bidiCmd: string, bidiParams: StringRecord): Promise<any> {
+    const [moduleName, methodName] = bidiCmd.split('.');
+
+    // if we don't get a valid format for bidi command name, reject
+    if (!moduleName || !methodName) {
+      throw new errors.UnknownCommandError();
```
mykola-mokhnach, line 397: Please provide proper error messages for easier debugging

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -376,4 +388,57 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
+      }
+    }
+    if (params?.optional?.length) {
+      for (const optionalParam of params.optional) {
+        args.push(bidiParams[optionalParam]);
+      }
+    }
+    this.log.debug(
+      `Executing bidi command '${bidiCmd}' with params ${JSON.stringify(
+        bidiParams,
```
mykola-mokhnach, line 434: Not sure if we want to print all params or abbreviate them if they are too long

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -376,4 +388,57 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
+      )} by passing to driver method '${command}'`,
+    );
+    let res = await this[command](...args);
+    if (_.isUndefined(res)) {
```
mykola-mokhnach, line 438: could be `const res = (await this[command](...args)) ?? null;

packages/base-driver/lib/protocol/errors.js
```diff
@@ -28,6 +28,20 @@ export class ProtocolError extends ES6Error {
   set stacktrace(value) {
     this._stacktrace = value;
   }
+
+  /**
+   * Get the Bidi protocol version of an error
+   * @param {string} id
```
mykola-mokhnach, line 34: Please describe the return type

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -308,11 +316,15 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
 
     if (!_.includes(validStrategies, strategy)) {
       throw new errors.InvalidSelectorError(
-        `Locator Strategy '${strategy}' is not supported for this session`
+        `Locator Strategy '${strategy}' is not supported for this session`,
       );
     }
   }
 
+  getBidiProxy(): string | null {
```
mykola-mokhnach, line 324: getBidiProxy sounds like we want to return an object. If we only need the proxy url then getBidiProxyUrl would probably match better

packages/appium/lib/appium.js
jlipps: since they're only used here, i prefer to keep them in their local context. if we ever use them in more than one place, we can factor them out.

packages/appium/lib/appium.js
jlipps: i can try to chunk it out some more. the fact that we use closures to maintain references to the websocket objects makes it nice, so we don't have to store too many references to them on drivers etc...

packages/appium/lib/appium.js
jlipps: for some reason this construction is not the same for typescript and causes the expression not to act as a guard for use of the sendToProxy function below. 🤷

packages/appium/lib/appium.js
jlipps: I can't tell from a brief look at the code. Do you think we should avoid the try/catch for now and then add it if errors ever occur? Seemed like a good idea to wrap third party library methods that we're not sure are safe. https://github.com/websockets/ws/blob/master/lib/websocket.js#L283

packages/appium/lib/appium.js
jlipps: yep good catch

packages/appium/lib/appium.js
jlipps: unfortunately the type of `dCaps.webSocketUrl` is set from the constraints, which only allows one type, not a disjunctive type, as far as i can tell.

packages/appium/lib/main.js
jlipps: I think I'll need help with this part. I'm just creating a virtual server object. It's not an actual server. I'm not sure we need to worry about ssl here since the server is already configured with ssl and this is just 'attaching' as an upgrade handler?

packages/base-driver/lib/basedriver/commands/bidi.ts
jlipps: `''` is the default context.

packages/base-driver/lib/basedriver/commands/bidi.ts
jlipps:
```
> [''].includes('')
true
```

packages/appium/test/e2e/driver.e2e.spec.js
jlipps: I'm not sure it would be worth the extra complexity in a test. The important thing is to wait for longer than the clock interval (500ms).

packages/base-driver/lib/basedriver/core.ts
jlipps: i think this abbreviation is fairly global. I'm open to changing it. Subscriptions is so long though :-/

packages/base-driver/lib/protocol/bidi-commands.js
```diff
@@ -0,0 +1,31 @@
+const SUBSCRIPTION_REQUEST_PARAMS = /** @type {const} */ ({
+  required: ['events'],
+  optional: ['contexts'],
+});
+
+const BIDI_COMMANDS = /** @type {const} */ ({
```
christian-bromann, line 6: You might want to use `@wdio/protocols` which provides a JSON object with various information about all bidi commands. If you require a certain format to be uphold you could reduce the object to what ever shape it needs to be in.

jlipps: oh this is a good idea! i'll have a look

packages/appium/lib/appium.js
jlipps: ok, i split it up into a number of methods.

packages/base-driver/lib/protocol/bidi-commands.js
jlipps: I don't see any way to reliably get parameter info from this JSON object, so I'm not sure it will work for us right now. Basically, we'd need to turn this:

```
'<pre>{<br />  events: string[];<br />  contexts?: BrowsingContextBrowsingContext[];<br />}</pre>'
```

into this:

```
{
  required: ['events'],
  optional: ['contexts'],
}
```

And I don't want to resort to regex parsing or TS reflection to do that.

christian-bromann: @jlipps gotcha, I think this should be possible to do with [`cddl`](https://www.npmjs.com/package/cddl) but requires further work.

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
     }
   }
 
+  /**
```
mykola-mokhnach, line 255: Perhaps it would make sense to extract bidi-related stuff into a mixin class? This class is already quite big and hard to read

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+   * Initialize a new bidi connection
+   * @param {import('ws').WebSocket} ws The websocket connection object
+   * @param {import('http').IncomingMessage} req The connection pathname, which might include the session id
+   */
+  initBidiSocket(ws, req) {
+    let outOfBandErrorPrefix = '';
+    const pathname = req.url;
+    if (!pathname) {
+      throw new Error('Invalid connection request');
```
mykola-mokhnach, line 296: I would prefer to have a more specific error message, for example: `Invalid connection request: request pathname is missing`

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+    }
+    const bidiSessionRe = new RegExp(`${BIDI_BASE_PATH}/([^/]+)$`);
```
mykola-mokhnach, line 298: these patterns could be extracted to global constants

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+    let bidiHandlerDriver;
+
+    /** @type {import('ws').WebSocket | null} */
+    let proxyClient = null;
+
+    if (sessionMatch) {
+      // If we found a session id, see if it matches an active session
+      const sessionId = sessionMatch[1];
+      bidiHandlerDriver = this.sessions[sessionId];
+      // @ts-ignore this is actually assigned in the guard block above but ts doesn't recognize it
```
mykola-mokhnach, line 324: I assume the comment is not relevant anymore

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+      if (!bidiHandlerDriver) {
+        // The session ID sent in doesn't match an active session; just ignore this socket
+        // connection in that case
+        throw new Error(
+          `Got bidi connection request for session with id ${sessionId} but this is not an active ` +
```
mykola-mokhnach, line 329: I would rather say "which is already closed or did not exist"

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+      // store this socket connection for later removal on session deletion. theoretically there
+      // can be multiple sockets per session
+      if (!this.bidiSockets[sessionId]) {
+        this.bidiSockets[sessionId] = [];
+      }
+      this.bidiSockets[sessionId].push(ws);
+
+      const bidiProxyUrl = bidiHandlerDriver.bidiProxyUrl;
+      if (bidiProxyUrl) {
+        if (!_.isString(bidiProxyUrl)) {
```
mykola-mokhnach, line 344: should we actually try to parse it as url instead of just a string?

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+          throw new Error(
+            `Got request to proxy bidi connections to upstream socket with url ` +
```
mykola-mokhnach, line 346: should we also print the driver class name to make it more explicit?

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+    const logSocketErr = (/** @type {Error} */ err) =>
+      this.log.error(`${outOfBandErrorPrefix}${err}`);
+
+    // This is a function which wraps the 'send' method on a web socket for two reasons:
+    // 1. Make it async-await friendly
+    // 2. Do some logging if there's a send error
+    const sendFactory = (/** @type {import('ws').WebSocket} */ socket) => {
+      const socketSend = B.promisify(socket.send, {context: socket});
+      return async (/** @type {string|Buffer} */ data) => {
+        const err = await socketSend(data);
```
mykola-mokhnach, line 370: I am confused. should it be try/catch instead?

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+        }
+      };
+    };
+
+    // Construct our send method for sending messages to the client
+    const send = sendFactory(ws);
+
+    // Construct a conditional send method for proxying messages from the client to an upstream
+    // bidi socket server (e.g. on a browser)
+    let sendToProxy = null;
```
mykola-mokhnach, line 382: this could be a ternary operator

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+   * @param {(data: string | Buffer) => Promise<void>} send - a method used to send data to the
+   * client
+   */
+  initBidiProxyHandlers(proxyClient, ws, send) {
+    // Set up handlers for events that might come from the upstream bidi socket connection if
+    // we're in proxy mode
+    if (proxyClient) {
+      // Here we're receiving a message from the upstream socket server. We want to pass it on to
+      // the client
+      proxyClient.on('message', async (/** @type {Buffer} */ data) => {
```
mykola-mokhnach, line 405: I assume data could also be a string or we don't care here?

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+        this.log.debug(
+          `--> BIDI Received data from client, sending to upstream bidi socket. Data: ${logData}`,
+        );
+        // if we're meant to proxy to an upstream bidi socket, just do that
+        // @ts-ignore sendToProxy is never null if proxyClient is truthy, but ts doesn't know
+        // that
+        await sendToProxy(data.toString('utf8'));
+      } else {
+        const res = await this.onBidiMessage(data, bidiHandlerDriver);
+        await send(JSON.stringify(res));
```
mykola-mokhnach, line 468: I assume `res` might be pretty large. Do we really want to load it all into RAM? Can it be streamed instead?

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+        return;
+      }
+
+      if (bidiHandlerDriver.bidiEventSubs[method]?.includes(context)) {
+        this.log.info(
+          `<-- BIDI EVENT ${method} (context: '${context}', params: ${JSON.stringify(params)})`,
+        );
+        // now we can send the event onto the socket
+        const ev = {type: 'event', context, method, params};
+        // TODO add conditional logic to send only if user is subscribed
```
mykola-mokhnach, line 531: without this todo this would greatly increase the amount of client/server traffic, right?

packages/appium/lib/appium.js
mykola-mokhnach: I would still prefer something like `Could not parse Bidi command '${_.truncate(data ...)}': ${err.message}`

packages/appium/lib/appium.js
```diff
@@ -241,6 +252,330 @@ class AppiumDriver extends DriverCore {
...
+      // https://w3c.github.io/webdriver-bidi/#protocol-definition
+      resMessage = {
+        id,
+        type: 'success',
+        result,
+      };
+    } catch (err) {
+      resMessage = err.bidiErrObject(id);
+    }
+    this.log.info(`<-- BIDI message #${id}`);
```
mykola-mokhnach, line 567: I believe `this.log` only includes the id of the ubmrella driver, not the id of the actual target driver. If we want to log proper id then we should retrieve the log instance from the actual target driver (e.g. bidiHandlerDriver.log)

packages/appium/lib/appium.js
```diff
@@ -410,6 +745,20 @@ class AppiumDriver extends DriverCore {
...
         await driverInstance.updateSettings(jwpSettings);
       }
+
+      // if the user has asked for bidi support, and the driver says that it supports bidi,
+      // send our bidi url back to the user. The inner driver will need to have already saved any
+      // internal bidi urls it might want to proxy to, cause we are going to overwrite that
+      // information here!
+      if (dCaps.webSocketUrl && driverInstance.doesSupportBidi) {
+        const {address, port, basePath} = this.args;
+        const scheme = 'secure' in this.server && this.server.secure ? 'wss' : 'ws';
```
mykola-mokhnach, line 755: 👍

packages/appium/lib/appium.js
```diff
@@ -535,6 +887,30 @@ class AppiumDriver extends DriverCore {
...
+  /**
+   * @param {string} sessionId
+   */
+  cleanupBidiSockets(sessionId) {
+    // clean up any bidi sockets associated with session
+    if (this.bidiSockets[sessionId]) {
+      try {
+        this.log.debug(`Closing bidi socket(s) associated with session ${sessionId}`);
+        for (const ws of this.bidiSockets[sessionId]) {
+          ws.close();
```
mykola-mokhnach, line 899: not sure if we want to use error code and reason to inform the client about the close reason

packages/base-driver/lib/basedriver/core.ts
mykola-mokhnach: how about `bidiEventSubscriptions`? `Subs` might be confused to `sub procedures`

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -376,4 +388,58 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
+    if (!this[command]) {
+      throw new errors.NotYetImplementedError();
+    }
+
+    const args: any[] = [];
+    if (params?.required?.length) {
+      for (const requiredParam of params.required) {
+        if (_.isUndefined(bidiParams[requiredParam])) {
+          throw new errors.InvalidArgumentError(
+            `The ${requiredParam} parameter was required but you omitted it`,
```
mykola-mokhnach, line 425: I believe we can handle this case better like we do for HTTP commands: https://github.com/appium/appium/blob/4dd95c096e4e4685c4f464b8251370ea001a562a/packages/base-driver/lib/protocol/protocol.js#L219

packages/base-driver/lib/protocol/errors.js
```diff
@@ -28,6 +28,27 @@ export class ProtocolError extends ES6Error {
...
+  /**
+   * Get the Bidi protocol version of an error
+   * @param {string|number} id - the id used in the request for which this error forms the response
+   * @see https://w3c.github.io/webdriver-bidi/#protocol-definition
+   * @returns The object conforming to the shape of a BiDi error response
+   */
+  bidiErrObject(id) {
+    // if we don't have an id, the client didn't send one, so we have nothing to send back.
+    // send back an empty string rather than making something up
+    if (!id && id !== 0) {
```
mykola-mokhnach, line 41: you may use !_.isNil

packages/appium/lib/appium.js
jlipps: sure, maybe i'll have you complete review before doing that

packages/appium/lib/appium.js
jlipps: i see no reason to make them global if they are only used locally?

packages/appium/lib/appium.js
jlipps: oops yeah refactor bug. good catch.

packages/appium/lib/appium.js
jlipps: i can update to include string

packages/appium/lib/appium.js
jlipps: `res` is whatever the driver has decided to return to the client. it will therefore already be in memory. i don't think this websocket protocol supports streaming.

packages/appium/lib/appium.js
jlipps: this was already handled by the conditional containing the TODO. i removed the TODO.

packages/appium/lib/appium.js
jlipps: ah yeah good point

packages/appium/lib/appium.js
mykola-mokhnach: it supports streaming for sure via pipes, for example https://stackoverflow.com/questions/22723671/streaming-data-in-node-js-with-ws-and-websocket-stream

Currently one of major limitations of the HTTP API is that we cannot stream big files, e.g. video recordings to the client, because these must be first fully stored in the memory

packages/appium/lib/appium.js
jlipps: just added as a TODO for a follow up PR to make it cleaner

packages/base-driver/lib/basedriver/core.ts
jlipps: I don't see any reason to be as complicated as we are there?

mykola-mokhnach: this helps users to fix their requests

mykola-mokhnach:
Seems like tests are not very happy

packages/appium/test/e2e/driver.e2e.spec.js
```diff
@@ -536,6 +536,61 @@ describe('FakeDriver via HTTP', function () {
       }
     });
   });
+
+  describe('Bidi protocol', function () {
```
mykola-mokhnach, line 540: would it be possible to add at least one test where we check BiDi stuff with https enabled? Check https://github.com/appium/appium/blob/d972a6a0eb3dbee0468700426be668807e20ca0c/packages/base-driver/test/e2e/express/server.e2e.spec.js#L110 for examples

packages/base-driver/lib/basedriver/core.ts
mykola-mokhnach: you could also leave a todo there, so I could improve the check later myself

packages/base-driver/lib/protocol/bidi-commands.js
```diff
@@ -0,0 +1,31 @@
+const SUBSCRIPTION_REQUEST_PARAMS = /** @type {const} */ ({
```
mykola-mokhnach, line 1: would it make sense to make it a .ts file instead as we need to declare multiple types?

packages/base-driver/lib/basedriver/core.ts
jlipps: 👍

packages/appium/test/e2e/driver.e2e.spec.js
jlipps: sure i will try that

packages/base-driver/lib/protocol/bidi-commands.js
jlipps: all we're doing is setting these to const so IDEs know they are immutable. not sure we need typescript for that

packages/appium/test/e2e/driver.e2e.spec.js
jlipps: ok i added a test but it looks like wdio has a bug in that it will not allow self signed certs for the bidi connection. hoping @christian-bromann will fix that soon :trollface:

mykola-mokhnach: have you tried [this](https://stackoverflow.com/questions/10888610/ignore-invalid-self-signed-ssl-certificate-in-node-js-with-https-request) workaround?

jlipps: i submitted a PR and it has been released. at long last @mykola-mokhnach CI is green!

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -308,11 +316,15 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
 
     if (!_.includes(validStrategies, strategy)) {
       throw new errors.InvalidSelectorError(
-        `Locator Strategy '${strategy}' is not supported for this session`
+        `Locator Strategy '${strategy}' is not supported for this session`,
       );
     }
   }
 
+  get bidiProxyUrl(): string | null {
```
mykola-mokhnach, line 324: a little doctring in the core class might be useful

mykola-mokhnach (approved):
great work

packages/base-driver/lib/basedriver/core.ts
jlipps: yep good idea

PR #19456, opened 2023-11-25
refactor: Drop @appium/docutils and @appium/typedoc-plugin-appium

mykola-mokhnach:
## Proposed changes

We do not (cannot) maintain these modules, and they pull us back by keeping vulnerable/outdated modules. 
So lets just cut them for good.


KazuCocoa:
Maybe after this, we'll keep markdown based files in https://github.com/appium/appium/blob/master/packages/appium/docs with adjusting links in the `docs` directory to follow markdown syntax on GH.
I was wondering how https://appium.io should be. I guess it will be just replaced with redirecting to the GitHub repository's docs directory (i'd like to get jlipps's knowledge/insight about documentation publishment)

mykola-mokhnach:
I am fine with any solution which we could support (e.g. it is easy enough to understand) and have resources to maintain it.

KazuCocoa:
yea, understood

jlipps:
I think we should close this PR. We shouldn't remove the entire docs system, which is working just fine. Only the typedoc autogeneration.

PR #19465, opened 2023-12-02
fix(docutils): remove `@appium/typedoc-plugin-appium` and all other uses of `typedoc`

eglitise:
## Proposed changes

This PR removes all uses of `typedoc` in the codebase, since we cannot maintain it any longer.
Included changes:
* Completely remove the `typedoc-plugin-appium` package
* Remove all `typedoc`-related code from the `docutils` package (including some CLI options for `appium-docs` that only made sense with `typedoc`)
  * The only exception is the `overwrite` option in [lib/scaffold.ts](https://github.com/appium/appium/blob/master/packages/docutils/lib/scaffold.ts) - I'm not really sure how this value gets passed further.
* Bump `docutils` to `1.0.0`. This is done to prevent breaking the docs in `appium-xcuitest-driver` (and anywhere else that uses the `docutils` package in a sensible semver range)
* Remove `typedoc`-related code everywhere else (`typedoc.json`, `tsconfig.json`, `package.json` for various packages, etc.)
* Remove `typedoc`-related documentation. This was only needed in one file - [the instructions for building extension docs](https://github.com/appium/appium/blob/master/packages/appium/docs/en/ecosystem/build-docs.md).
* Retain some documentation that was previously autogenerated using `typedoc` by adding it to VCS. This includes all the `commands` files (since they are currently linked in the docs navbar), but _not_ the `classes`, `enums`, `interfaces` and `modules` files. Links to any removed files have also been removed.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Documentation Update (if none of the other choices apply)

## Checklist

- [X] I have signed the CLA

## Further comments
I've prefixed this PR with `fix` so that `lerna` would include the name of the commit that modifies `docutils` in its changelog, but maybe this is not required if the commits are not squashed?


mykola-mokhnach (approved):
Thanks for picking this up @eglitise

KazuCocoa (approved):
Thanks for the clarification about the `ts` file and `md` ones. Looks good to me.

It could be nice to have a look by Jlipps as well.

jlipps (approved):
OK this LGTM! Bye bye typedoc 😢

PR #19478, opened 2023-12-06
fix(workflows): remove `tag=` prefix for action digest tags

eglitise:
## Proposed changes

Renovate appears to have forgotten how to parse GitHub Action digest tags with the `tag=` prefix. Given that tags without this prefix seem to be parsed just fine, let's see if removing the prefix fixes the problem.

## Types of changes

## Checklist

- [X] I have signed the CLA

## Further comments

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
       - name: Install dependencies
-        uses: bahmutov/npm-install@1a235c31658a322a3b024444759650ee6345c26d # tag=v1
+        uses: bahmutov/npm-install@1a235c31658a322a3b024444759650ee6345c26d # v1
         with:
           useRollingCache: true
           install-command: npm ci --foreground-scripts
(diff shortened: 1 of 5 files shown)


KazuCocoa (approved):

.github/workflows/build.yml
```diff
@@ -77,7 +77,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
       - name: Install dependencies
-        uses: bahmutov/npm-install@1a235c31658a322a3b024444759650ee6345c26d # tag=v1
+        uses: bahmutov/npm-install@1a235c31658a322a3b024444759650ee6345c26d # v1
```
KazuCocoa, line 80: or i guess we can use `@v1`  instead

eglitise: Then Dependabot would just create a PR to pin them (like #18866)

KazuCocoa: Maybe removing https://github.com/search?q=repo%3Aappium%2Fappium%20pinGitHubActionDigests&type=code would help...?
i haven't configured renovate well but maybe the helper does the work...?
https://docs.renovatebot.com/modules/manager/github-actions/

eglitise: This can of course be done, but is it not better to keep them pinned?

mykola-mokhnach: I personally get annoyed by these renovate messages. I am ok to keep actions manually pinned to a specific tag like we do in other projects where we don't use renovate

eglitise: Hmm, fair enough - updating only on major tag changes would still be fine. In that case I will close this PR and create a new one that removes the digests from all actions, and adjusts the Renovate config to not pin them.

PR #19512, opened 2023-12-10
feat(appium): Load drivers and plugins asynchronously

mykola-mokhnach:
## Proposed changes

Currently we load modules synchronously, which means the server load time equals to the SUM of all loading times. 
I think we may improve that by doing async load (using `async import`), which is available since node14, which changes the server load time to the MAX of module loading times.

I would be grateful if you could help with additional testing, because this change also highly depends on the actual hardware performance. Maybe it would also make sense to limit the maximum amount of modules being loaded in parallel 

As a bonus I've also aligned the drivers load procedure, so drivers are now loaded in the same sequence they are listed in the command line (previously they were always loaded in the sequence their names are stored in the manifest, basically, the install order)

## Types of changes


eglitise:
Since this PR adds new CLI args, I would propose changing the title to `feat`.

packages/appium/test/unit/extension/extension-config.spec.js
```diff
@@ -300,11 +300,8 @@ describe('ExtensionConfig', function () {
       });
 
       describe('when the extension is not actually installed', function () {
-        it('should throw', function () {
-          expect(() => config.require('fake')).to.throw(
-            ReferenceError,
-            /^could not find a driver installed at .+flotsam/i
-          );
+        it('should throw', async function () {
+          await config.requireAsync('fake').should.be.rejected;
```
eglitise, line 304: Is it possible to use `rejectedWith(/<error message>/)` in both tests to distinguish the error messages? Otherwise both of the changed tests are identical.

packages/appium/lib/extension/index.js
```diff
@@ -85,29 +115,30 @@ export function getActivePlugins(pluginConfig, usePlugins = []) {
...
-            );
-            log.debug(err.stack);
-          }
-        })
-    )
-  );
+export async function getActiveDrivers(driverConfig, maxParallelImports, useDrivers = []) {
+  /** @type {string[]} */
+  let filteredDriverNames = [];
+  if (useDrivers.length === 0) {
```
eglitise, line 125: Might be super picky, but the equivalent check in `getActivePlugins` uses `if (_.isEmpty(usePlugins))`, so this could be changed to also use lodash.

packages/fake-driver/package.json
```diff
@@ -23,14 +23,10 @@
   },
   "license": "Apache-2.0",
   "author": "https://github.com/appium",
-  "bin": {
-    "appium-fake-driver": "index.js"
-  },
```
eglitise, line 28: This entrypoint is also listed in `package-lock.json`, should that be updated as well?
Also I guess this is technically a breaking change, but it was never documented (and the driver also isn't meant to be used outside tests in the first place), so it might be okay.

eglitise (approved):
I admit I don't have much experience with async, but this looks good to me.
Also have yet to test this.

packages/fake-driver/package.json
mykola-mokhnach: yes, fake driver is only used for internal testing, so it is fine

packages/appium/test/unit/extension/extension-config.spec.js
mykola-mokhnach: updated

eglitise (approved):

eglitise:
Well, right now I'm on my Windows machine (Ryzen 5 3600), and it does not look good.
Just now I tried to test by installing 4 drivers and run `node .\packages\appium\index.js`.
The command seems to behave like so:
1) Do nothing for ~3 seconds
2) Load and print the manifest
3) Do nothing for ~2 seconds
4) Attempt to load 3 drivers at once (as expected) - unfortunately one immediately returns this error:
```bash
dbug Appium Requiring driver at C:\Users\edgar\.appium\node_modules\appium-uiautomator2-driver\build\index.js
ERR! Appium Could not load driver 'uiautomator2', so it will not be available. Error in loading the driver was: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
dbug Appium Error: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'      
dbug Appium     at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:236:11)
dbug Appium     at defaultLoad (node:internal/modules/esm/load:128:3)
dbug Appium     at ModuleLoader.load (node:internal/modules/esm/loader:409:13)
dbug Appium     at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:291:56)
dbug Appium     at new ModuleJob (node:internal/modules/esm/module_job:65:26)
dbug Appium     at ModuleLoader.#createModuleJob (node:internal/modules/esm/loader:303:17)
dbug Appium     at ModuleLoader.getJobFromResolveResult (node:internal/modules/esm/loader:260:34)
dbug Appium     at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:241:17)
dbug Appium     at ModuleLoader.import (node:internal/modules/esm/loader:328:23)
```
5) After the error, attempt to load the remaining 4th driver (again, as expected)
6) Return the same error as above for the 3 other drivers

On `master` branch, there are no errors, and it behaves like so:
1) Do nothing for ~0.5-1 second
2) Load and print the manifest
3) Start sequentially loading the drivers

For 4 drivers, the total length of these steps takes ~3 seconds, which is less than the ~5 seconds with parallel loading. But perhaps this time is also linked with the inability to load the drivers.

mykola-mokhnach:
> Well, right now I'm on my Windows machine (Ryzen 5 3600), and it does not look good. Just now I tried to test by installing 4 drivers and run `node .\packages\appium\index.js`. The command seems to behave like so:

This is good you use Windows to test it. We are all mac people:)
I've fixed the bug in the recent commit

mykola-mokhnach:
You may also want to set the `_LOG_TIMESTAMP` environment variable to `1` to enable log timestamps by default

eglitise:
Thanks for the fix, the error is no longer shown. I did a few measurements using `_LOG_TIMESTAMP` when loading 4 drivers:
* Current behavior: `1.998s` (average in 5 launches)
* New behavior: `1.884s` (average in 5 launches)
* New behavior, chunk size `4`: `1.892s` (average in 5 launches)
* New behavior, chunk size `1`: `1.901s` (average in 5 launches)

While 5 measurements isn't really enough for a reliable average, so far it seems like the core changes have a visible effect on load time (`~0.1s` saved), whereas changes in chunk size affect this a lot less.

jlipps (approved):

KazuCocoa (approved):
👍

PR #19513, opened 2023-12-10
refactor(support): Speed up findAppiumDependencyPackage API

mykola-mokhnach:
The previous implementation used `npm list` call, which is super slow, because it reads all package dependencies

### DIFF
--- a/packages/appium/test/e2e/local.e2e.spec.js
+++ b/packages/appium/test/e2e/local.e2e.spec.js
@@ -45,14 +45,6 @@ describe('when Appium is a dependency of the current project', function () {
   }
 
   /**
-   * Helper fn
-   * @returns {Promise<string>}
-   */
-  async function readHash() {
-    return await fs.readFile(hashPath, 'utf8');
-  }
-
-  /**
    * @template {CliExtensionSubcommand} ExtSubcommand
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):
lg, so is this when we start an appium process in a driver directory for example? Then appium loads the workspace.

mykola-mokhnach:
> lg, so is this when we start an appium process in a driver directory for example? Then appium loads the workspace.

yes, this whole logic is needed for appium to understand it is a dependency of an other project, which is anyway a minority of all deployments. I see a lot of complicated code there to support that scenario, which might be essentially not really needed as it covers edge cases. Even more, I don't want it to slow down the server startup for the majority of deployment cases.

jlipps (commented):
do all the cases still work after this change?

mykola-mokhnach:
> do all the cases still work after this change?

One case does not work - where appium is included as `file:` dependency. I assume it's ok to have it like that because:
- One can always provide APPIMU_HOME manually instead of autodetecting it
- It's an edge case

jlipps:
ok. as long as it works when appium is a dev/peer dep and we're in a driver repo doing driver development, that should be fine!

PR #19518, opened 2023-12-11
fix(appium): Respect the value of no-perms-check server arg

mykola-mokhnach:
## Proposed changes

It looks like this arg has been forgotten for a while, but it might be actually quite useful

## Types of changes

### DIFF
--- a/packages/appium/test/unit/config.spec.js
+++ b/packages/appium/test/unit/config.spec.js
@@ -9,7 +9,7 @@ import {
   getNonDefaultServerArgs,
   showBuildInfo,
   showConfig,
-  validateTmpDir,
+  requireDir,
 } from '../../lib/config';
 import {PLUGIN_TYPE} from '../../lib/constants';
 import {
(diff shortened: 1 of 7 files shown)


packages/appium/test/unit/config.spec.js
```diff
@@ -214,18 +214,21 @@ describe('Config', function () {
     });
   });
 
-  describe('validateTmpDir', function () {
-    it('should fail to use a tmp dir with incorrect permissions', function () {
-      validateTmpDir('/private/if_you_run_with_sudo_this_wont_fail').should.be.rejectedWith(
-        /could not ensure/
+  describe('requireDir', function () {
+    it('should fail to use a dir with incorrect permissions', function () {
+      requireDir('/private/if_you_run_with_sudo_this_wont_fail').should.be.rejectedWith(
```
eglitise, line 219: Well this is interesting. I had run unit tests on my local Windows machine recently, and a few days later discovered an empty directory at `C:\private\if_you_run_with_sudo_this_wont_fail`. I assume these tests were meant to be run on Unix machines, so I wonder if this behavior is expected?

mykola-mokhnach: yes, they are optimized for macOS
we tried to run them different platforms but never had time to stabilize 
Also on Windows they seem to be times slower 🤷

jlipps (approved):

KazuCocoa (approved):

Issue #18447, opened 2023-04-04
[Feat]: CLI command to list out an extension's registered scripts


jlipps:
I wanted to run the UiAutomator2 driver's reset/clean script but I didn't remember the precise name of it. Appium knows these things. So we should have a cli command that tells you what you can run, something like one of the following:

```
# with no script name, instead of running a script print out a list of available scripts
> appium driver run uiautomator2
The following scripts are exported by the uiautomator2 driver:
- reset
```

```
# or an entirely new cli subcommand
> appium driver list-scripts uiautomator2
The following scripts are exported by the uiautomator2 driver:
- reset
```

mykola-mokhnach:
Closed as fixed. The first option has been implemented.

PR #19539, opened 2023-12-16
feat(appium): Print the list of available extension scripts if no script name is provided

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/appium/appium/issues/18447

## Types of changes


KazuCocoa (approved):

packages/appium/lib/cli/extension-command.js
```diff
@@ -729,6 +729,24 @@ class ExtensionCliCommand {
       );
     }
 
+    if (!scriptName) {
+      const allScripts = _.toPairs(extScripts);
+      const root = this.config.getInstallPath(installSpec);
+      const existingScripts = await B.filter(
+        allScripts,
+        async ([, p]) => await fs.exists(path.join(root, p))
```
jlipps, line 737: do we really need to double check that extensions are telling the truth about script paths? can we not just report which keys are in the scripts object in package json?

jlipps (approved):

packages/appium/lib/cli/extension-command.js
mykola-mokhnach: I don't have any special preferences about that, also not sure why scripts are duplicated in the manifest
For example, doctor script paths I read from the package.json

jlipps: Sure. I guess it doesn't hurt. Wonder if path existence validation might be better done in manifest validation, then we can just rely on the keys being correct. But we can do that another time if you want to just merge this.

PR #19570, opened 2023-12-21
chore(appium): Ditch the redundant find-up dependency

mykola-mokhnach:
## Proposed changes

It is not necessary to run git in the parent folder if we just want to fetch the repo info.


KazuCocoa (approved):
Sounds reasonable

PR #19575, opened 2023-12-22
feat: Deny install/upgrade of packages which server dep does not meet the current Appium version

mykola-mokhnach:
## Proposed changes

We already have a similar check for locally installed extensions (that one only prints warnings). Although it does not seem logical to allow modules installation if these are not compatible.

The check is only applied to packages managed by NPM

## Types of changes

### DIFF
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -17,6 +17,8 @@ import {spawn} from 'child_process';
 import {inspect} from 'node:util';
 import {pathToFileURL} from 'url';
 import {Doctor} from '../doctor/doctor';
+import {npmPackage} from '../utils';
+import semver from 'semver';
 
 const UPDATE_ALL = 'installed';
 
@@ -31,6 +33,20 @@ class NoUpdatesAvailableError extends Error {}
  */
 function receiptToManifest(receipt) {
   return /** @type {ExtManifest<ExtType>} */ (_.omit(receipt, 'driverName', 'pluginName'));
+}
+
+/**
+ * Fetches the remote extension version requirement
+ *
+ * @param {string} installSpec Extension name (could also contain the version suffix)
+ * @returns {Promise<[string, string|null]>}
+ */
+async function getRemoteExtensionVersionReq(installSpec) {
+  const currentVersion = npmPackage.version;
+  const allDeps = await npm.showPackageDependencies(installSpec);
+  const requiredVersionPair = _.flatMap(_.values(allDeps).map(_.toPairs))
+    .find(([name]) => name === 'appium');
+  return [currentVersion, requiredVersionPair ? requiredVersionPair[1] : null];
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/appium/lib/cli/extension-command.js
```diff
@@ -225,6 +240,29 @@ class ExtensionCliCommand {
...
+      return;
+    }
+
+    await spinWith(this.isJsonOutput, `Checking if '${pkgName}' is compatible`, async () => {
+      const [serverVersion, extVersionRequirement] = await getRemoteExtensionVersionReq(installSpec);
+      if (serverVersion && extVersionRequirement && !semver.satisfies(serverVersion, extVersionRequirement)) {
+        throw this._createFatalError(
+          `'${installSpec}' cannot be installed because the server version it requires (${extVersionRequirement}) ` +
+          `does not meet the currently installed one (${serverVersion}). Please install ` +
+          `a compatible server version first.`
```
eglitise, line 260:
```suggestion
          `does not meet the currently installed one (${serverVersion}). Please install ` +
          `a compatible server version, or use a different version of '${pkgName}'.`
```

eglitise (approved):
Just to confirm, all official extensions have `"peerDependencies": { "appium": "^2.0.0" }` (or some earlier beta version), correct? So their installs should not be affected yet. It's only if I have e.g. Appium `2.0.0` and try to install an extension with `^2.2.0` or something, that an error would be raised?

Also I wonder if it's possible to add a CI test for this?

packages/appium/lib/cli/extension-command.js
mykola-mokhnach: > or use a different version of '${pkgName}'

This message is also shown on upgrade, so I was not sure if it wouldn't be confusing

mykola-mokhnach:
> Just to confirm, all official extensions have "peerDependencies": { "appium": "^2.0.0" } (or some earlier beta version), correct? So their installs should not be affected yet. It's only if I have e.g. Appium 2.0.0 and try to install an extension with ^2.2.0 or something, that an error would be raised?

Yes, that's correct. This feature is more relevant for cases where one is running an older Appium server, which does not have all the features that a newer extension version requires. I don't think driver or extension authors have any reasons to cap the minor or patch version of the server by default

mykola-mokhnach:
> Also I wonder if it's possible to add a CI test for this?

For that we would have to have some obsolete non matching extension available in NPM 🤷

packages/appium/lib/cli/extension-command.js
mykola-mokhnach: also, it is not clear if there is `a different version` of the package, which would be compatible to the current server version

KazuCocoa (approved):
lgtm

KazuCocoa:
Should we update `peerDependencies` to `@2.0.0` which still have `-beta.35`?
https://github.com/search?q=repo%3Aappium%2Fappium%20peerDependencies&type=code
(let me check semver.satisfies..)

KazuCocoa:
nvm, seems good:
```
> s.satisfies('2.0.0', '^2.0.0-beta.35')
true
```

PR #19621, opened 2024-01-07
docs(doctor): Add deprecation notice

mykola-mokhnach:


KazuCocoa (approved):

eglitise (approved):

packages/doctor/README.md
```diff
@@ -5,6 +5,12 @@
 [![NPM version](http://img.shields.io/npm/v/@appium/doctor.svg)](https://npmjs.org/package/@appium/doctor)
 [![Downloads](http://img.shields.io/npm/dm/@appium/doctor.svg)](https://npmjs.org/package/@appium/doctor)
 
+!!! warning
```
Dor-bl, line 8: I suggest using alert extension for github markdown.
In this case > [!WARNING]

mykola-mokhnach: I've tried. Not sure it changed anything

eglitise: The `!!! warning` syntax is used in MkDocs, whereas `[!WARNING]` is used in GitHub's Markdown. Since this is a readme, the latter would work better. You can check the [main readme](https://github.com/appium/appium/blob/master/README.md) for how it looks there.

Dor-bl: I've created a related PR for that:
https://github.com/appium/appium/pull/19622

PR #19655, opened 2024-01-15
chore(docutils): update consola to 3.2.3

eglitise:
This PR updates `consola` to v3. Most of the changes are copied from #18505, which this PR replaces. One difference: it seems like `consola` v3 `import` calls use ESM, so they have been changed to CJS `require()`.

I ran `./packages/docutils/bin/appium-docs.js` (after install) with various options and it seemed to work fine.

Also I'm not well versed in TS, so if something looks off or can be improved, please do point it out 🙏


packages/docutils/lib/cli/config.ts
```diff
@@ -46,9 +45,9 @@ export async function findConfig(argv: string[] = hideBin(process.argv)) {
 
   // if --verbose is used, set the log level to debug.
   // otherwise use --log-level or the default.
-  let logLevel: LogLevel | keyof typeof LogLevelMap;
+  let logLevel: keyof typeof LogLevelMap;
   if (preArgs.verbose) {
-    logLevel = LogLevel.Debug;
+    logLevel = 'debug';
```
mykola-mokhnach, line 50: can we use LogLevels.debug ?

mykola-mokhnach (approved):

packages/docutils/lib/cli/config.ts
eglitise: `logLevel` has the type `"silent" | "error" | "warn" | "info" | "debug"`, but `LogLevels.debug` is `number` . Similarly, `LogTypes.debug` is `Partial<LogObject>`.

packages/docutils/lib/constants.ts
```diff
@@ -2,7 +2,8 @@
  * Constants used across various modules in this package
  * @module
  */
-import {LogLevel} from 'consola';
+// eslint-disable-next-line @typescript-eslint/no-var-requires -- Consola 3 import call is ESM
+const {LogLevels} = require('consola');
```
jlipps, line 6: I don't think we have done this elsewhere in Appium so far (mixed require and import). I'm not sure of potential consequences. At least so far we have just not upgraded any modules which are ESM only.

jlipps (approved):

packages/docutils/lib/constants.ts
eglitise: I guess there's always the safe route of just blacklisting the package, but then I wonder why did the original PR have all of these changes added.

jlipps: hmm I'm not sure, do you know who was working on this earlier? was it @boneskull? maybe he can advise on the consequences of mixing import styles?

PR #19681, opened 2024-01-19
fix(docutils): change default alias type to `symlink`

eglitise:
Two `mike`-related fixes for `docutils`:

* Fix argument order. `--alias-type` was being passed after the version and alias names, which should be the last arguments. Also removed the equivalent code for `mike serve` because it does not accept these arguments anyway.
* Change the default alias type to `symlink`:
  * Current behavior
    * I go to `appium.io`
    * This redirects to `http://appium.io/docs/en/latest/`
    * The alias type is `redirect`, so this further redirects to `http://appium.io/docs/en/2.4/`
  * New behavior (at least, how I _assume_ it will work)
    * I go to `appium.io`
    * This redirects to `http://appium.io/docs/en/latest/`
    * The alias type is `symlink`, so the page contents are loaded from `http://appium.io/docs/en/2.4/`, but the URL stays as `http://appium.io/docs/en/latest/`

  In theory this is a breaking change, however, I would argue that:
    * the old URL is still accessible by using the version switcher (change to another version and back)
    * the new behavior is more desirable anyway, as it prevents unwanted sharing of version-specific links

### DIFF
--- a/packages/docutils/lib/builder/deploy.ts
+++ b/packages/docutils/lib/builder/deploy.ts
@@ -139,12 +139,6 @@ export async function deploy({
     const mikeArgs = [
       ...argify(_.pickBy(mikeOpts, (value) => _.isNumber(value) || Boolean(value))),
     ];
-    if (alias) {
-      mikeArgs.push('--update-aliases', version, alias);
-      mikeArgs.push('--alias-type', aliasType);
-    } else {
-      mikeArgs.push(version);
-    }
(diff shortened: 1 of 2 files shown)


mykola-mokhnach (approved):

KazuCocoa (approved):

jlipps:
Yeah, very nice for `latest` links not to redirect

PR #19724, opened 2024-01-27
chore: Remove commitlint and husky

mykola-mokhnach:


eglitise:
Is it worth adding a `format` script in `package.json` like for https://github.com/appium/appium-adb/pull/722?

mykola-mokhnach:
> Is it worth adding a `format` script in `package.json` like for [appium/appium-adb#722](https://github.com/appium/appium-adb/pull/722)?

Not sure if it makes sense to add it to the top level. Maybe for each package, but this could be done in a separate PR.

KazuCocoa (approved):

PR #19777, opened 2024-02-12
chore(fake-driver): add  @appium/support as missing the module

KazuCocoa:
## Proposed changes

fake driver imports "@appium/support" but it is not in the dependencies

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/packages/fake-driver/package.json
+++ b/packages/fake-driver/package.json
@@ -42,6 +42,7 @@
     "test:unit": "mocha --timeout 10s --slow 5s \"./test/unit/**/*.spec.js\""
   },
   "dependencies": {
+    "@appium/support": "^4.2.1",
     "@types/bluebird": "3.5.42",
     "@xmldom/xmldom": "0.8.10",
     "asyncbox": "3.0.0",


packages/fake-driver/package.json
```diff
@@ -42,6 +42,7 @@
     "test:unit": "mocha --timeout 10s --slow 5s \"./test/unit/**/*.spec.js\""
   },
   "dependencies": {
+    "@appium/support": "^4.2.1",
```
mykola-mokhnach, line 45: this driver has peer dependency to appium, which means we could use imports of `appium/support`. Not sure why this import is needed them

KazuCocoa: This driver has `@appium/support` and `appium/support`. I thought the peer was for `appium/support` and not `@appium/support`. Not sure the original intention to have both cases
https://github.com/search?q=repo%3Aappium%2Fappium+%22appium%2Fsupport%22+path%3Afake-driver&type=code , or suggesting to use only `appium/support` by replacing `@appium/support` instead of adding `@appium/support`?

It looks like only `appium/support` is also good tho.

mykola-mokhnach: If we use `appium/support` then the dependency is loaded from the peer `appium` folder, while `@appium/support` is loaded from `@appium/support` (a separate module). The agreement is we use the first option for first-class drivers that have the direct peer dependency to appium in package.json and the second one for all other modules.

KazuCocoa: then, looks like this fake driver implementation hasn't been updated, or implemented properly yet. Let me update them to use `appium/support`

PR #19784, opened 2024-02-13
Updated logic to handle for W3C and avoid for MJSONWP

Matt-Petersen:
## Proposed changes

Currently when using `MJSONWP` and when the status is a `7` it is getting caught in the following if statement in `appium/packages/base-driver/lib/jsonwp-proxy/proxy.js`
```
      if ( _.has(data, 'status') && parseInt(data.status, 10) !== 0) {
        // Some servers, like chromedriver may return response code 200 for non-zero JSONWP statuses
        throwProxyError(data);
      }
```
This throw proxy error is eating the commands and causing testing scripts to hang. Adding the `this.downstreamProtocol == W3C` still maintains the integrity of the check for W3C but prevents the script from hanging and allows the connection to continue.

## Types of changes

- [ x ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ x ] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [ x ] I have signed the CLA
- [ x ] I have added the necessary documentation (if appropriate)
- [ x ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/base-driver/lib/jsonwp-proxy/proxy.js
+++ b/packages/base-driver/lib/jsonwp-proxy/proxy.js
@@ -249,7 +249,7 @@ class JWProxy {
         this.downstreamProtocol = this.getProtocolFromResBody(data);
         this.log.info(`Determined the downstream protocol as '${this.downstreamProtocol}'`);
       }
-      if (_.has(data, 'status') && parseInt(data.status, 10) !== 0) {
+      if (this.downstreamProtocol == W3C && _.has(data, 'status') && parseInt(data.status, 10) !== 0) {
         // Some servers, like chromedriver may return response code 200 for non-zero JSONWP statuses
         throwProxyError(data);
       }


jlipps:
Sorry, what was the situation in which you encountered the hang? What client and what server/driver were you using?

packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -249,7 +249,7 @@ class JWProxy {
         this.downstreamProtocol = this.getProtocolFromResBody(data);
         this.log.info(`Determined the downstream protocol as '${this.downstreamProtocol}'`);
       }
-      if (_.has(data, 'status') && parseInt(data.status, 10) !== 0) {
+      if (this.downstreamProtocol == W3C && _.has(data, 'status') && parseInt(data.status, 10) !== 0) {
```
mykola-mokhnach, line 252: not sure if this is going to work. We don't expect to have the 'status' property on downstream responses if the protocol is W3C

Matt-Petersen:
Currently am using `"appium": "2.4.1"` and ` "appium-base-driver": "7.11.3"`.

When we are passing in a JSON that looks like 
```
  data: {
    value: 'Node was not in the scene tree or found under a different context.',
    status: 7,
```
This includes the status of 7 because it is for `MJSONWP `, since I am in the process of updating and migrating to W3C. But since it is not a 0 status it the `throwProxyError(data);` which stops any further communication. It looks like this if statement should only be applied to w3c cases and not `MJSONWP`.

Since the W3c would be already included in this if statement as is, it prevents `MJSONWP` with exception status other than 0 from being caught.

mykola-mokhnach:
> "appium-base-driver": "7.11.3"

Is out of date and should not be used. The above appium version is supplied with v9 of the base driver

Matt-Petersen:
Wow, I didn't catch the dependency name swap from `appium-base-driver` -> `@appium/base-driver` during all of my debugging 🤦  Thank you very much.

jlipps:
@Matt-Petersen what are you building such that you need to depend on `@appium/base-driver` directly?

Matt-Petersen:
Working on a project where there is a layer of abstraction between appium and the environment. So I am utilizing a websocket to make a connection with the enviroment then pass Json objects over the socket which I convert in the custom driver and pass back to the server.

jlipps:
OK. No doubt you know what you're doing and are doing it for a reason but a simpler architecture could be to just send or proxy HTTP commands to Appium directly.

PR #19851, opened 2024-03-04
docs(appium): add an example for getSessions endpoint and add Deprecated for endpoionts which have no vendor key in the path and not in w3c spec

KazuCocoa:
## Proposed changes

https://github.com/appium/appium-inspector/pull/1369/files#r1510701460

It looks like we should put an example of the response of `/sessions` since it is not in W3C webdriver spec. This example would help.

e.g. https://www.selenium.dev/documentation/legacy/json_wire_protocol/#sessions

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/packages/appium/docs/en/commands/base-driver.md
+++ b/packages/appium/docs/en/commands/base-driver.md
@@ -49,6 +49,8 @@ A session data object
 
 Returns capabilities for the session and event history (if applicable)
 
+**`Deprecated`**
+
 <!-- comment source: multiple -->
 
 #### Response
@@ -217,6 +219,8 @@ A possibly empty list of elements inside the shadow root matching the selector
 
 Get the log for a given log type.
 
+**`Deprecated`**
+
 <!-- comment source: method-signature -->
 
 #### Parameters
@@ -273,6 +277,8 @@ The event history for the session
 
 Get available log types as a list of strings
 
+**`Deprecated`**
+
 <!-- comment source: method-signature -->
 
 #### Response
@@ -314,6 +320,8 @@ The UI hierarchy in a platform-appropriate format (e.g., HTML for a web page)
 `GET` **`/sessions`**
 
 Get data for all sessions running on an Appium server
+
+**`Deprecated`**
 
 <!-- comment source: method-signature -->
 
--- a/packages/appium/docs/en/commands/base-driver.md
+++ b/packages/appium/docs/en/commands/base-driver.md
@@ -49,8 +49,6 @@ A session data object
 
 Returns capabilities for the session and event history (if applicable)
 
-**`Deprecated`**
-
 <!-- comment source: multiple -->
 
 #### Response
@@ -320,8 +318,6 @@ The UI hierarchy in a platform-appropriate format (e.g., HTML for a web page)
 `GET` **`/sessions`**
 
 Get data for all sessions running on an Appium server
-
-**`Deprecated`**
 
 <!-- comment source: method-signature -->
 
--- a/packages/appium/docs/en/commands/base-driver.md
+++ b/packages/appium/docs/en/commands/base-driver.md
@@ -319,8 +319,8 @@ Get data for all sessions running on an Appium server
 
 #### Response
 
-A list of session data objects. E
-ach data object must have `id` and the session's capabilities as `capabilities` key.
+A list of session data objects.
+Each session data object will be returned with `id` and the session's capabilities as `capabilities` keys like an example below:
 
 ```json
 [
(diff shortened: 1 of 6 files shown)


packages/appium/docs/en/commands/base-driver.md
```diff
@@ -319,9 +319,22 @@ Get data for all sessions running on an Appium server
 
 #### Response
 
-`MultiSessionData`<`C`\>[]
-
-A list of session data objects
+A list of session data objects.
```
mykola-mokhnach, line 322: we should mark this endpoint as deprecated and not recommended for use, same as getSession one as they are not part of the W3C spec.

mykola-mokhnach (approved):

packages/appium/docs/en/commands/base-driver.md
eglitise: Are there any replacements for those deprecated endpoints? They might be used in the Inspector.

KazuCocoa: No for now I believe.
I think the inspector can be just "attach to a session" and users should give explicit session id. Current attach to an existing session does not let a user set an arbitrary session id by themselves, but it can have such setting an arbitrary session id to "attach to" the session.

eglitise: That would be a lot more inconvenient than the current process of automatic session discovery. Maybe this information can be added in the `/status` endpoint response, similarly to Selenium Grid? And for GET `session/:sessionId`, maybe something like GET `/session/:sessionId/appium/info`?
Can't comment on `getLog` and `getLogTypes`, but those could also just be moved to the corresponding `/session/:sessionId/appium/...` endpoint.

KazuCocoa: Then, we might need to add `mobile: getSessions` as current base method to define additional commands not in w3c webdriver spec.

`/status` should be `ready` and `message` https://www.w3.org/TR/webdriver2/#status , so not for this usage.

KazuCocoa: ah, or maybe we could define the `/sessions` endpoint in a new plugin? but not in the base driver?

eglitise: `mobile: getSessions` sort of implies it isn't compatible with other Appium-supported platforms like Mac. And while having this in a separate plugin is feasible, from a user perspective it would feel quite arbitrary to separate this.

I think `/status` could still work here because while it has to include `ready` and `message`, it "may additionally include arbitrary meta information that is specific to the implementation". [Selenium Grid also seems to do this](https://www.selenium.dev/documentation/grid/advanced_features/endpoints/#grid-status), though I'm not sure if its `/status` endpoint is W3C-compliant.

KazuCocoa: then it looks like a new plugin for it would be nice. I don't know that response format by the grid, so we may need to check the response format first.
e.g. `selenium-grid-status` plugin for the new `status` format like the grid, `sessions-plugin` to keep `/sessions`

KazuCocoa: https://github.com/appium/appium/issues/19856 for the plugin

eglitise (commented):
At this point I would prefer if the deprecation labels were moved to a separate PR and merged after #19856/#19857/#19858

For the label format in MkDocs, it would also be better to use something like
```
!!! warning "Deprecated"

    Please use the `<endpoint>` endpoint instead
```
XCUITest driver docs already use this format.

KazuCocoa:
Sure, then this pr can drop deprecated marks from `sessions` and `session/<session_id>`. These should be ok, and will create another Pr to modify the format with existing ones as well.

Btw, other clients will drop/dropped these endpoint call (since dependent selenium already dropped them) so maybe the appium-inspector usage is the main discussion focus right now.

packages/appium/docs/en/commands/base-driver.md
```diff
@@ -217,6 +217,8 @@ A possibly empty list of elements inside the shadow root matching the selector
 
 Get the log for a given log type.
 
+**`Deprecated`**
```
eglitise, line 220: Can these warnings be changed to the MkDocs format and moved right below the command heading?
```
!!! warning "Deprecated"

    Please use the `<endpoint>` endpoint instead
```
[This format is already used in the XCUITest driver.](https://github.com/appium/appium-xcuitest-driver/blob/master/docs/reference/commands.md?plain=1#L332)

eglitise (changes requested):

packages/appium/docs/en/commands/base-driver.md
eglitise: Ok nevermind, this is already being done in the other PR.

eglitise (approved):

PR #19868, opened 2024-03-07
docs(appium): add blog system to docs

jlipps:
This way we can add persistent new/announcement/info posts when necessary.

Screenshot 2024-03-07 at 14 00 22

### DIFF
--- a/packages/appium/docs/base-mkdocs.yml
+++ b/packages/appium/docs/base-mkdocs.yml
@@ -10,6 +10,9 @@ theme:
   favicon: assets/images/appium-logo.png
 
 plugins:
+  blog:
+    enabled: true
+    post_excerpt: required
   redirects:
     redirect_maps:
       'intro/requirements.md': 'quickstart/requirements.md'
@@ -34,7 +37,7 @@ extra:
       link: https://twitter.com/AppiumDevs
   alternate:
     - name: English
-      link: /docs/en/ 
+      link: /docs/en/
       lang: en
     - name: 日本
       link: /docs/ja/
--- a/packages/appium/docs/en/blog/.authors.yml
+++ b/packages/appium/docs/en/blog/.authors.yml
@@ -1,0 +1,8 @@
+authors:
+  jlipps:
+    name: "Jonathan Lipps"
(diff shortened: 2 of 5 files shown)


eglitise (approved):
Great idea!
I wonder if it's worth adding a link on the docs' Welcome page, like for all other navbar items? Not sure how best to present that...
Also this will need to be added to the README, though only once the page is live.

packages/appium/docs/base-mkdocs.yml
```diff
@@ -10,17 +10,20 @@ theme:
...
-      'ecosystem/build-docs.md': 'developing/build-docs.md'
-      'reference/commands/base-driver.md': 'commands/base-driver.md'
-      'reference/commands/execute-driver-plugin.md': 'commands/execute-driver-plugin.md'
-      'reference/commands/images-plugin.md': 'commands/images-plugin.md'
-      'reference/commands/relaxed-caps-plugin.md': 'commands/relaxed-caps-plugin.md'
-      'reference/commands/universal-xml-plugin.md': 'commands/universal-xml-plugin.md'
+  - blog:
+      enabled: true
+      post_excerpt: required
+  - redirects:
```
eglitise, line 16: Almost missed this - this should not use the list syntax, otherwise no plugins can be inherited by e.g. Inspector/XCUITest docs.
https://www.mkdocs.org/user-guide/configuration/#configuration-inheritance

KazuCocoa (approved):
👍

packages/appium/docs/base-mkdocs.yml
jlipps: ah good catch, fixed

jlipps:
> I wonder if it's worth adding a link on the docs' Welcome page, like for all other navbar items? Not sure how best to present that...

I tried a few things but it made the layout look more ugly. Also it's there in the top bar. I think it's visible enough. I'm imagining most people will access it via direct links from places where blogs are published on e.g. social media.

jlipps:
it's live! https://appium.io/docs/en/latest/blog/

PR #19986, opened 2024-04-10
fix: do not print deprecation message for caps that is not provided

KazuCocoa:
## Proposed changes

https://github.com/appium/appium-xcuitest-driver/pull/2383#discussion_r1558901795 

`[BaseDriver] useSimpleBuildTest is a deprecated capability` message could print 3 times if the value was given. This PR fixes one reason.

With this pr, 

```
[XCUITestDriver@8e46] }
[BaseDriver] useSimpleBuildTest is a deprecated capability
[BaseDriver] useSimpleBuildTest is a deprecated capability
[BaseDriver] useSimpleBuildTest is a deprecated capability
[XCUITestDriver@8e46] The desired capabilities include neither an app nor a bundleId. WebDriverAgent will be started without the default app
[XCUITestDriver@8e46 (555c4c93)] Session created with session id: 555c4c93-b2d3-47b1-aaf4-a24783cb9ccc
```

with `useSimpleBuildTest` will be:

```
[XCUITestDriver@8e46] }
[BaseDriver] useSimpleBuildTest is a deprecated capability
[BaseDriver] useSimpleBuildTest is a deprecated capability
[XCUITestDriver@8e46] The desired capabilities include neither an app nor a bundleId. WebDriverAgent will be started without the default app
[XCUITestDriver@8e46 (555c4c93)] Session created with session id: 555c4c93-b2d3-47b1-aaf4-a24783cb9ccc
```

Another reason is `validateCaps`. `validateCaps` could be called multiple times in https://github.com/appium/appium/blob/4d48010a30339887f2138b3adbaf7dce7bb804a0/packages/base-driver/lib/basedriver/capabilities.ts and https://github.com/appium/appium/blob/4d48010a30339887f2138b3adbaf7dce7bb804a0/packages/base-driver/lib/basedriver/driver.ts in total. Ideally, we'd like to print once. This Pr does not include this.

## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

[...]

PR #19988, opened 2024-04-10
fix: do not print duplicated deprecation message

KazuCocoa:
## Proposed changes

This PR is a followup pr for https://github.com/appium/appium/pull/19986

> Another reason is validateCaps. validateCaps could be called multiple times in https://github.com/appium/appium/blob/4d48010a30339887f2138b3adbaf7dce7bb804a0/packages/base-driver/lib/basedriver/capabilities.ts and https://github.com/appium/appium/blob/4d48010a30339887f2138b3adbaf7dce7bb804a0/packages/base-driver/lib/basedriver/driver.ts in total.

This PR will prevent showing duplicated deprecation error messages by storing what keys have been logged.

After:
```
[BaseDriver] The 'useSimpleBuildTest' capability has been deprecated and must not be used anymore. Please check the driver documentation for possible alternatives.
[XCUITestDriver@b14f (f1dca87e)] Session created with session id: f1dca87e-0674-46c0-a0ad-9ef2fc367c41
[XCUITest] Current user: 'kazu'
```

Before:

```
[BaseDriver] The 'useSimpleBuildTest' capability has been deprecated and must not be used anymore. Please check the driver documentation for possible alternatives.
[BaseDriver] The 'useSimpleBuildTest' capability has been deprecated and must not be used anymore. Please check the driver documentation for possible alternatives.
[XCUITestDriver@b14f (f1dca87e)] Session created with session id: f1dca87e-0674-46c0-a0ad-9ef2fc367c41
[XCUITest] Current user: 'kazu'
```

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments


packages/base-driver/lib/basedriver/capabilities.ts
```diff
@@ -259,10 +267,12 @@ export function parseCaps<C extends Constraints>(
   let strippedRequiredCaps = stripAppiumPrefixes(requiredCaps);
   const strippedAllFirstMatchCaps: Capabilities<C>[] = allFirstMatchCaps.map(stripAppiumPrefixes);
 
-  // Validate the requiredCaps. But don't validate 'presence' because if that constraint fails on 'alwaysMatch' it could still pass on one of the 'firstMatch' keys
+  // Validate the requiredCaps. But don't validate 'presence' because if that constraint fails on 'alwaysMatch' it could still pass on one of the 'firstMatch' keys.
```
mykola-mokhnach, line 270: Probably we should just remember for which caps we have already printed the deprecation warning and avoid printing is twice

KazuCocoa: yea, maybe this logic needs to modify more. At least https://github.com/appium/appium/pull/19986 will prevent showing the deprecation logs for non-usage, thus this kind of fix can be later.

Let me take a look further

mykola-mokhnach:
shall we close this draft @KazuCocoa ?
I assume the original issue has been already addressed

KazuCocoa:
This pr is not yet. https://github.com/appium/appium/pull/19986 fixed 3 times lines without the caps name is not given case. 

This PR ordinally tried to prevent duplicated message when deprecated caps names are given.

e.g.

When `useSimpleBuildTest` is given with the latest based driver, it prints:
```
[BaseDriver] useSimpleBuildTest is a deprecated capability
[BaseDriver] useSimpleBuildTest is a deprecated capability
```
instead of one time `[BaseDriver] useSimpleBuildTest is a deprecated capability`. The issue itself is the same. The `validateDesiredCaps` is loaded twice at least. This PR tried to not handle caps validation as `deprecated` in one of the validation, but ideally we'd like to control the log better somehow.

e.g. give a parameter to the validation method to not print logs

KazuCocoa:
Updated the implementation

KazuCocoa:
(this is not per session. need to modify further)

packages/base-driver/lib/basedriver/capabilities.ts
```diff
@@ -259,10 +268,12 @@ export function parseCaps<C extends Constraints>(
...
+  // Do not check deprecated property to reduce duplicated logs.
   if (shouldValidateCaps) {
     strippedRequiredCaps = validateCaps(strippedRequiredCaps, constraints, {
       skipPresenceConstraint: true,
+      skipDeprecatedCheck: true,
```
mykola-mokhnach, line 276: I don't think this is going to help much.  We have at least 3 calls of validateCaps for different capability sets (e.g. parsed from firstMatch, alwaysMatch, etc). I expect some of them may have repeating cap names, so the message could be printed more than once, but it would be 'legit'. The problem there is that the validator itself is stateless, so we don't know from which context it's called and thus cannot keep any flags between calls to know which caps names we have already printed and which we did not in scope of the same driver.

KazuCocoa: yes, correct. The root cause is the validator package and the method is stateless, so it can keep appium process wide state like https://github.com/appium/appium/pull/19988/commits/abb241f166624ba4e39e5913f3ffef4906752d8e but cannot control it per session level.

I'm currently ok to close this and leave duplication messages (it should be better than the previous 3 lines bug).

PR #20059, opened 2024-05-02
ci: add back Node 18

eglitise:
Node.js 22 was recently released, and as a result, the current CI scripts are no longer running on Node.js 18 (which is still maintained). This PR adds it back.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,9 +55,9 @@ jobs:
     outputs:
       versions: ${{ steps.generate-matrix.outputs.versions }}
     steps:
-    - name: Select 2 most recent LTS versions of Node.js
+    - name: Select 3 most recent LTS versions of Node.js
       id: generate-matrix
-      run: echo "versions=$(curl -s https://endoflife.date/api/nodejs.json | jq -c '[[.[] | select(.lts != false)][:2] | .[].cycle | tonumber]')" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -55,9 +55,9 @@ jobs:
     outputs:
       versions: ${{ steps.generate-matrix.outputs.versions }}
     steps:
-    - name: Select 2 most recent LTS versions of Node.js
+    - name: Select 3 most recent LTS versions of Node.js
```
mykola-mokhnach, line 58: maybe instead of just hardcoding this number we could fetch version numbers that are not marked with `its: false` and whose `eol` timestamp is below the current date?

eglitise: I'm not fully sure how to add the current date in the jq command, but there's also a dedicated GH action that can do all of this: https://github.com/marketplace/actions/node-lts-versions
Maybe it's easier to just add that?

mykola-mokhnach: yes, lets try the action, maybe It already does all we need

eglitise: Okay, seems like it works as expected. It has options for `active` which is what we want, and also `lts` which excludes versions that are current but not yet LTS (in this case Node 22)

mykola-mokhnach (approved):
nice, I like it
Would you have some time later to port this solution to other CI scripts  in Appium repositories?

KazuCocoa:
oh, nice

Issue #20101, opened 2024-05-16
[Feat]: print appium relevant information with one command


KazuCocoa:
**Is your feature request related to a problem? Please describe.**

The main motivation here is how to help users to collect info for bug report etc. The best is to attach full appium log, but maybe we could prepare a command to copy and paste info to share env summary to bug report template or cloud vendors.

**Describe the solution you'd like**

An idea is:

`appium --info`

then:

- APPIUM_HOME
- current nodejs/npm versions
- installed drivers
- installed plugins

**Describe alternatives you've considered**

Current method to share these info is:

`node --version`
`npm --version`
`appium driver list`
`appium plugin list`

**Additional context**

mykola-mokhnach:
I like the idea

mykola-mokhnach:
Closed as fixed

PR #20108, opened 2024-05-17
feat(appium): Add --show-debug-info CLI argument

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/appium/appium/issues/20101

## Types of changes

### DIFF
--- a/packages/appium/lib/config.js
+++ b/packages/appium/lib/config.js
@@ -6,6 +6,7 @@ import {exec} from 'teen_process';
 import semver from 'semver';
 import os from 'node:os';
 import {npmPackage} from './utils';
+import B from 'bluebird';
 import {getDefaultsForSchema, getAllArgSpecs} from './schema/schema';
 
 export const APPIUM_VER = npmPackage.version;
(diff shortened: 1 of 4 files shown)


KazuCocoa:
Nice 🚀 
It makes sense to show up this command in help as well
(around: https://github.com/appium/appium/blob/master/packages/appium/lib/cli/parser.js#L72-L75 )

KazuCocoa:
oh, this is `appium server --show-debug-info`. I see the description in the server's one

KazuCocoa (approved):
lgtm, love this.

packages/appium/lib/config.js
```diff
@@ -50,11 +52,59 @@ const getFullGitPath = _.memoize(async function getFullGitPath() {
...
+      arch: os.arch(),
+      homedir: os.homedir(),
+      username: os.userInfo().username,
+    },
+    node: {
+      version: process.version,
+      arch: process.arch,
+      cwd: process.cwd(),
+      argv: process.argv,
+      env: process.env,
```
eglitise, line 87: Ran this flag on Windows just now and it worked fine, but I'm hesitant about returning `env`. I understand that it may point out issues, but it can also contain tons of either useless or sensitive vars, and that's not including whatever is in the Path variable. It can also contain a lot of entries, which can make it annoying for the end user to filter out the sensitive info, should they choose to do so.
(also this showed that I don't have `APPIUM_HOME` set at all, yet UiAutomator2 still works without issues. 🤷 )

eglitise (approved):

packages/appium/lib/config.js
mykola-mokhnach: This flag is created for debug usage, so it's ok to print the env. It won't be shared anywhere beside the local console, although is still useful to debug possible issues.

> (also this showed that I don't have APPIUM_HOME set at all, yet UiAutomator2 still works without issues. 🤷 )

it must not be set explicitly. The server can autodetect the value in most of the cases.

jlipps (commented):
this is great! we should update the issue template to tell people to paste the output of this.

PR #20202, opened 2024-06-05
feat(appium): Add session signature to all logs

mykola-mokhnach:
## Proposed changes

The idea has been borrowed from https://discuss.appium.io/t/proposal-for-logging-context-information/42710

We add a special middleware which parses the actual session signature from the request URL (actually first 8 characters of the session id) and stores it into the [AsyncLocalStorage](https://nodejs.org/api/async_context.html) instance attached to the global logger. Afterwards all logged strings retrieve this signature automatically from the storage and add it to the prefix.

## Types of changes


KazuCocoa:
lg overall. I will take a look further later

jlipps (approved):
Seems good to me. So the idea is that log lines for any session based request will get the session id prefixed in the log line for easier scanning?

mykola-mokhnach:
> Seems good to me. So the idea is that log lines for any session based request will get the session id prefixed in the log line for easier scanning?

exactly

PR #20203, opened 2024-06-06
feat(logger): Add the 'debug' level to the default logger

mykola-mokhnach:
## Proposed changes

It is a package owned by us now and there is no need to hack/patch the logger object anymore.

## Types of changes


KazuCocoa (approved):
oh, i didn't notice the patch

PR #20161, opened 2024-05-30
feat(logger): move appium-logger repo into packages/logger

KazuCocoa:
## Proposed changes

- Forked [npmlog](https://github.com/npm/npmlog) since it has been archived.
- Covert the code to TypeScript based code
- Rename the package from `npmlog` to `@appium/logger` in order to make it available for Appium project under appium org in npm
- Update the readme
- License keeps under ISC

## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [CONTRIBUTING](https://github.com/appium/appium/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added the necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #20207, opened 2024-06-06
fix(logger): fix license ref

chenrui333:
relates to https://github.com/appium/appium/pull/20161

https://github.com/appium/appium/blob/3fbc4f1fe07eadf9c7a2ef8fc9f4ba78dc3486a3/README.md?plain=1#L208

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -17,7 +17,7 @@
       "allowBranch": "master"
     }
   },
-  "ignoreChanges": ["**/test/**", "**/*.md"],
+  "ignoreChanges": ["**/test/**", "**/*.md", "**/LICENSE"],
   "useNx": false,
   "version": "independent"
 }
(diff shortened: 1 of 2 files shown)


KazuCocoa:
hm, our release script changed to Apache-2.0
https://github.com/appium/appium/commit/339acd345fafa1292c8d286261553fcc6b440e46#diff-6b07e442fc527c911a7841554308068a91cd3aaaf6dc9b34d3a19741d8d393fc

KazuCocoa:
This will bring back to ISC in package.json but next release will set as `Apache-2.0` again. We should fix script side

eglitise:
I'm not sure if configuring lerna to ignore license files will prevent it from changing the actual license line in package.json 🤔 would be great if it did, though

KazuCocoa:
https://github.com/appium/appium/pull/20217 will fix this root cause

PR #20213, opened 2024-06-07
feat(logger,support): Add the 'debug' level to the default logger (#20203)

KazuCocoa:
BREAKING CHANGE: Remove `patchLogger` exported method from `logger` in @appium/support

Add the 'debug' level to the default logger

---

Can we put `logger` and `support` in the `feat` prefix...?

This is the same of https://github.com/appium/appium/pull/20203 to bump support's major version.

### DIFF
--- a/packages/appium/lib/logsink.js
+++ b/packages/appium/lib/logsink.js
@@ -1,11 +1,10 @@
 import globalLog from '@appium/logger';
 import {createLogger, format, transports} from 'winston';
-import {fs, logger} from '@appium/support';
+import {fs} from '@appium/support';
 import { APPIUM_LOGGER_NAME } from './logger';
 import _ from 'lodash';
 
 // set up distributed logging before everything else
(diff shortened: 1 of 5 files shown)


mykola-mokhnach (approved):

eglitise:
Since this is a breaking change that includes changes in the `appium` package, I believe it will bump `appium` to 3.0. Maybe this can be split into 2 PRs, one that's not a breaking change and only changes to the new behavior (keeping the export), and the other being a breaking change that only removes the export?

KazuCocoa:
Correct. It looks like this bump appium as well.

Let me split this change into two

```
Changes:
 - appium: 2.9.0 => 3.0.0
...
```

KazuCocoa:
https://github.com/appium/appium/pull/20219 and https://github.com/appium/appium/pull/20220 (breaking change for support) will cover this change. The 20220 merge with breaching message will let appium/support v5 in the next publishment.

PR #20216, opened 2024-06-07
fix(appium): use fixed @appium deps as package.json

### LINKED ISSUE #20215: [Feat]: use ~ for dependencies in appium main package instead of ^
In the past, we changed to use `~` in appium/packages.json for appium packages as https://github.com/appium/appium/pull/19576 but I found that our release script changed the `~` to use `^`.... https://github.com/appium/appium/commit/76df4d600dcb7c867c37c3d7d302dc7fcc95ae09#diff-924340cfae1b8cf67b9e29eb9b6afb510ce67089dcb565083c35b11861f2fcd5

hm... can we control the deps to keep using `~`...?

KazuCocoa:
Closes https://github.com/appium/appium/issues/20215 hopefully.

In the past, we tried to use `~` as released package.json via lerna script to restrict the deps version to not be so aggressive, but I noticed it was removed and `^` was used forcefully https://github.com/appium/appium/commit/76df4d600dcb7c867c37c3d7d302dc7fcc95ae09#diff-924340cfae1b8cf67b9e29eb9b6afb510ce67089dcb565083c35b11861f2fcd5

This `^` could break out intentional deps relationship.

I'm trying to run dry run or something to test this out my local. So this is still draft


KazuCocoa:
i ran `npx lerna version --no-git-tag-version --no-push` without internet. Maybe it worked. Then, the diff had `^` based on this change... so this may not work expectedly. Closing,...

PR #20214, opened 2024-06-07
feat(appium): Add session signature to all logs (#20202)

KazuCocoa:
This is the same change in feat(appium): Add session signature to all logs #20202 for 20203 change. This should be after https://github.com/appium/appium/pull/20219 and https://github.com/appium/appium/pull/20220 merge to not break changes, thus currently this is draft to not merge before that

PR #20219, opened 2024-06-07
feat(logger): Add the debug level to the default logger

KazuCocoa:
Split https://github.com/appium/appium/pull/20213 to logger and appium packages update.

This PR won't include appium/support's removal.

### DIFF
--- a/packages/appium/lib/logsink.js
+++ b/packages/appium/lib/logsink.js
@@ -1,11 +1,10 @@
 import globalLog from '@appium/logger';
 import {createLogger, format, transports} from 'winston';
-import {fs, logger} from '@appium/support';
+import {fs} from '@appium/support';
 import { APPIUM_LOGGER_NAME } from './logger';
 import _ from 'lodash';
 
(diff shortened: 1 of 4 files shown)


eglitise:
Wouldn't this now cause `unleakString` to be called twice, and create two `debug` levels? I think the only changes that should be omitted from #20213 are the removal of the `patchLogger` method and its export.

KazuCocoa:
So far, this change and https://github.com/appium/appium/pull/20220 will be these new versions:

```
Changes:
 - appium: 2.9.0 => 2.10.0
...
...
```

so I haven't split them so strictly. Should we modify while these 2 prs will be in the same version right now?
(https://github.com/appium/appium/pull/20214 will also be in the 5.0.0 so far)

eglitise:
I guess it might be okay as long as both PRs come in the same version

mykola-mokhnach (approved):

eglitise (approved):

PR #20220, opened 2024-06-07
feat(support): Remove patchLogger from logger

KazuCocoa:
BREAKING CHANGE: Remove patchLogger exported method from logger

With these separate change made:

```
Changes:
 - appium: 2.9.0 => 2.10.0
...
...
```

with `npx lerna version --no-git-tag-version --no-push` on my local, thus I think these separated PRs will create expected changes.

### DIFF
--- a/packages/appium/lib/logsink.js
+++ b/packages/appium/lib/logsink.js
@@ -1,11 +1,10 @@
 import globalLog from '@appium/logger';
 import {createLogger, format, transports} from 'winston';
-import {fs, logger} from '@appium/support';
+import {fs} from '@appium/support';
 import { APPIUM_LOGGER_NAME } from './logger';
 import _ from 'lodash';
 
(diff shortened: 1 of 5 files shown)


eglitise:
Is the target branch correct?

KazuCocoa:
Yes. Once `20203-1` is merged into the master, then `20203-2` will be against the master. These can be separated but right now this branch is based on the `20203-1` to create the same diff with https://github.com/appium/appium/pull/20209.
So, https://github.com/appium/appium/pull/20214 will come after these merges. https://github.com/appium/appium/pull/20214  has removal of `unleakString` in packages/support/lib/logging.js

eglitise:
That PR removes the last remaining call of `unleakString` but not the method itself

mykola-mokhnach (approved):

eglitise (approved):

PR #20221, opened 2024-06-07
feat(support): remove unleakString as no usage

KazuCocoa:
BREAKING CHANGE: Remove unleakString since it is now in appium/logger

We'll bump `appium/support` with https://github.com/appium/appium/pull/20220 so it would be nice to include this removal as same.

Remaining references:

```
% git grep 'unleakString'              
packages/logger/lib/log.ts:import { unleakString } from './utils';
packages/logger/lib/log.ts:      prefix: unleakString(prefix || ''),
packages/logger/lib/log.ts:      message: unleakString(formattedMessage),
packages/logger/lib/utils.ts:export function unleakString(s: any): string {
```

This PR branch is against https://github.com/appium/appium/pull/20214 since the PR removes `unleakString` usage from appium/support.

(Just keeps draft until https://github.com/appium/appium/pull/20214 merge)


KazuCocoa:
Expected versioning after these relevant PR's merge:

```
Changes:
 - appium: 2.9.0 => 2.10.0
...
...
```

eglitise (approved):

mykola-mokhnach (approved):

mykola-mokhnach:
I have also another candidate for removal: node->linkGlobalPackage

eglitise:
>I have also another candidate for removal: node->linkGlobalPackage

This one seems to be called in `requirePackage` which is used by `initMJpegConsumer` in `mjpeg.js`

mykola-mokhnach:
> > I have also another candidate for removal: node->linkGlobalPackage
> 
> This one seems to be called in `requirePackage` which is used by `initMJpegConsumer` in `mjpeg.js`

hm, I still don't see a reason it was made public as it's only used internally...
And the Windows bug is still there...

KazuCocoa:
Sounds good. > https://github.com/appium/appium/pull/20221#issuecomment-2154520281

PR #20222, opened 2024-06-07
test(logger): add tests for unleakString

eglitise:
Adds the tests removed as part of #20221 to the logger package.

### DIFF
--- a/packages/logger/test/unit/basic-specs.js
+++ b/packages/logger/test/unit/basic-specs.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-console */
 import { Log } from '../../lib/log';
+import { unleakString } from '../../lib/utils';
 import {Stream} from 'node:stream';
 
 describe('basic', async function () {
@@ -420,6 +421,20 @@ describe('basic', async function () {
       log.enableColor();
       const o = log._format('test message', { inverse: true });
       o.includes('\u001b[7mtest message\u001b[0m').should.be.true;
+    });
+  });
+
+  describe('unleakString', function () {
+    it('should unleak a string', function () {
+      unleakString('yolo').should.eql('yolo');
+    });
+    it('should unleak a multiline string', function () {
+      unleakString(' yolo\nbolo ').should.eql(' yolo\nbolo ');
+    });
+    it('should convert an object to a string', function () {
+      for (const obj of [{}, null, undefined, [], 0]) {
+        unleakString(obj).should.eql(`${obj}`);
+      }
     });
   });
 });


packages/logger/test/unit/basic-specs.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable no-console */
 import { Log } from '../../lib/log';
+import { unleakString } from '../../lib/utils';
```
mykola-mokhnach, line 3: I was not quite sure about keeping this method in support
The reason for that is that support has dependencies on logger
And we need to unleakString to be used in logger, which creates a circular dependency.
I have copied unleakString to the logger because of that reason, although I can also observe this method is not used anywhere else except of logger.
Perhaps, it makes sense to remove it from support for good since we anyway bump the major version of it

mykola-mokhnach: cc @KazuCocoa

eglitise: This import pulls from the unleakString exported by the logger package. The unleakString in the support package will be removed as part of #20221.

mykola-mokhnach: why do we test it here then? it would make more sense to add tests in the logger module then

packages/logger/test/unit/basic-specs.js
```diff
@@ -422,4 +423,18 @@ describe('basic', async function () {
       o.includes('\u001b[7mtest message\u001b[0m').should.be.true;
     });
   });
+
+  describe('unleakString', function () {
+    it('should unleak a string', function () {
+      unleakString('yolo').should.eql('yolo');
```
mykola-mokhnach, line 429: we should probably need to add an assertion that both objects have different references.
For that strings should be assigned to variables. And then we could compare references using the corresponding API from `support.node` set of helpers

packages/logger/test/unit/basic-specs.js
eglitise: This is the logger module :)

packages/logger/test/unit/basic-specs.js
eglitise: I agree that more tests are needed here, but the aim of this PR was to simply copy over the existing tests from the support package.

packages/logger/test/unit/basic-specs.js
mykola-mokhnach: ah, yes. Sorry for the confusion

mykola-mokhnach (approved):

KazuCocoa (approved):

PR #20228, opened 2024-06-08
feat(support): Move SecureValuesPreprocessor to @appum/logger

mykola-mokhnach:
BREAKING CHANGE: Removed the 'log-internals' module from '@appium/support' 
BREAKING CHANGE: Removed the 'loadSecureValuesPreprocessingRules' API from '@appium/support -> logging' 

The SecureValuesPreprocessor has been moved directly to the `@appum/logger` module and set as a private logger instance property. 
The `loadSecureValuesPreprocessingRules` method is now public instance method of the Logger class. Unit tests were ported to the `@appium/logger` module as well.


KazuCocoa (approved):
lg.

This GitHub's squash and merge commit message must not include breaking change messages since this has `appium` package modification.

eglitise:
Like Kazu said, this would bump Appium to 3.0, so I would suggest separating the change in the appium package into its own PR.

packages/logger/lib/log.ts
```diff
@@ -1,13 +1,22 @@
...
+  StyleObject,
+  Logger,
+  LogLevel,
+  PreprocessingRulesLoadResult,
+  LogFiltersConfig
+} from './types';
 import type {Writable} from 'node:stream';
 import {AsyncLocalStorage} from 'node:async_hooks';
 import { unleakString } from './utils';
+import { SecureValuesPreprocessor } from './secure-values-prepreocessor';
```
eglitise, line 19: Typo in file name

mykola-mokhnach: thanks. good catch

PR #20230, opened 2024-06-08
chore(support): Simplify the logic for the AppiumLogger instance retrieval

mykola-mokhnach:
## Proposed changes

This is just a refactoring to make things easier to read and understand. The PR does not change any logic

### DIFF
--- a/packages/support/lib/logging.js
+++ b/packages/support/lib/logging.js
@@ -1,5 +1,3 @@
-// @ts-check
-
 import globalLog from '@appium/logger';
 import _ from 'lodash';
 import moment from 'moment';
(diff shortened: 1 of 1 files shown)


packages/support/lib/logging.js
```diff
@@ -116,10 +85,41 @@ function getLogger(prefix = null) {
...
+  // if is possible that there is a logger instance that is already around,
+  // in which case we want t o use that
+  const useGlobalLog = !!global._global_npmlog;
+  const logger = testingMode && !forceLogMode
+    // in testing mode, use a mock logger object that we can query
+    ? MOCK_LOG
+    // otherwise, either use the global, or a new `npmlog` object
+    : (global._global_npmlog || globalLog);
+  // The default value is 10000, which causes excessive memory usage
+  logger.maxRecordSize = MAX_LOG_RECORDS_COUNT;
```
eglitise, line 105: I guess this is now also being set if `testingMode && !forceLogMode == true`, but I'm not sure if this will affect anything

eglitise (approved):

packages/support/lib/logging.js
mykola-mokhnach: it does not make any difference as MOCK_LOG is just a noop wrapper

KazuCocoa (approved):

PR #20232, opened 2024-06-09
docs(appium): add strategic partner browserstack

jlipps:


KazuCocoa (approved):
Would the building docs for non en fail?

mykola-mokhnach (approved):
LGTM after docs building is fixed

eglitise:
I think I managed to fix the problem with docs building and asset paths [in my branch](https://github.com/appium/appium/compare/jlipps/strat-partner-announce...eglitise:appium:partner-announce). Basically, instead of using the custom `copy-assets.js` script, it's easier to rely on Material for MkDocs's own way of copying assets, by just placing them in the `overrides` directory, [as described here](https://squidfunk.github.io/mkdocs-material/customization/?h=override#setup-and-theme-structure). It worked fine, so I figured I would delete the custom script as well.

Other notes:
* The current implementation of color-specific images did not work for me. I changed to the [Material for MkDocs's documented approach](https://squidfunk.github.io/mkdocs-material/reference/images/?h=dark+mo#light-and-dark-mode) and it is now working.
* Color-specific images were removed from the footer, since the footer color stays the same regardless of the theme.

jlipps:
awesome, good finds and thanks @eglitise !

PR #20283, opened 2024-06-21
feat(appium): show extension update info message for newer major versions

eglitise:
## Proposed changes

If the user decides to update an extension using the standard command `appium <ext-type> update <ext-name>`, and a newer version does exist, the extension gets safely updated as expected. However, in case a newer _major_ version _also_ exists, the user is not informed about this right away. Currently this information is only shown if no newer minor or patch versions exist.
This PR adds an info message in such situations, so that the user is immediately informed of a newer major version.

I am also including a minor bugfix for the updating message, where it was always referring to drivers, even when a plugin was being updated.

## Types of changes

## Checklist

- [X] I have signed the CLA

### DIFF
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -642,6 +642,7 @@ class ExtensionCliCommand {
           `Updating ${this.type} '${e}' from ${update.current} to ${updateVer}`,
           async () => await this.updateExtension(e, updateVer)
         );
+        // if we're doing a safe update, but an unsafe update is also available, let the user know
         if (!unsafe && update.unsafeUpdate) {
           const newMajorUpdateMsg = `A newer major version ${update.unsafeUpdate} ` +
             `is available for ${this.type} '${e}', which could include breaking changes. ` +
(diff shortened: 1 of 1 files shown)


packages/appium/lib/cli/extension-command.js
```diff
@@ -639,9 +639,15 @@ class ExtensionCliCommand {
         const updateVer = unsafe && update.unsafeUpdate ? update.unsafeUpdate : update.safeUpdate;
         await spinWith(
           this.isJsonOutput,
-          `Updating driver '${e}' from ${update.current} to ${updateVer}`,
+          `Updating ${this.type} '${e}' from ${update.current} to ${updateVer}`,
           async () => await this.updateExtension(e, updateVer)
         );
+        if (!unsafe && update.unsafeUpdate) {
```
mykola-mokhnach, line 645: I'am a bit lost there. There is quite similar condition above, that prints same message. How it is different from this one?

eglitise: The above condition is only printed when there is no safe update available. For example, if I'm on uiautomator2 driver 2.45.0 and I request an update, the driver will get updated to the latest 2.x, specifically, 2.45.1 (as expected), but with no information that 3.x even exists. This PR aims to cover exactly this case.

mykola-mokhnach: could you please add it as an explanation comment to the source to avoid possible confusion?

eglitise: added

mykola-mokhnach (approved):

PR #20563, opened 2024-09-12
docs: about payout

KazuCocoa:
This is a note about steps to pay payouts via opencollective manually.

I need to investigate opencollevtive, but if we could automate some of these, it would be nice.


eglitise (approved):
I guess the `docs` folder is probably the most reasonable place for this document, since it doesn't need to be picked up by MkDocs.

PR #20590, opened 2024-09-18
fix(basedriver,appium): when rewriting socket urls, use a working hostname

jlipps:
I.e., the one that the client used rather than an internal one

Also, update the websocket test so that it works

### DIFF
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -772,9 +772,15 @@ class AppiumDriver extends DriverCore {
       // internal bidi urls it might want to proxy to, cause we are going to overwrite that
       // information here!
       if (dCaps.webSocketUrl && driverInstance.doesSupportBidi) {
-        const {address, port, basePath} = this.args;
+        const {port, basePath} = this.args;
+        // this is set by protocol.js when the client makes the request, so that we respond with the same host they used
+        const address = this.hostLastUsedByClient ?? this.args.address ;
         const scheme = `ws${this.server.isSecure() ? 's' : ''}`;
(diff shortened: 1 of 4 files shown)


packages/appium/lib/appium.js
```diff
@@ -772,9 +772,15 @@ class AppiumDriver extends DriverCore {
       // internal bidi urls it might want to proxy to, cause we are going to overwrite that
       // information here!
       if (dCaps.webSocketUrl && driverInstance.doesSupportBidi) {
-        const {address, port, basePath} = this.args;
+        const {port, basePath} = this.args;
+        // this is set by protocol.js when the client makes the request, so that we respond with the same host they used
+        const address = this.hostLastUsedByClient ?? this.args.address ;
```
mykola-mokhnach, line 777: extra space before semicolon

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -105,6 +105,8 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
 
   doesSupportBidi: boolean;
 
+  hostLastUsedByClient?: string;
```
mykola-mokhnach, line 108: not sure if this should be a property of the all drivers or the umbrella driver only

packages/appium/lib/appium.js
mykola-mokhnach: shall we actually try to use os.hostname() instead?

packages/base-driver/lib/basedriver/core.ts
mykola-mokhnach: I was actually thinking about whether it would make sense to use https://nodejs.org/api/async_context.html#new-asynclocalstorage to store various request properties

We currently only use this approach for logger (too add request if to each log line), but might also extend it further

PR #20594, opened 2024-09-19
fix(execute-driver-plugin): replace vm2 with Node's built-in vm

eglitise:
This PR drops the unsafe `vm2` module and replaces it with Node's built-in `vm` module. The change only affects `execute-driver-plugin`.
Since the affected code is responsible for executing arbitrary JS, I don't think it's possible to validate with 100% certainty that the new `vm` behavior exactly matches the previously configured `vm2` behavior, but they are identical as far as the tests go, at least.
Still, I did check `vm2`'s documentation. `NodeVM` (what was used until now) allowed to use Node built-ins, but none of them were actually enabled in the plugin code, so moving to `vm` does not reduce any functionality here.

### DIFF
--- a/packages/execute-driver-plugin/lib/execute-child.js
+++ b/packages/execute-driver-plugin/lib/execute-child.js
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import B from 'bluebird';
-import {NodeVM} from 'vm2';
+import vm from 'vm';
 import {logger, util} from 'appium/support';
 
 const log = logger.getLogger('ExecuteDriver Child');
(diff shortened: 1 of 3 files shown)


packages/execute-driver-plugin/lib/execute-child.js
```diff
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import B from 'bluebird';
-import {NodeVM} from 'vm2';
+import vm from 'vm';
```
mykola-mokhnach, line 3: node:vm

mykola-mokhnach (approved):

KazuCocoa (commented):
lgtm, let me test a few with this tonight. <- done. seems good to me with current my best knowledge...

btw, would `feat` be good for this...? Ideally patch as no behavior changes though. <- no strong opinion thing

jlipps (approved):
nice!

KazuCocoa (approved):

PR #20655, opened 2024-10-10
fix(types): Update the `setValue` method parameter type to reflect actual usage

mateusz-plociennik:
## Proposed changes

The Python Appium client sends both `text` and `value` arguments, where `text` is a string and `value` is an array of strings (keys to be pressed). Updated the parameter type for `text` in `setValue` from `string` to `string | string[]` to accommodate both use cases.

## Types of changes

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/packages/types/lib/driver.ts
+++ b/packages/types/lib/driver.ts
@@ -1095,10 +1095,10 @@ export interface ExternalDriver<
    * Send keystrokes to an element (or otherwise set its value)
    * @see {@link https://w3c.github.io/webdriver/#element-send-keys}
    *
-   * @param text - the text to send to the element
+   * @param text - the text (string) or value (string[]) to send to the element
    * @param elementId - the id of the element
    */
-  setValue?(text: string, elementId: string): Promise<void>;
+  setValue?(text: string | string[], elementId: string): Promise<void>;
 
   /**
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
Could you please link the spec where it's stated the `text` arg might be an array of strings?

mykola-mokhnach:
What I could find so far at https://w3c.github.io/webdriver/#dfn-element-send-keys is:

If text is not a [String](https://w3c.github.io/webdriver/#dfn-string), return an [error](https://w3c.github.io/webdriver/#dfn-error) with [error code](https://w3c.github.io/webdriver/#dfn-error-code) [invalid argument](https://w3c.github.io/webdriver/#dfn-invalid-argument).

jlipps:
this is for backwards compatibility with the old send keys (which expected an array of strings). i.e., if a w3c python client was talking to a non w3c backend. i don't think our w3c-only server now needs to care about arrays of strings sent in, hopefully all supported clients have now been updated to send a single string?

mateusz-plociennik:
@jlipps In that case the `value` (array of strings) should be ignored by the w3c-only server, right?

jlipps:
yes, that value won't come into the command handler at all

PR #20668, opened 2024-10-15
docs(appium): turn sponsors list into full list

jlipps:
OpenCollective generated section doesn't allow links to sponsor sites or sponsor provided logos


packages/appium/docs/en/sponsors.md
```diff
@@ -50,6 +51,11 @@ contributors!
 
 ## Bronze Sponsors
 
-<a href="https://opencollective.com/appium" target="_blank"><img src="https://opencollective.com/appium/tiers/sponsors.svg?avatarHeight=60&button=false&random=1"/></a>
+<div class="sponsor-section">
+  <a href="https://magicpod.com/" target="_blank"><img src="https://images.opencollective.com/magicpod/722b7e4/avatar/64.png" alt="MagicPod" /></a>
+  <a href="https://route4me.com/" target="_blank"><img src="../assets/images/sponsor-logo-route4me.png" alt="Route Planning and Route Mapping" /></a>
+  <a href="https://testingbot.com/" target="_blank"><img src="https://images.opencollective.com/testingbot1/aea09ee/logo/64.png" alt="TestingBot" /></a>
```
KazuCocoa, line 57: Looks like testingbot is not in the monthly sponsor for now so we should drop this.

Hm, if we want to maintain like this individual one, we need to find a way to keep syncing active sponsors from opencollective

jlipps: yeah I guess we need to keep an eye on it manually for now. removed testingbot. seems like even using open collective's widget does not correctly show only active recurring contributors.

KazuCocoa (approved):
The change itself looks good. To avoid unnecessary complaints, I'd like to leave a note about sync cycle to reflect current sponsorships in this page though

jlipps:
@KazuCocoa do you mean a comment for us? or a note on the page itself that everyone can see?

KazuCocoa:
I meant somewhere in a public space. Maybe... https://appium.io/docs/en/latest/sponsors/, https://opencollective.com/appium or https://github.com/appium/appium/blob/master/GOVERNANCE.md#sponsorship...? 

This is to clarify when we reflect the latest sponsorship relations to the https://appium.io/docs/en/latest/sponsors/ for external users, at least to show it is not immediate work.


> we could add this kind of check as monthly payment for contributions work at least

This work itself is our internal so it can be part of our opencollective work.

PR #20674, opened 2024-10-17
ci: Tune Python deps installation

mykola-mokhnach:
Addresses https://github.com/appium/appium/actions/runs/11378800825/job/31655205209

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -38,13 +38,19 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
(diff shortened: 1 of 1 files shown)


eglitise:
Well that's not great, this means `docutils` as a whole no longer works on any Python version below 3.12. It might be safer to revert my previous PR and explicitly set CI Python to 3.11, at least this way we won't have a breaking change.

mykola-mokhnach:
> Well that's not great, this means `docutils` as a whole no longer works on any Python version below 3.12. It might be safer to revert my previous PR and explicitly set CI Python to 3.11, at least this way we won't have a breaking change.

I'd say it's fine. I don't see any particular need there in supported older versions, especially from the perspective of a helper tool that is only used in our CI.

mykola-mokhnach:
Closed in favour of https://github.com/appium/appium/pull/20675

PR #20675, opened 2024-10-17
fix(docutils): restore support for Python <3.12

eglitise:
This PR adds back `docutils` support for `pip` `< 23.0`, which was unintentionally broken with #20666. Compatibility with `pip` `>= 23.0` is still preserved.

Original PR description as follows:

This PR reverts #20666, as it broke the installation of `docutils` prerequisites when using `pip < 23.0`.
At the same time, this PR also removes the support for `pip >= 23.0` on distributions that prevent modifying system packages (which was also added in #20666). It will be reintroduced in the future.

The original change was prompted by GitHub upgrading their CI Linux runners to Ubuntu 24.04, but the rollout has now been halted due to this and other breaking changes. To prevent similar issues in the future on the Appium side, the CI Linux runners for affected jobs are temporarily explicitly set to Ubuntu 22.04.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -35,7 +35,7 @@ on:
 jobs:
   docs:
     name: Docs
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04 # TODO: change this to latest after adding handling for PEP 668
     steps:
       - uses: actions/checkout@v4
       - name: Install dependencies (Node.js)
@@ -44,7 +44,7 @@ jobs:
           useRollingCache: true
           install-command: npm ci
       - name: Install dependencies (Python)
-        run: pip install -r packages/docutils/requirements.txt --break-system-packages
+        run: pip install -r packages/docutils/requirements.txt
       - name: Configure Git User
         run: |
           git config --local user.email "github-actions[bot]@users.noreply.github.com"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -35,7 +35,7 @@ on:
 jobs:
   docs:
     name: Docs
-    runs-on: ubuntu-22.04 # TODO: change this to latest after adding handling for PEP 668
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: Install dependencies (Node.js)
@@ -44,7 +44,7 @@ jobs:
           useRollingCache: true
           install-command: npm ci
       - name: Install dependencies (Python)
-        run: pip install -r packages/docutils/requirements.txt
+        run: PIP_BREAK_SYSTEM_PACKAGES=1 pip install -r packages/docutils/requirements.txt
       - name: Configure Git User
         run: |
           git config --local user.email "github-actions[bot]@users.noreply.github.com"
--- a/packages/appium/docs/zh/contributing/develop.md
+++ b/packages/appium/docs/zh/contributing/develop.md
@@ -76,7 +76,7 @@ Our documentation system uses [MKDocs](https://www.mkdocs.org/) and therefore re
 
 ```sh
 # installing needed Python dependencies
-pip install -r packages/docutils/requirements.txt --break-system-packages
+pip install -r packages/docutils/requirements.txt
 # build the project
 npm run build
 # run dev server
--- a/packages/appium/docs/zh/contributing/develop.md
+++ b/packages/appium/docs/zh/contributing/develop.md
@@ -76,7 +76,7 @@ Our documentation system uses [MKDocs](https://www.mkdocs.org/) and therefore re
 
 ```sh
 # installing needed Python dependencies
-pip install -r packages/docutils/requirements.txt
+PIP_BREAK_SYSTEM_PACKAGES=1 pip install -r packages/docutils/requirements.txt
 # build the project
 npm run build
 # run dev server
--- a/packages/docutils/lib/init.ts
+++ b/packages/docutils/lib/init.ts
@@ -5,7 +5,13 @@
  */
 
 import * as JSON5 from 'json5';
-import {NAME_MKDOCS_YML, NAME_TSCONFIG_JSON, NAME_PYTHON, REQUIREMENTS_TXT_PATH} from './constants';
+import {
+  NAME_MKDOCS_YML,
+  NAME_TSCONFIG_JSON,
+  NAME_PYTHON,
+  PIP_MIN_VERSION_NUMBER_FOR_FLAG,
+  REQUIREMENTS_TXT_PATH,
+} from './constants';
 import YAML from 'yaml';
 import {exec} from 'teen_process';
 import {Simplify} from 'type-fest';
@@ -145,6 +151,10 @@ export async function initPython({
   if (upgrade) {
     args.push('--upgrade');
   }
+  const pipMajorVersion = await getPipMajorVersion(pythonPath);
+  if (pipMajorVersion >= PIP_MIN_VERSION_NUMBER_FOR_FLAG) {
+    args.push('--break-system-packages');
+  }
   if (dryRun) {
     dryRunLog.info('Would execute command: %s %s', pythonPath, args.join(' '));
   } else {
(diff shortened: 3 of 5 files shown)


mykola-mokhnach:
I would rather leave it as is and strive for python 3.12 as the current solution could backfire. So we remain stuck with an obsolete environment.
Could you also please share your point on this topic @jlipps and @KazuCocoa ?

eglitise:
I agree that the solution in this PR is not permanent, but the main goal is to revert an unintentional breaking change. A proper adjustment to enable support for those distributions can be done separately (and ideally without breaking backwards compatibility).
It's worth pointing out that the original issue was _only_ affecting distributions that prevent modifying system packages (such as CI). I have Python 3.12 + pip 24.0 on my local machine, and it does not require using the `--break-system-packages` flag even outside `venv`.

mykola-mokhnach:
> I agree that the solution in this PR is not permanent, but the main goal is to revert an unintentional breaking change. A proper adjustment to enable support for those distributions can be done separately (and ideally without breaking backwards compatibility). It's worth pointing out that the original issue was _only_ affecting distributions that prevent modifying system packages (such as CI). I have Python 3.12 + pip 24.0 on my local machine, and it does not require using the `--break-system-packages` flag even outside `venv`.

Backward compatibility is a major concern for user-facing parts of the framework. The docgen parts thou seems rather internal to me.

eglitise:
The breaking change does actually affect end users of the `docutils` package - they can no longer use Python <3.12 to install the required dependencies with `npm run install-docs-deps`.

KazuCocoa:
or adding the arg with env var? So `BREAK_SYSTEM_PACKAGE=1 <command>` when the docutil command will add the flag. Afaik, https://github.com/AppiumTestDistribution uses this to generate their documentation. So I think it would be nice to have backward compatibility as not only this repository usage.

eglitise:
It is _exactly because_ other orgs like ATD are using this to generate their docs, that this revert is required. In the current state, `docutils` simply does not work anymore on GitHub's runners without explicitly specifying Python 3.12 - you can see this on the Inspector repo (and the XCUITest repo before the PR that added it).
As for the CLI flag, like I said, that can be added in a separate PR. I find it more pressing to restore backwards compatibility to something that worked for years (Python <3.12) than worry about backwards compatibility for something that worked for 3 days (Python 3.12+ on specific distributions).

jlipps:
I'd vote for reverting, patching, then reapplying as a breaking change to the docutils package so that consumers can opt in

eglitise:
It's also possible that there's some way to reapply the change without breaking backwards compatibility. The original PR was pushed out quickly because it broke CI, but had it been tested on Python <3.12, it would have failed and I would have looked for another approach.

KazuCocoa (approved):

packages/docutils/lib/init.ts
```diff
@@ -165,6 +175,30 @@ export async function initPython({
...
 
+/**
+ * Retrieves the pip major version number
+ */
+async function getPipMajorVersion(pythonPath: string): Promise<number> {
+  const pipVersionArgs = ['-m', 'pip', '--version'];
+  try {
+    const result = await exec(pythonPath, pipVersionArgs, {shell: true});
+    const {code, stdout} = result;
+    if (code !== 0) {
```
mykola-mokhnach, line 186: exec always throws an error if code is not zero

packages/docutils/lib/init.ts
```diff
@@ -165,6 +175,30 @@ export async function initPython({
...
+      throw new DocutilsError(`Could not retrieve pip version. Reason: ${stdout}`);
+    }
+    const pipVersionString = stdout.split(' ')[1];
+    const pipMajorVersionInt = parseInt(pipVersionString, 10);
```
mykola-mokhnach, line 190: We could also use regexp/semver for this purpose

mykola-mokhnach: `util.compareVersions` helper from `appium/support` might also be good

.github/workflows/docs.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           useRollingCache: true
           install-command: npm ci
       - name: Install dependencies (Python)
-        run: pip install -r packages/docutils/requirements.txt --break-system-packages
+        run: pip install -r packages/docutils/requirements.txt
```
mykola-mokhnach, line 47: we might follow @KazuCocoa advice here by providing `BREAK_SYSTEM_PACKAGE=1` environment variable. This should work for both older and newer pip versions

packages/docutils/lib/init.ts
```diff
@@ -141,10 +147,14 @@ export async function initPython({
...
   if (upgrade) {
     args.push('--upgrade');
   }
+  const pipMajorVersion = await getPipMajorVersion(pythonPath);
+  if (pipMajorVersion >= PIP_MIN_VERSION_NUMBER_FOR_FLAG) {
+    args.push('--break-system-packages');
+  }
   if (dryRun) {
     dryRunLog.info('Would execute command: %s %s', pythonPath, args.join(' '));
   } else {
```
mykola-mokhnach, line 160: we might simply provide `BREAK_SYSTEM_PACKAGE=1` environment variable instead, so there would be no need to compare versions

eglitise: The user would then need to check whether they need this flag. Why complicate the user flow if the code can do it automatically?

.github/workflows/docs.yml
eglitise: Here CI is calling pip directly, without invoking docutils, so we can't really add any flags

mykola-mokhnach: weird.. wouldn't `BREAK_SYSTEM_PACKAGE=1 pip install -r packages/docutils/requirements.txt` work as expected?

packages/docutils/lib/init.ts
mykola-mokhnach: not sure what you mean. Why would they need to check it? Could you please provide an example scenario where it would be needed?

eglitise:
I think I understand the problem. From this discussion I had assumed that `BREAK_SYSTEM_PACKAGE=1` would be a custom `docutils`-specific flag, but what was meant was actually `pip`'s own `PIP_BREAK_SYSTEM_PACKAGES=1` flag. Since this flag is an environment variable version of `--break-system-packages`, older `pip` versions will safely ignore it instead of throwing an error, which ensures full compatibility.
I've now adjusted the code to remove the `pip` version detection and replaced it with this environment variable.

packages/appium/docs/zh/contributing/develop.md
```diff
@@ -76,7 +76,7 @@ Our documentation system uses [MKDocs](https://www.mkdocs.org/) and therefore re
 
 ```sh
 # installing needed Python dependencies
-pip install -r packages/docutils/requirements.txt --break-system-packages
+pip install -r packages/docutils/requirements.txt
```
mykola-mokhnach, line 79: should we also add the env var prefix to this doc?

packages/docutils/lib/init.ts
```diff
@@ -140,18 +146,19 @@ export async function initPython({
   upgrade = false,
 }: InitPythonOptions = {}): Promise<void> {
   pythonPath = pythonPath ?? (await findPython()) ?? NAME_PYTHON;
+  const pythonPathWithFlag = [PIP_BREAK_SYSTEM_PACKAGES_FLAG, pythonPath].join(' ');
```
mykola-mokhnach, line 149: this is not really necessary. You may pass custom environment variables using the `env` option of exec

packages/appium/docs/zh/contributing/develop.md
eglitise: I guess we could. This doc as a whole is a bit outdated compared to the `en` version, which uses `npm run install-docs-deps`.

packages/docutils/lib/init.ts
eglitise: Removing this would result in the logging methods not reflecting the use of the environment variable. I can still add the `env` option but that wouldn't make the code any cleaner.

mykola-mokhnach: I'm not sure this would work at all in Windows. The above syntax where env variables are provided from the command line is only supported by standard *nix shells

eglitise: Thanks, I checked this on Windows and it indeed failed. Changed the code to use exec's `env` option.

packages/docutils/lib/init.ts
```diff
@@ -141,17 +141,17 @@ export async function initPython({
...
   pythonPath = pythonPath ?? (await findPython()) ?? NAME_PYTHON;
 
-  const args = ['-m', 'pip', 'install', '-r', REQUIREMENTS_TXT_PATH, '--break-system-packages'];
+  const args = ['-m', 'pip', 'install', '-r', REQUIREMENTS_TXT_PATH];
   if (upgrade) {
     args.push('--upgrade');
   }
   if (dryRun) {
-    dryRunLog.info('Would execute command: %s %s', pythonPath, args.join(' '));
+    dryRunLog.info('Would execute command: %s %s (environment variables: %s)', pythonPath, args.join(' '), PIP_ENV_VARS);
```
mykola-mokhnach, line 149: Please split long lines to improve readability

mykola-mokhnach (approved):

KazuCocoa (approved):

Issue #20117, opened 2024-05-19
[Feat]: Prepare 'reset' action for 'setup'


KazuCocoa:
**Is your feature request related to a problem? Please describe.**

Next to https://github.com/appium/appium/issues/20097

The issue adds `setup` like `appium setup xxxx` command to help installing preset appium drivers.

This issue aims to discuss "reset" action like `appium setup reset`. My initial thought for the action in https://github.com/appium/appium/pull/20102 was full cleanup as removing APPIUM_HOME so that the user can start the setup from scratch. The typical usage is in user help. e.g. request users to run `appium setup reset` and run `appium setup mobile` and if some issue still continues instead of saying "please remove APPIUM_HOME from your local and install drivers/plugins again", like that.

https://github.com/appium/appium/pull/20102#discussion_r1606088748 had a conversation about the reset could be driver/plugin uninstallation but not removing APPIUM_HOME entirely. Some could be interaction basis, so a user can select uninstall which driver. (It is more `appium setup uninstall`, like that?)

So, to proceed to prepare such a full reset behavior, maybe it would be nice to clarify the usage and what we should implement.

Original idea

```
appium setup reset
```

Then remove APPIUM_HOME to start setup from scratch. The typical usage case is in an issue to ask a user to start appium env setup from scratch for some installation error claims.

Another ideas from https://github.com/appium/appium/pull/20102

- Uninstall installed drivers/plugins and extension.yml but keeps APPIUM_HOME
- Let user select what preinstalled modules install

cc @mykola-mokhnach @jlipps

PR #20747, opened 2024-11-13
feta(appium): Add a subcommand to reset all installed drivers and plugins

mykola-mokhnach:
https://github.com/appium/appium/issues/20117


KazuCocoa (approved):
lgtm

PR #20793, opened 2024-11-30
feat(base-driver): Allow to prefix feature names with automation names

mykola-mokhnach:
## Proposed changes

This is part of Appium 3 migration preparation. The PR fine-tunes insecure features logic by allowing prefixes in denied/allowed feature names. For example: 
Previously we would have to provide something like `foo,bar,baz` to the list of allow-insecure features in order to enable them. This applies to all installed drivers, which verify these feature names. This PR allows to provide the list of features looking like `xcuitest:foo,uiautomator2:bar,*:baz`, which would ONLY enable feature `foo` for the `xcuitest` driver, feature `bar` - for `uiautomator2` driver, and feature `baz` for all installed drivers. For now providing a feature name without a colon would be parsed as this name would have the `*:` prefix, thus maintaining the backward compatibility. In Appium3 we MIGHT remove this check, thus enforcing users to always provide the prefix explicitly.

## Types of changes

### DIFF
--- a/packages/base-driver/lib/basedriver/core.ts
+++ b/packages/base-driver/lib/basedriver/core.ts
@@ -312,7 +312,7 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
    * not
    *
    * @param name - name of feature/command
-   * @deprecated
+   * @deprecated Use {@link assertFeatureEnabled} instead
    */
   ensureFeatureEnabled(name: string) {
     this.assertFeatureEnabled(name);
--- a/packages/base-driver/lib/basedriver/core.ts
+++ b/packages/base-driver/lib/basedriver/core.ts
@@ -20,7 +20,11 @@ import AsyncLock from 'async-lock';
 import _ from 'lodash';
 import {EventEmitter} from 'node:events';
 import os from 'node:os';
-import {DEFAULT_BASE_PATH, PROTOCOLS, MAX_LOG_BODY_LENGTH} from '../constants';
+import {
+  DEFAULT_BASE_PATH,
+  PROTOCOLS,
+  MAX_LOG_BODY_LENGTH,
+} from '../constants';
 import {errors} from '../protocol';
 import DeviceSettings from './device-settings';
 import helpers, {BASEDRIVER_VER} from './helpers';
@@ -29,7 +33,8 @@ import {BIDI_COMMANDS} from '../protocol/bidi-commands';
 const NEW_COMMAND_TIMEOUT_MS = 60 * 1000;
 
 const ON_UNEXPECTED_SHUTDOWN_EVENT = 'onUnexpectedShutdown';
-const ALL_DRIVERS = '*';
+
+const ALL_DRIVERS_MATCH = '*';
 const FEATURE_NAME_SEPARATOR = ':';
 
 export class DriverCore<const C extends Constraints, Settings extends StringRecord = StringRecord>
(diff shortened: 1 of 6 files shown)


eglitise (approved):

KazuCocoa (approved):

mykola-mokhnach:
I have updated the logic to perform argument value validation/patching on server startup rather than when a secure feature is actually invoked
Also added a note to the documentation

packages/driver-test-support/lib/unit-suite.js
```diff
@@ -580,24 +580,11 @@ export function driverUnitTestSuite(
       d = new DriverClass();
     });
 
-    it('should say a feature is enabled when it is explicitly allowed', function () {
```
eglitise, line 583: I think we should have a test to validate that arguments without a prefix are still parsed correctly. The test can then be removed for Appium 3.

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -258,13 +265,35 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
    */
   isFeatureEnabled(name: string): boolean {
+    // automationName comparison is case-insensitive,
+    // while feature name is case-sensitive
+    const currentAutomationName = _.toLower(this.opts.automationName);
+
+    const parseFullName = (fullName: string) => {
+      const separatorPos = fullName.indexOf(FEATURE_NAME_SEPARATOR);
+      if (separatorPos < 0) {
+        // This should not happen as we preprocess cotresponding server arguments before
```
eglitise, line 275: Typo

packages/driver-test-support/lib/unit-suite.js
mykola-mokhnach: I believe this is already covered by existing functional tests. See, for example https://github.com/appium/appium/blob/master/packages/execute-driver-plugin/test/e2e/plugin.e2e.spec.js

packages/base-driver/lib/basedriver/core.ts
mykola-mokhnach: fixed

eglitise (approved):

KazuCocoa (approved):

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -258,13 +265,35 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
+      }
+      return [
+        _.toLower(fullName.substring(0, separatorPos)),
+        fullName.substring(separatorPos + 1)
+      ];
+    };
+    const parseFullNames = (fullNames: string[]) => fullNames.map(parseFullName);
+    const matches = ([automationName, featureName]: [string, string]) =>
+      [currentAutomationName, ALL_DRIVERS_MATCH].includes(automationName) && featureName === name;
+
```
KazuCocoa, line 289: It looks like we still need https://github.com/appium/appium/pull/20793/files#diff-bb56f5d13605c22db7ab1c490736e963c7c1449a8056c6c509fbc470fcbd2dfcR472-R494 into the base driver since environment which doesn't have this latest appium version such as `appium@2.5.4` installs this driver but the appium package doesn't have `*` workaround by this PR, so this behaves a breaking change for such environment.

I think our intention is for future preparation so still we won't require this prefix kind. Or we needed to bump this base driver's major version to not affect such old Appium packages

PR #20800, opened 2024-12-03
docs(appium): Enable the documentation to be translated via Crowdin

mykola-mokhnach:
## Proposed changes

The full process is described in the corresponding guide document included into this PR

## Types of changes


packages/appium/docs/en/developing/documents-translation.md
```diff
@@ -0,0 +1,25 @@
+---
```
KazuCocoa, line 1: Does https://github.com/appium/appium/blob/master/packages/appium/docs/mkdocs-en.yml need to have this file reference?

KazuCocoa (approved):
Left one index-related small comment otherwise, lgtm

I think it's a great idea to use Crowdin. I haven't used Crowdin to handle diff (when we add/remove some lines to/from existing files), but if it handles such diff as well, it will be more brilliant.

mykola-mokhnach:
> Left one index-related small comment otherwise, lgtm
> 
> I think it's a great idea to use Crowdin. I haven't used Crowdin to handle diff (when we add/remove some lines to/from existing files), but if it handles such diff as well, it will be more brilliant.

AFAIK it splits .md file content into paragraphs and then verifies for changes based on such units (e.g. unchanged paragraphs remain translated). There is still one issue though related to how we are going to import existing ja and zh translations into Crowdin. There we'd need someone with actual language knowledge to perform the import.

I've also forgoten about mkdocs. I assume there also should be some code to synchronize TOC between English and other languages?

packages/appium/docs/en/developing/documents-translation.md
eglitise: I would instead suggest adding the contents of this file to `contributing.md`, as a new level-2 section. I don't think this has much to do with development.

eglitise:
> There is still one issue though related to how we are going to import existing ja and zh translations into Crowdin. There we'd need someone with actual language knowledge to perform the import.

As a starting point it might be fine to just copy over the currently translated files. The only issue with that is files that have been changed or removed in the English version since then (more relevant for `ja` I suppose).

mykola-mokhnach:
> > There is still one issue though related to how we are going to import existing ja and zh translations into Crowdin. There we'd need someone with actual language knowledge to perform the import.
> 
> As a starting point it might be fine to just copy over the currently translated files. The only issue with that is files that have been changed or removed in the English version since then (more relevant for `ja` I suppose).

I really want to avoid that as Crowdin should be the primary source of truth. That it why it makes more sense to import what we have into crowdin first rather than trying to hack it for various cases

eglitise:
> I really want to avoid that as Crowdin should be the primary source of truth. That it why it makes more sense to import what we have into crowdin first rather than trying to hack it for various cases

Yes, that's what I mean. The Crowdin files for these languages are currently not translated, so all that's needed is to update them using the existing translations in the repo, by just copying the file contents.

packages/appium/docs/en/developing/documents-translation.md
mykola-mokhnach: done

packages/appium/docs/en/contributing.md
```diff
@@ -121,3 +121,30 @@ npm run dev:docs
 ```
 
 You can then view the documentation at `http://127.0.0.1:8000/docs/en`.
+
+### Translating Appium Documentation
```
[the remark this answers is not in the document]
mykola-mokhnach: changed

KazuCocoa (approved):
This pr lgtm, while exactly the index page could be considered to find a better way to sync. (I don't have good ideas right now, so just let me 👍 for this pr's current scope)

mykola-mokhnach:
It looks like Crowdin fails to import the main index file (perhaps due to the fact it contains HTML). I have reported an issue about that: https://community.crowdin.com/t/an-unknown-error-is-shown-if-i-try-to-import-a-markdown-file-containing-html/11055

PR #20805, opened 2024-12-03
feat: Remove @appium/doctor from the monorepo

mykola-mokhnach:


eglitise:
Generally looks good 👍 

There is one more remaining reference to this package, in the main [CHANGELOG.md](https://github.com/appium/appium/blob/master/CHANGELOG.md) file. I can probably make a PR afterwards to move its link to a dedicated section for removed plugins (along with the previously removed `typedoc-plugin-appium`).

Also, this package has 2 dependencies not used anywhere else in this monorepo: `appium-adb` and `inquirer`. The latter is part of [the Renovate config's exclusion list](https://github.com/appium/appium/blob/master/renovate/default.json), so it can be removed from there as well.

KazuCocoa (approved):

mykola-mokhnach:
> Generally looks good 👍
> 
> There is one more remaining reference to this package, in the main [CHANGELOG.md](https://github.com/appium/appium/blob/master/CHANGELOG.md) file. I can probably make a PR afterwards to move its link to a dedicated section for removed plugins (along with the previously removed `typedoc-plugin-appium`).
> 
> Also, this package has 2 dependencies not used anywhere else in this monorepo: `appium-adb` and `inquirer`. The latter is part of [the Renovate config's exclusion list](https://github.com/appium/appium/blob/master/renovate/default.json), so it can be removed from there as well.

Thanks, I've removed it from the renovate config

PR #20809, opened 2024-12-04
chore: Bump webdriverio to v9

mykola-mokhnach:
This should also allow to address various security issues with dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,7 +60,7 @@
         "tsd": "0.31.2",
         "typescript": "5.7.2",
         "validate.js": "0.13.1",
-        "webdriverio": "8.40.6",
+        "webdriverio": "9.4.1",
         "ws": "8.18.0",
         "yaml-js": "0.3.1"
       },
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
> WDIO v9 drops Node 16, which would be a breaking change for `execute-driver-plugin`. Maybe the target branch could be changed to `appium3`?

node16 was EOLed long ago. eventually, we might also simply bump the major version for the plugin itself as it does not affect the server. the latter only uses wdio to run tests

eglitise:
We can bump the plugin's major version along with its Node version range, but `lerna` will just reset it back to the monorepo-level Node version range - that's why I'm suggesting to do this in the `appium3` branch. Plus we'd need to merge this change into that branch anyway.

mykola-mokhnach:
It looks like wdio bump broke e2e tests in some weird way as it just exists without actually printing any useful stacktrace: https://github.com/appium/appium/actions/runs/12155382666/job/33897551759?pr=20809#step:5:6323 :(

PR #20816, opened 2024-12-04
chore(github): update issue templates

eglitise:
This PR updates the GitHub issue templates used when reporting a bug or feature request. I took some inspiration from the existing templates in the Inspector and WDIO repos.

Most of the changes are in the bug report issue form:
- Replace checkmark about the latest component updates with a mandatory field for the Appium version. I think this provides more information at a glance.
- Replace checkmark about officially supported components with an optional field for third-party components. Same as above - more information at a glance.
- Add a new mandatory dropdown about the Appium server host type. This is primarily here to quickly identify if a cloud provider is being used, and most users will be able to leave the default option.
- Swap order of minimal required example and Appium logs. The latter is more valuable (at least in my opinion), so it is requested first.
- Replace several entries in Environment Details with a single entry for debug details, using `appium --show-debug-details`.

Additionally, the feature request file was changed from simple Markdown to an issue form - its contents were basically left unchanged.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,51 +1,94 @@
...
 body:
+
   - type: markdown
     attributes:
-      value: ':stop_sign: _For support questions, please visit [http://discuss.appium.io](http://discuss.appium.io) instead._'
+      value: ':stop_sign: _Make sure you are reporting an issue. For general questions, please use [http://discuss.appium.io](http://discuss.appium.io)._'
+
   - type: checkboxes
     attributes:
-      label: 'Do I have the most recent component updates?'
```
mykola-mokhnach, line 11: I would prefer to keep this checkbox. Or change its text to something like `I tried to reproduce this issue with the most recent server and driver/plugin versions`

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,51 +1,94 @@
...
+      description: 'Please [search :mag: the issues](https://github.com/appium/appium/issues) to check if this bug has already been reported.'
       options:
-      - label: 'I use the most recent available driver/plugin and server versions'
+      - label: I have searched the existing issues
         required: true
-  - type: checkboxes
+  
+  - type: input
     attributes:
-      label: 'Is the component officially supported by the Appium team?'
```
mykola-mokhnach, line 18: Not sure I want to remove this checkbox as well

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,51 +1,94 @@
...
-      description: 'Make sure you report an issue for a component which is officially supported/maintained by the Appium team.'
-      options:
-      - label: 'I have verified the component repository is present under the Appium organization in GitHub'
-        required: true
-  - type: checkboxes
+      label: Appium Version
```
mykola-mokhnach, line 21: Perhaps it is a good time to simplify that and ask people to just provide the output of `appium server --show-debug-info` (with sensitive data sanitized)

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,33 @@
+name: 💡 Feature Request
+description: Suggest an idea for this project.
+title: "💡 feat: <title>"
+labels: [Enhancement]
+body:
```
mykola-mokhnach, line 5: Maybe here there should also be a checkbox about has checked for possible duplicates

.github/ISSUE_TEMPLATE/bug_report.yml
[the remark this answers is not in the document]
eglitise: That is true, but the debug info field is optional, whereas this one is mandatory.

.github/ISSUE_TEMPLATE/feature_request.yml
eglitise: I guess we could, but I don't think we get feature requests often enough for this to be an issue. At least, I don't recall ever seeing a duplicate feature request.

.github/ISSUE_TEMPLATE/bug_report.yml
eglitise: Makes sense. I suppose it should be optional, in case the user is already on the latest component versions?

mykola-mokhnach: maybe. We should also provide information on how to verify if component version is the latest one (e.g. `npm info appium` for the server and `appium X update` CLI for drivers/plugins)

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,51 +1,94 @@
...
     attributes:
-      label: 'Minimal Reproducible Example'
+      label: Environment Details
       description: |
-        Please provide a the _smallest, complete code snippet_ that Appium's maintainers can run to reproduce the issue ([read more about what this entails](https://stackoverflow.com/help/minimal-reproducible-example)).  Failing this, any sort of reproduction steps are better than nothing!
+        Please provide additional information about your environment; feel free to remove any items which are not relevant.
+
+        :warning: _Remember to redact or remove any sensitive information!_
+      value: |
+          - Debug details (`appium --show-debug-info`):
```
KazuCocoa, line 79: 👍 

This value prints environment variables. Some env could include sensitive info on that so perhaps `:warning: _Remember to redact or remove any sensitive information!_` reminder would also be helpful here too (if possible)

.github/ISSUE_TEMPLATE/bug_report.yml
eglitise: I like the command for the server, but the `appium <driver/plugin> update` CLI will also actually update the package, which the user may not want to do right away. But I guess the `npm info <driver/plugin> version` command works here as well?

KazuCocoa (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
eglitise: It's right there 2 lines higher :)
(for the record, I don't think the command output should include env vars, though I do agree that some of them may indeed be relevant for debugging purposes)

.github/ISSUE_TEMPLATE/bug_report.yml
mykola-mokhnach: yes, it should do

mykola-mokhnach (approved):

PR #20821, opened 2024-12-05
fix(base-driver): add * always for non-separator for backward compatibility

KazuCocoa:
## Proposed changes

Base-driver change in https://github.com/appium/appium/pull/20793/files#diff-f1838792b23d8d547095fc6ce7e449672c87a7688d1372ee172ebcea14cf7d38R274-R280  affected old appium packages such as `appium@2.11.x` etc since that was minor version update.
It means they are currently required to add `*` etc to make the insecure work.

e.g.
```
appium --allow-insecure chromedriver_autodownload
```

So, The original PR works only with latest appium package, `2.13.0`.

It would be nice to bump the base-driver's major version update to avoid possible unexpected base-driver deps update in such lower environment as us.

note: this breaking change handling can be later like only in `appium3` branch for example.

So, let's add the workaround in the base-driver as a fix right now, and let me create a followup PR for the base driver with this change's revert with breaking change message to bump the base-driver's major version update to apply the change only for newer appium packages.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/packages/base-driver/lib/basedriver/core.ts
+++ b/packages/base-driver/lib/basedriver/core.ts
@@ -272,9 +272,7 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
     const parseFullName = (fullName: string) => {
       let separatorPos = fullName.indexOf(FEATURE_NAME_SEPARATOR);
       if (separatorPos < 0) {
-        // Workaround for lower appium packages
-        fullName = `${ALL_DRIVERS_MATCH}${FEATURE_NAME_SEPARATOR}${fullName}`;
-        separatorPos = fullName.indexOf(FEATURE_NAME_SEPARATOR);
+        return [ALL_DRIVERS_MATCH, fullName];
       }
       return [
         _.toLower(fullName.substring(0, separatorPos)),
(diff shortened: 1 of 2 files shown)


packages/base-driver/lib/basedriver/core.ts
```diff
@@ -270,13 +270,11 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
+      let separatorPos = fullName.indexOf(FEATURE_NAME_SEPARATOR);
       if (separatorPos < 0) {
-        // This should not happen as we preprocess corresponding server arguments in advance
-        throw new Error(
-          `The full feature name must include both the driver name/wildcard and the feature ` +
-          `name split by a colon, got '${fullName}' instead`
-        );
+        // Workaround for lower appium packages
+        fullName = `${ALL_DRIVERS_MATCH}${FEATURE_NAME_SEPARATOR}${fullName}`;
+        separatorPos = fullName.indexOf(FEATURE_NAME_SEPARATOR);
```
KazuCocoa, line 277: We need this workaround to keep the insecure flag work. Removing the Error is not sufficient since it doesn't work with secure flags.

This change is to minimize the code change purpose only.

I have tested this works with `appium --allow-insecure chromedriver_autodownload` on my local.

KazuCocoa:
I also tested this won't cause regression with current master. The appium package's `validateFeatures` adds `*`

eglitise:
I would suggest still keeping the error message, just commented out, and with a `TODO` for removal with Appium 3, similarly to the original PR's change in `validateFeatures`.

eglitise (approved):
Let's also wait for input from @mykola-mokhnach

mykola-mokhnach (approved):

PR #20824, opened 2024-12-06
feat(execute-driver-plugin): Bump webdriverio to v9

mykola-mokhnach:
BREAKING CHANGE: webdriverio major version bumped to 9

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2414,10 +2414,30 @@
         "node": ">=14"
       }
     },
+    "node_modules/@promptbook/utils": {
+      "version": "0.69.5",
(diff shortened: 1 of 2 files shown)


eglitise:
Should the version bump also be included in the monorepo-level `package.json`?

mykola-mokhnach:
> Should the version bump also be included in the monorepo-level `package.json`?

No, this PR is only to bump the major version of the particular plugin and to bump wdio there to avoid security issues related to obsolete transitive dependencies.

jlipps (approved):
yep LGTM

PR #20825, opened 2024-12-06
fix(appium): update CLI help messages

eglitise:
## Proposed changes

This PR updates the CLI help messages. Some of the information there was outdated, incorrect, or even missing. The behavior of the CLI commands themselves remains unchanged.

* Fix all instances of help text for the driver/plugin name parameter due to incorrect ternary statements
* ~~Rename the driver/plugin name parameter from `driver`/`plugin` to `driverName`/`pluginName`. This is to distinguish the parameter from the `driver`/`plugin` subcommand keyword~~
* Change the example plugin name from `find_by_image` to `images` (we do not have a plugin named `find_by_image`)
* Change the order of subcommands when running `appium -h` (`setup` is now last instead of first)
* Add the missing installed plugins for all `appium setup` presets
* Add the missing `reset` sub-subcommand for `appium setup`
* Add `description` fields to the top-level subcommands (`server`/`driver`/`plugin`/`setup`)
* Clarify most other help messages

## Types of changes

- [X] Documentation Update (if none of the other choices apply)

### DIFF
--- a/packages/appium/lib/cli/args.js
+++ b/packages/appium/lib/cli/args.js
@@ -108,7 +108,7 @@ function makeInstallArgs(type) {
   return new Map([
     ...globalExtensionArgs,
     [
-      [`${type.toUpperCase()}NAME`],
+      [`${type}Name`],
       {
         type: 'str',
         help:
@@ -152,7 +152,7 @@ function makeUninstallArgs(type) {
   return new Map([
     ...globalExtensionArgs,
     [
-      [`${type.toUpperCase()}NAME`],
+      [`${type}Name`],
       {
         type: 'str',
         help:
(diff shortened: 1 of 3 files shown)


packages/appium/lib/cli/args.js
```diff
@@ -108,13 +108,12 @@ function makeInstallArgs(type) {
   return new Map([
     ...globalExtensionArgs,
     [
-      [type],
+      [`${type}Name`],
```
mykola-mokhnach, line 111: would this change the actual argument behaviour?

eglitise: Hmm, seems like it. I guess I will revert this for now, but I'd still prefer to change it somehow - I just can't yet figure out where exactly it's used.

mykola-mokhnach (approved):

PR #20839, opened 2024-12-14
feat(appium): Add session.status BiDi command

mykola-mokhnach:
Also removed exception throwing from event listener as it might break the whole process due to unhandled async exception


KazuCocoa (approved):
lg

PR #20845, opened 2024-12-17
fix(base-driver): Return an empty object if the corresponding API response is undefined

mykola-mokhnach:
## Proposed changes

This is related to a java-client-related discussion where we figured out that chrome returns an empty object ({}) in response to various BiDi method like session.subscribe, whereas appium returns null, thus breaking the client logic.

## Types of changes


KazuCocoa (approved):
Makes sense to me

PR #20856, opened 2024-12-24
chore: Migrate to flat ESLint config (v9)

mykola-mokhnach:
**breaking**: Removed the eslint-config-appium module from the monorepo as obsolete
**breaking**: The eslint-config-appium-ts module is now the only module to provide a common flat eslint config. 
**breaking**: The eslint-config-appium-ts module is now ESM. 
**breaking**: The eslint-config-appium-ts module requires all clients to migrate their linting configs to the [flat config model](https://eslint.org/docs/latest/use/migrate-to-9.0.0)

It was also necessary to update various other sources to address linting issues caused by the above updates. 
Neverteless we'd need to make sure that this particular change only bumps the majot version of the eslint-config-appium-ts package and nothing else while publishing. That is why I've also bumped the version number manually for the above in the package.json .


package.json
```diff
@@ -85,6 +85,8 @@
   },
   "devDependencies": {
     "@colors/colors": "1.6.0",
+    "@eslint/eslintrc": "^3.1.0",
+    "@eslint/js": "^9.10.0",
```
eglitise, line 89: These dependencies should probably be set to an exact version for consistency

packages/base-driver/test/unit/protocol/routes.spec.js
```diff
@@ -53,11 +51,6 @@ describe('Protocol', function () {
       cmdName.should.equal('timeouts');
     });
 
-    it('should properly lookup correct command name for endpoint with session', function () {
```
eglitise, line 56: Is this test no longer necessary?

package.json
mykola-mokhnach: not really. These are peer dependencies and it makes sense to have them as a range to avoid unnecessary conflicts with a container package

packages/eslint-config-appium-ts/index.mjs
```diff
@@ -0,0 +1,231 @@
...
+
+  {
+    name: 'Default Ignores',
+    ignores: [
+      ...(fs.existsSync(gitignorePath) ? includeIgnoreFile(gitignorePath).ignores : []),
+      '**/*-d.ts',
+      '**/build/**',
+      '**/*.min.js',
+      '**/coverage/**',
+      '**/node_modules/**',
```
eglitise, line 227: This line may not be necessary since [it's one of the default patterns](https://eslint.org/docs/latest/use/configure/ignore#ignoring-files)

package.json
eglitise: The `eslint-config-appium-ts` `package.json` file does indeed use ranges just like before, but this particular file is the monorepo-level `package.json`, where all other dependencies are set to exact versions.

eslint.config.mjs
```diff
@@ -0,0 +1,35 @@
...
+    files: [
+      'packages/test-support/lib/**',
+      'packages/driver-test-support/lib/**',
+      'packages/plugin-test-support/lib/**',
+    ],
+  },
+  {
+    ignores: [
+      'examples/javascript-wd',
+      'sample-code',
```
eglitise, line 27: It looks like these paths no longer exist in the monorepo

packages/eslint-config-appium-ts/index.mjs
```diff
@@ -0,0 +1,231 @@
...
+      'mocha/no-setup-in-describe': 'off',
+      'mocha/no-exports': 'off',
+      'import/no-named-as-default-member': 'off',
+      'mocha/no-skipped-tests': 'off',
+    },
+  },
+
+  {
+    name: 'Default Ignores',
+    ignores: [
```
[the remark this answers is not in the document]
mykola-mokhnach: added

package.json
mykola-mokhnach: fixed, this was a copypaste issue

packages/eslint-config-appium-ts/index.mjs
mykola-mokhnach: removed

eglitise:
Unsure whether it's better to merge `eslint-config-appium` into `eslint-config-appium-ts`, or the other way around. I'm leaning more towards the latter option. Yes, it would require updating dependents of `eslint-config-appium-ts` to use `eslint-config-appium` instead, but the package name would be simpler and more obvious for all dependents, and the manual version bump done here can be avoided (it could probably be done in a separate PR with a minor change labeled as breaking), thereby also avoiding the need to manually update the changelog afterwards.

eslint.config.mjs
mykola-mokhnach: Looks like it's some legacy stuff from the Appium 1 era. Removed

mykola-mokhnach:
> Unsure whether it's better to merge `eslint-config-appium` into `eslint-config-appium-ts`, or the other way around. I'm leaning more towards the latter option. Yes, it would require updating dependents of `eslint-config-appium-ts` to use `eslint-config-appium` instead, but the package name would be simpler and more obvious for all dependents, and the manual version bump done here can be avoided (it could probably be done in a separate PR with a minor change labeled as breaking), thereby also avoiding the need to manually update the changelog afterwards.

I don't have strong preferences on either option. @KazuCocoa @jlipps which one would you prefer?

KazuCocoa:
I don't have any strong opinions about that either, but the merge idea seems good for making deps simple. Then, it was probably a reasonable idea to keep one package and make the other deprecate to me.


Btw, below is on this PR's version updates:
```
% npx lerna version --no-git-tag-version --no-push --allow-branch eslint_bump
lerna notice cli v8.1.9
lerna info versioning independent
lerna info Looking for changed packages since @appium/base-driver@9.13.1
lerna info ignoring diff in paths matching [ '**/test/**', '**/*.md' ]
lerna WARN version Skipping working tree validation, proceed at your own risk
lerna info getChangelogConfig Successfully resolved preset "conventional-changelog-conventionalcommits"

Changes:
 - appium: 2.13.1 => 2.14.0
 - @appium/base-driver: 9.13.1 => 9.14.0
 - @appium/base-plugin: 2.2.50 => 2.2.51
 - @appium/docutils: 1.0.26 => 1.1.0
 - @appium/driver-test-support: 0.7.1 => 0.8.0
 - @appium/eslint-config-appium-ts: 1.0.0 => 1.1.0
 - @appium/execute-driver-plugin: 3.0.36 => 4.0.0
 - @appium/fake-driver: 5.5.7 => 5.6.0
 - @appium/fake-plugin: 3.1.54 => 3.2.0
 - @appium/images-plugin: 3.0.24 => 3.1.0
 - @appium/plugin-test-support: 0.3.46 => 0.3.47
 - @appium/support: 6.0.0 => 6.1.0
 - @appium/test-support: 3.1.0 => 3.2.0
 - @appium/types: 0.22.3 => 0.23.0
 - @appium/universal-xml-plugin: 1.0.26 => 1.1.0

? Are you sure you want to create these versions? No
```

just a note

mykola-mokhnach:
yes, I'm ok with these versions. Perhaps, we could skip changing the major package version manually in this PR and do it in the next one to just properly keep the list of breaking changes.

I'm more interested about what you think regarding the package deprecation: e.g. should we deprecate `eslint-config-appium` or `eslint-config-appium-ts`?

KazuCocoa:
Then, the deprecation idea itself is good to me. I like to keep `-ts` to show it supports typescript (personally) but not sure this naming is helpful today for JS world :P

eglitise:
> Btw, below is on this PR's version updates:

FYI this also includes changes from other PRs since the last Appium release, which explains the major version bump for `execute-driver-plugin`.

> Perhaps, we could skip changing the major package version manually in this PR and do it in the next one to just properly keep the list of breaking changes.

Agreed - everything other than the `eslint-config-appium[-ts]` changes can just be `chore`.
Note that if `eslint-config-appium-ts` is selected to remain, in the second PR it would still need a manual version bump, since Lerna cannot bump a package from 0.x to 1.0 (meanwhile this will not be an issue with `eslint-config-appium`).

mykola-mokhnach:
Ok, then I change this PR to be a `chore` change and we keep the `eslint-config-appium-ts` one

KazuCocoa (approved):
lgtm, in my best knowledge

eglitise:
Checked the actual rules list just now, for a reference on what exactly was changed.
These rules were modified:
* `@typescript-eslint/no-var-requires`: 'error' ->'off'
* `no-redeclare`: 1 ->'off'
* `import/no-unresolved` no longer set to 'off' for `*.test-d.ts` files

These rules were added for all files:
* `@typescript-eslint/no-require-imports`: 'off'
* `@typescript-eslint/no-empty-object-type`: 'off'
* `no-dupe-class-members`: 'off'
* `import/named`: 'warn'

These rules were added for test files:
* `mocha/max-top-level-suites`: 'off'
* `mocha/consistent-spacing-between-blocks`: 'off'
* `@typescript-eslint/no-unused-expressions`: 'off'
* `mocha/no-setup-in-describe`: 'off'
* `mocha/no-exports`: 'off'
* `import/no-named-as-default-member`: 'off'
* `mocha/no-skipped-tests`: 'off'

For ease of reference, would it perhaps be possible to group/sort the list of rules?

mykola-mokhnach:
> For ease of reference, would it perhaps be possible to group/sort the list of rules?

Yes, this would be useful. Would you like to do it in another PR?

PR #20876, opened 2025-01-03
feat: allow plugins to define custom bidi commands and emit bidi events

jlipps:


packages/appium/docs/en/developing/build-plugins.md
```diff
@@ -368,7 +371,58 @@ clients won't have nice client-side functions designed to target these endpoints
 to create and release client-side plugins for each language you want to support (directions or
 examples can be found at the relevant client docs).
 
-An alternative to this way of doing things is to overload a command which all WebDriver clients
+#### Extending the BiDi Protocol
+
+You can also make new commands accessible via the WebDriver BiDi (WebSocket-based) protocol. BiDi
+commands come in two parts: a "module", which is basically a container or namespace, and
```
mykola-mokhnach, line 377: According to what we have discussed with @christian-bromann BiDi modules must also have a vendor prefix if they are not standard. For example, we follow the below naming strategy:

`${vendorPrefix}:${domain}.${eventName}`

An example of such event would be the custom `contextUpdated` event in xcuitest driver: `appium:xcuitest.contextUpdated`

packages/appium/lib/appium.js
```diff
@@ -140,17 +140,6 @@ class AppiumDriver extends DriverCore {
     })();
   }
 
-  /**
```
mykola-mokhnach, line 143: Why this method has been removed?

packages/appium/docs/en/developing/build-drivers.md
```diff
@@ -320,10 +320,6 @@ your driver of the appropriate name, and it will be called when the BiDi command
 the client. To see which specific names you should use for BiDi commands, have a look at
 [bidi-commands.js](https://github.com/appium/appium/blob/master/packages/base-driver/lib/protocol/bidi-commands.js)
 
-Currently, you also need to define a `doesSupportBidi` field on your driver instances, and ensure
```
mykola-mokhnach, line 323: removing `doesSupportBidi` is a breaking change, isn't it?

packages/appium/lib/bidi.ts
```diff
@@ -95,14 +100,27 @@ export function onBidiConnection(this: AppiumDriver, ws: WebSocket, req: Incomin
   }
 }
 
+function wrapCommandWithPlugins(driver: ExtensionCore, plugins: ExtensionCore[], method: string, params: StringRecord) {
+    const [moduleName, methodName] = method.split('.');
+    let next = async () => await driver.executeBidiCommand(method, params);
+    for (const plugin of plugins.filter((p) => p.doesBidiCommandExist(moduleName, methodName))) {
+      next = ((_next) => async () => {
+        return await plugin.executeBidiCommand(method, params, _next, driver);
```
mykola-mokhnach, line 108: curly brackets and `return` are redundant

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -106,14 +98,10 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
-  bidiEventSubs: Record<string, string[]>;
-
-  doesSupportBidi: boolean;
-
-  bidiCommands: BidiModuleMap = BIDI_COMMANDS as BidiModuleMap;
-
   constructor(opts: InitialOpts = <InitialOpts>{}, shouldValidateCaps = true) {
-    this._log = logger.getLogger(helpers.generateDriverLogPrefix(this as Core<C>));
+    super();
+    this._log = this.log; // TODO: remove references to this._log and use this.log instead
```
mykola-mokhnach, line 103: is this change related to the one in PR title? would it be better to split them?

packages/base-driver/lib/basedriver/extension-core.ts
```diff
@@ -0,0 +1,118 @@
+import {logger} from '@appium/support';
```
mykola-mokhnach, line 1: 👍

packages/base-driver/lib/basedriver/extension-core.ts
```diff
@@ -0,0 +1,118 @@
...
+  bidiEventSubs: Record<string, string[]>;
+  bidiCommands: BidiModuleMap = BIDI_COMMANDS as BidiModuleMap;
+  log: AppiumLogger;
+  // used to handle driver events
+  eventEmitter: NodeJS.EventEmitter;
+
+
+  constructor(logPrefix?: string) {
+    logPrefix ??= helpers.generateDriverLogPrefix(this);
+    this.log = logger.getLogger(logPrefix);
```
mykola-mokhnach, line 27: I expect it to be a getter/readonly field

packages/base-plugin/lib/plugin.js
```diff
@@ -45,9 +45,10 @@ class BasePlugin {
    * @param {Record<string,unknown>} [cliArgs]
    */
   constructor(name, cliArgs = {}) {
+    super(`Plugin [${name}]`);
     this.name = name;
     this.cliArgs = cliArgs;
-    this.logger = logger.getLogger(`Plugin [${name}]`);
+    this.logger = this.log; // TODO deprecate 'logger' in favor of 'log'
```
mykola-mokhnach, line 51: you could already mark the property as `@deprecated`

packages/types/lib/driver.ts
```diff
@@ -610,7 +610,6 @@ export interface Core<C extends Constraints, Settings extends StringRecord = Str
   isCommandsQueueEnabled: boolean;
   eventHistory: EventHistory;
   bidiEventSubs: Record<string, string[]>;
-  doesSupportBidi: boolean;
```
mykola-mokhnach, line 613: this property must not be removed as deleting it may break existing drivers, but instead marked as deprecated.

packages/base-plugin/lib/plugin.js
```diff
@@ -45,9 +45,10 @@ class BasePlugin {
    * @param {Record<string,unknown>} [cliArgs]
    */
   constructor(name, cliArgs = {}) {
+    super(`Plugin [${name}]`);
```
mykola-mokhnach, line 48: I would prefer if plugin logger could also somehow share the the driver hash prefix, so it would be visible in logs which driver has initiated an action.

Also, usually people already have the `Plugin` suffix in their names, so I would just put the class name instead: `super(this.constructor.name)`

jlipps:
hmm, tests are timing out, need to figure that out.

jlipps:
ok @mykola-mokhnach, CI is now green, please feel free to do another review after updates from your comments.

packages/appium/docs/en/developing/build-plugins.md
```diff
@@ -368,7 +371,60 @@ clients won't have nice client-side functions designed to target these endpoints
...
+```
+
+In this imaginary example, we have defined two new BiDi commands: `appium:video.startFramerateCapture` and
+`appium:video.stopFramerateCapture`. The first command takes a required and an optional parameter, and the
+second does not. When you have implemented the `startFrameCap` and `stopFrameCap` methods on your
+plugin class, they will be called whenever the BiDi commands are triggered by a client. The
+signatures for these methods would look as follows:
+
+```js
+async startFrameCap(next, driver, videoSource, showOnScreen): Promise<any>;
```
mykola-mokhnach, line 410: I assume we also need typedefs for args

packages/appium/lib/appium.js
```diff
@@ -626,12 +615,16 @@ class AppiumDriver extends DriverCore {
...
-   * @returns {Array} - array of plugin instances
+   * @returns {Array<import('@appium/types').Plugin>} - array of plugin instances
    */
   pluginsForSession(sessionId = null) {
     if (sessionId) {
       if (!this.sessionPlugins[sessionId]) {
-        this.sessionPlugins[sessionId] = this.createPluginInstances();
+        const driverId = generateDriverLogPrefix(this.sessions[sessionId]);
+        if (!driverId) {
+          throw new Error(`no driver id`);
```
mykola-mokhnach, line 625: can we have a better error message?

packages/appium/lib/appium.js
```diff
@@ -662,14 +655,16 @@ class AppiumDriver extends DriverCore {
...
    */
-  createPluginInstances() {
+  createPluginInstances(driverId = null) {
     /** @type {Plugin[]} */
     const pluginInstances = [];
     for (const [PluginClass, name] of this.pluginClasses.entries()) {
       const cliArgs = this.getCliArgsForPlugin(name);
-      const plugin = new PluginClass(name, cliArgs);
+      const plugin = new PluginClass(name, cliArgs, driverId);
+      /** @type {Plugin & ExtensionCore} */(plugin).updateBidiCommands(PluginClass.newBidiCommands ?? {});
```
mykola-mokhnach, line 667: this would probably fail if an older base plugin is used. Need to explicitly check if updateBidiCommands fmethod is present on the instance

packages/appium/lib/bidi.ts
```diff
@@ -95,14 +100,25 @@ export function onBidiConnection(this: AppiumDriver, ws: WebSocket, req: Incomin
   }
 }
 
+function wrapCommandWithPlugins(driver: ExtensionCore, plugins: ExtensionCore[], method: string, params: StringRecord) {
```
mykola-mokhnach, line 103: can we also have the return type declared?

packages/appium/lib/bidi.ts
```diff
@@ -381,56 +407,63 @@ function initBidiEventListeners(
...
+        const ev = {type: 'event', context, method, params};
+        await send(JSON.stringify(ev));
+      }
+    };
+    return eventListener;
   };
-  driverEe.on(BIDI_EVENT_NAME, eventListener);
+  bidiHandlerDriver.eventEmitter.on(BIDI_EVENT_NAME, eventListenerFactory('driver', bidiHandlerDriver as ExtensionCore));
+  for (const plugin of bidiHandlerPlugins) {
+    plugin.eventEmitter.on(BIDI_EVENT_NAME, eventListenerFactory('plugin', plugin));
```
mykola-mokhnach, line 467: we should explicitly check for eventEmitter property presence

packages/base-plugin/lib/plugin.js
```diff
@@ -43,11 +49,16 @@ class BasePlugin {
   /**
    * @param {string} name
    * @param {Record<string,unknown>} [cliArgs]
+   * @param {string|null} driverId
    */
-  constructor(name, cliArgs = {}) {
+  constructor(name, cliArgs = {}, driverId = null) {
+    super();
+    if (driverId) {
+      this.updateLogPrefix(`${generateDriverLogPrefix(this)} (for driver ${driverId})`);
```
mykola-mokhnach, line 57: I would prefer shorter log prefixes, otherwise this `(for driver ` part would be repeated for each log line thus making less space for useful message payload. What if we format it as `${generateDriverLogPrefix(this)}<${driverId}>`?

packages/appium/lib/appium.js
jlipps: I don't think this would fail. You can send extra parameters to functions in JS if you want, without error.

packages/appium/lib/appium.js
jlipps: oops, this is debug stuff i accidentally left in.

packages/appium/lib/appium.js
jlipps: oh, you're referring to the `updateBidiCommands`, yes you're right.

mykola-mokhnach: It's not about parameters, but rather about the presence of `updateBidiCommands` instance method

packages/appium/lib/bidi.ts
jlipps: TS is smart enough to figure out the return type implicitly, is there a benefit in adding it explicitly?

mykola-mokhnach: yes, it helps the linter to assert that the variable returned by function has the expected type

packages/base-driver/lib/basedriver/extension-core.ts
```diff
@@ -0,0 +1,128 @@
...
+import _ from 'lodash';
+import {generateDriverLogPrefix} from './helpers';
+
+export class ExtensionCore {
+  bidiEventSubs: Record<string, string[]>;
+  bidiCommands: BidiModuleMap = BIDI_COMMANDS as BidiModuleMap;
+  _logPrefix?: string;
+  protected _log: AppiumLogger;
+  // used to handle driver events
+  eventEmitter: NodeJS.EventEmitter;
```
mykola-mokhnach, line 23: consider using readonly for properties that are not going to be reassigned  during instance lifecycle

packages/base-driver/lib/basedriver/extension-core.ts
```diff
@@ -0,0 +1,128 @@
...
+
+
+  constructor(logPrefix?: string) {
+    this._logPrefix = logPrefix;
+    this.bidiEventSubs = {};
+    this.eventEmitter = new EventEmitter();
+  }
+
+  get log() {
```
mykola-mokhnach, line 32: Please provide return types

mykola-mokhnach (approved):

packages/base-plugin/lib/plugin.js
```diff
@@ -43,11 +49,16 @@ class BasePlugin {
   /**
    * @param {string} name
    * @param {Record<string,unknown>} [cliArgs]
+   * @param {string|null} driverId
    */
-  constructor(name, cliArgs = {}) {
+  constructor(name, cliArgs = {}, driverId = null) {
+    super();
+    if (driverId) {
+      this.updateLogPrefix(`${generateDriverLogPrefix(this)} <${driverId}>`);
```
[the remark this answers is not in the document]
jlipps: I don't think so.

Issue #20878, opened 2025-01-05
🐞 bug: Original error: driverInstance.updateBidiCommands is not a function is thrown when starting UiAutomator2 session with Appium inspector using Appium server version v.2.14.0


AndriyShykin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Appium Version

2.14.0

### Appium Host Type

Virtual Machine/Docker

### Are you using Appium components not maintained by the Appium team?

-

### Is this issue reproducible using the latest components?

- [X] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

Original error: driverInstance.updateBidiCommands is not a function is thrown when starting UiAutomator2 session with Appium inspector v. 2024.12.1 and Appium server version v.2.14.0 installed from NPM. UiAutomator2 driver version is 3.9.6.

<img width="1095" alt="Screenshot 2025-01-05 at 10 31 57" src="https://github.com/user-attachments/assets/2b537a05-b3a4-430d-8cb8-b02527b3c318" />

### Expected Behavior

The session starts or fails with a different error.

### Appium Log

https://gist.github.com/AndriyShykin/58055ac37867f973d41932e8d6e91e1a

### Environment Details

- Debug details (`appium --show-debug-info`):{
  "os": {
    "platform": "linux",
    "release": "6.8.0-51-generic",
    "arch": "x64",
    "homedir": "/home/appuser",
    "username": "appuser"
  },
  "node": {
    "version": "v23.5.0",
    "arch": "x64",
    "cwd": "/home/appuser",
    "argv": [
      "/usr/local/bin/node",
      "/home/appuser/node_modules/appium/build/lib/main.js",
      "--show-debug-info"
    ],
    "env": {
      "PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "HOSTNAME": "dde761956faf",
      "NODE_VERSION": "23.5.0",
      "YARN_VERSION": "1.22.22",
      "ANDROID_HOME": "/opt/android",
      "JAVA_HOME": "/usr",
      "HOME": "/home/appuser",
      "NODE_PATH": "/home/appuser/node_modules",
      "APPIUM_OMIT_PEER_DEPS": "1"
    },
    "npm": {
      "location": "unknown",
      "version": "unknown"
    }
  },
  "appium": {
    "location": "/home/appuser/node_modules/appium",
    "homedir": "/home/appuser",
    "build": {
[...]

PR #20879, opened 2025-01-05
fix(appium): Do not update BiDi commands for drivers which don't support it

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/20878


eglitise (approved):

jlipps (commented):
yikes, good catch! I guess this was a breaking basedriver change!

PR #20887, opened 2025-01-06
fix(eslint-config-appium-ts): add Prettier config, update typescript-eslint imports

eglitise:
## Proposed changes

I noticed 2 things with the recently changed `eslint-config-appium-ts`:
- There was an unused import for `eslint-config-prettier`. Since the `eslintrc` version of `eslint-config-appium-ts` was extending this config, it has been added back.
- `typescript-eslint` had changed its suggested imports for use with flat config, but this config was still using the old imports. They have now been updated.

## Types of changes


KazuCocoa (approved):
👀

eslint.config.mjs
```diff
@@ -1,4 +1,13 @@
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+import fs from 'node:fs';
+
 import tsConfig from '@appium/eslint-config-appium-ts';
+import {includeIgnoreFile} from '@eslint/compat';
+
+const __filename = fileURLToPath(import.meta.url);
```
mykola-mokhnach, line 8: I think this would be not really necessary after https://github.com/appium/appium/pull/20888 is merged

packages/eslint-config-appium-ts/index.mjs
```diff
@@ -1,24 +1,17 @@
...
-import {includeIgnoreFile} from '@eslint/compat';
-
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
-const gitignorePath = path.resolve(__dirname, '.gitignore');
+import {configs as tsConfigs, parser as tsParser, plugin as tsPlugin} from 'typescript-eslint';
 
 export default [
   js.configs.recommended,
+  eslintConfigPrettier,
```
mykola-mokhnach, line 11: might this change raise new linter alerts after the module version is bumped?

packages/eslint-config-appium-ts/index.mjs
```diff
@@ -1,24 +1,17 @@
...
   pluginPromise.configs['flat/recommended'],
   importPlugin.flatConfigs.recommended,
+  ...tsConfigs.recommended,
```
mykola-mokhnach, line 14: might this change raise new linter alerts after the module version is bumped?

eslint.config.mjs
eglitise: Agreed, I'll update this PR afterwards

mykola-mokhnach (approved):

Issue #20881, opened 2025-01-05
💡 feat: add new driver endpoints to retrieve available commands and extensions


eglitise:
### Is your feature request related to a problem?

Appium Inspector has a Commands tab, which allows the user to execute various driver commands and extensions. However, different drivers support different commands and extensions, and currently there is no way to identify the commands and extensions supported by the active driver.

### Describe the solution you'd like.

Add two new endpoints:
* `GET /session/:sessionId/appium/commands`
* `GET /session/:sessionId/appium/extensions`

### `GET /session/:sessionId/appium/commands`

This endpoint should return all commands defined as endpoints in the active driver, the base driver, and any active plugins.

The response format should be largely based around the data structure format currently used for [the base driver's method map](https://github.com/appium/appium/blob/master/packages/base-driver/lib/protocol/routes.js#L24). The same format should also be used by the active driver and plugins. The endpoint response will then combine all of these responses into one object.

In short, the format of each response should be an object where each key-value pair corresponds to a command. The key should be the command endpoint URL, and its value should also be an object, where each key-value pair corresponds to a supported HTTP method. For this command value object, the key should be the HTTP method name, and the value should be an object with the following properties:
* `info` (string, optional) - description with additional information about the command, e.g. 'Simulator only'
* `deprecated` (boolean, optional) - indicates whether the command is deprecated. `false` if omitted
* `payloadParams` (object, optional) - an object containing all the supported command parameters. It supports the following keys:
  * `required` (array, optional) - an array listing the names of all required parameters
  * `optional` (array, optional) - an array listing the names of all optional parameters

Example response for `GET /session/:sessionId/appium/commands`:
```json
{
  "basedriver": {
    "/session/:sessionId/url": {
      "GET": {"command": "getUrl"},
      "POST": {"command": "setUrl", "payloadParams": {"required": ["url"]}},
    },
    ...
  },
  "endpointdriver": {
    "/session/:sessionId/appium/device/lock": {
      "POST": {"command": "lock", "payloadParams": {"optional": ["seconds"]}},
    },
    ...
  },
  "plugins": {
    "images": {
      "/session/:sessionId/appium/compare_images": {
        "POST": {
          "command": "compareImages",
          "payloadParams": {
            "required": ["mode", "firstImage", "secondImage"],
            "optional": ["options"],
          }
        }
      }
      ...
    }
    ...
  }
}
```

Remaining questions:
* The method map cannot be returned as-is and will require additional processing:
[...]

PR #20914, opened 2025-01-17
feat(base-driver): Add an API to list commands

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/20881


packages/base-driver/lib/protocol/routes.js
```diff
@@ -868,7 +868,15 @@ export const METHOD_MAP = /** @type {const} */ ({
     POST: {command: 'executeCdp', payloadParams: {required: ['cmd', 'params']}},
   },
 
-  //region Webauthn
+  // #region Inspector
+
+  '/session/:sessionId/appium/commands': {
+    POST: { command: 'listCommands'},
```
KazuCocoa, line 874: (minor)
```suggestion
    POST: {command: 'listCommands'},
```

KazuCocoa (commented):
I haven't test this on my local, otherwise lgtm so far

mykola-mokhnach:
Open questions:
- Not all drivers support extension commands (currently only XCUITest driver does)
- Current command type definition doesn't have the `info` field. Shall we add it first?
- It is unclear how to detect if a command is `readonly`
- It is unclear where to fetch an argument type
- It is unclear what to do with BiDi commands

packages/appium/lib/appium.js
```diff
@@ -565,6 +568,31 @@ class AppiumDriver extends DriverCore {
...
+  listCommands(sessionId) {
+    if (!sessionId) {
+      return {};
+    }
+
+    // @ts-ignore It's ok if the newMethodMap property is not there
+    const driverMethodMap = this.driverForSession(sessionId)?.newMethodMap ?? {};
+    const pluginMethodMaps = _.flatMap(
+      // @ts-ignore It's ok if the newMethodMap property is not there
+      this.pluginsForSession(sessionId).map((p) => p.newMethodMap ?? {})
```
eglitise, line 584: Should this be `p?.newMethodMap`?

packages/appium/lib/appium.js
```diff
@@ -940,10 +968,15 @@ class AppiumDriver extends DriverCore {
...
+/**
+ * Help decide which commands should be proxied to sub-drivers and which
+ * should be handled by this, our umbrella driver
+ * @param {string} cmd
+ * @returns {boolean}
+ */
 function isAppiumDriverCommand(cmd) {
-  return !isSessionCommand(cmd) || cmd === DELETE_SESSION_COMMAND;
+  return !isSessionCommand(cmd)
+    || _.includes([DELETE_SESSION_COMMAND, LIST_DRIVER_COMMANDS_COMMAND], cmd);
```
eglitise, line 979: We may want to define a new const that aggregates these driver-level commands, since the goal is to eventually add another one for listing available extension commands

packages/base-driver/lib/basedriver/commands/inspector.ts
```diff
@@ -0,0 +1,23 @@
+import type {Constraints, IInspectorCommands, CommandsMap} from '@appium/types';
```
eglitise, line 1: Should this file be added to the imports of `base-driver/lib/basedriver/commands/index.ts`?

eglitise:
> Not all drivers support extension commands (currently only XCUITest driver does)

I'm not sure I understand this. UiA2 and Espresso drivers both also have `mobile:` commands, Mac2 driver has `macos:` commands, etc.

> Current command type definition doesn't have the info field. Shall we add it first?

Sure - it should be optional anyway, so no risk of breaking anything.

> It is unclear how to detect if a command is readonly

Looking at the method map, I think we can omit this and just use the method name for this purpose: `GET` methods should not trigger a refresh, whereas `POST` should.

> It is unclear where to fetch an argument type

I think this is something that could ideally also be added to the method map.

> It is unclear what to do with BiDi commands

This one I'm not really sure about. Their implementation seems different enough from regular commands that it could be moved to its own endpoint.

> There a commands with duplicated names, which makes it impossible to make the name to a mapping key. Such commands may differ by their paths or method names

True, but aren't command names specifically meant to be used by end users calling them? It doesn't really make much sense to have them be duplicated. Identical command names for different method names (e.g. for `/session/:sessionId/element/active`) should probably be changed, whereas identical command names for different paths are (from what I can see) used for JSONWP commands, which should also be removed in the future.

> Shall we somehow differentiate commands belonging to the base driver/endpoint driver/plugin extensions? If we should then how to do it best?

Good idea - we could modify the response to split such groups into their own key-value pairs, e.g.
```
{
  "basedriver": {...baseDriverCommands},
  "endpointdriver": {...endpointDriverCommands},
  "plugins": {
    "pluginName": {...pluginCommands},
    ...
  }
}
```

eglitise:
I also realised that it's probably fine to adjust the response to more closely resemble the current method map format. It already distinguishes `GET` and `POST` methods for the same endpoint, and it could solve the issue with parameter names being arrays, by just returning the data in the exact same format.

packages/base-driver/lib/basedriver/commands/inspector.ts
mykola-mokhnach: why should it? it's just a mixin. All its methods are available through bae driver

mykola-mokhnach:
> I'm not sure I understand this. UiA2 and Espresso drivers both also have mobile: commands, Mac2 driver has macos: commands, etc.

This is the mobile commands implementation for the XCUITest driver: https://github.com/appium/appium-xcuitest-driver/blob/master/lib/execute-method-map.ts

This is the one for mac2 driver: https://github.com/appium/appium-mac2-driver/blob/master/lib/commands/execute.js

As you could see the above driver stores all extension commands into the `executeMethodMap` static property of the driver, which then allows us to fetch these from the base driver, while the latter neither exposes the mapping in that way nor makes it unified, which makes it impossible to compile this data from the base/umbrella  driver. Basically, to make that work we'd need to make sure all drivers use the unified interface via `executeMethodMap`. Which is a major refactoring

mykola-mokhnach:
> I think this is something that could ideally also be added to the method map.

This would be a major breaking change. Previously we did not do any type validations on params (at least the majority of them). All necessary validations are in API handlers instead.

mykola-mokhnach:
> True, but aren't command names specifically meant to be used by end users calling them? It doesn't really make much sense to have them be duplicated. Identical command names for different method names (e.g. for /session/:sessionId/element/active) should probably be changed, whereas identical command names for different paths are (from what I can see) used for JSONWP commands, which should also be removed in the future.

I feel like it would be a risky deal to put a bet on the fact that all command names must be unique. Duplicate names exist not only because JWP/W3C, but some commands also have POST/GET aliases, since the latter does not allow to pass parameters (yes, one could pass GET params in the query, but none of Selenium clients allows that, thus only POST).

mykola-mokhnach:
```
{
  "basedriver": {...baseDriverCommands},
  "endpointdriver": {...endpointDriverCommands},
  "plugins": {
    "pluginName": {...pluginCommands},
    ...
  }
}
```

Splitting commands by plugins makes sense, although I am not sure if vanilla users even understand the concept of basedriver or need to know about it.

eglitise:
> Basically, to make that work we'd need to make sure all drivers use the unified interface via `executeMethodMap`

This does seem likely. Well, it doesn't need to be implemented right away, but we could define it as a guideline somewhere in the documentation and adjust the other drivers over time.

> This would be a major breaking change. Previously we did not do any type validations on params (at least the majority of them). All necessary validations are in API handlers instead.

In theory it would make sense to have the expected parameter types be listed along with the parameters themselves, but I agree it would be a breaking change. Maybe we could specify the types without utilising them, and then require their usage with Appium 3's major basedriver version?

> some commands also have POST/GET aliases, since the latter does not allow to pass parameters

Then I'm not sure why the GET alias is necessary - though that's beside the point. As far as the end user is concerned, only the command name is relevant, and the endpoints or method names are abstracted away.
I guess the issue here is that the command names may differ depending on the client. We can work around it by using the endpoint as a key, just like in the current structure.

> although I am not sure if vanilla users even understand the concept of basedriver or need to know about it.

I agree, and I don't see a reason to separate them in the Inspector GUI either - but it may still be useful for other reasons in the future.

mykola-mokhnach:
@eglitise I would like to agree the response structure of the listCommands API based on the above conversation.

Currently the output looks like

```js
{"rest":[
  {"path":"/status","method":"GET","name":"getStatus"},  
  {"path":"/session","method":"POST","name":"createSession","params": [
    {"name":"desiredCapabilities","required":false},
    {"name":"requiredCapabilities","required":false},
    {"name":"capabilities","required":false}
  ]} ...
```

Could you please provide the updated json structure example?

eglitise:
@mykola-mokhnach I've now updated the issue description for #20881 based on the above discussion. I have also listed several remaining questions there.

eglitise (approved):

jlipps (approved):

Issue #20880, opened 2025-01-05
💡 feat: replace non-W3C endpoints used in Inspector


eglitise:
### Is your feature request related to a problem?

Appium Inspector currently relies on two non-W3C endpoints that we'd like to remove in Appium 3:
### `GET /sessions`
Purpose: discover existing sessions and attach to them.

Response format: list of session objects, where each object has `id` and `capabilities` keys, for example:
```json
[
  {
    "id":"ba30c6da-c266-4734-8ddb-c16f5bb53e16",
    "capabilities":{ "platformName":"iOS","automationName":"XCUITest","platformVersion":"17.2","deviceName":"iPhone 15" }
  },
  {
    "id":"1441110c-1ece-4e45-abbf-ebf404f45f0a",
    "capabilities":{ "platformName":"Android","automationName":"UiAutomator2" }
  },
  ...
]
```

### `GET /session/:sessionId`
Purposes:
* retrieve the currently used capabilities while in a session.
* retrieve event timings: http://appium.io/docs/en/latest/guides/event-timing/

Response format: object where each entry corresponds to a capability and its value. Additional keys for event timings can also be present, such as `commands` and event type names, for example:
```json
{
  "platformName":"Android",
  "automationName":"UiAutomator2",
  "commands": [
    {
      "cmd": "click",
      "startTime": 1736092760,
      "endTime": 1736092762
    },
    ...
  ]
}
```

### Describe the solution you'd like.

Both endpoints should be replaced with new W3C-compliant endpoints.

### `GET /sessions` -> `GET /appium/sessions`
Purpose: same as before - discover existing sessions and attach to them. However, server owners may not want to expose this information, so this functionality should be disabled by default, and only available behind the `discover_sessions` feature flag.

Response format: same as before (list of session objects, where each object has `id` and `capabilities` keys), with a new additional key `created`. This key is required, and its value should be an integer matching the Unix timestamp (in milliseconds) of the session creation time. For example:
```json
[
  {
    "id":"ba30c6da-c266-4734-8ddb-c16f5bb53e16",
    "created": 1736092760555,
    "capabilities":{ "platformName":"iOS","automationName":"XCUITest","platformVersion":"17.2","deviceName":"iPhone 15" }
  },
  {
    "id":"1441110c-1ece-4e45-abbf-ebf404f45f0a",
    "created": 1736092460555,
    "capabilities":{ "platformName":"Android","automationName":"UiAutomator2" }
  },
  ...
]
```

[...]

PR #20936, opened 2025-01-25
feat: add /appium/sessions, /session/:sessionId/appium/capabilities and deprecated marks will be removed in the future

KazuCocoa:
## Proposed changes

This doesn't include breaking changes.

- https://github.com/appium/appium/issues/20880
    - Added `/appium/sessions`, which is the same as exiting `/sessions`
    - Added `/session/:sessionId/appium/capabilities`, which is the same as existing `GET` of `/session/:sessionId`
- Added `deprecated` property for non-vendor prefix endpoints and have alternatives in W3C, `/appium` prefix or extentions
    - Not in W3C spec (to drop them in Appium 3)
- Moved non-`/appium` endpoints but still used by Appium in Appium only comment section
    - These are candidates we need to define extensions, or keep them
    - I don't add `deprecated` for them

URLs used to check if endpoints are deprecated or not in w3c
https://www.w3.org/TR/webdriver1/
https://www.w3.org/TR/webdriver2/

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments


packages/base-driver/lib/protocol/routes.js
```diff
@@ -104,39 +101,23 @@ export const METHOD_MAP = /** @type {const} */ ({
...
-  '/session/:sessionId/ime/deactivate': {
-    POST: {command: 'deactivateIMEEngine', deprecated: true},
-  },
-  '/session/:sessionId/ime/activate': {
-    POST: {
-      command: 'activateIMEEngine',
-      payloadParams: {required: ['engine']},
-      deprecated: true,
-    },
-  },
```
KazuCocoa, line 139: Moved to used only by Appium sectio.

packages/base-driver/lib/protocol/routes.js
```diff
@@ -206,7 +199,7 @@ export const METHOD_MAP = /** @type {const} */ ({
   },
   '/session/:sessionId/element/active': {
     GET: {command: 'active'}, // W3C: https://w3c.github.io/webdriver/webdriver-spec.html#dfn-get-active-element
-    POST: {command: 'active'},
```
KazuCocoa, line 209: `GET` is in W3C. `POST` is MJSONWP.

packages/base-driver/lib/protocol/routes.js
```diff
@@ -316,10 +312,6 @@ export const METHOD_MAP = /** @type {const} */ ({
   'session/:sessionId/element/:elementId/computedlabel': {
     GET: {command: 'getComputedLabel'},
   },
-  '/session/:sessionId/rotation': {
-    GET: {command: 'getRotation'},
-    POST: {command: 'setRotation', payloadParams: {required: ['x', 'y', 'z']}},
-  },
```
KazuCocoa, line 322: Moved to Appium only section

packages/base-driver/lib/protocol/routes.js
```diff
@@ -377,10 +369,6 @@ export const METHOD_MAP = /** @type {const} */ ({
       deprecated: true,
     },
   },
-  '/session/:sessionId/location': {
-    GET: {command: 'getGeoLocation'},
-    POST: {command: 'setGeoLocation', payloadParams: {required: ['location']}},
-  },
```
KazuCocoa, line 383: Moved to Appium only section

packages/base-driver/lib/protocol/routes.js
```diff
@@ -763,85 +871,6 @@ export const METHOD_MAP = /** @type {const} */ ({
       payloadParams: {required: ['vendor', 'event']},
     },
   },
-
-  /*
-   * The W3C spec has some changes to the wire protocol.
-   * https://w3c.github.io/webdriver/webdriver-spec.html
-   * Begin to add those changes here, keeping the old version
-   * since clients still implement them.
-   */
```
KazuCocoa, line 772: Moved to the upper side used by W3C specs.

packages/base-driver/lib/protocol/routes.js
```diff
@@ -878,55 +901,30 @@ export const METHOD_MAP = /** @type {const} */ ({
...
-
-  '/session/:sessionId/webauthn/authenticator/:authenticatorId': {
-    DELETE: {
-      command: 'removeVirtualAuthenticator',
-    },
-  },
-
-  '/session/:sessionId/webauthn/authenticator/:authenticatorId/credential': {
+  /*
+   * The W3C spec has some changes to the wire protocol.
```
KazuCocoa, line 905: The remaining items are not used in W3C spec

packages/base-driver/lib/protocol/routes.js
mykola-mokhnach: you may use the `// #region X ... // #endregion` comments to create collapsible regions in various editors, like VSCode.

packages/base-driver/lib/protocol/routes.js
```diff
@@ -465,6 +567,12 @@ export const METHOD_MAP = /** @type {const} */ ({
       payloadParams: {required: ['status', 'value']},
     },
   },
+  '/appium/sessions': {
```
mykola-mokhnach, line 570: I would prefer to define a new handler for this endpoint.
I remember we anyway wanted to put it under a security flag.
Plus we might want to change the actual response format, which would not be possible by using existing command handlers.

It is anyway possible to call the legacy handler from the newly added one.

eglitise: Yes, both this and the capabilities endpoint should have different responses from the current `/sessions` and `/session/:sessionId` endpoints, so the handlers would also need to be different. But I think this implementation may also be added in separate PRs, and this one can focus more on tidying up the existing routes.

packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -111,7 +111,12 @@ class JWProxy {
   }
 
   endpointRequiresSessionId(endpoint) {
-    return !_.includes(['/session', '/sessions', '/status'], endpoint);
+    return !_.includes([
+      '/session',
+      '/sessions',
+      '/status',
+      '/appium/sessions'
```
eglitise, line 118: Do we perhaps want to extend this check to include all possible endpoints that starts with `/appium` (since any endpoint with a session id will start with `/session/:sessionId`)?

eglitise:
Just a suggestion, but I think we can reorganise this file even more and clearly separate deprecated endpoints from the rest:
1) W3C standard endpoints
2) W3C-compliant Appium extension endpoints
3) W3C-compliant extension endpoints from other vendors
4) Non-W3C-compliant endpoints still used in Appium
5) Deprecated endpoints

eglitise:
Other things I noticed:
* 3 endpoints are configured to support both MJSONWP and W3C. Do we want to add comments that we would like to remove the non-W3C parameter support in the future?
  * `/session` should only allow `capabilities`
  * `/session/:sessionId/window` should only allow `handle`
  * `/session/:sessionId/element/:elementId/value` should only allow `value`
* `/session/:sessionId/orientation` isn't in the W3C spec, so should probably be moved to the appropriate section
* Deprecation marks could be added to all endpoints in the `Applications Management` region that don't already have them, plus the next five (with the last one being `/session/:sessionId/appium/device/pull_folder`)
* `/session/:sessionId/element/:elementId/displayed` is an odd endpoint in that it's not explicitly mentioned in the [W3C Endpoints list](https://w3c.github.io/webdriver/#endpoints), but _is_ mentioned in the [Element Displayedness appendix](https://w3c.github.io/webdriver/#element-displayedness). I guess that means it's part of the spec?

KazuCocoa:
For the `displayed`, yes, the endpoint was mentioned in the page, so I haven't left a deprecation mark in this PR for safe because not sure we're confident to remove it. v2 (https://www.w3.org/TR/webdriver2/) also keeps it.

KazuCocoa:
> Deprecation marks could be added to all endpoints in the Applications Management region that don't already have them, plus the next five (with the last one being /session/:sessionId/appium/device/pull_folder)

I haven't checked all our drivers (and clients), so I haven't marked deprecated marks for appium-related endpoints yet. Perhaps we could do it in a separate PR by checking existing drivers'(clients) behavior.

packages/base-driver/lib/jsonwp-proxy/proxy.js
KazuCocoa: Could be? I set explicit endpoints only for now as not sure if we'll add session ids in `/appium` prefix in the future.

KazuCocoa:
I have moved entire deprecated endpoints to the bottom of each section. Some endpoints have partial deprecations (e.g. only POST), so they remain in current place

packages/appium/lib/appium.js
```diff
@@ -195,6 +195,16 @@ class AppiumDriver extends DriverCore {
     }));
   }
 
+  async getAppiumSessions () {
+    throw new errors.NotImplementedError('Not implemented yet. ' +
+      'Please check https://github.com/appium/appium/issues/20880 for more details.');
+  }
+
+  async getAppiumSessionCapabilities () {
```
mykola-mokhnach, line 203: this command name should be explicitly added to the umbrella driver commands list

KazuCocoa: like https://github.com/appium/appium/pull/20936/commits/289bbab0237f0c1070028c45059a00c80d01a782?

packages/appium/lib/appium.js
```diff
@@ -195,6 +196,16 @@ class AppiumDriver extends DriverCore {
...
+    throw new errors.NotImplementedError('Not implemented yet. ' +
+      'Please add getAppiumSessionCapabilities method in the unbrella driver.');
```
mykola-mokhnach, line 206: it's already there %)

packages/appium/lib/appium.js
mykola-mokhnach: yes

packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -111,7 +111,12 @@ class JWProxy {
   }
 
   endpointRequiresSessionId(endpoint) {
-    return !_.includes(['/session', '/sessions', '/status'], endpoint);
+    return !_.includes([
```
mykola-mokhnach, line 114: not sure if we just need to verify if the endpoint contains `:sessionId` substring to avoid hardcodes

packages/base-driver/lib/protocol/protocol.js
```diff
@@ -19,6 +19,7 @@ import { generateDriverLogPrefix } from '../basedriver/helpers';
 export const CREATE_SESSION_COMMAND = 'createSession';
 export const DELETE_SESSION_COMMAND = 'deleteSession';
 export const GET_STATUS_COMMAND = 'getStatus';
+export const GET_APPIUM_SESSION_CAPABILITIES = 'getAppiumSessionCapabilities';
```
mykola-mokhnach, line 22: COMMAND suffix is missing

packages/base-driver/lib/jsonwp-proxy/proxy.js
eglitise: Agreed, this would be a more future-proof solution.

eglitise:
> I have moved entire deprecated endpoints to the bottom of each section

👍 
At the moment there are still 2 separate sections for deprecated endpoints, and since for many endpoints it's not immediately clear whether they are JSONWP endpoints or custom Appium endpoints, I would still suggest placing all of them in one section at the very end of the list. But regardless, the current order is already much better.

packages/base-driver/lib/jsonwp-proxy/proxy.js
KazuCocoa: Do we want include such a change in this PR? Just thought to add the change with test code in another pr. Just my preference for this PR though.

packages/appium/lib/appium.js
KazuCocoa: oh, nice

KazuCocoa: oh, nice

KazuCocoa:
Sure, moved to the bottom (https://github.com/appium/appium/pull/20936/commits/011a5980ca1d9ac13d9e989854f8b9b312849268). I like each section has valid and deprecated endpoints style, but either is fine

packages/base-driver/lib/protocol/protocol.js
KazuCocoa: https://github.com/appium/appium/pull/20936/commits/12a1cb3123d72333fee7725ded098fc77cb07737

packages/base-driver/lib/jsonwp-proxy/proxy.js
mykola-mokhnach: another PR is OK

packages/appium/lib/appium.js
mykola-mokhnach: I meant the error message itself is confusing , since this method is already defined in the umbrella driver. It makes sense to keep the definition though

KazuCocoa: I see.. this method needs a way to call the same name's one from destination driver like delete session since not this method is part of https://github.com/appium/appium/pull/20936/files#diff-ad0e9402cbbaa56d0519b416feaff4f66815b49e7cef7518c049c12fc0c1cadcR966 ... I didn't catch up with the method call route.
So, this method is not well implemented yet. Probably here needs similar get destination driver instance one like delete session does, whcih probably nice to extract similar method and re-use for both.


I'd like to do this implementation in a followup PR with `endpointRequiresSessionId`'s change (to not grow this pr diff)

eglitise (approved):

mykola-mokhnach (approved):

PR #20947, opened 2025-01-30
chore: use RegExp to check session id requirement

KazuCocoa:
## Proposed changes

A followup PR of https://github.com/appium/appium/pull/20936

We'd like to avoid hardcode of `endpointRequiresSessionId`

## Types of changes

refactoring

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/packages/base-driver/lib/jsonwp-proxy/proxy.js
+++ b/packages/base-driver/lib/jsonwp-proxy/proxy.js
@@ -111,8 +111,8 @@ class JWProxy {
   }
 
   /**
-   * Return true if the given endpoint starts with '/session' and
-   * possibly it requires session id after the path.
+   * Return true if the given endpoint started with '/session' and
+   * it could have session id after the path.
    * e.g.
    * - should return true
    *   - /session/82a9b7da-faaf-4a1d-8ef3-5e4fb5812200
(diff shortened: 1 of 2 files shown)


packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -110,13 +110,24 @@ class JWProxy {
...
+   * @returns {boolean}
+   */
   endpointRequiresSessionId(endpoint) {
-    return !_.includes([
-      '/session',
-      '/sessions',
-      '/status',
-      '/appium/sessions'
-    ], endpoint);
+    return new RegExp(/^\/session\/[^/.]+(|\/.*)$/).test(endpoint);
```
mykola-mokhnach, line 130: I think that using path-to-regexp->parse would be more reliable than regular expressions. After we have parsed tokens we might verify one of these tokens is `:sessionId` parameter. WDYT?

eglitise: Yes, since `base-driver` already has `path-to-regexp` in dependencies, this seems like a better solution.

KazuCocoa: Oh, didn't know of `path-to-regex`.
I also noticed that https://github.com/appium/appium/blob/96a08630e44183cb4970d7acbe1b03faa95d6c93/packages/base-driver/lib/jsonwp-proxy/proxy.js#L147-L170 around these lines should be treated well. I'll modify these lines as well

KazuCocoa: Btw, the incoming url is `/session/82a9b7da-faaf-4a1d-8ef3-5e4fb5812200` format, here doesn't include `:sessionId`

mykola-mokhnach (approved):
nice

PR #20973, opened 2025-02-10
chore(appium): return getAppiumSessionCapabilities from umbrella driver

KazuCocoa:
## Proposed changes

Another followup PR of https://github.com/appium/appium/pull/20936 for https://github.com/appium/appium/issues/20880

This PR does:
- Let `getAppiumSessionCapabilities` call in `appium.js` execute  `getAppiumSessionCapabilities` in the destination's driver
     - In my understanding here is the implementation will depend on umbrella drivers. We'll be able to related the error condition later
- Add `getAppiumSessionCapabilities` in the fake driver
> $ curl http://127.0.0.1:4723/session/8b406c82-1db8-4974-afaf-7f5e950a5b60/appium/capabilities
{"value":{"automationName":"fake","someInfo":"fakedriver specific stuff","sessionId":"8b406c82-1db8-4974-afaf-7f5e950a5b60"}}%

I think that we haven't decided the return value and format yet, so this doesn't specify the type yet.

## Types of changes

## Checklist

- [ ] I have read the [Contributing Guide](https://appium.io/docs/en/latest/contributing/)
- [ ] I have signed the CLA

## Further comments


packages/appium/lib/appium.js
```diff
@@ -201,10 +201,29 @@ class AppiumDriver extends DriverCore {
...
 
-  async getAppiumSessionCapabilities () {
-    throw new errors.NotImplementedError('TODO: Implement this method body to call getAppiumSessionCapabilities ' +
-      'from the destination driver instance in a session.'
-    );
+
+  /**
+   * Return session capabilities info related to the destination driver.
+   * @param sessionId - the id of the session to return capabilities info.
+   * @returns {Promise<{value: AppiumSessionCapabilities, error?: Error, protocol: string|undefined}>}
```
mykola-mokhnach, line 208: protocol?: string

packages/appium/lib/appium.js
```diff
@@ -201,10 +201,29 @@ class AppiumDriver extends DriverCore {
...
+   */
+  async getAppiumSessionCapabilities (sessionId) {
+    const dstSession = this.sessions[sessionId];
```
mykola-mokhnach, line 211: we could use `driverForSession`

packages/appium/lib/appium.js
```diff
@@ -201,10 +201,29 @@ class AppiumDriver extends DriverCore {
...
+    if (!dstSession) {
+      throw new Error(`The session with id '${sessionId}' does not exist`);
+    }
+
+    // to extend the newCommandTimeout for session id related endpoint call.
+    await dstSession.startNewCommandTimeout();
+
+    if (!_.isFunction(dstSession.getAppiumSessionCapabilities)) {
+      throw new errors.UnsupportedOperationError('The driver does not support the endpoint.');
```
mykola-mokhnach, line 220: We could improve this error message: `${dstSession.constructor.name} does not support session capabilities retrieval yet`

packages/fake-driver/lib/driver.js
```diff
@@ -152,6 +152,14 @@ export class FakeDriver extends BaseDriver {
     return await super.deleteSession(sessionId);
   }
 
+  async getAppiumSessionCapabilities(sessionId) {
+    return {
```
mykola-mokhnach, line 156: should we returns this.caps?

packages/types/lib/driver.ts
```diff
@@ -469,6 +474,13 @@ export interface ISessionHandler<
    * @returns A session data object
    */
   getSession(): Promise<SingularSessionData<C, SessionData>>;
+
+  /**
+   * Return session capabilities info.
+   * @param sessionId - the id of the session to return capabilities info.
+   * @returns A session capabilities object
+   */
+  getAppiumSessionCapabilities?(sessionId?: string): Promise<AppiumSessionCapabilities>;
```
mykola-mokhnach, line 483: getAppiumSessionCapabilities?(sessionId?: string) -> getAppiumSessionCapabilities(sessionId?: string)

packages/types/lib/driver.ts
```diff
@@ -189,7 +189,12 @@ export interface MultiSessionData<C extends Constraints = Constraints> {
 }
 
 /**
- * Data returned by {@linkcode ISessionCommands.getSession}.
+ * Data returned by {@linkcode ISessionHandler.getAppiumSessionCapabilities}.
+ */
+export interface AppiumSessionCapabilities {}
```
mykola-mokhnach, line 194: not sure if we need a new type for that. Usually each driver already has a type for its caps

mykola-mokhnach: do we generally know how the output of this API should look like?

KazuCocoa: I think we haven't decided yet, so I haven't specified details yet. This PR just replaced better error and response from the umbrella driver

mykola-mokhnach: lets remove this typedef for now then and return `any` until we have a clear decision

packages/fake-driver/lib/driver.js
KazuCocoa: Returned caps and a few extras as an example for now. No strong opinion but just an example to return an arbitrary values as example response value. (not only this.caps)

packages/appium/lib/appium.js
```diff
@@ -201,10 +201,30 @@ class AppiumDriver extends DriverCore {
...
-    );
+
+  /**
+   * Return session capabilities info related to the destination driver.
+   * https://github.com/appium/appium/issues/20880
+   *
+   * @param sessionId - the id of the session to return capabilities info.
+   * @returns {Promise<{value: object, error?: Error, protocol?: string}>}
+   */
+  async getAppiumSessionCapabilities (sessionId) {
```
mykola-mokhnach, line 212: after looking at this implementation for the second time it does not really make sense to have this method in the umbrella driver since the only purpose of it is to call the corresponding driver-specific implementation

packages/types/lib/driver.ts
```diff
@@ -469,7 +469,17 @@ export interface ISessionHandler<
...
-}
+
+  /**
+   * Return session capabilities info.
+   * FIXME: Please define better type after deciding the response format in
+   * https://github.com/appium/appium/issues/20880
+   * @param sessionId - the id of the session to return capabilities info.
+   * @returns A session capabilities object
+   */
+  getAppiumSessionCapabilities?(sessionId: string): Promise<{
```
mykola-mokhnach, line 480: I still don't understand why we need a question mark after the method name

packages/types/lib/driver.ts
```diff
@@ -469,7 +469,17 @@ export interface ISessionHandler<
...
+    value: object, error?: Error, protocol?: string}>
```
mykola-mokhnach, line 481: I believe this declaration should only return the type of the returned result. This protocol and error stuff is something that is specific for the umbrella implementation, but not for the base driver one

packages/types/lib/driver.ts
KazuCocoa: Without `?`, driver build like fake driver with this type failed until such a driver implements `getAppiumSessionCapabilities`. I thought we won't block driver build with this update, so I added this `?`.

Btw, I was confused but I'm wondering if https://github.com/appium/appium/pull/20979 is a proper one at this time instead of this.

packages/appium/lib/appium.js
KazuCocoa: Yea, i was confused with the same thing. I think right now this endpoint can return an error message to refer to https://github.com/appium/appium/issues/20880 until we'll define concrete return values.

PR #20979, opened 2025-02-11
chore(appium): returns this.caps for /session/:sessionId/appium/capabilities

KazuCocoa:
## Proposed changes

Alternative of https://github.com/appium/appium/pull/20973

I wondered if this was proper one for now. Concrete implementation should be done after deciding response format in https://github.com/appium/appium/issues/20880 .

## Types of changes

## Checklist

- [ ] I have read the [Contributing Guide](https://appium.io/docs/en/latest/contributing/)
- [ ] I have signed the CLA

## Further comments


packages/appium/lib/appium.js
```diff
@@ -202,9 +202,8 @@ class AppiumDriver extends DriverCore {
   }
 
   async getAppiumSessionCapabilities () {
```
mykola-mokhnach, line 204: Maybe we can remove the method from the umbrella driver completely as we are not going to implement it there anyway

KazuCocoa: Then, we should remove GET_APPIUM_SESSION_CAPABILITIES_COMMAND stuff as well, correct?
https://github.com/appium/appium/pull/20979/commits/f327d31a62835e34cd2317d5a1a43d21e6ce3d62

mykola-mokhnach: I assume we must make a decision here whether we want this method to be implemented in each driver or in the umbrella one. I would still look for the umbrella one, since we have all the necessary properties known in the BaseDriver class, unless you need to put something special into the output, which we cannot unify

eglitise:
The original goal for this new endpoint was to keep the response the same as the current [`getSessions`](https://github.com/appium/appium/blob/a1e389068164c37486c67c48f7008154ecc3f01b/packages/base-driver/lib/basedriver/driver.ts#L363) , just remove the event timing information from it. So basically it can be as simple as
```js
  async getAppiumSessionCapabilities () {
    return this.caps as SingularSessionData<C, SessionData>;
  }
```
There is nothing driver-specific here, so it can be easily added in the umbrella driver.

KazuCocoa:
Do we want to drop `this.caps.eventTimings ? {...this.caps, events: this.eventHistory`  behavior?

eglitise:
> Do we want to drop this.caps.eventTimings ? {...this.caps, events: this.eventHistory behavior?

Yes, this is described in #20880. This data should be moved to a different endpoint, e.g. the existing `GET /session/:sessionId/appium/events`.

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -359,8 +359,16 @@ export class BaseDriver<
 
   /**
    * Returns capabilities for the session and event history (if applicable)
+   * @deprecated Use getAppiumSessionCapabilities instead
    */
   async getSession() {
+    return await this.getAppiumSessionCapabilities();
```
eglitise, line 365: `getSession` and `getAppiumSessionCapabilities` are not the same. Both include the capabilities, but only `getSession` has the event timings. It should stay unchanged, only deprecated.

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -359,13 +359,22 @@ export class BaseDriver<
...
+   * Use {@linkcode EventCommands.getLogEvents} instead to get the event history.
    */
   async getSession() {
     return (
       this.caps.eventTimings ? {...this.caps, events: this.eventHistory} : this.caps
     ) as SingularSessionData<C, SessionData>;
   }
 
+  /**
+   * Returns capabilities for the session and event history (if applicable)
```
eglitise, line 372: No need to mention event history here

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -359,13 +359,22 @@ export class BaseDriver<
 
   /**
    * Returns capabilities for the session and event history (if applicable)
+   * @deprecated Use {@linkcode ISessionCommands.getAppiumSessionCapabilities} instead for getting the capabilities.
+   * Use {@linkcode EventCommands.getLogEvents} instead to get the event history.
```
eglitise, line 363: I'm not sure if we want to deprecate `getSession` without first extending `getLogEvents` with the event timing

KazuCocoa: `this.eventHistory` itself was available via https://github.com/appium/appium/blob/e934b61856babab084148102e78060603aab3549/packages/base-driver/lib/basedriver/commands/event.ts#L31 and our clients can call the endpoint already. But hm, we addresses this getSessions to get events in https://github.com/appium/appium/blob/master/packages/appium/docs/en/guides/event-timing.md so probably we need to update the docs with the eventHistory endpoint like `driver.logs.events` (in Ruby) instead


e.g.
```
> d.logs.events
=> {"commands"=>
  [{"cmd"=>"getLogEvents", "startTime"=>1739350220976, "endTime"=>1739350220976},
   {"cmd"=>"getLogEvents", "startTime"=>1739350227281, "endTime"=>1739350227283},
   {"cmd"=>"getLogEvents", "startTime"=>1739350229548, "endTime"=>1739350229548}]}
```

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -359,13 +359,22 @@ export class BaseDriver<
...
   async getSession() {
     return (
       this.caps.eventTimings ? {...this.caps, events: this.eventHistory} : this.caps
     ) as SingularSessionData<C, SessionData>;
   }
 
+  /**
+   * Returns capabilities for the session
+   */
+  async getAppiumSessionCapabilities() {
```
mykola-mokhnach, line 374: should we add the explicit return type to this method?

packages/base-driver/lib/basedriver/driver.ts
eglitise: The docs show that the getSession endpoint also included `"<event_type>": [<occurence_timestamp_1>, ...]` - not sure if the dedicated events endpoint has this information

KazuCocoa: Yes, like below, you can add custom events and get them via `d.logs.events`
```
d.logs.events
=> {"commands"=>[{"cmd"=>"logCustomEvent", "startTime"=>1739353503907, "endTime"=>1739353503907}],
 "appium:customEvent"=>[1739353503907]}
[267] pry(main)> d.logs.event = { vendor: 'appium', event: 'customEvent' }
=> {:vendor=>"appium", :event=>"customEvent"}
> d.logs.events
=> {"commands"=>
  [{"cmd"=>"logCustomEvent", "startTime"=>1739353503907, "endTime"=>1739353503907},
   {"cmd"=>"getLogEvents", "startTime"=>1739353509792, "endTime"=>1739353509793},
   {"cmd"=>"logCustomEvent", "startTime"=>1739353524414, "endTime"=>1739353524416}],
 "appium:customEvent"=>[1739353503907, 1739353524415]}
```

The events endpoint also can get a specific type only such as:
```
d.logs.events "appium:customEvent"
=> {"appium:customEvent"=>[1739353592089, 1739353592935]}
```
which is not available in the getSessions.

eglitise (approved):

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -359,13 +359,22 @@ export class BaseDriver<
...
     return (
       this.caps.eventTimings ? {...this.caps, events: this.eventHistory} : this.caps
     ) as SingularSessionData<C, SessionData>;
   }
 
+  /**
+   * Returns capabilities for the session
+   */
+  async getAppiumSessionCapabilities(): Promise<DriverCaps<C>> {
+    return this.caps;
```
mykola-mokhnach, line 375: just a note for the future: It might be hard to extend the output if necessary. To make that easier we could use the following format: `return {capabilities: this.caps}`

eglitise: Sure, sounds good 👍 would we also want to create a new type for this return value?

KazuCocoa: https://github.com/appium/appium/pull/20979/commits/2f178b8c313c239ac675f41a8ad20758f77d8f1f ?

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -359,13 +360,22 @@ export class BaseDriver<
...
   async getSession() {
     return (
       this.caps.eventTimings ? {...this.caps, events: this.eventHistory} : this.caps
     ) as SingularSessionData<C, SessionData>;
   }
 
+  /**
+   * Returns capabilities for the session
+   */
+  async getAppiumSessionCapabilities(): Promise<AppiumSessionCapabilities> {
```
mykola-mokhnach, line 375: should we provide generic types?

eglitise (approved):

mykola-mokhnach:
Would it be possible to add a basic integration test for the new endpoint?

KazuCocoa:
Added, it simply calls the endpoint and check the response

mykola-mokhnach (approved):

packages/base-driver/test/e2e/basedriver/driver.e2e.spec.js
```diff
@@ -1,13 +1,59 @@
...
+import {TEST_HOST, getTestPort} from '@appium/driver-test-support';
 
 const DEFAULT_CAPS = {
-  platformName: 'iOS',
-  'appium:deviceName': 'Delorean',
+  platformName: 'fake',
+  'appium:automationNAme': 'fake',
 };
 
-driverE2ETestSuite(BaseDriver, DEFAULT_CAPS);
```
eglitise, line 9: Is it okay to drop this...? Though I realise this is also being used for `fake-driver` E2E tests

KazuCocoa: i read the code more, maybe we could leave the suite as-is and this new ones can be an extra. let me run tests on ci

KazuCocoa: I confirmed the latest one keeps the suite as well

Issue #20760, opened 2024-11-18
bug: websocket close unexpectedly


lelemka0:
### Do I have the most recent component updates?

- [X] I use the most recent available driver/plugin and server versions

### Is the component officially supported by the Appium team?

- [X] I have verified the component repository is present under the Appium organization in GitHub

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Websocket connection close unexpectedly after 5 minutes of connection establishment.

### Expected Behavior

WebSocket connection should not be closed unexpectedly.

### Minimal Reproducible Example

```
appium server --use-plugins=devtools
```
Since some problems in the original plugin will cause the first message to be lost and puppeteer will fail to connect, please use this fork: https://github.com/lelemka0/appium-devtools-plugin
```
import { remote } from 'webdriverio';
import Puppeteer from 'puppeteer-core';
import * as URL from 'node:url'
(async () => {
    const driver = await remote(wdOpts);
    const cdpList = await driver.executeScript('devtools: listTargets', []);
    const proxyResult = await driver.executeScript('devtools: proxyTarget', [{ name: cdpList.targets[0].name }]);
    const puppeteer = await Puppeteer.connect({ browserWSEndpoint: URL.resolve(proxyResult.root + '/', 'devtools/browser') });
    const startTime = Date.now();
    const duration = 6 * 60 * 1000;
    const endTime = startTime + duration;
    while (Date.now() <= endTime) {
        console.log(await driver.getDeviceTime()); //Avoid closing the session due to timeout "newCommandTimeout"
        console.log(await puppeteer.version());
        await new Promise(resolve => setTimeout(resolve, 10e3));
    }
})();
```

### Environment

- Operating system: Windows 10
- Appium server version (output of `appium --version`): 2.12.1
- Appium driver(s) and their version(s) (`appium driver list`): uiautomator2@3.8.2
- Appium plugin(s) and their version(s) (`appium plugin list`): devtools
- Node.js version (output of `node --version`): 23.2.0
- `npm` version (output of `npm --version`): 10.9.0

### Link to Appium Logs

https://gist.github.com/lelemka0/73fd72a4c0eef665eabc5f1abfbd7d79

### Further Information

I am using [appium-devtools-plugin](https://github.com/mykola-mokhnach/appium-devtools-plugin) and puppeteer to process WebView automation. This plugin exposes a ws endpoint and forwards the CDP message.
After about five minutes after the connection is established, the connection is always closed.
[...]

PR #21003, opened 2025-02-18
feat(appium): Add a command line parameter to configure HTTP server request timeout

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/20760

Basically, this timeout is also unexpectedly applied to all web sockets we create in the server, including BiDi ones. It makes them to disconnect after 5 minutes timeout (the default Node.js timeout). I have bumped this value to 1 hour until we have a proper solution from Node.js API


KazuCocoa (approved):

eglitise:
Would it be possible to also add this parameter to the CLI docs?
(https://github.com/appium/appium/blob/master/packages/appium/docs/en/cli/args.md)

PR #21011, opened 2025-02-20
chore(schema,tsconfig,universal-xml-plugin): fix repository field in package.json

eglitise:
I noticed that the links to PRs in the `universal-xml-plugin` changelog were leading to a non-existent page. This was due to a misconfiguration in the `repository` field in `package.json`, which is fixed in this PR.
I then checked this field for all other packages and found that `schema` and `tsconfig` were also not aligned, so those are also adjusted here.

### DIFF
--- a/packages/schema/package.json
+++ b/packages/schema/package.json
@@ -19,7 +19,7 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/appium/appium.git",
-    "directory": "schema"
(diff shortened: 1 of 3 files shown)


mykola-mokhnach (approved):

eglitise:
Strange that the PR title job is failing 🤔

KazuCocoa (approved):

PR #21014, opened 2025-02-20
fix(renovate): use chore prefix for all PRs

eglitise:
Currently Renovate PRs seems to sometimes have the `chore` prefix, and sometimes `fix`, which is inconsistent and adds unnecessary commits to the package changelogs. The `fix` prefix is applied due to `:semanticPrefixFixDepsChoreOthers`, which is included in `config:js-app` (see https://github.com/renovatebot/config-help/issues/712).
This PR ensures that Renovate PRs all have the `chore` prefix, as was originally intended according to the README.


KazuCocoa (approved):
makes sense

mykola-mokhnach (approved):

Issue #21015, opened 2025-02-20
UnknownCommandError appium/settings


Phoenix124:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Appium Version

2.12.2

### Appium Host Type

Command Line

### Are you using Appium components not maintained by the Appium team?

no

### Is this issue reproducible using the latest components?

- [x] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

Try to update session settings and got error
uiautomator2@3.9.8
```
[38;5;-94m[AndroidUiautomator2Driver@179b] Proxying [POST /appium/settings] to [POST http://127.0.0.1:8206/appium/settings] with body: {"settings":{"waitForIdleTimeout":100,"allowInvisibleElements":true}}
2025-02-20 09:39:16:941 - [a4371f5f]
[38;5;-94m[AndroidUiautomator2Driver@179b] Got response with status 404: {"sessionId":null,"value":{"error":"unknown command","message":"The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource","stacktrace":"io.appium.uiautomator2.common.exceptions.UnknownCommandException: The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource\n\tat io.appium.uiautomator2.http.ServerHandler.channelRead(ServerHandler.java:84)\n\tat 
.2025-02-20 09:39:16:941 - [a4371f5f]
[W3C]
 Matched W3C error code 'unknown command' to UnknownCommandError
2025-02-20 09:39:16:941 - [AppiumDriver@db94] Event 'newSessionStarted' logged at 
2025-02-20 09:39:16:941 - [AppiumDriver@db94]
 Encountered internal error running command: io.appium.uiautomator2.common.exceptions.UnknownCommandException: The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource
	at io.appium.uiautomator2.http.ServerHandler.channelRead(ServerHandler.java:84)
2025-02-20 09:39:16:942 - 
[a4371f5f]
[HTTP]
```

### Expected Behavior

No errors

### Environment Details

- Debug details (`appium --show-debug-info`):
- Last component version(s) which did _not_ exhibit the problem:
- Platform and version of device under test: android 13
- Real device or emulator/simulator: emulator

### Minimal Reproducible Example

`AppiumDriverRunner.getMobileDriver().setSetting(Setting.ALLOW_INVISIBLE_ELEMENTS, true);`

[...]

PR #21018, opened 2025-02-20
fix(base-driver): Optimize the logic of getUrlForProxy

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/appium/appium/issues/21015

I've also added basic caching to command names detection function, so it saves CPU cycles.

## Types of changes

### DIFF
--- a/packages/base-driver/lib/jsonwp-proxy/proxy.js
+++ b/packages/base-driver/lib/jsonwp-proxy/proxy.js
@@ -21,7 +21,7 @@ import nodeUrl from 'node:url';
 
 const DEFAULT_LOG = logger.getLogger('WD Proxy');
 const DEFAULT_REQUEST_TIMEOUT = 240000;
-const LOOSE_COMMAND_MATCHER = pathToRegexMatch('{/*prefix}/session/:sessionId{/*command}');
+const COMMAND_WITH_SESSION_ID_MATCHER = pathToRegexMatch('/session/:sessionId{/*command}');
 
 const {MJSONWP, W3C} = PROTOCOLS;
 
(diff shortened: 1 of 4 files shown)


packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -140,67 +123,61 @@ class JWProxy {
...
+    );
+    const commandName = normalizedPathname
+      ? routeToCommandName(
+        normalizedPathname,
+        /** @type {import('@appium/types').HTTPMethod | undefined} */ (method)
+      )
+      : '';
+    const requiresSessionId = !commandName || (commandName && isSessionCommand(commandName));
+    const proxyPrefix = `${this.scheme}://${this.server}:${this.port}${this.base}`;
+    let proxySuffix = normalizedPathname ? `/${_.trimStart(normalizedPathname, '/')}` : '';
```
KazuCocoa, line 158: oh, nice. i didn't know of the `trimStart` usage in lodash

packages/base-driver/lib/jsonwp-proxy/proxy.js
```diff
@@ -140,67 +123,61 @@ class JWProxy {
...
+    if (parsedUrl.search) {
+      proxySuffix += parsedUrl.search;
     }
```
KazuCocoa, line 161: do you know if we have this test case? I didn't come up with this query param case, so would be nice to have this pattern as well

KazuCocoa (approved):
lgtm overall (for my understanding as well)

packages/base-driver/lib/jsonwp-proxy/proxy.js
mykola-mokhnach: added a test

KazuCocoa (approved):

Issue #21043, opened 2025-02-25
🐞 bug: <Unable to find detect Web view in Hybrid app in Real iPhone running iOS 18+>


sockerzero:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Appium Version

2.16.2

### Appium Host Type

Command Line

### Is this issue reproducible using the latest components?

- [x] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

Hi team,

I have an app with webView.isInspectable = true. When testing on an iPhone 14 running iOS 17.1, the web view context is detected successfully (with Appium inspector and via code). However, on an iPhone 13 (iOS 18.3) and an iPhone 16 (iOS 18.1), the web view isn’t detected in Appium Inspector, or calling driver.getContexts() via code only returns a single context (NATIVE).

I was able to use Safari on macOS to open the Web Inspector tool and inspect elements within the app

I confirmed I turned ON the same settings in all 3 devices:
- Settings -> Develop -> "Enable UI Automation'
- Settings -> App -> Safari -> Web Inspector = ON and Remote Automation = ON

Attached are a screenshot from an iOS 17 device and the corresponding Appium logs when attempting to switch contexts via Appium Inspector.

<img width="895" alt="Image" src="https://github.com/user-attachments/assets/57ed70f8-ef2b-4a6c-b8a9-9382b01b4b54" />

[iP14PM-17.1.txt](https://github.com/user-attachments/files/18961169/iP14PM-17.1.txt)

And below is a screenshot from an iOS 18 device along with the corresponding Appium logs captured when attempting to switch contexts using Appium Inspector.

<img width="746" alt="Image" src="https://github.com/user-attachments/assets/2d243d42-1103-4fb0-93e8-0827679202c2" />

[IP16PM-18.1.txt](https://github.com/user-attachments/files/18961173/IP16PM-18.1.txt)

### Expected Behavior

I expect both Appium Inspector and my code to detect the web view context so that I can switch to it.

### Environment Details

Appium version: 2.16.2
uiautomator2: 3.9.7
xcuitest: 8.3.2
webdriverio: 9.5.1
node: v20.18.3

### Minimal Reproducible Example

Using Appium Inspector 2024.12.1 or via code (driver.getContexts())

mykola-mokhnach:
Sometimes it is necessary to interact with a web view first before the remote debugger can pick it up. Try to click it and then fetch contexts.

mykola-mokhnach:
[...]

PR #21044, opened 2025-02-25
fix(base-driver): Tune capabilities array parsing

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/21043#issuecomment-2681751065

### DIFF
--- a/packages/base-driver/lib/basedriver/helpers.js
+++ b/packages/base-driver/lib/basedriver/helpers.js
@@ -376,26 +376,29 @@ export function duplicateKeys(input, firstKey, secondKey) {
  * Takes a desired capability and tries to JSON.parse it as an array,
  * and either returns the parsed array or a singleton array.
  *
- * @param {string|Array<String>} cap A desired capability
+ * @param {string[]} cap A desired capability
  */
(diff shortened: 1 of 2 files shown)


KazuCocoa (approved):

packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -376,26 +376,29 @@ export function duplicateKeys(input, firstKey, secondKey) {
  * Takes a desired capability and tries to JSON.parse it as an array,
  * and either returns the parsed array or a singleton array.
  *
- * @param {string|Array<String>} cap A desired capability
+ * @param {string[]} cap A desired capability
```
eglitise, line 379: Shouldn't this be `string|string[]`? If we always expect it to be an array, it will always return on line 383.

mykola-mokhnach: nope, this method should always return an array of strings

mykola-mokhnach: err, I had to change the return type, but changed the param type instead %)

PR #21055, opened 2025-03-01
feat(images-plugin): supports image elements included in actions.

kkb912002:
## Proposed changes

It allows using image elements in the PerformActions command.

## Types of changes

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/packages/images-plugin/lib/plugin.js
+++ b/packages/images-plugin/lib/plugin.js
@@ -89,7 +89,13 @@ export default class ImageElementPlugin extends BasePlugin {
     for (const actionSequence of actionSequences) {
       for (const action of actionSequence.actions) {
         // The actions that can have an Element as the origin are "pointerMove" and "scroll".
-        if (!_.has(action, 'origin') || !_.isPlainObject(action.origin)) {
+        if (
+          !_.isPlainObject(
+            /** @type {{origin?: "viewport" | "pointer" | import('@appium/types').Element}} */ (
+              action
+            ).origin,
+          )
+        ) {
           continue;
         }
 
--- a/packages/images-plugin/lib/plugin.js
+++ b/packages/images-plugin/lib/plugin.js
@@ -1,7 +1,6 @@
-
-
 import _ from 'lodash';
 import {errors} from 'appium/driver';
+import {util} from '@appium/support';
 import {BasePlugin} from 'appium/plugin';
 import {compareImages} from './compare';
 import ImageElementFinder from './finder';
@@ -81,6 +80,36 @@ export default class ImageElementPlugin extends BasePlugin {
     }
 
     // otherwise just do the normal thing
+    return await next();
+  }
+
+  async performActions(next, driver, ...args) {
+    // Replace with coordinates when ActionSequence includes image elements.
+    const [actionSequences] = /** @type {[import('@appium/types').ActionSequence[]]} */ (args);
+    for (const actionSequence of actionSequences) {
+      for (const action of actionSequence.actions) {
+        // The actions that can have an Element as the origin are "pointerMove" and "scroll".
+        if (
+          (action.type === 'pointerMove' || action.type === 'scroll') &&
+          !_.isNil(action.origin) &&
+          !_.isString(action.origin)
+        ) {
+          const elId = util.unwrapElement(action.origin);
+          if (_.isString(elId) && elId.startsWith(IMAGE_ELEMENT_PREFIX)) {
+            const imgEl = this.finder.getImageElement(elId);
+            if (!imgEl) {
+              throw new errors.NoSuchElementError();
+            }
+            // Add the element's center coordinates to the offset value.
+            action.x += imgEl.center.x;
+            action.y += imgEl.center.y;
+            // Set the origin to the viewport so that the external driver can process it using coordinates.
+            delete action.origin;
+          }
+        }
+      }
+    }
+
     return await next();
   }
 }
--- a/packages/images-plugin/lib/plugin.js
+++ b/packages/images-plugin/lib/plugin.js
@@ -89,24 +89,27 @@ export default class ImageElementPlugin extends BasePlugin {
     for (const actionSequence of actionSequences) {
       for (const action of actionSequence.actions) {
         // The actions that can have an Element as the origin are "pointerMove" and "scroll".
-        if (
-          (action.type === 'pointerMove' || action.type === 'scroll') &&
-          !_.isNil(action.origin) &&
-          !_.isString(action.origin)
-        ) {
-          const elId = util.unwrapElement(action.origin);
-          if (_.isString(elId) && elId.startsWith(IMAGE_ELEMENT_PREFIX)) {
-            const imgEl = this.finder.getImageElement(elId);
-            if (!imgEl) {
-              throw new errors.NoSuchElementError();
-            }
-            // Add the element's center coordinates to the offset value.
-            action.x += imgEl.center.x;
-            action.y += imgEl.center.y;
-            // Set the origin to the viewport so that the external driver can process it using coordinates.
-            delete action.origin;
-          }
+        if (!_.has(action, 'origin') || !_.isPlainObject(action.origin)) {
+          continue;
         }
+
+        const actionWithEl = /** @type {import('@appium/types').PointerMoveAction | import('@appium/types').ScrollAction} */ (action);
+
+        const elId = util.unwrapElement(/** @type {import('@appium/types').Element} */ (actionWithEl.origin));
+        if (!_.startsWith(elId, IMAGE_ELEMENT_PREFIX)) {
+          continue;
+        }
+
+        const imgEl = this.finder.getImageElement(elId);
+        if (!imgEl) {
+          throw new errors.NoSuchElementError();
+        }
+
+        // Add the element's center coordinates to the offset value.
+        actionWithEl.x += imgEl.center.x;
+        actionWithEl.y += imgEl.center.y;
+        // Set the origin to the viewport so that the external driver can process it using coordinates.
+        delete actionWithEl.origin;
       }
     }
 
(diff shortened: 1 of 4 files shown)


packages/images-plugin/test/unit/plugin.spec.js
```diff
@@ -133,6 +133,89 @@ describe('ImageElementPlugin#handle', function () {
         },
       ]);
     });
+    it('should replace with coords of the image elements in PerformActions parameters', async function () {
```
eglitise, line 136: This test can be split into two (one to confirm that pointerMove+scroll are modified, and one to confirm that none of the other types are modified), and placed under a new `describe('performActions')` group

packages/images-plugin/test/unit/plugin.spec.js
```diff
@@ -133,6 +133,89 @@ describe('ImageElementPlugin#handle', function () {
...
+      let called = false;
+      const next = () => {
```
eglitise, line 138: Not sure if this needs to be redeclared if the goal with these tests is to check if the action sequence is modified

eglitise:
Please also add the `performActions` command to the commands reference document: https://github.com/appium/appium/blob/master/packages/appium/docs/en/commands/images-plugin.md

packages/images-plugin/test/unit/plugin.spec.js
kkb912002: Fixed as suggested. Thank you.

packages/images-plugin/test/unit/plugin.spec.js
kkb912002: I removed it.

packages/images-plugin/test/e2e/plugin.e2e.spec.js
```diff
@@ -89,6 +89,20 @@ describe('ImageElementPlugin', function () {
     width.should.eql(80);
     height.should.eql(91);
     await imageEl.click();
+
+    const origin = util.wrapElement(imageEl.elementId);
```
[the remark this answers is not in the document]
kkb912002: I made the changes, but do they align with what you intended?

packages/images-plugin/test/unit/plugin.spec.js
```diff
@@ -177,4 +178,97 @@ describe('ImageElementPlugin#handle', function () {
...
+    });
+    it('should replace with coords of the image elements in pointerMove, scroll actions', async function () {
+      const actionSequences = [
+        {
+          type: 'pointer',
+          id: 'mouse',
+          parameters: {pointerType: 'touch'},
+          actions: [
+            {type: 'pointerMove', x: 0, y: 0, duration: 0, origin: imageEl},
+            {type: 'pause', duration: 0},
```
eglitise, line 197: Can we remove the `pause` actions from this line and line 205, to ensure separation from the other unit test?

packages/images-plugin/test/unit/plugin.spec.js
```diff
@@ -177,4 +178,97 @@ describe('ImageElementPlugin#handle', function () {
...
+            {type: 'pointerDown', button: 0},
+            {type: 'pause', duration: 125},
+            {type: 'pointerUp', button: 0},
+          ],
+        },
+        {
+          type: 'wheel',
+          id: 'wheel',
+          actions: [
+            ..._.range(7).map(() => ({type: 'pause', duration: 0})),
```
eglitise, line 252: Do we need 7 instances of pause here...? Same for line 263

packages/images-plugin/test/unit/plugin.spec.js
kkb912002: The pauses are there to create a natural action sequence where each type of action does not overlap, but they don’t seem strictly necessary for this test. Should we remove all non-essential null actions (pause 0)?

packages/images-plugin/test/e2e/plugin.e2e.spec.js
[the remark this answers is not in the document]
kkb912002: Yes, I've applied it!

packages/images-plugin/test/unit/plugin.spec.js
eglitise: Yes, since these are unit tests, they don't need to be realistic user scenarios

kkb912002: I've removed them.

eglitise (approved):

packages/images-plugin/lib/plugin.js
```diff
@@ -83,6 +82,36 @@ export default class ImageElementPlugin extends BasePlugin {
...
+    for (const actionSequence of actionSequences) {
+      for (const action of actionSequence.actions) {
+        // The actions that can have an Element as the origin are "pointerMove" and "scroll".
+        if (
+          (action.type === 'pointerMove' || action.type === 'scroll') &&
+          !_.isNil(action.origin) &&
+          !_.isString(action.origin)
+        ) {
+          const elId = util.unwrapElement(action.origin);
+          if (_.isString(elId) && elId.startsWith(IMAGE_ELEMENT_PREFIX)) {
```
mykola-mokhnach, line 98: this might be reduced to ```
if (!_.startsWith(elId, IMAGE_ELEMENT_PREFIX)) { 
  continue; 
}
...
```

Comment (mykola-mokhnach) [reply packages/images-plugin/lib/plugin.js:98]: same comments could be applied to the above condition. In general, it is easier to scan through guards rather than log nested blocks, for example:

```
for (const item in items) {
   if (!condition(item)) {
      continue;
   }
   // do something with item
}
```

over

```
for (const item in items) {
   if (condition(item)) {
      // do something with item
   }
}
```

File: packages/images-plugin/lib/plugin.js [hunk 7]
```diff
@@ -83,6 +82,36 @@ export default class ImageElementPlugin extends BasePlugin {
...
+
+  async performActions(next, driver, ...args) {
+    // Replace with coordinates when ActionSequence includes image elements.
+    const [actionSequences] = /** @type {[import('@appium/types').ActionSequence[]]} */ (args);
+    for (const actionSequence of actionSequences) {
+      for (const action of actionSequence.actions) {
+        // The actions that can have an Element as the origin are "pointerMove" and "scroll".
+        if (
+          (action.type === 'pointerMove' || action.type === 'scroll') &&
+          !_.isNil(action.origin) &&
```
mykola-mokhnach, line 94: perhaps it would make more sense to only verify `_.isPlainObject(action.origin)` instead of checking for nils and strings

packages/images-plugin/lib/plugin.js
```diff
@@ -83,6 +82,36 @@ export default class ImageElementPlugin extends BasePlugin {
...
   }
+
+  async performActions(next, driver, ...args) {
+    // Replace with coordinates when ActionSequence includes image elements.
+    const [actionSequences] = /** @type {[import('@appium/types').ActionSequence[]]} */ (args);
+    for (const actionSequence of actionSequences) {
+      for (const action of actionSequence.actions) {
+        // The actions that can have an Element as the origin are "pointerMove" and "scroll".
+        if (
+          (action.type === 'pointerMove' || action.type === 'scroll') &&
```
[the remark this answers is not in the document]
kkb912002: I also think that's good code. However, the type guard isn't working correctly. 😞
Do you know any workarounds?
image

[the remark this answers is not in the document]
mykola-mokhnach: sure, you can enforce the typecast if typescript is unable to autodetect it, for example:

```
const elId = util.unwrapElement(/** @type {import('@appium/types').Element} */ (action.origin));
```

packages/images-plugin/lib/plugin.js
```diff
@@ -83,6 +82,42 @@ export default class ImageElementPlugin extends BasePlugin {
...
     return await next();
   }
+
+  async performActions(next, driver, ...args) {
+    // Replace with coordinates when ActionSequence includes image elements.
+    const [actionSequences] = /** @type {[import('@appium/types').ActionSequence[]]} */ (args);
+    for (const actionSequence of actionSequences) {
+      for (const action of actionSequence.actions) {
+        // The actions that can have an Element as the origin are "pointerMove" and "scroll".
+        if (!_.has(action, 'origin') || !_.isPlainObject(action.origin)) {
```
mykola-mokhnach, line 92: the _.has check is redundant as the second one also returns false for undefined values

mykola-mokhnach (approved):

packages/images-plugin/lib/plugin.js
```diff
@@ -83,6 +82,42 @@ export default class ImageElementPlugin extends BasePlugin {
     // otherwise just do the normal thing
     return await next();
   }
+
+  async performActions(next, driver, ...args) {
+    // Replace with coordinates when ActionSequence includes image elements.
+    const [actionSequences] = /** @type {[import('@appium/types').ActionSequence[]]} */ (args);
```
mykola-mokhnach, line 88: just a small comment to clear out this typescript behaviour.
there is a subtle difference between these two syntaxes:

```js
/** @type {number} */
const x = y;
```

and 

```js
const x = /** @type {number} */ (y);
```

the first one assumes the type of `x` is number and tells the TS compiler to verify if `y` matches this type or raise an error if it does not.

The second one tells TS that `y` must be a number. If `x` was not given a type previously or it is any/unknown then the expression will be compiled successful.

Conclusion: it is always preferable to use the first syntax where possible as it helps to avoid more potential issues. The second instrument is powerful, but sometimes might be breaking if wrong assumptions about types are made, so use with care.

packages/images-plugin/lib/plugin.js
kkb912002: I've removed it. The type guard is a bit troublesome. 😞

PR #21070, opened 2025-03-05
chore: create beta versions of breaking change appium 3 packages

jlipps:


mykola-mokhnach:
It looks like something is broken in the CI now

KazuCocoa:
shall we bump the min node/npm versions as well as https://github.com/appium/appium/pull/20790 ?

jlipps:
The remaining test errors are due to the fact that they rely on published versions of appium-fake-driver and appium-uiautomator2-driver. These published versions do not have appium@^3.0.0-beta.0 in their peerDep requirements, so `appium driver install --source=npm @appium/fake-driver` will fail. Possible solutions:

1. Merge this anyway with broken tests
2. Publish new versions of these packages with a more relaxed version spec that includes appium 3
3. Quarantine the tests and merge this, then publish new beta versions, and make those new beta versions the target of the test (not sure what to do about appium-uiautomator2-driver though since we are not planning on making that appium 3 compatible yet)

eglitise:
I think `fake-driver` already has Appium 3 in its peerDeps (on the appium3 branch), maybe it's possible to publish that one in beta first, would that fix it?

Also I'm still not sure why there's a huge amount of changes in `package-lock.json`, is it still on lockfile version 2? I can't imagine any other reason than lockfile v3.

jlipps:
> is it still on lockfile version 2

No, it's moved to version 3

> maybe it's possible to publish that one in beta first, would that fix it?

This PR exists in order to publish that one in beta 😆

KazuCocoa (approved):
lg.
Remaining failure is fake-driver accepts `"appium": "^2.4.1 || ^3.0.0-beta.0"` in appium3 branch, but not in current regular version.
I think as same as ui2, we can add that to master in fake driver

KazuCocoa:
oh, https://github.com/appium/appium/pull/21070#issuecomment-2704988194 already addressed https://github.com/appium/appium/pull/21076. uia2 already has that, so fake-driver also can have the same thing

KazuCocoa:
I have merged https://github.com/appium/appium/pull/21076 to master, once we release the current master once, fake driver related test failures will be fixed, I assume

mykola-mokhnach:
I observe the following failed tests:

>      1) should show updates for installed drivers with --updates
>      2) should install a driver from npm
>      3) should install a driver from npm and a local driver
>      4) should install _two_ drivers from npm

maybe it would make sense to explicitly skip them with a TODO or FIXME comment to keep the CI always green. Having false positives there might create possible incidents because we'd pay less attention to cI results if it is constantly red.

jlipps:
OK, I will downgrade to npm v8 to keep lockfile version at 2

jlipps:
> maybe it would make sense to explicitly skip them with a TODO or FIXME

sure, I added some skips until we can try again after publish of drivers.

jlipps:
weird, looks like CI is hanging on or after nx reset. Not sure what's going on there.

packages/appium/test/e2e/cli-driver.e2e.spec.js
```diff
@@ -111,41 +111,43 @@ describe('Driver CLI', function () {
...
-      const updateVersion = String(fake.updateVersion || fake.unsafeUpdateVersion);
-      util.compareVersions(
-        String(updateVersion),
-        '>',
-        penultimateFakeDriverVersionAsOfRightNow
-      ).should.be.true;
-      // TODO: this could probably be replaced by looking at updateVersion in the JSON
-      const {stderr} = await runAppiumRaw(appiumHome, [DRIVER_TYPE, LIST, '--updates'], {});
-      stderr.should.match(new RegExp(`fake.+[${updateVersion} available]`));
+      // FIXME turn back on when fake-driver supports appium 3 beta
```
mykola-mokhnach, line 114: would it help if we just hack it as
```
if (process.env.CI) {
  return this.skip();
}
```

so the rest of the test shouldn't be commented out?

jlipps: but the tests will also fail locally...

mykola-mokhnach: I assume it is ok if we add a proper explanation comment %)

jlipps: ok, I published an appium server version and a new fake driver, and unquarantined the tests so all the tests but one pass (that test downgrades fake driver and then updates it, which doesn't work because we would need multiple published versions of fake driver that support appium 3 🙄 ). so that one test is quarantined and we'll need to update after the next server version publish i guess.

mykola-mokhnach (approved):

Issue #13156, opened 2019-08-29
Support 'SetFileDetector' for Mobile devices


ckzgraphics:
## The problem

There is **no support for uploading a file** from the local machine to a website opened on a **remotely attached real device** using '[SetFileDetector](https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/remote/RemoteWebDriver.html#setFileDetector-org.openqa.selenium.remote.FileDetector-)'

## Scenario

- The test case is to verify if a file is being uploaded on a website using a web browser on a mobile device

## Setup

- Suppose I have two machines- Machine A and Machine B
- On Machine A I have my test script and a file (_.png, etc_)
- On Machine B I have set up an Appium server with a real device

## Expected Behaviour

- The user should be able to use '_SetFileDetector_' and upload the file on a remote device's web browser

## Actual Behaviour

- Currently, it is necessary to have the file on the real device

ckzgraphics:
Hello @jlipps , @mykola-mokhnach  - This is a valid use case. Could you please share your thoughts on this?

mykola-mokhnach:
Currently there is a possibility to push a file to a remote device. Although I'm not sure the browser  debug protocol itself supports anything like this in both mobile chrome and mobile safari

ckzgraphics:
I believe you are referring to the following command (_just to be on the same page_) -

`driver.pushFile("/path/to/device/foo.bar", new File("/Users/johndoe/files/foo.bar"));`

Document- [Appium Push File Command](http://appium.io/docs/en/commands/device/files/push-file/)

vikramvr20:
@mykola-mokhnach @KazuCocoa  - I am trying to automate file upload in ios mobile web. Since it's IOS, the test runs on the safari browser and the safari driver does not support file upload using sendKeys. I am trying to use the push command in Appium to push files into "Files" App in Ios devices. 

But how can we use driver.push file in IOS devices. Currently am using it like below and I am not able to see the document getting uploaded in the Files App. 
    
File classpathRoot = new File(System.getProperty("user.dir"));
File assetDir = new File(classpathRoot, "/resources");
File txt = new File(assetDir.getCanonicalPath(), "output.txt");
driver.pushFile("@com.apple.DocumentsApp:documents", txt);

If the push command does not work for the Files App. What other options do I have?

mykola-mokhnach:
I don't think such feature would be possible with neither iOS or Android browser.

[...]

Issue #12390, opened 2019-03-27
[Feature Request] install_app able to install apk/ipa from file local to the client


Jiurong-Yang:
The install_app method does not work unless the apk/ipa file is located on the server side machine (Using file path). It would be nice if I can use this method with the file on the client side and appium transfers the api/ipa to the server node and create a temp file, install it then delete the temp file. I believe the push_file method is doing something like this as I am able to push a file that is on the client machine to the phone that is connected to the server machine. 

Right now I'm working around this by uploading the apk/ipa to a database and using the attachment url as the path in install_app which works great. But for a simpler setup it would be nice not to have to have an additional service like a db as part of the solution.

mykola-mokhnach:
I think it would be a good idea to write a server plugin, which implements a storage concept.
This storage might be controlled over BiDi commands. The following commands should be available:
- Upload file
- List files
- Delete file
- Get file info (?)

The upload one will be the most complicated one as we don't won't to send the whole file in a single chunk (as it might be too big and won't fit into the process memory). That is why the upload API must accept limited in size chunks along with the file hashsum (sha1?), so the server could verify the file integrity after all chunks are sent to it. 
For example:

```js
await driver.send({method: 'appium:xcuitest.uploadToStorage', params: {
  name: 'yolo.ipa',
  hash: '1234567',
  size: 1234567,
  chunk: 'base64encodedChunkBelow64KB',
  index: 0,
}});
....
// continue calling the above API until all chunks are sent
```

Upon server exit the temporary storage should be deleted.

jlipps:
I think this is a good idea. It's complicated enough to use however that it might be nice for it to ship with some client side plugins as well, at least for Python and WebdriverIO

mykola-mokhnach:
Storage plugin: https://github.com/appium/appium/tree/master/packages/storage-plugin
Java client: https://github.com/appium/java-client/pull/2275

PR #21075, opened 2025-03-06
feat: Add storage plugin

mykola-mokhnach:
## Proposed changes

Related to https://github.com/appium/appium/issues/13156 and https://github.com/appium/appium/issues/12390

## Types of changes


packages/storage-plugin/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "@appium/base-plugin": "^2.3.3",
+    "@appium/support": "^6.0.6",
+    "async-lock": "1.4.1",
+    "bluebird": "3.7.2",
+    "lodash": "4.17.21",
+    "rimraf": "5.0.10",
+    "source-map-support": "0.5.21"
+  },
+  "peerDependencies": {
+    "appium": "^2.16.0"
```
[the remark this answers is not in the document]
mykola-mokhnach: I don't think so. Older server versions don't support bidi extensions

saikrishna321:
@mykola-mokhnach Can throw some light on when this plugin will come to use ?

KazuCocoa (commented):
lgtm, overall. so, this plugin's primay usage is when appium server works remotely from the client and over bidi?

mykola-mokhnach:
> @mykola-mokhnach Can you please throw some light on when this plugin will come to use ?

Kindly check the PR changes. There is a README file plus the ticket description contains links to related user issues

mykola-mokhnach:
> lgtm, overall. so, this plugin's primay usage is when appium server works remotely from the client and over bidi?

yes, that's correct. Before that one must either upload files over some other protocols to the server where Appium is running or host them elsewhere and make available through HTTP(S)

packages/base-driver/package.json
```diff
@@ -62,6 +62,7 @@
     "path-to-regexp": "8.2.0",
     "serve-favicon": "2.5.0",
     "source-map-support": "0.5.21",
+    "teen_process": "2.3.1",
```
eglitise, line 65: I don't think this is needed, as `base-driver` only imports `teen_process` in tests. Instead it can be added to the monorepo-level dev dependencies, similarly to how `asyncbox` is used.

mykola-mokhnach: removed

packages/storage-plugin/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+```bash
+curl -X POST --header "Content-Type: application/json" --data '{"name":"app.ipa","sha1":"ccc963411b2621335657963322890305ebe96186"}' http://127.0.0.1:4723/storage/add
+```
+
+The purpose of this command is to upload files to the remote storage partially
+to avoid excessive memory usage. In order to upload a file from a local file system
+to the Appium server's file system it is necessary to perform following steps:
+
+- Calculate [SHA1](https://en.wikipedia.org/wiki/SHA-1#:~:text=In%20cryptography%2C%20SHA%2D1%20(,rendered%20as%2040%20hexadecimal%20digits.) hash of the source file
```
KazuCocoa, line 60:
```suggestion
- Calculate <a href="https://en.wikipedia.org/wiki/SHA-1#:~:text=In%20cryptography%2C%20SHA%2D1%20(,rendered%20as%2040%20hexadecimal%20digits.">SHA1</a> hash of the source file
```

Or I think this link simply can be `https://en.wikipedia.org/wiki/SHA-1`?

packages/storage-plugin/lib/storage.ts
```diff
@@ -0,0 +1,320 @@
...
+      `'${name}' has been added to the server storage within ` +
+      `${timer.getDuration().asMilliSeconds}ms. Verifying hashes.`
+    );
+    if (_.toLower(actualHashDigest) !== _.toLower(sha1)) {
+      throw new StorageArgumentError(
+        `The actual SHA1 hash value '${actualHashDigest}' must be equal ` +
+        `to the expected hash value of '${sha1}' for '${name}'`
+      );
+    }
+    await fs.mv(fullPath, path.join(this._root, name));
```
KazuCocoa, line 285: Would it be nice to address that if the same name file already exists, the new one will override the existing one. If the already existing file is dir, an exception would be raised, so this finalizing will fail in this README?

(based on fs.rename in the `mv` behavior)

packages/storage-plugin/lib/plugin.ts
```diff
@@ -0,0 +1,184 @@
...
+const STORAGE_ADDITIONS_CACHE: LRUCache<string, () => any> = new LRUCache({
+  max: 50,
+  ttl: WS_TTL_MS,
+  dispose: (f: () => any) => f(),
+});
+
+export class StoragePlugin extends BasePlugin {
+
+  static async updateServer(expressApp: Express, httpServer: AppiumServer): Promise<void> {
+    const buildHanlder = (methodName: string) => async (req: Request, res: Response) => {
```
jlipps, line 28: `buildHanlder` => `buildHandler`

jlipps (approved):

packages/storage-plugin/lib/storage.ts
mykola-mokhnach: added a comment

packages/storage-plugin/lib/plugin.ts
mykola-mokhnach: fixed

packages/storage-plugin/README.md
mykola-mokhnach: changed

KazuCocoa (approved):
lgtm

PR #21110, opened 2025-03-14
feat(base-driver): Make extension name prefix mandatory

mykola-mokhnach:


packages/base-driver/package.json
```diff
@@ -52,7 +52,7 @@
     "axios": "1.8.3",
     "bluebird": "3.7.2",
     "body-parser": "1.20.3",
-    "express": "4.21.2",
+    "express": "5.0.1",
```
eglitise, line 55: I think we can bump this in a separate PR since it may affect many tests

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -253,14 +253,12 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
       const separatorPos = fullName.indexOf(FEATURE_NAME_SEPARATOR);
-      if (separatorPos < 0) {
-        // TODO: This is for the backward compatibility with Appium 2.
-        // TODO: We should bring back to raise an Error below for Appium 3.
-        // throw new Error(
-        //   `The full feature name must include both the driver name/wildcard and the feature ` +
-        //   `name split by a colon, got '${fullName}' instead`
-        // );
-        return [ALL_DRIVERS_MATCH, fullName];
+      if (separatorPos <= 0) {
```
eglitise, line 256: `appium/lib/utils.js` uses `separatorPos < 0` - should both be aligned?

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -253,14 +253,12 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
...
+        throw new Error(
+          `The full feature name must be prefixed by the driver name` +
+          `or a wildcard character ('*') and the feature ` +
```
eglitise, line 259: This message can also be aligned with the one used in `appium/lib/utils.js`, and use `ALL_DRIVERS_MATCH`. Personally I think this one is more easily understandable, so I would change the one in the `appium` package.

packages/base-driver/package.json
mykola-mokhnach: this was actually a regression after some merge. I've upgraded express as the first thing for appium3

packages/base-driver/lib/basedriver/core.ts
mykola-mokhnach: no, because we verify if automationName is not empty in the other function.
Actually, I don't expect this condition is ever going to be satisfied because we do parse arguments. So it's rather just a precaution/defensive programming

packages/base-driver/lib/basedriver/core.ts
mykola-mokhnach: unified error messages

KazuCocoa (approved):

packages/base-driver/package.json
eglitise: It doesn't seem like the branch has used Express 5 at any point, although the tests do pass, so I guess it should be safe

eglitise (approved):

PR #21117, opened 2025-03-16
docs: update zh contributing documentation

zhangxh075:


eglitise (approved):

mykola-mokhnach:
The translation should be done in crowdin. By merging such PRs we are making further differences between the original and translated content

eglitise:
Ah, I genuinely forgot about the Crowdin integration.
@zhangxh075 would you be able to add these translations via Crowdin? I believe it should be here: https://crowdin.com/project/appium-documentation/zh-CN

mykola-mokhnach:
> @eglitise Sure, but I need some time to understand Crowdin. Also, I have already translated a lot of content before. Can those contents be migrated to Crowdin?

I believe crowdin has the import feature, but it might be complicated to do it in bulk mode because of the nature of .md documents (e.g. they are not strictly formatted). This is the main reason we haven't done this sync on our own, - it must be performed by qualified translator, who is able to notice accident synchronisation issues between the original and translated text, and correct them.

As for using Crosdin itself, - they have pretty decent documentation and youtube tutorials. Check, for example, https://support.crowdin.com/for-translators/

PR #21118, opened 2025-03-16
feat(base-driver): Use `markSensitive` to mask password input field value

dprevost-LMI:
## Proposed changes

appium-log

As shown in the image above, when entering a value in a field like a password, appium leaks the value at the debug and info level. Since the above can also happen in a cloud provider like BrowserStack, leaking it even at the debug log level can compromise an actual production user account.

Since this approach would be the simplest and quickest, I would like to explicitly exclude it before going into full implementation of the dynamic regular expression to filter the logs as specified [here](https://github.com/appium/appium/issues/20844)

By using `markSensitive` to leverage the header `X-Appium-Is-Sensitive` from the clever solution of PR https://github.com/appium/appium/pull/21123. We can now mask the password value and safely enter the passwords in input fields without worrying about leaks in logs (when no plugins are used 😉 ).

Pros: 
- Built-in with Appium
- Easy to use by users: need to pass a simple header
- No need to keep sensitive data (or part of it as a RegExp) in memory, which is a security OWASP point. [See here](https://owasp.org/www-project-secure-coding-practices-quick-reference-guide/stable-en/02-checklist/05-checklist)

Cons:
- A new line of code could expose the sensitive data without using the new `markSensitive` function.
- A third-party plugin could not check this header and leak the password values in the logs.

Remaining tasks:
- [ ] Add unit tests
- [ ] Test in my enterprise app
- [ ] Reveiw wdio PR to use the new header
- [ ] Finish the below template
- [ ] Set PR to ready

These are draft changes, and if the idea is accepted, I will implement all required unit tests

## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I have read the [Contributing Guide](https://appium.io/docs/en/latest/contributing/)
- [ ] I have signed the CLA
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)
[...]

PR #21123, opened 2025-03-17
feat: Add a possibility to mask sensitive log values depending on request headers

mykola-mokhnach:
## Proposed changes

Related to the conversation initiated by @dprevost-LMI in https://github.com/appium/appium/pull/21118.

The idea is that if the server gets the request header `X-Appium-Is-Sensitive` set to `true` then the logger automatically replaces all values being marked as sensitive (via `support.logging.markSensitive` API) with the default mask text. 

## Types of changes

### DIFF
--- a/packages/base-driver/lib/express/middleware.js
+++ b/packages/base-driver/lib/express/middleware.js
@@ -82,8 +82,15 @@ export function handleLogContext(req, res, next) {
 
   const sessionId = SESSION_ID_PATTERN.exec(req.url)?.[1];
   const sessionInfo = sessionId ? {sessionId, sessionSignature: calcSignature(sessionId)} : {};
+  const doesContainSensitiveData = ['true', '1', 'yes'].includes(
+    _.toLower(_.first(req.headers['x-appium-is-sensitive']))
+  );
 
-  log.updateAsyncContext({requestId, ...sessionInfo}, true);
+  log.updateAsyncContext({
+    requestId,
+    ...sessionInfo,
+    isSensitive: doesContainSensitiveData,
+  }, true);
 
   return next();
 }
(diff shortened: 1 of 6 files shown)


dprevost-LMI:
Thanks. I would have never been able to come up with a solution so simple, but that I don't fully understand yet, either 😅.
In the next few days, I'll take some time to plug it into wdio and test it out in my enterprise app!

mykola-mokhnach:
> Thanks. I would have never been able to come up with a solution so simple, but that I don't fully understand yet, either 😅. In the next few days, I'll take some time to plug it into wdio and test it out in my enterprise app!

Just to make it cleaner, this solution does 99% of the task. The rest would be to wrap appropriate log arguments that you'd like to make sensitive into the `markSensitive` function, for example:

```
getLogger(driver, req.params.sessionId).debug(
        `Calling ` +
          `${driver.constructor.name}.${spec.command}() with args: ` +
          _.truncate(JSON.stringify(args), {length: MAX_LOG_BODY_LENGTH}
)
```

becomes

```
getLogger(driver, req.params.sessionId).debug(
        `Calling ` +
          `${driver.constructor.name}.${spec.command}() with args: ` +
          markSensitive(_.truncate(JSON.stringify(args), {length: MAX_LOG_BODY_LENGTH})
)
```

then if the corresponding server request has the above header enabled then the server should log something like `Calling  driver.command with args: **SECURE**`. If the magic header is not present this log should look normal and contain usual arguments.

eglitise (approved):

KazuCocoa (approved):
lgtm

Maybe... we could address this header thing in.. https://appium.io/docs/en/latest/guides/log-filters and around https://appium.io/docs/en/latest/developing/build-drivers/#log-messages-to-the-appium-log ...?

(Can be another pr)

dprevost-LMI:
For your information, the arg needs to be separated from the string to be correctly processed; otherwise, it is interpolated into the string.

To take the same example:
```
getLogger(driver, req.params.sessionId).debug(
        `Calling ` +
          `${driver.constructor.name}.${spec.command}() with args: ` +
          _.truncate(JSON.stringify(args), {length: MAX_LOG_BODY_LENGTH}
)
```

becomes

```
getLogger(driver, req.params.sessionId).debug(
        `Calling ` +
          `${driver.constructor.name}.${spec.command}() with args: `,
          markSensitive(_.truncate(JSON.stringify(args), {length: MAX_LOG_BODY_LENGTH})
)
```

PR #21134, opened 2025-03-20
feat(base-driver): Remove deprecated routes

mykola-mokhnach:


packages/base-driver/lib/protocol/routes.js
```diff
@@ -41,41 +28,27 @@ export const METHOD_MAP = /** @type {const} */ ({
...
-        optional: ['desiredCapabilities', 'requiredCapabilities', 'capabilities'],
+        optional: ['capabilities', 'capabilities', 'capabilities'],
       },
     },
   },
   '/session/:sessionId': {
-    GET: {command: 'getSession', deprecated: true},
+    // Even though this command is not present in the official W3C protocol
+    // we find it useful and keep it for good in Appium
+    GET: {command: 'getSession'},
```
eglitise, line 38: Isn't this replaced by `GET /session/:sessionId/appium/capabilities`?

packages/appium/test/unit/utils.spec.js
```diff
@@ -16,36 +16,23 @@ import logger from '../../lib/logger';
...
     });
     it('should return W3C caps unchanged if only W3C caps were provided', function () {
-      let {desiredCaps, processedJsonwpCapabilities, processedW3CCapabilities, protocol} =
-        parseCapsForInnerDriver(undefined, W3C_CAPS);
-      desiredCaps.should.deep.equal(BASE_CAPS);
-      should.not.exist(processedJsonwpCapabilities);
-      processedW3CCapabilities.should.deep.equal(W3C_CAPS);
-      protocol.should.equal('W3C');
-    });
-    it('should return JSONWP and W3C caps if both were provided', function () {
```
eglitise, line 42: We should probably still have a test for the case when the user provides JWP+W3C - but should we only accept W3C & ignore JWP, or raise an error?

mykola-mokhnach: ok, I'll update this test to check for an error instead if only jwp caps are provided

packages/appium/test/unit/appiumdriver.spec.js
```diff
@@ -123,9 +123,9 @@ describe('AppiumDriver', function () {
         mockFakeDriver
           .expects('createSession')
           .once()
-          .withExactArgs(undefined, null, W3C_CAPS, [])
+          .withExactArgs(W3C_CAPS, W3C_CAPS, W3C_CAPS, [])
```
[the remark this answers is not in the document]
eglitise: We would have to update all drivers regardless, due to the minimum Node version bump. But I'm fine if we decide to keep this for now, and change it in the next major release.

packages/base-driver/lib/protocol/routes.js
eglitise: Alright, if we want to handle it in a separate PR, that sounds reasonable.

packages/appium/test/unit/utils.spec.js
mykola-mokhnach: actually now this method only accepts w3c caps and the case where unacceptable caps are provided is already covered

packages/base-driver/lib/protocol/routes.js
```diff
@@ -110,22 +82,7 @@ export const METHOD_MAP = /** @type {const} */ ({
...
-          if (util.hasValue(jsonObj.name) && !util.hasValue(jsonObj.handle)) {
-            return [jsonObj.name, jsonObj.name];
-          }
-          return [jsonObj.name, jsonObj.handle];
-        },
-        validate: (jsonObj) =>
-          !util.hasValue(jsonObj.name) &&
-          !util.hasValue(jsonObj.handle) &&
-          'we require one of "name" or "handle" to be set',
+        required: ['name'],
```
KazuCocoa, line 85: “handle”?

mykola-mokhnach: good catch. Updated

packages/appium/test/unit/appiumdriver.spec.js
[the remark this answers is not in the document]
jlipps: yeah I think we should probably make this breaking change for v3. It's been super annoying and we're already making breaking changes so...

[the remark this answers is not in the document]
mykola-mokhnach: I don't mind. It should be done in a separate PR though, this one is already quite large

KazuCocoa (approved):
lg

PR #21149, opened 2025-03-27
fix: adjust parameter for W3C element send keys endpoint

eglitise:
Noticed a small bug that snuck in during removal of deprecated routes - the parameter name for the [W3C Element Send Keys endpoint](https://www.w3.org/TR/webdriver1/#dfn-element-send-keys) (`POST /session/:sessionId/element/:elementId/value`) was only accepting `value`, but the spec actually expects `text`.


mykola-mokhnach (approved):
good catch
thanks

Issue #20321, opened 2024-07-03
bug: remove the obsolete "status" field from W3C error responses


joerg1985:
### Do I have the most recent component updates?

- [X] I use the most recent available driver/plugin and server versions

### Is the component officially supported by the Appium team?

- [X] I have verified the component repository is present under the Appium organization in GitHub

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Appium 2.10.3 does still use none W3C encoded errors.

E.g. calling http://127.0.0.1:4723/test/ does result in this error:
```
{
    "status": 9,
    "value": {
        "error": "unknown command",
        "message": "The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource",
        "stacktrace": ""
    }
}
```

### Expected Behavior

Appium should use W3C encoded errors.
https://www.w3.org/TR/webdriver2/#example-3

### Minimal Reproducible Example

1. run appium
2. call http://127.0.0.1:4723/test/

### Environment

- Operating system: Windows 10 x64
- Appium server version (output of `appium --version`): 2.10.3
- Appium driver(s) and their version(s) (`appium driver list`):
- Appium plugin(s) and their version(s) (`appium plugin list`):
- Node.js version (output of `node --version`):
- `npm` version (output of `npm --version`):
- Last component(s) version which did _not_ exhibit the problem:
- Platform and version under test:
- Real device or emulator/simulator:

### Further Information

I created a [PR](https://github.com/SeleniumHQ/selenium/pull/14195) some days ago to remove the support of none W3C encoded errors from Selenium. It is totally open whether this gets merged or not.

mykola-mokhnach:
Could you please provide an example of the expected response? 

In the provided link the above error message exists and is valid:

> unknown command 	404 	unknown command 	A [command](https://www.w3.org/TR/webdriver2/#dfn-commands) could not be executed because the [remote end](https://www.w3.org/TR/webdriver2/#dfn-remote-ends) is not aware of it.

joerg1985:
Sorry i did not look too deep into this before opening, the encoding is correct.
The title of this issue should be: "status" could be removed from error responses.

e.g. the selenium server does respond in this case with:
```
{
  "value": {
    "stacktrace": "",
    "error": "unknown command",
    "message": "Unable to find handler for (GET) \u002fstatusdas"
  }
}
```
vs. the current response of appium: 
```
{
    "status": 9,
    "value": {
        "error": "unknown command",
[...]

PR #21150, opened 2025-03-27
chore(base-driver): Remove the legacy JWP-related stuff

mykola-mokhnach:
## Proposed changes

Also deleted the hacky Python middleware. I assume it has been not relevant anymore for quite a while.

### DIFF
--- a/packages/base-driver/lib/express/middleware.js
+++ b/packages/base-driver/lib/express/middleware.js
@@ -5,6 +5,9 @@ export {handleIdempotency} from './idempotency';
 import {match} from 'path-to-regexp';
 import {util} from '@appium/support';
 import {calcSignature} from '../helpers/session';
+import {getResponseForW3CError} from '../protocol/errors';
+
+const SESSION_ID_PATTERN = /\/session\/([^/]+)/;
 
 /**
  *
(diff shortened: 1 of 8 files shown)


mykola-mokhnach:
Linking https://github.com/appium/appium/issues/20321

packages/base-driver/lib/express/middleware.js
```diff
@@ -5,6 +5,9 @@ export {handleIdempotency} from './idempotency';
 import {match} from 'path-to-regexp';
 import {util} from '@appium/support';
 import {calcSignature} from '../helpers/session';
+import {getResponseForW3CError} from '../protocol/errors';
+
+const SESSION_ID_PATTERN = /\/session\/([^/]+)/;
```
eglitise, line 10: This constant appears to be unused now

eglitise: Nevermind, it is used on line 62.

eglitise (approved):
If I understand these changes correctly:
* Error messages will no longer include the `status` or `sessionId` fields, instead keeping them in `value` for a more standard W3C response
* Errors will also no longer include the `An unknown server-side error occurred while processing the command`/`The requested resource could not be found` parts and instead use the common `getResponseForW3CError` handler
* The `fixPythonContentType` hack has been removed and the header is instead explicitly set in `lib/protocol/protocol.js`

mykola-mokhnach:
> If I understand these changes correctly:
> 
> * Error messages will no longer include the `status` or `sessionId` fields, instead keeping them in `value` for a more standard W3C response

not all error responses had these extra fields, i've just cleaned then up from ones that still contained them

> * Errors will also no longer include the `An unknown server-side error occurred while processing the command`/`The requested resource could not be found` parts and instead use the common `getResponseForW3CError` handler

I've just removed the copy from middlewares as it was the same as the actual error message for corresponding errors types. I was going to create another PR afterwards to improve typing for these errors

> * The `fixPythonContentType` hack has been removed and the header is instead explicitly set in `lib/protocol/protocol.js`

headers tuning is not related to the middleware removal.

KazuCocoa (approved):
lg

PR #21157, opened 2025-03-29
feat(base-driver): Migrate errors to typescript

mykola-mokhnach:
Breaking changes:
- Removed BadParametersError from errors as it is only used internally and changed its signature
- Made w3c and jsonwp properties of ProtocolError private
- Fixed UnknownError constructor signature to accept same parameters as other protocol errors

### DIFF
--- a/packages/base-driver/lib/protocol/errors.ts
+++ b/packages/base-driver/lib/protocol/errors.ts
@@ -1,51 +1,63 @@
 import _ from 'lodash';
 import {util, logger} from '@appium/support';
 import {StatusCodes as HTTPStatusCodes} from 'http-status-codes';
+import type { ErrorBiDiCommandResponse, Class } from '@appium/types';
 
 const mjsonwpLog = logger.getLogger('MJSONWP');
 const w3cLog = logger.getLogger('W3C');
 
-const W3C_UNKNOWN_ERROR = 'unknown error';
-
 class BaseError extends Error {
-  constructor(message = '') {
+  public cause: Error | undefined;
+  public message: string;
+  public name: string;
+  public stack: string | undefined;
+
+  constructor(message: string = '', cause?: Error) {
     super(message);
-    /** @type {[string, () => any][]} */
-    const propsMap = [
-      ['message', () => message],
-      ['name', () => this.constructor.name],
-      ['stack', () => (new Error(message)).stack],
-    ];
+    this.cause = cause;
+    this.message = message;
+    this.name = this.constructor.name;
+    this._formatStack();
+  }
+
+  private _formatStack(): void {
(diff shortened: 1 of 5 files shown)


packages/base-driver/lib/protocol/errors.ts
```diff
@@ -402,16 +399,15 @@ export class InsecureCertificateError extends ProtocolError {
...
-  /**
-   * @param {string} [message] error message
-   */
-  constructor(message) {
+
+  constructor(message: string = '', cause?: Error) {
     super(
       message ||
         'Navigation caused the user agent to hit a certificate warning, which is usually the result of an expired or invalid TLS certificate',
       ElementIsNotSelectableError.code(),
```
eglitise, line 407: Should we perhaps define `code`/`w3cStatus`/`error` in every error class? `InsecureCertificateError`, `NoSuchContextError` and `InvalidContextError` are using one or more params from other error classes. I feel like the error codes should certainly be unique.

mykola-mokhnach: I have updated InsecureCertificateError constructor. As for the other two, I cannot map them to existing W3C error signatures at https://w3c.github.io/webdriver/#dfn-error-code 
Do you have some proposals regarding the latter?

KazuCocoa (approved):
lgtm

packages/base-driver/lib/protocol/errors.ts
eglitise: I believe the error codes used in this file were added [according to this document](https://github.com/appium/appium/blob/master/packages/base-driver/docs/mjsonwp/errors.md), which in turn references [the Selenium JSONWP codes](https://www.selenium.dev/documentation/legacy/json_wire_protocol/#responses). I checked the Selenium code and it seems [these codes are labeled as legacy](https://github.com/SeleniumHQ/selenium/blob/trunk/javascript/selenium-webdriver/lib/error.js#L368). So I guess it may be fine to leave the current error code for `InsecureCertificateError`.
`NoSuchContextError` and `InvalidContextError` aren't present in Selenium at all, which makes this trickier. Ideally this is something that should be aligned with W3C, perhaps to allow extensions for errors, but for now I guess we can leave them as `UnknownError`.
It's worth noting that because `NoSuchContextError` and `InvalidContextError` don't exist in Selenium, their code values are also non-standard and should probably also be replaced with the `UnknownError` code (especially because Selenium seems to already use code 35 as `SQL_DATABASE_ERROR`)

mykola-mokhnach: for w3c they are already mapped to unknown errors. The static `code()` method is only used for mjsonwp.

PR #21160, opened 2025-03-30
fix(docutils): add & adjust various log messages

eglitise:
This PR adjusts validation for the `docutils` executable:
* Add missing error messages:
  * if `appium-docs init` is called without Python being installed
  * if `appium-docs build` is called without `mkdocs` being installed
* Add missing debug messages:
  * when calling `appium-docs validate`
  * when calling `appium-docs build --deploy --serve`
* Adjust info message when calling `appium-docs init` with an existing `tsconfig.json` file 
* Adjust debug messages when calling `appium-docs validate`

### DIFF
--- a/packages/docutils/lib/init.ts
+++ b/packages/docutils/lib/init.ts
@@ -5,7 +5,7 @@
  */
 
 import * as JSON5 from 'json5';
-import {NAME_MKDOCS_YML, NAME_TSCONFIG_JSON, NAME_PYTHON, PIP_ENV_VARS, REQUIREMENTS_TXT_PATH} from './constants';
+import {NAME_MKDOCS_YML, NAME_TSCONFIG_JSON, PIP_ENV_VARS, REQUIREMENTS_TXT_PATH} from './constants';
 import YAML from 'yaml';
 import {exec} from 'teen_process';
 import {Simplify} from 'type-fest';
@@ -15,7 +15,7 @@ import type { ScaffoldTask } from './scaffold';
 import {getLogger} from './logger';
 import {MkDocsYml, TsConfigJson} from './model';
 import _ from 'lodash';
-import {findPython, stringifyJson5, stringifyYaml} from './fs';
+import {requirePython, stringifyJson5, stringifyYaml} from './fs';
 
 /**
  * Data for the base `mkdocs.yml` file
@@ -139,7 +139,7 @@ export async function initPython({
   dryRun = false,
   upgrade = false,
 }: InitPythonOptions = {}): Promise<void> {
-  pythonPath = pythonPath ?? (await findPython()) ?? NAME_PYTHON;
+  const foundPythonPath = await requirePython(pythonPath);
 
   const args = ['-m', 'pip', 'install', '-r', REQUIREMENTS_TXT_PATH];
   if (upgrade) {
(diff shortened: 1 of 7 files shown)


packages/docutils/lib/init.ts
```diff
@@ -139,7 +139,10 @@ export async function initPython({
   dryRun = false,
   upgrade = false,
 }: InitPythonOptions = {}): Promise<void> {
-  pythonPath = pythonPath ?? (await findPython()) ?? NAME_PYTHON;
+  const foundPythonPath = pythonPath ?? (await findPython());
+  if (!foundPythonPath) {
+    throw new DocutilsError(`Could not find Python in PATH. Is it installed?`);
```
mykola-mokhnach, line 144: I see the same message is repeated. Perhaps, we could move it to a separate function called `requirePython` ?

packages/docutils/lib/validate.ts
```diff
@@ -358,24 +368,24 @@ export class DocutilsValidator extends EventEmitter {
...
   protected async validateNpmVersion() {
+    log.debug(`Validating ${NAME_NPM} version`);
+
     const npmEngineRange = DOCUTILS_PKG.engines?.npm;
     if (!npmEngineRange) {
-      throw new DocutilsError('Could not find property engines.npm in package.json. This is a bug');
+      throw new DocutilsError(`Could not find property 'engines.npm' in ${NAME_PACKAGE_JSON}. This is a bug`);
+    }
+
+    const npmPath = this.npmPath ?? (await whichNpm());
```
mykola-mokhnach, line 378: not sure if it makes sense in this context, but on Windows npm is a batch executable

mykola-mokhnach (approved):

packages/docutils/lib/validate.ts
eglitise: Yes, I was actually struggling quite a bit with this...
`teen_process`'s `exec` calls `spawn`, which doesn't work for `bat`/`cmd` files on Windows, unless `shell: true` is set.
Adding this flag then caused it to start failing for all paths that contain a space (node/npm is installed in Program Files by default).
I tried wrapping the path using `util.quote`, but that additionally escaped the backslashes in the path, causing it to still fail.
For now I've left it as is - hopefully it can be investigated more in the future.

Actually, the `exec` call for retrieving the `tsc` version was also failing on Windows, but I have not investigated that one yet.

mykola-mokhnach: try to enable `windowsVerbatimArguments` option if on windows, similar to https://github.com/appium/appium-adb/blob/ae59cfb7f1c89e2c3f869c54fd3e70cc19b22f03/lib/tools/apk-signing.js#L53

packages/docutils/lib/init.ts
eglitise: I've extracted the code that raises an error to a separate method, but for the code that returns `this.fail` I only extracted the message string, since those would still need a return call.

packages/docutils/lib/validate.ts
eglitise: Unfortunately this doesn't help. In fact, Node automatically sets this option "when `shell` is specified and is CMD".

eglitise: I managed to fix both the `npm` and `tsc` checks by using `npm.exec` from `@appium/support`. A similar approach is already used in [`appium/lib/config.js`](https://github.com/appium/appium/blob/master/packages/appium/lib/config.js#L68).
Both checks now work as expected on Windows, but I will also validate it on macOS tomorrow.

eglitise: Confirmed on macOS that the fix still keeps the expected behavior 👍

KazuCocoa (approved):

Issue #21171, opened 2025-04-02
validate.js Regular Expression Denial of Service vulnerability


sanya-jolly:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Appium Version

2.17.1

### Appium Host Type

Command Line

### Is this issue reproducible using the latest components?

- [x] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

![Image](https://github.com/user-attachments/assets/58a83d2e-579e-4aa8-ae7a-ccb80647b2a4)
![Image](https://github.com/user-attachments/assets/cdce077a-e864-4d6e-a97b-2d4e412910f4)
![Image](https://github.com/user-attachments/assets/c48ff0c2-663f-4d2e-8f4b-86cc2b571c35)

### Expected Behavior

Regular expressions are patterns used to match character combinations in strings. Certain poorly constructed patterns can lead to excessive backtracking, especially when processing crafted malicious inputs. This excessive backtracking can cause the application to hang or crash, resulting in a denial of service.

Thus, patch should be available for this.

### Environment Details

- Debug details (`appium --show-debug-info`):
- Last component version(s) which did _not_ exhibit the problem:
- Platform and version of device under test:
- Real device or emulator/simulator:

### Minimal Reproducible Example

just run yarn audit on RN .77 in any mobile project

mykola-mokhnach:
This package has not been maintained for years. It looks like we just need to get rid of it completely

KazuCocoa:
Upcoming appium 3 will drop the deps. https://github.com/appium/appium/pull/21175

PR #21175, opened 2025-04-03
feat(base-driver): Remove dependency to validate.js

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/21171


packages/base-driver/lib/basedriver/validation.ts
```diff
@@ -0,0 +1,120 @@
...
+      return null;
+    },
+    inclusion: (value: any, options?: any): string | null => {
+      if (_.isUndefined(value) || !options) {
+        return null;
+      }
+      if (_.isArray(options) && options.some((opt) => opt === value)) {
+        return null;
+      }
+      return `-> '${value}' not part of '${options ?? []}'`;
```
eglitise, line 69: Not sure if this method and `inclusionCaseInsensitive` even make sense if we allow `options` to be empty. I guess the original code wasn't checking that though.
Also is the `->` necessary?

packages/base-driver/test/unit/basedriver/capability.spec.js
```diff
@@ -344,66 +344,25 @@ describe('Desired Capabilities', function () {
...
-        presence: true,
-      },
-    };
-
-    await d
-      .createSession({
-        alwaysMatch: {
-          platformName: 'iOS',
-          'appium:foo': null,
+  for (const capValue of [null, '', {}, [], ' ',]) {
```
eglitise, line 347: Extra comma at the end of the array (but I guess it would make sense in multi-line format)

packages/appium/scripts/parse-yml-commands.js
```diff
@@ -5,55 +5,55 @@
...
-
-  if (!_.isArray(value)) {
-    value = [value];
-  }
+const validators = {
+  isArray: (value, options) => {
+    if (_.isUndefined(value) || _.isArray(value) || !options) {
+      return null;
+    }
+    return ` must be an array`;
```
eglitise, line 20: Extra space at the start of message

packages/appium/scripts/parse-yml-commands.js
```diff
@@ -5,55 +5,55 @@
...
+  },
+  presence: (value, options) => {
+    if (_.isUndefined(value) && options) {
+      return `must be present`;
+    }
+    return null;
+  },
+  hasAttributes: (value, options) => {
+    if (!value) {
```
eglitise, line 29: Is it maybe worth checking if `options` is non-null here as well, like in `isArray`?

packages/appium/scripts/parse-yml-commands.js
```diff
@@ -5,55 +5,55 @@
...
-  }
-
-  // if just a bare value, allow it through
-  if (!_.isArray(value)) {
-    return;
-  }
+    return null;
+  },
+  hasPossibleAttributes(value, options) {
+    if (!value || !_.isArray(value)) {
```
eglitise, line 43: Same here

packages/base-driver/lib/basedriver/validation.ts
```diff
@@ -0,0 +1,120 @@
...
+    },
+    presence: (value: any, options?: any): string | null => {
+      if (_.isUndefined(value) && options) {
+        return `is required to be present`;
+      }
+      if (
+        !options?.allowEmpty &&
+        (!_.isUndefined(value) && _.isEmpty(value) || _.isString(value) && !_.trim(value))
```
eglitise, line 86: This condition is difficult to to understand without additional brackets

packages/base-driver/lib/basedriver/validation.ts
mykola-mokhnach: I've updated the message and the logic there. 
In general, one may put anything as `options` value, the question is only how we are going to handle it. I want this handling to be safe though.

packages/base-driver/test/unit/basedriver/capability.spec.js
mykola-mokhnach: removed it

packages/appium/scripts/parse-yml-commands.js
mykola-mokhnach: fixed

packages/appium/scripts/parse-yml-commands.js
mykola-mokhnach: not sure what exactly you mean. Could you provide an example code?

packages/base-driver/lib/basedriver/validation.ts
mykola-mokhnach: added

packages/appium/scripts/parse-yml-commands.js
```diff
@@ -5,55 +5,55 @@
...
+    }
 
-  for (const item of value) {
-    for (const key of _.keys(item)) {
-      if (!options.includes(key)) {
-        return `must not include '${key}'. Available options: ${options}`;
+    for (const item of value) {
+      for (const key of _.keys(item)) {
+        if (!_.includes(options, key)) {
+          return `must not include '${key}'. Available options: ${options}`;
```
eglitise, line 51: This should probably be `must include '${key}'`

packages/appium/scripts/parse-yml-commands.js
eglitise: I assume that we can just skip the loops if `options` is falsy:
```js
if (!value || !options) {
  return null;
}
```
For `hasPossibleAttributes` I guess a falsy `options` value should raise an error?

packages/base-driver/lib/basedriver/validation.ts
```diff
@@ -0,0 +1,125 @@
...
+import log from './logger';
+import _ from 'lodash';
+
+export class Validator {
+  private readonly _validators: Record<
+    keyof Constraint,
+    (value: any, options?: any, key?: string) => string | null
+  > = {
+    isString: (value: any, options?: any): string | null => {
+      if (_.isString(value) || _.isUndefined(value) || !options) {
```
eglitise, line 11: I wonder about the `options` check here. If we pass e.g. an integer to this method, but do not provide `options`, it will still return `null` - is that okay? The previous code didn't check `options` at all for most of these methods

packages/appium/scripts/parse-yml-commands.js
mykola-mokhnach: changed

packages/appium/scripts/parse-yml-commands.js
mykola-mokhnach: changed

packages/base-driver/lib/basedriver/validation.ts
mykola-mokhnach: I've updated the logic, so now it actually considers both possible alternatives, e.g.

```
foo: {
  isString: true
}
```

would verify is the given cap value is a valid string

and 

```
foo: {
  isString: false
}
```

would verify is the given cap value is NOT a string

Same applies to other is... validators

KazuCocoa (approved):
lgtm overall

packages/base-driver/lib/basedriver/validation.ts
```diff
@@ -8,50 +8,70 @@ export class Validator {
...
+      return options ? 'must be of type string' : null;
     },
     isNumber: (value: any, options?: any): string | null => {
-      if (_.isNumber(value) || _.isUndefined(value) || !options) {
+      if (_.isUndefined(value) || _.isNil(options)) {
         return null;
       }
 
+      if (_.isNumber(value)) {
+        return options ? null : 'most not be of type string';
```
eglitise, line 27: `must not be of type number`
Other validators also have the same typo (`most`->`must`)

mykola-mokhnach: good catch. Updated

packages/appium/scripts/parse-yml-commands.js
```diff
@@ -5,55 +5,55 @@
...
-  }
-
-  // if just a bare value, allow it through
-  if (!_.isArray(value)) {
-    return;
-  }
+    return null;
+  },
+  hasPossibleAttributes(value, options) {
+    if (!value || !_.isArray(value) || !options) {
```
eglitise, line 43: If I understand this method correctly, `options` _must_ define _all_ the attributes of `value`, so if `options` is empty, then the only accepted case should be if `value` has no attributes at all

mykola-mokhnach: updated the condition

eglitise (approved):

PR #21176, opened 2025-04-03
fix(logger): Error stack logging

mykola-mokhnach:


packages/logger/lib/log.ts
```diff
@@ -237,8 +237,11 @@ export class Log extends EventEmitter implements Logger {
     let stack: string | undefined;
     for (const arg of [message, ...args]) {
       const result = this._formatLogArgument(arg);
-      stack = result.stack;
-      messageArguments.push(result.arg);
+      if (result.stack) {
+        stack = result.stack;
+      } else {
+        messageArguments.push(result.arg);
```
eglitise, line 243: Shouldn't this line be executed regardless of the value of `result.stack`? At least, `npmlog` seems to work like that.

mykola-mokhnach: not really. The actual stack trace gets added to the log twice if such condition is not present

eglitise (approved):

PR #21185, opened 2025-04-06
fix(ci): run on all Node LTS versions instead of only active

eglitise:
I noticed that CI has only been running on Node 22 as of a few days ago. Turns out that the `node-lts-versions` action had been configured only for `active` versions.
From the `node-lts-versions` README:

```
At the time of writing, active=[22] and lts=[18,20,22].
```
This PR ensures that CI is running on all Node LTS versions, by changing `active` to `lts`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,7 +12,7 @@ jobs:
   prepare_matrix:
     runs-on: ubuntu-latest
     outputs:
-      versions: ${{ steps.generate-matrix.outputs.active }}
+      versions: ${{ steps.generate-matrix.outputs.lts }}
     steps:
     - name: Select all active LTS versions of Node.js
       id: generate-matrix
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -12,7 +12,7 @@ jobs:
   prepare_matrix:
     runs-on: ubuntu-latest
     outputs:
-      versions: ${{ steps.generate-matrix.outputs.active }}
+      versions: ${{ steps.generate-matrix.outputs.lts }}
     steps:
     - name: Select all active LTS versions of Node.js
```
mykola-mokhnach, line 17: we should also update this step name then

mykola-mokhnach (approved):

.github/workflows/deploy.yml
eglitise: well, it already says LTS, so arguably it had been incorrect until now, but fair enough

KazuCocoa:
oh, nice. Maybe we need to apply the same change to various repos as well

PR #21198, opened 2025-04-11
chore: Remove various deprecated driver methods

mykola-mokhnach:


KazuCocoa (approved):

packages/types/lib/driver.ts
```diff
@@ -1526,174 +1357,6 @@ export interface ExternalDriver<
...
-  equalsElement?(elementId: string, otherElementId: string): Promise<boolean>;
-  /**
-   * Get the list of IME engines
-   *
-   * @returns The list of IME engines
-   *
-   * @deprecated
-   * @privateRemarks Not implemented in `appium-xcuitest-driver`
-   */
-  availableIMEEngines?(): Promise<string[]>;
```
eglitise, line 1661: These 5 IME methods should not be removed yet, since their endpoints still exist in `base-driver`

packages/types/lib/driver.ts
```diff
@@ -1708,105 +1371,6 @@ export interface ExternalDriver<
...
-
-  /**
-   * Click the current mouse location
-   *
-   * @param button - the button ID
-   *
-   * @deprecated Use the Actions API instead
-   * @privateRemarks Not implemented in `appium-xcuitest-driver`
-   */
-  clickCurrent?(button?: number): Promise<void>;
```
eglitise, line 1739: There's still a `clickCurrent` method in `base-driver` and `fake-driver`, is it related to this one?

packages/types/lib/driver.ts
```diff
@@ -1708,105 +1371,6 @@ export interface ExternalDriver<
...
-
-  /**
-   * Double-click the current mouse location
-   *
-   * @deprecated Use the Actions API instead
-   * @privateRemarks Not implemented in `appium-xcuitest-driver`
-   */
-  doubleClick?(): Promise<void>;
```
eglitise, line 1747: `fake-driver` also has `doubleClick`, unsure if it's related

packages/types/lib/driver.ts
mykola-mokhnach: removed it for good

packages/types/lib/driver.ts
mykola-mokhnach: it has its own mapping defined, which is independent from base driver stuff, so let it be there

packages/types/lib/driver.ts
mykola-mokhnach: removed them for good

eglitise: I don't think we should remove them yet, as they haven't been deprecated. Same for `receive_async_response`. We can remove them in the next major version.

packages/base-driver/lib/protocol/validators.ts
```diff
@@ -1,38 +1,27 @@
...
-  asyncScriptTimeout: (ms) => {
-    msValidator(ms);
-  },
-  clickCurrent: (button) => {
-    if (!(isNumber(button) || _.isUndefined(button)) || button < 0 || button > 2) {
-      throw new Error('Click button must be 0, 1, or 2');
-    }
-  },
-  setNetworkConnection: (type) => {
+  implicitWait: msValidator,
```
eglitise, line 10: I wonder if `implicitWait` and `asyncScriptTimeout` are still relevant, given that the methods/endpoints with the same name were removed

packages/types/lib/driver.ts
mykola-mokhnach: - They are not used anywhere except of the android driver, where they are in the execute method map anyway, which means removing them won't break anything
- All these endpoints belong to the obsolete JWP protocol

packages/base-driver/lib/protocol/validators.ts
mykola-mokhnach: removed

packages/types/lib/driver.ts
eglitise: The routes file also has several other JWP endpoints that have execute method equivalents - that's not the issue. I just don't think any routes should be removed without first deprecating them.

packages/base-driver/lib/protocol/validators.js
```diff
@@ -1,38 +0,0 @@
...
-}
-
-function msValidator(ms) {
-  if (!_.isNumber(ms) || ms < 0) {
-    throw new Error('Wait ms must be a number equal to 0 or greater');
-  }
-}
-
-const validators = {
-  setUrl: (url) => {
```
eglitise, line 14: `setUrl` is called by `POST /session/:sessionId/url`, which is a W3C endpoint, so there is value in retaining its validation

packages/base-driver/lib/protocol/validators.js
```diff
@@ -1,38 +0,0 @@
...
-  },
-  asyncScriptTimeout: (ms) => {
-    msValidator(ms);
-  },
-  clickCurrent: (button) => {
-    if (!(isNumber(button) || _.isUndefined(button)) || button < 0 || button > 2) {
-      throw new Error('Click button must be 0, 1, or 2');
-    }
-  },
-  setNetworkConnection: (type) => {
```
eglitise, line 31: While this method is called by the MJSONWP endpoint `POST /session/:sessionId/network_connection`, it was not deprecated and should still be retained in Appium 3, so the validation should also be retained

packages/base-driver/lib/protocol/validators.js
mykola-mokhnach: this validation does not make sense, it must be performed by drivers which implement this endpoint rather than on the router level

packages/base-driver/lib/protocol/validators.js
mykola-mokhnach: same as above

packages/types/lib/driver.ts
mykola-mokhnach: @jlipps @KazuCocoa I would like to know your opinion on that - does it make sense to still keep these routes?

packages/base-driver/lib/protocol/routes.js
```diff
@@ -250,21 +250,6 @@ export const METHOD_MAP = /** @type {const} */ ({
...
-  },
-  '/session/:sessionId/ime/activated': {
-    GET: {command: 'isIMEActivated'},
-  },
-  '/session/:sessionId/ime/deactivate': {
-    POST: {command: 'deactivateIMEEngine'},
-  },
-  '/session/:sessionId/ime/activate': {
-    POST: {command: 'activateIMEEngine', payloadParams: {required: ['engine']}},
-  },
```
KazuCocoa, line 267: For this file, I'd like to keep these (for now) and put `deprecated` first before removing them like other removals because we haven't marked as `deprecated` yet in current `master`. Instead, I'm good at removing these method from types to imply we don't expect to be implemented these endpoints

packages/types/lib/driver.ts
KazuCocoa: for the endpoints themselves in packages/base-driver/lib/protocol/routes.js, i'd like to remove them after marking `deprecated` first.
In terms of types in this file, I think we can remove and put this typehint in Android driver repo only to show these IMEs are not expected to be called except for android driver

packages/base-driver/lib/protocol/routes.js
mykola-mokhnach: restored

packages/base-driver/lib/express/server.js
```diff
@@ -167,8 +166,6 @@ export function configureServer({
   app.use(handleUpgrade(webSocketsMapping));
   if (allowCors) {
     app.use(allowCrossDomain);
-  } else {
-    app.use(allowCrossDomainAsyncExecute(basePath));
```
eglitise, line 171: Do we want to keep this handler as well, while the `receive_async_response` endpoint remains?

packages/base-driver/lib/protocol/validators.ts
eglitise: I assume `implicitWait` and `asyncScriptTimeout` were added back unintentionally

packages/base-driver/test/e2e/protocol/protocol.e2e.spec.js
```diff
@@ -229,22 +229,6 @@ describe('Protocol', function () {
       });
     });
 
-    it('should return the correct error even if driver does not throw', async function () {
```
eglitise, line 232: Would this test still be relevant even after the `receive_async_response` is removed? It does not seem to be specific to that endpoint, so I'm not sure if it should be removed

packages/base-driver/lib/express/server.js
mykola-mokhnach: restored

packages/base-driver/test/e2e/protocol/protocol.e2e.spec.js
mykola-mokhnach: updated the test

eglitise: The updated test is not the same. The original test queried an endpoint whose implementation does not throw an error, but this one just queries a non-existent endpoint. The simplest way to update the test would probably be to change the endpoint to a different one that has a real equivalent (e.g. `/session/foo/appium/settings`), and therefore also rename the method in `base-driver/test/e2e/protocol/fake-driver.js` to match the one used by that endpoint (e.g. `updateSettings`)

mykola-mokhnach: Unfortunately I cannot make it work other ways and I don't want to spend more time on it. If you have a better solution that works then I would be happy to replace this one

eglitise: Sure:
```js
    it('should return the correct error even if driver does not throw', async function () {
      const {status, data} = await axios({
        url: `${baseUrl}/session/foo/appium/settings`,
        method: 'GET',
        validateStatus: null,
      });
      status.should.equal(500);
      data.value.error.should.eql('unknown error');
      data.value.message.should.eql('Mishandled Driver Error');
      should.not.exist(data.sessionId);
    })
```

And in `packages/base-driver/test/e2e/protocol/fake-driver.js`, replace the removed `receiveAsyncResponse` method with:
```js
  async getSettings() {
    // this is here to test a failing command that does not throw an error
    return {status: 13, value: 'Mishandled Driver Error'};
  }
```

mykola-mokhnach: unfortunately I cannot make it work locally:

```

  Protocol
    via express router
info HTTP --> GET /session/foo/appium/settings {}
dbug FakeDriver@55d0 Calling FakeDriver.getSettings() with args: ["foo"]
dbug FakeDriver@55d0 Responding to client with driver.getSettings() result: {}
info HTTP <-- GET /session/foo/appium/settings 200 4 ms - 12 
      1) should return the correct error even if driver does not throw
info HTTP Closing Appium HTTP server
info HTTP Appium HTTP server has been succesfully closed after 0ms


  0 passing (43ms)
  1 failing

  1) Protocol
       via express router
         should return the correct error even if driver does not throw:

      AssertionError: expected 200 to equal 500
      + expected - actual

      -200
      +500
      
      at Context.<anonymous> (test/e2e/protocol/protocol.e2e.spec.js:238:21)
      at processTicksAndRejections (node:internal/process/task_queues:105:5)
```

eglitise: That's odd, I also ran it locally but without issues 🤔 though I did not run a single test, but the entire suite:
```
cd packages/base-driver
npx mocha --exit --timeout 20s --slow 10s ./test/e2e/protocol/protocol.e2e.spec.js
```
This returned 32 failures, but this test was not one of them.
Might the difference be due to me running it on Windows?

mykola-mokhnach: you may try to create another draft branch and let CI run on it

packages/fake-driver/lib/driver.js
```diff
@@ -166,6 +166,11 @@ export class FakeDriver extends BaseDriver {
     return ['1'];
   }
 
+  async getSettings() {
```
eglitise, line 169: This should be in `packages/base-driver/test/e2e/protocol/fake-driver.js` 😅

eglitise (approved):

PR #21232, opened 2025-04-28
docs(appium): Describe how to log sensitive data

mykola-mokhnach:


packages/appium/docs/en/developing/sensitive.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Sensitive Log Data
+---
+
+Since Appium server version 2.18.0 there is a possibility to mask sensitive
```
eglitise, line 5: The version here can probably be omitted since the docs themselves are versioned

packages/appium/docs/en/developing/sensitive.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Sensitive Log Data
```
eglitise, line 2: Maybe 'Masking Sensitive Data'?

packages/appium/docs/en/developing/sensitive.md
```diff
@@ -0,0 +1,46 @@
...
+becomes
+
+```js
+import {logger} from '@appium/support';
+
+this.log.info('Value: %s', logger.markSensitive(value));
+```
+
+The formatting happens via the standard Node.js's
+[util.format](https://nodejs.org/api/util.html#utilformatformat-args) API.
```
eglitise, line 39: All of these lines should be indented 4 spaces (yes, 2 spaces isn't enough) in order for them to be correctly aligned with the bulletpoint on line 24

packages/appium/docs/en/developing/sensitive.md
mykola-mokhnach: I would prefer to still keep it for people that don't check docs version

packages/appium/docs/en/developing/sensitive.md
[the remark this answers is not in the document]
mykola-mokhnach: changed

mykola-mokhnach: changed

packages/appium/docs/en/developing/sensitive.md
```diff
@@ -0,0 +1,49 @@
...
+    import {logger} from '@appium/support';
+
+    this.log.info('Value: %s', logger.markSensitive(value));
+    ```
+
+    The formatting happens via the standard Node.js's
+    [util.format](https://nodejs.org/api/util.html#utilformatformat-args) API.
+
+- While sending the appropriate server request, where this log line is used and should be masked,
+  add the custom header `X-Appium-Is-Sensitive` with its value set to `1` or `true`.
```
KazuCocoa, line 45: Headers are case-insnsitive, so I believe not necessary to address but here could address the header key is case-insensitive

KazuCocoa (approved):

packages/appium/docs/en/developing/sensitive.md
mykola-mokhnach: added

PR #21233, opened 2025-04-29
feat(appium): implement GET /appium/sessions and deprecate GET /sessions

eglitise:
## Proposed changes

This PR adds an implementation for the `GET /appium/sessions` endpoint, according to #20880.
The target branch is intentionally set to `master`, _not_ `appium3` - seeing as `GET /sessions` has already been removed in Appium 3, it is beneficial for Appium 2 users to already have access to its replacement endpoint, for ease of migration.

I confirmed that these changes work using the Inspector (which already has support for this endpoint). Still, it would also be beneficial to add new tests specifically for validating the new `created` field - any suggestions welcome.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/packages/appium/docs/en/guides/security.md
+++ b/packages/appium/docs/en/guides/security.md
@@ -27,9 +27,9 @@ Appium when starting it from the command line:
 
 |<div style="width:10em">Parameter</div>|Description|
 |---------------------------------------|-----------|
-|`--relaxed-security`                   |Setting this flag turns on _all_ insecure features (unless blocked by `--deny-insecure`; see below)|
-|`--allow-insecure`                     |Setting this flag to a comma-separated list of feature names or a path to a file containing a feature list (each name on a separate line) will allow _only_ the features listed. For example, `--allow-insecure=adb_shell` will cause _only_ the ADB shell execution feature to be enabled. This is true _unless_ `--relaxed-security` is also used, in which case all features will still be enabled. It makes no sense to combine this flag with `--relaxed-security`.|
-|`--deny-insecure`                      |This flag can likewise be set to a comma-separated list of feature names, or a path to a feature file. Any features listed here will be _disabled_, regardless of whether `--relaxed-security` is set and regardless of whether the names are also listed with `--allow-insecure`.|
+|`--relaxed-security`|Turns on _all_ insecure features (unless blocked by `--deny-insecure`; see below)|
+|`--allow-insecure`|Turns on _only_ specified features. Features can be provided as a comma-separated list of feature names, or in the [Appium Configuration file](./config.md). For example, `--allow-insecure=adb_shell` will cause _only_ the ADB shell execution feature to be enabled. Has no effect when used in combination with `--relaxed-security`.|
+|`--deny-insecure`|Explicitly turns _off_ specified features, overriding `--relaxed-security` and any features specified using `--allow-insecure`. Like `--allow-insecure`, features can be provided as a comma-separated list of feature names, or in the [Appium Configuration file](./config.md).|
 
 ## Insecure Features
 
@@ -44,17 +44,15 @@ might use. Here is an incomplete list of examples from some of Appium's official
 |`record_audio`|Allow recording of host machine audio inputs|XCUITest|
 |`execute_driver_script`| Allows to send a request which has multiple Appium commands.|Execute Driver Plugin|
 
-## Driver-scope security
+Some insecure features operate on the server level, and do not require a driver session:
 
-Since Appium server version 2.12.3 there is a possibility to prefix each feature name with the corresponding automationName where it is going to be applied. For example, if you provide the following features to allow:
-`uiautomator2:adb_shell,xcuitest:get_server_logs,*:record_audio`, then the `adb_shell` would only be allowed
-for the UiAutomator2 driver, `get_server` - for the XCUITest driver and the `record_audio` one - for
-all installed drivers. Feature names provided without explicit automation name prefix are equivalent to
-these having the `*:` prefix. The same feature naming rule applies to the list of denied features.
+|<div style="width:12em">Feature Name</div>|Description|
+|------------|-----------|
+|`session_discovery`|Allows retrieving the list of active server sessions via `GET /appium/sessions`|
 
 ## Examples
 
-To turn on the `get_server_logs` feature for my Appium server, I could start it like this:
+To turn on the `get_server_logs` feature, the Appium server could be started like this:
 
 ```bash
 appium --allow-insecure=get_server_logs
@@ -71,3 +69,29 @@ To allow all features except one:
 ```bash
 appium --relaxed-security --deny-insecure=adb_shell
 ```
+
+## Driver-scope security
+
+Since Appium server version `2.13`, it is possible to apply features on a per-driver basis. This
+can be achieved by prefixing each feature name with the `automationName` of the driver for which
+the feature should be applied. To apply a feature for all drivers, or to apply a server-level
+feature, the wildcard (`*`) prefix should be used.
+
+For example, the server could be started as follows:
+```bash
+appium --allow-insecure=uiautomator2:adb_shell,xcuitest:get_server_logs,*:record_audio
+```
+
+This would result in the following:
+
+* The `adb_shell` feature would be enabled only for the UiAutomator2 driver
+* The `get_server_logs` feature would be enabled only for the XCUITest driver
+* The `record_audio` feature would be enabled for all drivers
+
+Feature provided without an explicit prefix are equivalent to having the wildcard prefix.
+These prefix rules apply to both the `--allow-insecure` and `--deny-insecure` server arguments.
+
+!!! warning
+
+    Starting from Appium 3, the scope prefix is required, and features provided without a scope will
+    raise an error. Note that the behavior of the `--relaxed-security` flag remains unchanged.
--- a/packages/appium/lib/appium.js
+++ b/packages/appium/lib/appium.js
@@ -171,17 +171,19 @@ class AppiumDriver extends DriverCore {
       const globalAllowedFeatures = filterInsecureFeatures(this.allowInsecure);
       if (!_.isEmpty(globalAllowedFeatures)) {
         logger.info('Explicitly enabling insecure features:');
-        globalAllowedFeatures.map((a) => logger.info(`    ${a}`));
+        globalAllowedFeatures.forEach((a) => logger.info(`    ${a}`));
       }
     }
-    if (!_.isEmpty(this.args.denyInsecure)) {
-      this.denyInsecure = validateFeatures(this.args.denyInsecure);
-      const globalDeniedFeatures = filterInsecureFeatures(this.denyInsecure);
-      if (!_.isEmpty(globalDeniedFeatures)) {
-        logger.info('Explicitly disabling insecure features:');
-        globalDeniedFeatures.map((a) => logger.info(`    ${a}`));
-      }
+    if (_.isEmpty(this.args.denyInsecure)) {
+      return;
     }
+    this.denyInsecure = validateFeatures(this.args.denyInsecure);
+    const globalDeniedFeatures = filterInsecureFeatures(this.denyInsecure);
+    if (_.isEmpty(globalDeniedFeatures)) {
+      return;
+    }
+    logger.info('Explicitly disabling insecure features:');
+    globalDeniedFeatures.forEach((a) => logger.info(`    ${a}`));
   }
 
   sessionExists(sessionId) {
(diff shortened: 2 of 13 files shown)


packages/types/lib/driver.ts
```diff
@@ -183,13 +183,29 @@ export interface IExecuteCommands {
...
   capabilities: DriverCaps<C>;
 }
 
 /**
- * Data returned by {@linkcode ISessionCommands.getSession}.
+ * Data returned by {@linkcode ISessionHandler.getAppiumSessions}.
+ *
+ * @typeParam C - The driver's constraints
+ */
+export interface DatedMultiSessionData<C extends Constraints = Constraints> {
```
mykola-mokhnach, line 201: not sure about the Dated prefix, sounds confusing to me

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -41,6 +41,8 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
 
   sessionId: string | null;
 
+  sessionCreationTime: number;
```
mykola-mokhnach, line 44: it's rather a `timestamp`

Also it makes sense to add a measurement unit (seconds or milliseconds ) either into the var name or as a comment

packages/appium/lib/appium.js
```diff
@@ -188,16 +188,28 @@ class AppiumDriver extends DriverCore {
...
+  /**
+   * Retrieve information about all active sessions
+   * @returns {Promise<import('@appium/types').DatedMultiSessionData[]>}
+   */
   async getAppiumSessions () {
-    throw new errors.NotImplementedError('Not implemented yet. ' +
-      'Please check https://github.com/appium/appium/issues/20880 for more details.');
+    return _.toPairs(this.sessions).map(([id, driver]) => ({
+      id,
+      created: driver.sessionCreationTime,
```
mykola-mokhnach, line 210: `created` sounds link a boolean. How about createdAt or createdTimestamp?

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -358,9 +365,26 @@ export class BaseDriver<
     return ret;
   }
 
+  /**
+   * Returns the session id, creation time and capabilities for the session
+   */
+  async getAppiumSessions() {
+    const ret: DatedMultiSessionData<C>[] = [];
```
mykola-mokhnach, line 372: I'm not sure about this implementation. I assume the endpoint must be handled by the umbrella driver instead if we want to list all sessions.

packages/appium/lib/appium.js
```diff
@@ -188,16 +188,28 @@ class AppiumDriver extends DriverCore {
...
   }
 
+  /**
+   * Retrieve information about all active sessions
+   * @returns {Promise<import('@appium/types').DatedMultiSessionData[]>}
+   */
   async getAppiumSessions () {
-    throw new errors.NotImplementedError('Not implemented yet. ' +
-      'Please check https://github.com/appium/appium/issues/20880 for more details.');
+    return _.toPairs(this.sessions).map(([id, driver]) => ({
```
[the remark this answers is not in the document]
eglitise: True, I haven't looked at this part yet. It's actually a bit confusing since we now need to use driver name prefixes, but this particular feature works on the server, not driver level. I guess the only way is to use the `*` prefix then?

packages/base-driver/lib/basedriver/core.ts
eglitise: Changed to `sessionCreationTimestampMs`

packages/appium/lib/appium.js
eglitise: I kind of agree, but I'm not very keen on changing it now that the Inspector already includes checks for the `created` field (https://github.com/appium/appium-inspector/pull/1927). This name was also specified in #20880.

packages/base-driver/lib/basedriver/driver.ts
eglitise: I got confused by the types. Looks like there is currently no interface for methods that are only implemented in `AppiumDriver` and not `BaseDriver`. I have now removed `getAppiumSessions` from `ISessionHandler` and `BaseDriver`, and only left it in `AppiumDriver`.

packages/appium/lib/appium.js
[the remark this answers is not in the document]
mykola-mokhnach: actually the umbrella driver also has a name (AppiumDriver), but * would make more sense to me

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -345,6 +346,11 @@ export class BaseDriver<
 
     return [this.sessionId, caps] as CreateResult;
   }
+
+  /**
+   * Returns the session id and capabilities for the session
+   * @deprecated Use AppiumDriver.getAppiumSessions instead for getting the session data.
+   */
   async getSessions() {
     const ret: MultiSessionData<C>[] = [];
```
mykola-mokhnach, line 355: I'm not even sure this code is used anywhere. Shell we just delete it for good?

packages/types/lib/driver.ts
```diff
@@ -183,13 +183,29 @@ export interface IExecuteCommands {
...
 
 /**
- * Data returned by {@linkcode ISessionCommands.getSession}.
+ * Data returned by `AppiumDriver.getAppiumSessions`
+ *
+ * @typeParam C - The driver's constraints
+ */
+export interface TimestampedMultiSessionData<C extends Constraints = Constraints> {
+  id: string;
+  created: number;
```
mykola-mokhnach, line 203: maybe add a comment that this property should contain UNIX UTC timestamp in seconds

packages/base-driver/lib/basedriver/driver.ts
eglitise: It's probably due to the `getSessions` type in `ISessionHandler`, I guess it could be removed as well, but since it's removing an exported type, it's safer to do it in the Appium 3 branch

packages/base-driver/test/unit/basedriver/capability.spec.js
```diff
@@ -174,7 +174,7 @@ describe('Desired Capabilities', function () {
       logWarnSpy.called.should.be.true;
 
       let sessionCaps = await d.getAppiumSessionCapabilities();
-      sessionCaps.capabilities?.noReset?.should.eql(true);
+      sessionCaps.capabilities.noReset?.should.eql(true);
```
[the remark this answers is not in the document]
mykola-mokhnach: I understand. I mean the whole assertion might not be executed if the variable is not defined, which we don't want to happen. We still want it to run successfully instead.
As a workaround we may either use ts-ignore/ts-expect comment here or use `!` to provide Typescript with a hint that we require this field to be always defined and want an exception to be raised if it's not

packages/base-driver/lib/basedriver/driver.ts
mykola-mokhnach: fine with me

packages/appium/lib/appium.js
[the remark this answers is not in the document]
eglitise: Added this feature. I had to also implement functionality to apply global-scope features to AppiumDriver, which could be useful in the future.

packages/base-driver/test/unit/basedriver/capability.spec.js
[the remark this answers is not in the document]
eglitise: Added `ts-expect-error` comments

packages/appium/docs/en/guides/security.md
```diff
@@ -27,9 +27,9 @@ Appium when starting it from the command line:
...
-|`--deny-insecure`                      |This flag can likewise be set to a comma-separated list of feature names, or a path to a feature file. Any features listed here will be _disabled_, regardless of whether `--relaxed-security` is set and regardless of whether the names are also listed with `--allow-insecure`.|
+|`--relaxed-security`|Turns on _all_ insecure features (unless blocked by `--deny-insecure`; see below)|
+|`--allow-insecure`|Turns on _only_ specified features. Features can be provided as a comma-separated list of feature names, or in the [Appium Configuration file](./config.md). For example, `--allow-insecure=adb_shell` will cause _only_ the ADB shell execution feature to be enabled. Has no effect when used in combination with `--relaxed-security`.|
```
mykola-mokhnach, line 31: maybe we could also update the examples in docs to include the prefix (adb_shell -> `uiautomator2:adb_shell`)

packages/appium/lib/appium.js
```diff
@@ -188,16 +189,35 @@ class AppiumDriver extends DriverCore {
...
+   * @returns {Promise<import('@appium/types').TimestampedMultiSessionData[]>}
+   */
   async getAppiumSessions () {
-    throw new errors.NotImplementedError('Not implemented yet. ' +
-      'Please check https://github.com/appium/appium/issues/20880 for more details.');
+    // We are using `isFeatureEnabled` instead of `assertFeatureEnabled`,
+    // otherwise users that don't want to expose their session list
+    // will get an error every time someone tries to request them
+    if (!this.isFeatureEnabled(SESSION_DISCOVERY_FEATURE)) {
+      return [];
```
mykola-mokhnach, line 214: we must throw an error here. Consider using assertFeatureEnabled instead

packages/appium/lib/appium.js
```diff
@@ -188,16 +189,35 @@ class AppiumDriver extends DriverCore {
...
+   * Retrieve information about all active sessions.
+   * Results are returned only if the `session_discovery` insecure feature is enabled.
+   * @returns {Promise<import('@appium/types').TimestampedMultiSessionData[]>}
+   */
   async getAppiumSessions () {
-    throw new errors.NotImplementedError('Not implemented yet. ' +
-      'Please check https://github.com/appium/appium/issues/20880 for more details.');
+    // We are using `isFeatureEnabled` instead of `assertFeatureEnabled`,
+    // otherwise users that don't want to expose their session list
+    // will get an error every time someone tries to request them
```
mykola-mokhnach, line 212: this is ok, otherwise it won't be possible to know if the feature needs to be enabled or we just have no active session

packages/appium/docs/en/guides/security.md
eglitise: I'd prefer to do this in the Appium 3 branch where the scope is mandatory, since it would also require removing the driver scope security section and integrating it into the rest of the document

packages/appium/lib/appium.js
mykola-mokhnach: that's ok. We may also prevent the call from being logged multiple times in a row, but it feels like a preliminary optmization to me. 
wdyt @KazuCocoa ?

KazuCocoa: Returning an error for the not-enabled case is preferred for me, as same as other security flags to tell callers they should enable it to use the endpoint.
We don't recommend exposing Appium to a public internet as-is. Cloud vendors can filter out the specific endpoint in the front server of their appium if they want to avoid returning errors. Then, the error amount's concern is probably for users who use appium locally or in their private network. Then, calling the endpoint repeatedly could be accidental by internal members

eglitise: Alright, I have changed it to raise an error - but this revealed an issue where the feature validation is only done upon session creation, rather than upon server launch. Looks like this will take a bit more work.

packages/appium/lib/appium.js
```diff
@@ -151,6 +153,37 @@ class AppiumDriver extends DriverCore {
...
+        `Enabling relaxed security. All insecure features will be ` +
+          `enabled unless explicitly disabled by --deny-insecure`,
+      );
+      this.relaxedSecurityEnabled = true;
+    } else if (!_.isEmpty(this.args.allowInsecure)) {
+      this.allowInsecure = validateFeatures(this.args.allowInsecure);
+      const globalAllowedFeatures = filterInsecureFeatures(this.allowInsecure);
+      if (!_.isEmpty(globalAllowedFeatures)) {
+        logger.info('Explicitly enabling insecure features:');
+        globalAllowedFeatures.map((a) => logger.info(`    ${a}`));
```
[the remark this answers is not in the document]
mykola-mokhnach: I agree to Copilot here. A "classic" for-of/forEach would make sense here

packages/appium/lib/appium.js
```diff
@@ -151,6 +153,37 @@ class AppiumDriver extends DriverCore {
...
+      }
+    }
+    if (!_.isEmpty(this.args.denyInsecure)) {
```
mykola-mokhnach, line 177: consider flipping the condition to make have short-circuit return here

packages/appium/lib/appium.js
```diff
@@ -151,6 +153,37 @@ class AppiumDriver extends DriverCore {
     return false;
   }
 
+  /**
+   * Configures insecure features according to the values in `args.relaxedSecurityEnabled`,
+   * `args.allowInsecure`, and `args.denyInsecure`, and informs the user about any
+   * globally-applied features.
+   * Uses `logger` instead of `this.log` to reduce user confusion.
```
mykola-mokhnach, line 160: why using this.log is confusing?

packages/appium/lib/appium.js
```diff
@@ -327,18 +374,31 @@ class AppiumDriver extends DriverCore {
...
         this.log.info(
-          `Applying relaxed security to '${InnerDriver.name}' as per ` +
-            `server command line argument. All insecure features will be ` +
-            `enabled unless explicitly disabled by --deny-insecure`,
+          `Enabling relaxed security for this session as per the server configuration. ` +
+            `All insecure features will be enabled unless explicitly disabled by --deny-insecure`,
         );
         driverInstance.relaxedSecurityEnabled = true;
       }
+      const allowedDriverFeatures = filterInsecureFeatures(this.allowInsecure, driverName);
```
mykola-mokhnach, line 386: maybe it would make sense to extract this block into a separate method to improve readability

packages/appium/lib/appium.js
eglitise: It's a minor thing, but when the server is launched, all the default log messages have the `[Appium]` prefix. Using `this.log` would cause these messages to use the `[AppiumDriver@<hash>]` prefix.

mykola-mokhnach: it's not the only log line that looks like that. Not sure why it should be confusing

eglitise: From what I can tell, it would be the only one that has this prefix out of the lines shown upon server launch. Once a session is created, the log is filled with all sorts of prefixes, including this one, but all lines in the default state (without a session) only have the `[Appium]` prefix.

mykola-mokhnach: maybe then it might be a good idea to isolate this security-related stuff in a separate mixin module, similarly to https://github.com/appium/appium/blob/master/packages/appium/lib/bidi-commands.ts  instead of putting helpers to utils and to the driver main module?

mykola-mokhnach (approved):
Nicely done @eglitise 
I think it's ok to merge the PR is the current state if you are happy about it.

packages/appium/lib/appium.js
eglitise: Moved them to `insecure-features.ts`. It could still be improved by reducing duplication, though.

jlipps (approved):

KazuCocoa (approved):

PR #21240, opened 2025-05-09
chore(base-driver): Tune the deprecation warning message

mykola-mokhnach:
Currently the message mentions the internal method name instead of mentioning the actual METHOD PATH combination. This leads to unnecessary confusion, because we actually deprecate APIs rather than method handlers. For example, let say we have the legacy GET /log/types API, which we'd like to replace with GET /se/log/types one. Currently we mention the name of the method, which is `getLogTypes` in the warning message even if this particular method must still remain, We only want to deprecate one of REST APIs referencing that method (since both the new and the old one rely on the same method)


eglitise (approved):
I'd probably use `endpoint` instead of `API`, but this is much clearer regardless.

KazuCocoa (approved):

PR #21245, opened 2025-05-10
fix(base-driver): Drop unknown method arguments

mykola-mokhnach:
Related to https://github.com/appium/ruby_lib_core/actions/runs/14946904200/job/41991386002?pr=611

The PR fixes the unexpected driver behavior where it was throwing an error for unknown method arguments instead of dropping them with a log message.

### DIFF
--- a/packages/base-driver/lib/protocol/protocol.ts
+++ b/packages/base-driver/lib/protocol/protocol.ts
@@ -155,6 +155,13 @@ export function checkParams(
   // go through the required parameters and check against our arguments
   let matchedReqParamSet: string[] = [];
   for (const requiredParamsSet of requiredParams) {
+    if (!_.isArray(requiredParamsSet)) {
+      log.warn(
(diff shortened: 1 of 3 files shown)


packages/base-driver/lib/protocol/protocol.ts
```diff
@@ -140,25 +147,36 @@ export function checkParams(paramSets: PayloadParams, jsonObj: any, protocol?: k
     optionalParams.push('id');
   }
 
+  if (_.isEmpty(requiredParams)) {
+    // if we don't have any required parameters, then just filter out unknown ones
+    return pickKnownParams(args, _.difference(actualParamNames, [], optionalParams));
```
eglitise, line 152: The second argument can be simplified to `_.difference(actualParamNames, optionalParams)`

mykola-mokhnach: true, changed

KazuCocoa (approved):

PR #21264, opened 2025-05-14
chore: Bump latest changes from master

mykola-mokhnach:


packages/universal-xml-plugin/test/unit/plugin.spec.js
```diff
@@ -1,4 +1,4 @@
-import {UniversalXMLPlugin} from '../..';
```
eglitise, line 1: Wonder how this one was working before 🤔

eglitise (approved):

packages/universal-xml-plugin/test/unit/plugin.spec.js
mykola-mokhnach: I assume it was using the default index import. Nevertheless explicit is always better than implicit and linter was also complaining about it

KazuCocoa (approved):

PR #21265, opened 2025-05-17
chore(appium): Exclude engines from the list of package fields to sync

mykola-mokhnach:


KazuCocoa (approved):

eglitise:
I'm not sure why this was needed...? Isn't this exactly the field that we would want to sync for Appium 3?

mykola-mokhnach:
> I'm not sure why this was needed...? Isn't this exactly the field that we would want to sync for Appium 3?

I would like to avoid the situation where the main manifest has the minimum version set, but some packages don't have their major version bumped yet, and still get their engines updated. I expect that not all packages in the monorepo  would have their major version bumped after Appium3 release happens.

eglitise:
I was under the assumption that all packages would get a major version bump, if only due to bumping the engines (which would naturally be a breaking change). I don't see a reason to keep any package with the old engines, as this would simply increase the maintenance efforts with zero benefits.

mykola-mokhnach:
> I was under the assumption that all packages would get a major version bump, if only due to bumping the engines (which would naturally be a breaking change). I don't see a reason to keep any package with the old engines, as this would simply increase the maintenance efforts with zero benefits.

From the pure maintenance cost perspective it is ok to keep various subpackages in their current state as we can always bump them separately. Also bumping all packages in monorepo at once might be more error-prone.

jlipps:
What's the sort of error you could imagine popping up @mykola-mokhnach? Like some random package depending on a subpackage in our monorepo, that we didn't intend to force a node version bump on? Or requiring work in downstream packages to incorporate a major version change manually? IMO if we're marking it as a major bump, there's not a lot of errors that could sneak in...?

mykola-mokhnach:
> What's the sort of error you could imagine popping up @mykola-mokhnach? Like some random package depending on a subpackage in our monorepo, that we didn't intend to force a node version bump on? Or requiring work in downstream packages to incorporate a major version change manually? IMO if we're marking it as a major bump, there's not a lot of errors that could sneak in...?

All I want to reduce the size of the scope. The previous experience confirmed that it is more beneficial and less error-prone to do multiple smaller changes rather than a big one. Also smaller changes can be distributed better.

PR #21315, opened 2025-06-03
ci: Enable automatic sync of translated docs

mykola-mokhnach:


KazuCocoa (approved):

eglitise:
Wouldn't this result in the removal of all current Chinese translations, since they aren't migrated to Crowdin yet?

mykola-mokhnach:
> Wouldn't this result in the removal of all current Chinese translations, since they aren't migrated to Crowdin yet?

It will, but this is also the original intention to encourage people to do translations in Crowdin. Otherwise by having the current outdated content we may never reach a properly translated state.

KazuCocoa:
Enabling this is fine to me. If we merge all, or get rid of ch for a while from the automatically created PR will depend on our review.

I personally think we'll need to create a separate PR only with ch resources and ping accounts who added them to check if the Crowdin is fine. I was thinking to create such a PR, but not yet though.

mykola-mokhnach:
> Enabling this is fine to me. If we merge all, or get rid of ch for a while from the automatically created PR will depend on our review.
> 
> I personally think we'll need to create a separate PR only with ch resources and ping accounts who added them to check if the Crowdin is fine. I was thinking to create such a PR, but not yet though.

I assume under "separate PR" you mean importing existing translations to Crowdin, because translated documents would anyway be overridden after the next sync.

I would expect that the proofreading process in Crowdin is more convenient for an end-user.

KazuCocoa:
Yea. Some pages should match with this repo's existing ones, like the past JP resources. So, for a while, possibly we need to drop ch resources from the automatically created ones, but this doesn't block this enabling schedule pr.

eglitise:
Fair enough. At the very least, all of the existing Chinese translations will still be available on the `appium2` branch, and can then be used by Crowdin contributors.

eglitise (approved):

PR #21391, opened 2025-06-27
Remove JWP status

AutomatedTester:
## Proposed changes

This removes the JWP status field and cleans up some other related code.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have signed the CLA

## Further comments


packages/base-driver/lib/index.js
```diff
@@ -10,30 +10,22 @@ try {
 }
 
 // BaseDriver exports
-export {ExtensionCore} from './basedriver/extension-core';
-import {BaseDriver} from './basedriver/driver';
-export {DriverCore} from './basedriver/core';
-export {DeviceSettings} from './basedriver/device-settings';
+export { ExtensionCore } from './basedriver/extension-core';
```
mykola-mokhnach, line 13: Please do not mix styling changes with actual PR changes. Extract them into a separate PR is possible

packages/base-driver/lib/jsonwp-proxy/protocol-converter.js
```diff
@@ -1,317 +0,0 @@
-import _ from 'lodash';
```
mykola-mokhnach, line 1: unfortunately we cannot delete this proxy as there are still some subdrivers, like windows one or ios-webview client that use JWP protocol. The first one is just abandoned by MS and the second one requires major atom updates to fully support w3c

mykola-mokhnach (changes requested):
unfortunately I don't see how these jwp leftovers could be removed now taking current limitations into account

eglitise:
Maybe it's possible to move the JWP code into a new plugin, or a new package, that would be auto-installed for extensions that require JWP?
(non-plugin package would probably be better, since it wouldn't need to be explicitly activated when launching the server)

packages/base-driver/lib/index.js
AutomatedTester: Sorry about that, it's the tooling doing it automatically

AutomatedTester:
I can move this out into different package. if people want.

For the iOS-webview, I am pretty sure that we can get rid of the atoms for nearly everything. For visibility we need Safari 17.4+ for  [element checkvisibility](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility).

Are people still using windows one **and** updating the main server? I doubt that people would be mixing versions like that without other weird bugs appearing

mykola-mokhnach:
JSONWP removal has been addressed in appium4 branch and will be released in the upcoming beta

PR #21394, opened 2025-06-28
chore!: set minimum Node.js version to v20.19.0

eglitise:
BREAKING CHANGE: Appium now requires Node.js in the version range `^20.19.0 || ^22.12.0 || >=24.0.0`
BREAKING CHANGE: Appium now requires `npm` version `>=10`

This PR only changes the `node` and `npm` fields according to the above. Versions for the current pre-release packages are unchanged, since it is possible to instruct Lerna to do this.


mykola-mokhnach:
Thanks @eglitise for changing it. I assume lerna is not going to bump the major version of the `appium` package after we merge it (e.g. 3.x.x -> 4.x.x). Otherwise we probably don't need to mark this PR as breaking, at least not for base driver, appium, etc. packages, which already have their major versions bumped.

mykola-mokhnach (approved):

KazuCocoa (approved):

PR #21416, opened 2025-07-11
fix(base-driver,types): For extension commands that use prefix "mobile", ensure …

pujagani:
…logEvents() has the name of executed script

Related to https://github.com/appium/java-client/issues/2219

## Proposed changes

When the client tries to retrieve event logs for extension commands (commands prefixed with "mobile:"), which essentially calls the execute script command, the event logs reflect the command name as "execute" instead of the script name. 

Example:

When calling: driver.activateApp()

The event logs reflect this call as "execute" instead of "activateApp".

The changes fix this by renaming the command in case it is prefixed with "mobile:".

## Types of changes

## Checklist

- [X] I have read the [Contributing Guide](https://appium.io/docs/en/latest/contributing/)
- [X] I have signed the CLA

## Further comments


packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -161,6 +161,18 @@ export class BaseDriver<
 
     // log timing information about this command
     const endTime = Date.now();
+
+    if (cmd === 'execute') {
+      const firstArg = args?.[0];
+      const mobileCommandMatch = typeof firstArg === 'string' && firstArg.startsWith('mobile:')
```
mykola-mokhnach: this logic should not be in the base driver as the `mobile:` prefix is not common. Different drivers might have different prefixes. A proper way to address this issue would be to extract event logging logic to a separate protected method with a default implementation. This method should be overridden/extended in appropriate drivers inherited from BaseDriver, like xcuitest one.

mykola-mokhnach (changes requested):
see the comment above

pujagani:
I have addressed the comments.

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -171,6 +174,11 @@ export class BaseDriver<
     return res;
   }
 
+  protected modifyCommandName(cmd: string, args: any[]): string {
```
mykola-mokhnach: how about clarifyCommandName?

packages/base-driver/test/e2e/protocol/mock-execute-driver.js
```diff
@@ -0,0 +1,29 @@
+
+import {BaseDriver} from '../../../lib';
+import {PROTOCOLS} from '../../../lib/constants';
+
+class MockExecuteDriver extends BaseDriver {
```
mykola-mokhnach, line 5: You may use the existing FakeDriver module for this purpose, just add the necessary methods there

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -161,6 +161,9 @@ export class BaseDriver<
 
     // log timing information about this command
     const endTime = Date.now();
+
+    cmd = this.modifyCommandName(cmd, args);
```
mykola-mokhnach: I was thinking about the implementation and it looks like there is a way to retrieve a name of an execute extension. You may check the static `executeMethodMap`  property of the driver. It should contain commands mapping as in https://github.com/appium/appium-xcuitest-driver/blob/master/lib/execute-method-map.ts
I assume we could use the `command` property there to fetch the actual name, so then we don't need to modify inherited drivers

packages/base-driver/test/e2e/protocol/mock-execute-driver.js
pujagani: FakeDriver has a implementation of "executeCommand" that is different than what is required for the test here hence created a new driver.

packages/base-driver/lib/basedriver/driver.ts
pujagani: Thank you for that. That sounds good. Will keep in mind when making changes in the xcuitest-driver once these changes are merged.

mykola-mokhnach: my point is that if this approach is implemented then it won't be necessary to do any changes in xcuitest driver

pujagani: Ahh I understand better now. Apologies. Thank you so much for clarifying that. I will revisit this PR with the approach suggested.

packages/base-driver/lib/helpers/extension-command-name.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Resolves the name of extension method corresponding to an `execute` command string
+ * based on the driver's `executeMethodMap`.
+ *
+ * @param driverInstance - The instance of the driver which may have a command mapping.
+ * @param commandName - The command name to resolve.
+ * @returns The resolved extension command name if a mapping exists. Otherwise, the original command name.
+ */
+export function resolveExecuteExtensionName(driverInstance: unknown, commandName: string): string {
+  const DriverClass = Object.getPrototypeOf(driverInstance)?.constructor;
+  const methodMap = (DriverClass as any)?.executeMethodMap;
```
mykola-mokhnach: I believe we could use better type declarations here rather than unknown. Check, for example, https://github.com/appium/appium/blob/70f22a4b9f66a53da266991f9cdc5b9cb8ab49a4/packages/base-driver/lib/basedriver/commands/execute.ts#L26

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -162,7 +163,12 @@ export class BaseDriver<
     // log timing information about this command
     const endTime = Date.now();
 
-    cmd = this.clarifyCommandName(cmd, args);
+    if (cmd === 'execute') {
+      const firstArg = args?.[0];
+      if (typeof firstArg === 'string' && firstArg.includes(':')) {
```
mykola-mokhnach: we must not make any assumptions about whether an argument contains a colon.checking if it's a string (and maybe not empty) should suffice

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -162,7 +163,12 @@ export class BaseDriver<
     // log timing information about this command
     const endTime = Date.now();
 
-    cmd = this.clarifyCommandName(cmd, args);
```
mykola-mokhnach: it would still make sense to keep this logic in a separate method just in case we'd need to patch it in descendant classes

packages/base-driver/lib/helpers/extension-command-name.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  if (methodMap && typeof methodMap === 'object' && commandName in methodMap) {
```
mykola-mokhnach: `typeof methodMap === 'object'` -> consider replacing with `_.isPlainObject(methodMap)`

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -171,6 +175,17 @@ export class BaseDriver<
     return res;
   }
 
+  protected clarifyCommandName(cmd: string, args: string[]) {
```
mykola-mokhnach: Please also provide the return type

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -171,6 +175,17 @@ export class BaseDriver<
...
+    if (cmd === 'execute') {
+      const firstArg = args?.[0];
+      if (typeof firstArg === 'string' && firstArg.trim().length > 0) {
```
mykola-mokhnach: nit: I find lodash's _.isString more flexible in comparison to typeof

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -171,6 +175,17 @@ export class BaseDriver<
...
+        cmd = resolveExecuteExtensionName.call(this, firstArg);
```
mykola-mokhnach: there is no need to reassign the function arg. Simply return it in place

packages/base-driver/lib/helpers/extension-command-name.ts
```diff
@@ -0,0 +1,27 @@
+import _ from 'lodash';
+import {Constraints, Driver, DriverClass} from '@appium/types';
```
mykola-mokhnach: use `import type` to import types

packages/base-driver/lib/helpers/extension-command-name.ts
```diff
@@ -0,0 +1,27 @@
...
+import BaseDriver from '../basedriver/driver';
```
mykola-mokhnach: consider replacing with `import type {BaseDriver} from '../basedriver/driver';`

mykola-mokhnach (approved):

mykola-mokhnach:
Unfortunately I could observe some [test failures](https://github.com/appium/appium/actions/runs/16717325815/job/47346892579?pr=21416). @pujagani could you please take a look at them?

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -171,6 +175,17 @@ export class BaseDriver<
     return res;
   }
 
+  protected clarifyCommandName(cmd: string, args: string[]): string {
```
KazuCocoa: just a small nits:
Wondered if we could move this method into `lib/helpers/extension-command-name.ts` (or resolveExecuteExtensionName can include this `cmd` check part also) as this has no `this` usage

pujagani: My earlier commit did have that. But the goal is to have this method here to allow the implementing drivers to override it if required from what I understand.

KazuCocoa: Got it. I didn't notice that the intentionally allow subclasses to override it

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -163,7 +163,9 @@ export class BaseDriver<
     // log timing information about this command
     const endTime = Date.now();
 
-    cmd = this.clarifyCommandName(cmd, args);
+    if (_.isFunction((this as any).clarifyCommandName)) {
+      cmd = (this as any).clarifyCommandName(cmd, args);
```
KazuCocoa: I think we can define

`clarifyCommandName?(cmd: string, args: string[]): string;`

in the type definition, then here can avoid “as any” I assume

KazuCocoa: https://github.com/appium/appium/blob/3b11b102f4ccb27473b703069e7180bea3706d2c/packages/types/lib/driver.ts#L759 Like this

KazuCocoa (approved):

Issue #21453, opened 2025-08-01
🐞 bug: appium/support uses unmaintained 'mv' package which uses outdated/unsupported rimraf package


BryceLTaylor:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Appium Version

2.19.0

### Appium Host Type

Command Line

### Is this issue reproducible using the latest components?

- [x] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

running npm install with `appium-uiautomator2-driver` and/or `appium-xcuitest-driver` causes a warning `npm warn deprecated rimraf@2.4.5: Rimraf versions prior to v4 are no longer supported` 

This is because both use the package `@appium/support`.  `@appium/support` uses the `mv` npm package which has not been updated in 10 years.  `mv` uses an outdated version of the `rimraf` package, causing the warning listed above.  [This issue was filed](https://github.com/andrewrk/node-mv/issues/34) on the mv package a year ago to update rimraf, but besides a few thumbs up there is no response.

### Expected Behavior

No unmaintained packages are used, especially ones that throw errors or warnings when running npm install.  The `mv` package should be removed or replaced.

### Environment Details

- Debug details (`appium --show-debug-info`):
- Last component version(s) which did _not_ exhibit the problem:
- Platform and version of device under test:
- Real device or emulator/simulator:

### Minimal Reproducible Example

Using npm version 11.4.2 include `appium-uiautomator2-driver` and/or `appium-xcuitest-driver` as a dependency (or dev dependency) in your package.json and run npm install.  Note warning `npm warn deprecated rimraf@2.4.5: Rimraf versions prior to v4 are no longer supported`

### Further Information

This issue has been mentioned as part of other issues in a few issues that are now closed: https://github.com/appium/appium/issues/6286  lists many packages and was closed after a partial fix.  https://github.com/appium/appium/issues/20154 was about another package, but someone mentions that `mv` was a root cause in the comments.

mykola-mokhnach:
Thanks for the report. I am going to replace the deprecated package soon

mykola-mokhnach:
The PR has been merged. Will be available after the next version publish

mykola-mokhnach:
Appium 3 has been published

PR #21454, opened 2025-08-02
feat(support): Drop the usage of the obsolete "mv" package

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/21453


packages/support/lib/fs.js
```diff
@@ -404,10 +445,16 @@ export default fs;
...
 /**
  * @typedef {import('glob')} glob
- * @typedef {import('mv')} mv
  * @typedef {import('fs').PathLike} PathLike
  */
 
+/**
+ * @typedef {Object} mv
+ * @property {boolean} [mkdirp=false] Whether to automatically create the destination folder structure
+ * @deprecated @property {boolean} [clobber=false] Legacy deprecated property, not used anymore
```
eglitise: Seems like `mv` had the default set to `clobber=true`, so changing the default to `false` is technically a breaking change. Could we add a test for the expected behavior if the destination already exists?

mykola-mokhnach: Good catch. Updated the code and tests

eglitise (approved):

Issue #21457, opened 2025-08-05
🐞 bug: RangeError: Maximum call stack size exceeded witb BiDi logs


KazuCocoa:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Appium Version

current master: 7103126a1bc57bf346b4605548e2978dabce2014

### Appium Host Type

Command Line

### Is this issue reproducible using the latest components?

- [x] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

Run BiDi log execution such as https://github.com/appium/python-client/pull/1154

Then, the `Maximum call stack size exceeded` error occurred below.

```
RangeError: Maximum call stack size exceeded
    at Log.<anonymous> (/Users/kazu/github/appium/packages/appium/lib/logsink.js:97:17)
    at Log.emit (node:events:536:35)
    at Log.log (/Users/kazu/github/appium/packages/logger/lib/log.ts:259:10)
    at Log.debug (/Users/kazu/github/appium/packages/logger/lib/log.ts:179:10)
    at Object.wrappedLogger.<computed> [as debug] (/Users/kazu/github/appium/packages/support/lib/logging.js:72:22)
    at EventEmitter.eventListener (/Users/kazu/github/appium/packages/appium/lib/bidi-commands.ts:484:20)
    at EventEmitter.emit (node:events:524:28)
    at Log.listener (/Users/kazu/.appium/node_modules/appium-uiautomator2-driver/node_modules/appium-android-driver/lib/commands/log.js:175:23)
    at Log.emit (node:events:536:35)
    at Log.log (/Users/kazu/github/appium/packages/logger/lib/log.ts:259:10)
    at Log.debug (/Users/kazu/github/appium/packages/logger/lib/log.ts:179:10)
    at Object.wrappedLogger.<computed> [as debug] (/Users/kazu/github/appium/packages/support/lib/logging.js:72:22)
    at EventEmitter.eventListener (/Users/kazu/github/appium/packages/appium/lib/bidi-commands.ts:484:20)
    at EventEmitter.emit (node:events:524:28)
    at Log.listener (/Users/kazu/.appium/node_modules/appium-uiautomator2-driver/node_modules/appium-android-driver/lib/commands/log.js:175:23)
    at Log.emit (node:events:536:35)
    at Log.log (/Users/kazu/github/appium/packages/logger/lib/log.ts:259:10)
    at Log.debug (/Users/kazu/github/appium/packages/logger/lib/log.ts:179:10)
[...]

PR #21461, opened 2025-08-05
fix(appium): Avoid recursive log calls in BiDi event listeners

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/21457

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24741,7 +24741,7 @@
       "resolved": "https://registry.npmjs.org/ajv-formats/-/ajv-formats-3.0.1.tgz",
       "integrity": "sha512-8iUql50EUR+uUcdRQ3HDqa6EVyo3docL8g5WJ3FNcWmu62IbkGUue/pEyLBW8VGKKucTPgqeks4fIU1DA4yowQ==",
       "requires": {
-        "ajv": "^8.0.0"
+        "ajv": "8.17.1"
       }
     },
     "ansi-colors": {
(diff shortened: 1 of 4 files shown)


KazuCocoa (approved):

KazuCocoa:
I'll try this out on my local also. I'll leave a new comment here after testing

KazuCocoa:
In this case, nothing was emitted so no `BIDI EVENT` related logs was logged. I mean nothing was logged like `[AndroidUiautomator2Driver@6da0] <-- BIDI EVENT log.entryAdded (context: 'NATIVE_APP', params: {"type":"server","level":"info","source":{"realm":""},"text":"[AndroidUiautomator2Driver@6da0] <-- BIDI EVENT log.entryAdded (context: 'NATIVE_APP', params: {\"type\":\"server\",\"level\":\"info\",\"source\":{\"realm\":\"\"},\"text\":\"[AndroidUiautomator2Driver@6da0] <-- BIDI EVENT log.entryAdde...). All further similar events won't be logged.`. Expected one is... we want to print the log once (without recursive call)


Or... if this implementation wants to avoid logging the log (with `markNonEmittable()`) itself, this works expectedly while then maybe we have no way to log logs with `markNonEmittable()`...? (Then, simply we probably could remove the log itself)

KazuCocoa:
mmm, it's wired, but once I rebooted my host machine, the issue https://github.com/appium/appium/issues/21457 didn't occur with current master. But then, BiDi logs didn't logs further

> [42a99c02][AndroidUiautomator2Driver@00ae] <-- BIDI EVENT log.entryAdded (context: 'NATIVE_APP', params: {"type":"syslog","level":"info","source":{"realm":""},"text":"08-06 21:36:42.691 17893 17927 I appium  : channel read: GET /session/4cb24f5c-c8cd-4400-912b-8a1d8455cc0a/source","timestamp":1754483861855}). All further similar events won't be logged.

KazuCocoa:
So... so far, I think we can just close this as my local machine state was bad

PR #21477, opened 2025-08-10
chore: bump all non-rc packages to rc

eglitise:
Replaces #21476.

* Bump the major version of all non-RC packages to `rc.0`
* Bump the versions of all Appium dependencies to the next major
* Set the published tag for all packages to `rc`
* For packages using Appium 3 as a peer dependency, ~~bump it to `rc.1`, since there were a few notable changes since `beta.1`~~ drop the Appium 2 compatibility

After this PR, running a publish on the entire monorepo will release all the packages as `rc.1`, except the 3 that are already published in RC, which will be released as `rc.2`.


packages/plugin-test-support/package.json
```diff
@@ -41,22 +41,23 @@
...
-    "@appium/types": "^0.26.0",
+    "@appium/types": "^1.0.0-rc.0",
     "get-port": "5.1.1",
     "log-symbols": "4.1.0",
     "source-map-support": "0.5.21",
     "teen_process": "2.3.3"
   },
   "peerDependencies": {
-    "appium": "^2.0.0-beta.43 || ^3.0.0-beta.0",
+    "appium": "^2.0.0-beta.43 || ^3.0.0-rc.1",
```
mykola-mokhnach: I'd say we keep `^3.0.0-beta.0` everywhere for better compatibility and remove 2.0.0

eglitise: We should still keep 2.0.0 if it's still compatible, though the question is how should we test that 🤔

mykola-mokhnach: Appium 2 is not maintained anyway. I don't see a point in keeping it after bumping major versions of corresponding packages

mykola-mokhnach (approved):

KazuCocoa (approved):

PR #21535, opened 2025-08-30
docs(appium): enhance CLI pages

eglitise:
This is an enhancement to the CLI documentation, making it more consistent and hopefully also easier to understand.

Additional changes in terms of the included information:
* Env vars: removed unused `APPIUM_PREFER_SYSTEM_UNZIP`, `APPIUM_HOST` and `APPIUM_PORT`, added `APPIUM_STORAGE_KEEP_ALL` and `APPIUM_STORAGE_ROOT` in a separate section
* Args: renamed to Server, added missing `--shutdown-timeout` option, removed unused `--trace-dir` option, moved `--help` to the index page
* Extensions: added more examples

PR #21540, opened 2025-09-03
docs(appium): enhance API pages

eglitise:
This is a major enhancement to the API documentation:
* Adjust the formatting of all endpoints to follow #21535
* Combine all plugin endpoint pages into one
* Split the base-driver endpoints page into protocol-specific pages (similarly to how WDIO does it)
* Add missing information, update outdated information
* Small fix for JSDoc comments in `@appium/opencv`

Note that I am only including changes for the endpoints that are currently listed in the docs. I am aware that there's a lot more endpoints missing from the docs, but these will be added in future PRs.


packages/appium/docs/en/reference/api/appium.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Parameters
+
+|Name|Description|Type|
+|--|--|--|
+|`settings`|Object containing the names and values of the settings to be updated|object|
+
+#### Response
+
+`null`
```
mykola-mokhnach, line 79: in some driver implementations this commands actually returns all settings together with updated ones

packages/appium/docs/en/reference/api/webdriver.md
```diff
@@ -0,0 +1,269 @@
+---
+title: WebDriver Protocol
+---
+<style>
+  ul[data-md-component="toc"] .md-nav {
+    display: none;
+  }
+</style>
+
+The following is a list of [W3C WebDriver protocol](https://w3c.github.io/webdriver/) endpoints
```
mykola-mokhnach, line 10: I believe there are more endpoints. Should we list them all?

mykola-mokhnach: Maybe we could just provide links to supported parts of the original spec: https://www.w3.org/TR/webdriver2/

eglitise: I plan to add the remaining ones in future PRs. This PR only changes formatting for the endpoints currently listed in the Appium docs.

packages/appium/docs/en/reference/api/appium.md
eglitise: I was unaware of this, but even so, I feel like this documentation should list the default server-provided configuration, and any changes to this configuration should be documented for the drivers that change them.

mykola-mokhnach (approved):

KazuCocoa (approved):
lg

PR #21563, opened 2025-09-08
fix(appium): Return a non-zero exit code if an extension script execution fails

mykola-mokhnach:
Currently it just prints the error message, but the exit code is still zero, which is confusing. See, for example, https://github.com/appium/appium-xcuitest-driver/actions/runs/17533887861/job/49794425622 (step "Prepare Server")


KazuCocoa (approved):
looks reasonable to me

eglitise (approved):

PR #21564, opened 2025-09-08
chore(types): refactor command-related types

eglitise:
This PR is a refactor of types related to driver/plugin commands. Currently the `driver.ts` file is by far the largest type file, making it difficult to understand, and most of its size is taken up by typings for endpoint-related commands.
As part of this PR, all command-related types have been moved to separate files under a new `commands` directory, with a new top-level interface to group the commands of each protocol, similarly to the API endpoints documentation.
As for the commands that already had interfaces due to them being implemented in BaseDriver, I've also grouped them under a new top-level interface, but otherwise left its constituent interfaces and types unchanged.

Other changes include:
* Rename `command.ts` to `command-maps.ts` for clarity
* Move capability constraint-related types from `driver.ts` to `constraints.ts`
* Fix `Driver` interface not extending `IBidiCommands`, by extending them in `IImplementedCommands`

Other than the aforementioned `Driver` interface change, no other type values were changed.


packages/types/lib/driver.ts
```diff
@@ -490,83 +52,8 @@ export type SettingsUpdateListener<T extends Record<string, unknown> = Record<st
   curValue: unknown,
 ) => Promise<void>;
 
-// WebDriver
-
-export interface Rect {
```
mykola-mokhnach, line 495: is it safe to delete exported types from here? Are they being referenced directly anywhere?

eglitise: All of the types have been moved to files in the `commands` directory - none have been deleted. In this case the `Rect` interface has been moved to [`webdriver.ts`](https://github.com/appium/appium/blob/3266f8ddc701e5c264a353786b434e7ce0ffed81/packages/types/lib/commands/webdriver.ts#L453). It is then exported by `commands/index.ts` which in turn is exported by `index.ts`, so all exports remain the same as well.

mykola-mokhnach (approved):

PR #21685, opened 2025-10-29
test: Update integration tests for compatibility with Node24

mykola-mokhnach:
What was done:

- Split tests execution, so it's easier to detect what exactly has failed and there is no need to scan one huge log
- Updated some base driver e2e tests to reflect changes in configureApp that were introduced by Appium3 (not sure why these tests were not failing before)
- Skipped some tests explicitly for node24+. It looks like this node version has some breaking changes that would require to dedicate more time finding possible resolutions. Perhaps, there are no solutions at all and we'd need to just deprecate some functionality (like SPDY-related stuff).
- Replaced usage of `should` to `expect` in some test modules. It does not seem to affect the actual result, but does not hurt anyway, so I ddi not revert them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,20 +78,22 @@ jobs:
             package_json="${package_dir}package.json"
             if [[ -f "$package_json" ]]; then
               # Check if package.json has test:e2e script and it's not a no-op
-              pkg_info=$(node -e "
+              has_e2e=$(node -e "
                 try {
                   const pkg = require('./$package_json');
                   const script = pkg.scripts && pkg.scripts['test:e2e'];
                   if (script && !script.includes('echo') && !script.includes('No e2e tests')) {
-                    const pkgName = pkg.name || '${package_dir%/}';
-                    console.log(pkgName);
+                    console.log('true');
                   }
                 } catch(e) {
                   // Ignore errors
                 }
               ")
-              if [[ -n "$pkg_info" ]]; then
-                packages+=("$pkg_info")
+              if [[ "$has_e2e" == "true" ]]; then
+                # Use the directory path without packages/ prefix (remove trailing slash)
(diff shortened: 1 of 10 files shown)


packages/base-driver/test/e2e/express/server.e2e.spec.js
```diff
@@ -64,26 +65,16 @@ describe('server', function () {
 
   it('should start up with our middleware', async function () {
     const {data} = await axios.get(`http://${TEST_HOST}:${port}/`);
-    data.should.eql('Hello World!');
-  });
-  it('should fix broken context type', async function () {
```
eglitise, line 69: Was this test also failing?

mykola-mokhnach: yep, it verifies the middleware that has been deleted in Appium 3 as obsolete

.github/workflows/build.yml
```diff
@@ -64,19 +64,134 @@ jobs:
...
+          node-version: ${{ matrix.node-version }}
+          cache: 'npm'
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+        with:
+          useRollingCache: true
+          install-command: npm ci --foreground-scripts
+      - name: Build packages
+        run: npm run build
+      - name: Clear V8 code cache (Node.js 24+ compatibility)
```
KazuCocoa: Perhaps we could use `&` and `*` to reduce duplication

https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases

KazuCocoa (approved):
otherwise lg

.github/workflows/build.yml
mykola-mokhnach: refactored

PR #21706, opened 2025-11-06
fix(base-driver): Use shouldUpgradeCallback for proper upgrades processing

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/20760 for newer Node versions

### DIFF
--- a/packages/base-driver/lib/express/server.js
+++ b/packages/base-driver/lib/express/server.js
@@ -14,6 +14,7 @@ import {
   allowCrossDomainAsyncExecute,
   handleIdempotency,
   handleUpgrade,
+  tryHandleWebSocketUpgrade,
   catch404Handler,
   handleLogContext,
 } from './middleware';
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):

mykola-mokhnach:
@pimterry Could you please take a quick look at this PR if your time allows?

packages/base-driver/lib/express/server.js
```diff
@@ -294,29 +340,23 @@ async function startServer({
...
-  // is empty meaning no base path at all
-  if (basePath !== '' && !basePath.startsWith('/')) {
-    basePath = `/${basePath}`;
+function hasShouldUpgradeCallback(server) {
+  // Check if shouldUpgradeCallback is available on http.Server
+  // This is a runtime check that works regardless of TypeScript types
+  try {
+    // Use bracket notation to access property that may not exist in type definitions
+    // eslint-disable-next-line dot-notation
+    return typeof server['shouldUpgradeCallback'] !== 'undefined';
```
pimterry, line 354: I suspect this will work, but that field isn't an officially supported API, so you might find issues with this in future. The 'correct' way to do this would really be to check the Node version and compare to the versions that include this feature.

pimterry (approved):
I don't know anything about the appium internals really so I can't comment on the specifics much, but overall the general structure looks like reasonable usage of `shouldUpgradeCallback` & upgrade events to me, yes.

packages/base-driver/lib/express/server.js
mykola-mokhnach: Eventually (probably in a year) we are anyway going to drop older node versions in package engines, so this check won't be necessary anymore.

Thank you for checking the PR.

PR #21797, opened 2025-12-11
chore(renovate): update config

eglitise:
A few updates for the Renovate config:
* Remove explicit `workarounds:typesNodeVersioning` preset (already included in `config:js-app` -> `config:recommended` -> `workarounds:all`)
* Remove uses of `groupSlug` (Renovate automatically generates the slug from`groupName`)
* Cleanup the ESM-only package ban list by removing the packages not used in this monorepo. Since Appium 3 supports a semver range that allows `require`'ing ESM packages, this blocklist is obsolete - however, the packages that _are_ used may still require additional code changes due to major version updates, so these will be handled separately.


mykola-mokhnach (approved):

mykola-mokhnach:
should the PR prefix be changed to `ci:` ?

eglitise:
I think `chore` is probably more appropriate, given it was also used in the past for similar PRs.

PR #21806, opened 2025-12-11
chore(deps): update pkg-dir to v8 and rename it

eglitise:
Updates `pkg-dir` from `v5` to `v8`:
* `v6` required Node 12, became ESM-only, and changed from default export to named exports
* `v7` required Node 14
* `v8` required Node 18
* `v8.1.0` renamed the package to `package-directory`

### DIFF
--- a/packages/docutils/lib/fs.ts
+++ b/packages/docutils/lib/fs.ts
@@ -6,7 +6,7 @@
 import {fs} from '@appium/support';
 import _ from 'lodash';
 import path from 'node:path';
-import _pkgDir from 'pkg-dir';
+import {packageDirectory} from 'package-directory';
 import readPkg, {NormalizedPackageJson, PackageJson} from 'read-pkg';
 import {JsonValue} from 'type-fest';
 import * as YAML from 'yaml';
@@ -30,7 +30,7 @@ const log = getLogger('fs');
  *
  * Caches result
  */
-const findPkgDir = _.memoize(_pkgDir);
+const findPkgDir = _.memoize(packageDirectory);
 
 /**
  * Stringifies a thing into a YAML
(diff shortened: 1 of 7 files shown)


packages/docutils/lib/fs.ts
```diff
@@ -5,8 +5,9 @@
 
 import {fs} from '@appium/support';
 import _ from 'lodash';
+import crypto from 'node:crypto';
```
mykola-mokhnach: you could use util.uuidV4

mykola-mokhnach (approved):

packages/docutils/lib/fs.ts
```diff
@@ -30,7 +31,7 @@ const log = getLogger('fs');
  *
  * Caches result
  */
-const findPkgDir = _.memoize(_pkgDir);
+const findPkgDir = _.memoize(packageDirectory, (opts) => opts?.cwd ?? crypto.randomUUID());
```
mykola-mokhnach: why random uuid is used as a default here?

packages/docutils/lib/fs.ts
eglitise: I think at some point we should instead just drop all of the `uuidV*` helpers from `util`. The only one of those that's used in the entire org is `uuidV4`, which is equivalent to `crypto.randomUUID`.

packages/docutils/lib/fs.ts
eglitise: I thought about it some more and now removed them. Originally I added them so that Lodash has a cache key to use for storing the result, in case `findPkgDir` is called without an argument. However, neither file exports the memoized function, and inside both files the function is always called with `{cwd}`, so the key will never be `undefined`.

PR #21810, opened 2025-12-12
chore: fix all lint issues

eglitise:
Follow-up to #21781, resolving all the linting issues caused by the updated rules.
This includes replacing instances of the deprecated `url.parse` with `new URL`, although we do use `url.parse` for relative URLs (which is not supported by `new URL`), so this was not possible to apply everywhere. Ideally this should be fixed in the future.


packages/base-driver/lib/basedriver/helpers.js
```diff
@@ -354,12 +353,16 @@ export function generateDriverLogPrefix(obj, sessionId = null) {
...
+    : username
+      ? href.replace(`${username}@`, '')
+      : href;
   /** @type {import('axios').AxiosBasicCredentials|undefined} */
-  const axiosAuth = auth ? {
-    username: auth.substring(0, auth.indexOf(':')),
-    password: auth.substring(auth.indexOf(':') + 1),
+  const axiosAuth = username ? {
+    username,
+    password: password ?? '',
```
eglitise: This could also be considered a bug fix: previously this code was improperly handling the case where the URL only contained the username (without password), which would result in the axios username being empty, and the password set to the parsed username.

KazuCocoa (approved):

Issue #16798, opened 2022-04-18
enhancement: appium driver/plugin list should show package name for known extensions


boneskull:
Running `appium driver list` or `appium plugin list` displays the "short" driver name, which is good, but it'd also be helpful if the package name was also displayed (linked via URL, if supported by terminal).

```
✔ Listing available drivers
- fake@3.2.6 [installed (linked from /Users/chrishiller/projects/appium/packages/fake-driver)]
- uiautomator2 [not installed]
- xcuitest [not installed]
- youiengine [not installed]
- windows [not installed]
- mac [not installed]
- mac2 [not installed]
- espresso [not installed]
- tizen [not installed]
- flutter [not installed]
- safari [not installed]
- gecko [not installed]
```

I wanted to use this list to grab a working copy of e.g., the gecko driver, but I didn't know the package name, so I didn't have easy access to the URL.

If we consider this to be too much information for end users, perhaps a `--details` or `--verbose` flag would be appropriate here; we could list the npm package name and the repository URL?

jlipps:
yeah, agree this should be available somewhere, a --details or --verbose would be a good place to hide it

mykola-mokhnach:
closed as fixed

PR #21813, opened 2025-12-14
feat(appium): Add repository links for extensions list

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/appium/appium/issues/16798

## Types of changes

Github links are now included into both detailed and normal list output. The normal output looks like:

```
- xcuitest@10.1.6 [installed (linked from /Users/test/code/appium-xcuitest-driver)] https://github.com/appium/appium-xcuitest-driver
- gecko@2.0.6 [installed (linked from /Users/test/code/appium-geckodriver)] https://github.com/appium/appium-geckodriver
- uiautomator2@6.0.2 [installed (linked from /Users/test/code/appium-uiautomator2-driver)] https://github.com/appium/appium-uiautomator2-driver
- espresso (appium-espresso-driver) [not installed] https://github.com/appium/appium-espresso-driver
- mac2 (appium-mac2-driver) [not installed] https://github.com/appium/appium-mac2-driver
- windows (appium-windows-driver) [not installed] https://github.com/appium/appium-windows-driver
- safari (appium-safari-driver) [not installed] https://github.com/appium/appium-safari-driver
- chromium (appium-chromium-driver) [not installed] https://github.com/appium/appium-chromium-driver
```

### DIFF
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -218,7 +218,7 @@ class ExtensionCliCommand {
       for (const data of _.values(listData)) {
         const repoUrl = await this._getRepositoryUrl(data);
         if (repoUrl) {
-          /** @type {any} */ (data).repositoryUrl = repoUrl;
+          data.repositoryUrl = repoUrl;
         }
       }
     });
@@ -237,8 +237,8 @@ class ExtensionCliCommand {
     /** @type {Record<string, string>} */
     const repoUrlMap = {};
     for (const [name, data] of _.toPairs(listData)) {
-      if (/** @type {any} */ (data).repositoryUrl) {
-        repoUrlMap[name] = /** @type {any} */ (data).repositoryUrl;
+      if (data.repositoryUrl) {
+        repoUrlMap[name] = data.repositoryUrl;
       }
     }
 
@@ -1235,6 +1235,7 @@ export {ExtensionCliCommand as ExtensionCommand};
  * @property {string|null} unsafeUpdateVersion - Same as above, but a major version bump
  * @property {string} [updateError] - Update check error message (if present)
  * @property {boolean} [devMode] - If Appium is run from an extension's working copy
+ * @property {string} [repositoryUrl] - Repository URL for the extension (if available)
  */
 
 /**
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -189,21 +189,25 @@ class ExtensionCliCommand {
       return;
     }
     await spinWith(this.isJsonOutput, lsMsg, async () => {
-      for (const [ext, data] of _.toPairs(listData)) {
-        if (!data.installed || data.installType !== INSTALL_TYPE_NPM) {
-          // don't need to check for updates on exts that aren't installed
-          // also don't need to check for updates on non-npm exts
-          continue;
-        }
-        try {
-          const updates = await this.checkForExtensionUpdate(ext);
-          data.updateVersion = updates.safeUpdate;
-          data.unsafeUpdateVersion = updates.unsafeUpdate;
-          data.upToDate = updates.safeUpdate === null && updates.unsafeUpdate === null;
-        } catch (e) {
-          data.updateError = e.message;
-        }
-      }
+      // Filter to only extensions that need update checks (installed npm packages)
+      const extensionsToCheck = _.toPairs(listData).filter(
+        ([, data]) => data.installed && data.installType === INSTALL_TYPE_NPM
+      );
+
+      await B.map(
+        extensionsToCheck,
+        async ([ext, data]) => {
+          try {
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):
lg

Double checked a module from npm but not our default list as well:

```
- flutter@3.2.0 [installed (npm)] https://github.com/appium/appium-flutter-driver
- ```

**Comment by eglitise** [comment]
I like the idea of adding the repo URL, but I'm not sure if it needs to be in the default output. IMO it'd be fine to leave the default output as-is (excluding the package name as well), but change the title to something like `Listing available drivers (rerun with --verbose for more info)`. The verbose output (which I assume did not exist at the time of the original issue) should of course include everything.

**Comment by mykola-mokhnach** [comment]
> I like the idea of adding the repo URL, but I'm not sure if it needs to be in the default output. IMO it'd be fine to leave the default output as-is (excluding the package name as well), but change the title to something like `Listing available drivers (rerun with --verbose for more info)`. The verbose output (which I assume did not exist at the time of the original issue) should of course include everything.

I was actually not sure how to do it better. Do you think adding the link to the default output may potentially break anything?

@jlipps Could you please share your input on the above?

**Comment by eglitise** [comment]
I'm not concerned about anything breaking; I just find that it adds clutter and looks a bit out of place in the default output.

**Comment by jlipps** [comment]
I think I agree with @eglitise that this is useful information but better to hide behind a `--verbose` flag (both for visual complexity as well as the time taken to run more network requests)

**Comment by mykola-mokhnach** [comment]
Updated

File: packages/appium/lib/cli/extension-command.js [hunk 1]
```diff
@@ -126,13 +127,39 @@ class ExtensionCliCommand {
...
-    let lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} ${this.type}s`;
+    const listData = this._buildListData(showInstalled);
+
+    const lsMsg =
+      `Listing ${showInstalled ? 'installed' : 'available'} ${this.type}s` +
+      (verbose ? ' (verbose mode)' : '');
+    await this._checkForUpdates(listData, showUpdates, lsMsg);
+
+    // Fetch repository URLs for all extensions (needed for both JSON and display output)
+    await this._addRepositoryUrlsToListData(listData);
```
eglitise: Is it possible to call the fetch action only in verbose mode?

packages/appium/lib/cli/extension-command.js
```diff
@@ -126,13 +127,39 @@ class ExtensionCliCommand {
    * @return {Promise<ExtensionList<ExtType>>} map of extension names to extension data
    */
   async list({showInstalled, showUpdates, verbose = false}) {
-    let lsMsg = `Listing ${showInstalled ? 'installed' : 'available'} ${this.type}s`;
+    const listData = this._buildListData(showInstalled);
+
+    const lsMsg =
+      `Listing ${showInstalled ? 'installed' : 'available'} ${this.type}s` +
+      (verbose ? ' (verbose mode)' : '');
```
eglitise: Suggest adding the `(rerun with --verbose for more info)` part when in non-verbose mode

packages/appium/lib/cli/extension-command.js
```diff
@@ -148,99 +175,256 @@ class ExtensionCliCommand {
...
+   * @param {ExtensionListData<ExtType>} data
+   * @returns {string}
+   * @private
+   */
+  _formatNonInstalledExtensionLine(name, data) {
+    const installTxt = ' [not installed]'.grey;
+    let pkgNameTxt = '';
+
+    if (data.pkgName) {
+      pkgNameTxt = ` (${data.pkgName})`.cyan;
```
eglitise: Since the package name is printed for non-installed extensions, but not for installed extensions, it feels like a bit of a mismatch. I'm not sure if it's even needed, given that it's also present in the verbose output

packages/appium/lib/cli/extension-command.js
mykola-mokhnach: ofc it is

packages/appium/lib/cli/extension-command.js
mykola-mokhnach: I've added it, but the label itself is shown for a very short time, I can barely see it with my internet connection speed

packages/appium/lib/cli/extension-command.js
mykola-mokhnach: updated

packages/appium/lib/cli/extension-command.js
eglitise: Ah, this seems to be because the `if (!showUpdates)` guard was moved outside of the `await spinWith` callback. I suggest moving it back inside the callback, as currently the line is only properly logged with the `--updates` flag, which feels inconsistent.
There is also a missing space before `(rerun`.

mykola-mokhnach: fixed both

eglitise (approved):

PR #21819, opened 2025-12-15
fix(scripts): properly import log-symbols in scripts

eglitise:
The recent update to `log-symbols` `v7` broke the symbols for script files (files with ES6 imports were not affected):
symbols-old

This PR fixes the imports, plus the symbols are now colored (which was not the case with `log-symbols` `v4`):
symbols-new


eglitise:
From the linked Copilot session:
> The changes look correct and complete. I don't see any issues to report.

PR #21863, opened 2026-01-17
feat(eslint-config-appium-ts): check node prefix for standard modules

KazuCocoa:
## Proposed changes

I wondered if it would make sense to add a lint to require/fix `node:` for standard node modules.

We did review to add this `node:` in several places. To avoid that kind of review and allow us to apply it with `--fix`, what do you think of adding `eslint-plugin-unicorn`?

The primary changes are:

- packages/eslint-config-appium-ts/package.json
- packages/eslint-config-appium-ts/index.mjs

Other changes were applied with `npm run lint:fix`.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments


KazuCocoa:
Hm, this is `feat` for the eslint package, but others are chore or fix... CI fails without their changes

mykola-mokhnach:
I would split this PR.
Do not make the rule to an error, make it a warning instead.

So the first PR would be to add the rule to the lint config, do not change other modules
The second PR  would be to update other modules

The third PR MIGHT be to make it to an error, although it would be still a breaking change
Also, do we already have ESLint updated in other appium modules after the recent major bump? I would rather focus on that frist.

KazuCocoa:
Makes sense. It is what I considered after this pr

eglitise:
> do we already have ESLint updated in other appium modules after the recent major bump?

That change isn't released yet, so I think it's fine to add more breaking changes due to this plugin, as we'll have a major version bump anyway.

PR #21864, opened 2026-01-17
feat(eslint-config-appium-ts): add unicorn/prefer-node-protocol with warn

KazuCocoa:
## Proposed changes

We did review to add this `node:` in several places. To avoid that kind of review and allow us/contributors to apply it with `--fix`, what do you think of adding `eslint-plugin-unicorn`?

Added `unicorn/prefer-node-protocol` to show warnings for `node:`. This will help our future review to reduce poining out the `node:`.

https://github.com/appium/appium/pull/21863 is previous one.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/packages/eslint-config-appium-ts/index.mjs
+++ b/packages/eslint-config-appium-ts/index.mjs
@@ -157,7 +157,6 @@ export default defineConfig([
     files: ['**/test/**', '*.spec.*js', '-specs.*js', '*.spec.ts'],
     plugins: {
       mocha: mochaPlugin,
-      unicorn
     },
     extends: [mochaPlugin.configs.recommended],
     rules: {
--- a/packages/eslint-config-appium-ts/index.mjs
+++ b/packages/eslint-config-appium-ts/index.mjs
@@ -13,6 +13,7 @@ import {importX} from 'eslint-plugin-import-x';
 import mochaPlugin from 'eslint-plugin-mocha';
 import nodePlugin from 'eslint-plugin-n';
 import {configs as tsConfigs} from 'typescript-eslint';
+import unicorn from 'eslint-plugin-unicorn';
 
 const gitignorePath = path.resolve(process.cwd(), '.gitignore');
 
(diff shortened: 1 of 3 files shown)


mykola-mokhnach (approved):

packages/eslint-config-appium-ts/index.mjs
```diff
@@ -154,6 +157,7 @@ export default defineConfig([
     files: ['**/test/**', '*.spec.*js', '-specs.*js', '*.spec.ts'],
     plugins: {
       mocha: mochaPlugin,
+      unicorn
```
eglitise: Not sure if this line is needed, since we are not defining any test file-specific rules from this plugin. Test files are still subject to the  `unicorn/prefer-node-protocol` rule, since its parent configuration applies to all JS/TS files.

eglitise:
The unicorn plugin is quite large and powerful, so it feels like a bit of a waste to only use it for one rule. I would suggest adding more rules (whether in this PR or a new one), seeing as the next release will be a major version bump anyway.
To reduce code adjustment efforts in other repos, I think it's fine if we only limit ourselves to the rules with autofix. I'm particularly interested in the `prefer-module` rule, if we can even apply it.

packages/eslint-config-appium-ts/index.mjs
KazuCocoa: Ok, let me remove this. I also wondered if this would be needed actually

KazuCocoa:
I didn't have other ideas except for this `node:` which is frequently reviewed. We can add more rules later if we have preference :)

packages/eslint-config-appium-ts/index.mjs
KazuCocoa: https://github.com/appium/appium/pull/21864/commits/373119e3ccac5fcfd3e77cadfac6fa8d4949662c

eglitise:
We can certainly add them later, but they would technically be a breaking change. This is merely a good opportunity to do so, since we're already making other breaking changes.

eglitise (approved):

PR #21865, opened 2026-01-17
feat(images-plugin): Migrate to typescript

mykola-mokhnach:


eglitise:
The `ImageElementPlugin` default export seems to be removed, so this would technically be a breaking change (which I'm fine with, but it should be marked as such).

mykola-mokhnach:
> The `ImageElementPlugin` default export seems to be removed, so this would technically be a breaking change (which I'm fine with, but it should be marked as such).

I assume the import was not used. In order for appium to load the plugin only main plugin class export is necessary...

packages/images-plugin/package.json
```diff
@@ -23,19 +23,19 @@
...
-  "types": "./build/lib/plugin.d.ts",
+  "main": "./build/lib/index.js",
+  "types": "./build/lib/index.d.ts",
   "files": [
     "build",
     "docs",
     "lib",
-    "index.js",
     "tsconfig.json",
     "!build/tsconfig.tsbuildinfo"
```
eglitise, line 33: Are all of these needed in the final bundle? I'm guessing that just `build/lib` would suffice.

eglitise (approved):
Did a quick GitHub-wide search and could not find any results for imports of the default `ImageElementPlugin` export.

packages/images-plugin/package.json
mykola-mokhnach: theoretically yes, however we used to also include sources for published modules for easier debugging. Also they don't take much space

PR #21868, opened 2026-01-18
test(images-plugin): Migrate tests to typescript

mykola-mokhnach:


eglitise:
I see there's a lot of casting for the driver object `d` as `any`; is it not possible to make TS see the correct object signature? Or is it perhaps that the `BaseDriver` object isn't yet defined in TS?

KazuCocoa (approved):

mykola-mokhnach:
> I see there's a lot of casting for the driver object `d` as `any`; is it not possible to make TS see the correct object signature? Or is it perhaps that the `BaseDriver` object isn't yet defined in TS?

`click` expects ExternalDriver interface, while BaseDriver does not really implement it. This is still ok for tests, that is why we cast to any to workaround possible typescript complaints

eglitise (approved):

PR #21908, opened 2026-01-31
feat(relaxed-caps-plugin): Migrate to typescript

mykola-mokhnach:


packages/relaxed-caps-plugin/package.json
```diff
@@ -23,24 +23,22 @@
...
   },
   "files": [
-    "index.js",
+    "build/lib",
     "lib",
-    "build",
-    "tsconfig.json",
-    "!build/tsconfig.tsbuildinfo",
-    "!build/test"
+    "tsconfig.json"
```
KazuCocoa, line 34: This tsconfig.json also might not be necessary

KazuCocoa (commented):
Important might fail (CI)

mykola-mokhnach:
I'll fix tests in another PR to avoid unnecessary version bump

packages/relaxed-caps-plugin/package.json
mykola-mokhnach: We do include it to the dist in other modules. Lets maybe standardise which files we'd like to have in the released dist and which ones are not necessary

PR #21935, opened 2026-02-08
feat(fake-driver): Migrate to typescript

mykola-mokhnach:


packages/fake-driver/lib/commands/contexts.ts
```diff
@@ -1,84 +1,59 @@
...
+}
 
-  /**
-   * Get the list of available contexts
-   */
-  async getContexts(this: FakeDriver) {
-    return _.keys(this.getRawContexts());
-  },
+export async function getContexts(this: FakeDriver): Promise<string[]> {
+  return _.keys(this.getRawContexts());
```
eglitise, line 29: I'm not really sure why the syntax here is `this.localMethod()`, while line 49 uses `localMethod.call(this)`. Wouldn't just `localMethod(this)` work in both cases (and in other files)?

packages/fake-driver/lib/commands/find.ts
```diff
@@ -1,161 +1,142 @@
...
+    if (existingElId) {
+      return {ELEMENT: existingElId, [W3C_WEB_ELEMENT_IDENTIFIER]: existingElId};
+    }
+  } else {
+    // raw node: reuse id if we already have an element for this node
+    for (const [id, el] of _.toPairs(this.elMap)) {
+      if (el.node === node) {
+        return {ELEMENT: id, [W3C_WEB_ELEMENT_IDENTIFIER]: id};
+      }
+    }
```
eglitise, line 43: Can probably reuse `getExistingElementForNode`

packages/fake-driver/lib/driver.ts
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  async getFakeDriverArgs(): Promise<typeof this.cliArgs> {
+    await B.delay(1);
+    return this.cliArgs;
+  }
+
+  async getDeprecatedCommandsCalled(): Promise<string[]> {
+    await B.delay(1);
+    return [];
```
eglitise, line 216: Can we keep the TODO comment?
In general I would suggest keeping most if not all the removed comments

packages/fake-driver/lib/index.ts
```diff
@@ -4,19 +4,16 @@ import {startServer} from './server';
...
 
-export {FakeDriver, startServer, main};
-
-/**
- * @typedef {import('./types').W3CFakeDriverCaps} W3CFakeDriverCaps
- * @typedef {import('./types').FakeDriverCaps} FakeDriverCaps
- */
+export {FakeDriver, startServer};
+export type {FakeDriverCaps, W3CFakeDriverCaps} from './types';
+export default FakeDriver;
```
eglitise: I wonder about this default export. Currently the default export is specified directly in `driver.js`, and `index.js` only has the named export.

packages/fake-driver/lib/fake-app.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  get width(): number {
+    if (this._width === null) {
+      this.setDims();
+    }
+    const w = this._width;
+    if (w === null) {
+      throw new Error('Cannot fetch app dimensions');
+    }
```
eglitise, line 80: This looks like a behavioral change; is it needed as part of this PR?
Same for lines 88-90

packages/fake-driver/lib/fake-app.ts
```diff
@@ -0,0 +1,232 @@
...
+      fragment,
+      XMLDom.MIME_TYPE.XML_TEXT
+    ) as XMLDocument;
+  }
+
+  deactivateFrame(): void {
+    this.activeFrame = null;
+    if (this.activeWebview) {
+      this.activateWebview(this.activeWebview);
+    }
```
eglitise, line 158: This also seems like a behavioral change

packages/fake-driver/lib/fake-app.ts
```diff
@@ -0,0 +1,232 @@
...
+    const prompt = this.activeAlert?.getAttr('prompt');
+    if (prompt) {
+      return prompt;
+    }
+    const fromAttrs = this.activeAlert?.nodeAttrs?.text;
+    if (fromAttrs) {
+      return fromAttrs;
+    }
+    const node = this.activeAlert?.node as {getAttribute?(name: string): string | null} | undefined;
+    return node?.getAttribute?.('text') ?? '';
```
eglitise, line 217: Same here

packages/fake-driver/lib/fake-element.ts
```diff
@@ -0,0 +1,124 @@
...
+  getSize(): {width: number; height: number} {
+    return {
+      width: parseFloat(this.nodeAttrs.width || '0'),
+      height: parseFloat(this.nodeAttrs.height || '0'),
+    };
+  }
+
+  click(): void {
+    const curClicks = Number(this.getAttr('clicks') || 0);
+    this.setAttr('clicks', String(curClicks + 1));
```
eglitise, line 97: I'm guessing this was changed for consistency with all other attributes being strings

packages/fake-driver/lib/commands/find.ts
mykola-mokhnach: not sure if that would work here as they return different values

packages/fake-driver/lib/fake-element.ts
mykola-mokhnach: exactly, otherwise we cannot pass typechecks

packages/fake-driver/lib/driver.ts
mykola-mokhnach: restored TODOs

packages/fake-driver/lib/index.ts
mykola-mokhnach: do you think it may create any issues?

packages/fake-driver/lib/fake-app.ts
mykola-mokhnach: this is necessary to ensure nullability assertions. Previously they were not strictly ensured

packages/fake-driver/lib/fake-app.ts
mykola-mokhnach: the original typedef requires this method to return a string. Previously this was not checked strictly

packages/fake-driver/lib/index.ts
eglitise: Likely no, but if it's being moved anyway, then I'm not sure there's a reason to keep it at all

mykola-mokhnach: ok, removed

packages/fake-driver/lib/commands/contexts.ts
mykola-mokhnach: Because assertWebviewContext is not attached to the FakeDriver class, so calling it on this would create a typecheck error. It's a common trick to use a function with a particular instance context without explicitly adding it to the class interface

packages/fake-driver/lib/commands/find.ts
mykola-mokhnach: not sure if that would work here as they return different values

eglitise: Ah, you're right, I missed that one compares against `node`, while the other against `node.node`

packages/fake-driver/lib/commands/contexts.ts
eglitise: Maybe I'm missing something, but the same seems to apply to `getRawContexts` , so I'm not sure why `this.getRawContexts()` here doesn't throw an error

mykola-mokhnach: you are right, both methods are included into the instance interface, so they could be called with `this.` notation. I somehow thought that only one is there, because it does not really make sense to have `assertWebviewContext` as public API, but it was already there before the migration. Updated

packages/fake-driver/lib/fake-app.ts
mykola-mokhnach: activateWebview does not expect optional arguments and would anyway throw an exception if the webview argument is undefined.

mykola-mokhnach: typescript checks have just uncovered a potential bug

packages/fake-driver/lib/commands/contexts.ts
eglitise: `alert.ts` also has a few `assertAlert.call` and `postAcceptAlert.call` instances, and both of those methods are in public API as well

mykola-mokhnach: updated that module as well

eglitise (approved):

PR #21937, opened 2026-02-08
fix(base-driver): normalize sessionId to avoid string[] type errors

Frangitsco:
Changes
This PR fixes TypeScript build failures in protocol.ts caused by req.params.sessionId being typed as string | string[]. Several internal calls expect a plain string, which caused the monorepo build to fail on a clean install.

A helper function (normalizeSessionId) is introduced to normalize the value, and all internal uses inside asyncHandler now use the normalized version.

Types of changes
Bugfix (nonbreaking change which fixes an issue)

Checklist
I have tested the changes locally with npm run build

I ensured no unrelated changes are included

I have followed the code style and conventions of the project

### DIFF
--- a/packages/base-driver/lib/protocol/protocol.ts
+++ b/packages/base-driver/lib/protocol/protocol.ts
@@ -297,6 +297,10 @@ export function routeConfiguringFunction(driver: Core<any>): RouteConfiguringFun
   };
 }
 
+function normalizeSessionId(id: string | string[] | undefined): string | undefined {
+  return Array.isArray(id) ? id[0] : id;
+}
+
 function buildHandler(
   app: Application,
   method: string,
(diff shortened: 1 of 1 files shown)


KazuCocoa:
Could you also share the type error message? Also, for example, what repository did you get the error will also help

Frangitsco:
@KazuCocoa Thanks for reviewing!

I reproduced the issue on a clean clone of the main Appium monorepo: https://github.com/appium/appium

From the repo root:
npm install
npm run build

These are the errors I received before applying the fix:

packages/base-driver/lib/protocol/protocol.ts:313:51 - error TS2345: 
Argument of type 'string | string[]' is not assignable to parameter of type 'string | null | undefined'.
  Type 'string[]' is not assignable to type 'string'.

313     let currentProtocol = extractProtocol(driver, req.params.sessionId);
                                                      ~~~~~~~~~~~~~~~~~~~~

packages/base-driver/lib/protocol/protocol.ts:319:27 - error TS2345:
Argument of type 'string | string[]' is not assignable to parameter of type 'string | null | undefined'.

319         getLogger(driver, req.params.sessionId).warn(
                              ~~~~~~~~~~~~~~~~~~~~

…and so on (16 similar errors in total), all caused by req.params.sessionId being typed as string | string[].

KazuCocoa:
I see. I guess you installed newer version of https://www.npmjs.com/package/@types/express-serve-static-core on your local instead of this appium's lock file. What happens with `npm ci` instead of `npm install`. It will ensure your env gets deps from https://github.com/appium/appium/blob/master/package-lock.json.

So, `npm ls @types/express-serve-static-core` on your local would show non-5.0.0. e.g. your local would have 5.1.1.
https://github.com/appium/appium/blob/master/package-lock.json#L4221-L4232

Frangitsco:
Thanks again, you were right. I reinstalled the repo using `npm ci` so that dependencies exactly match the lockfile, and the monorepo builds successfully with no TypeScript errors.

I checked my previous environment using:
npm explain @types/express-serve-static-core
and found that my local install had resolved:

`@types/express@5.0.6`
 → which pulled in `@types/express-serve-static-core@5.1.1` (via `^5.0.0`)

With that newer version, TypeScript types `req.params.sessionId` as `string | string[]`, which is what caused the build errors I reported. I can confirm using `npm ci` keeps the versions pinned to the lockfile and avoids this issue. So the failure was caused by dependency drift in my local environment, I can adjust or close the PR if useful.

KazuCocoa:
I understood. So, newer express type can have such an array case. Only `/session/:sessionId(.*)` like this endpoint definition could be such a case, but Appium doesn't have/expect it.

I have created https://github.com/appium/appium/pull/21942 to raise an error for an invalid pattern with the necessary dependency update in the lock file.

Issue #21940, opened 2026-02-09
[xcuitest] does not follow log level


amedvedjev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Appium Version

3.2.0

### Appium Host Type

Command Line

### Is this issue reproducible using the latest components?

- [x] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

appium@3.2.0
xcuitest@10.21.1

While starting Appium server with any log level (ERROR or DEBUG checked) xcuitest driver always sending logs in INFO level like
```
info XCUITestDriver@d353 Starting screen capture on the device '00008030-001D79403C84202E' with command: 'ffmpeg -f mjpeg -.... Will timeout in 600000ms
info HTTP <-- POST /session/b89c5c63-bd37-4d3a-bafc-c6d37ac86854/appium/start_recording_screen 200 3325 ms - 12 
```

### Expected Behavior

Expected xcuitest to follow Appium server log level settings.

### Environment Details

- Debug details (`appium --show-debug-info`):
- Last component version(s) which did _not_ exhibit the problem:
- Platform and version of device under test:
- Real device or emulator/simulator:

### Minimal Reproducible Example

I am starting appium as
```
appium --log-level error --port 6009 --keep-alive-timeout 1800  --session-override --log-timestamp --tmp /Users/..., --use-plugins execute-driver --relaxed-security
```
and in code open driver with caps
```
appium:app: /Users/..., appium:automationName: XCuiTest, appium:bundleId: com.xxx.dev, appium:clearSystemFiles: true, appium:derivedDataPath: /Users/..., appium:deviceName: iPhone QA9 old, appium:forceAppLaunch: true, appium:fullContextList: true, appium:fullReset: false, appium:iosInstallPause: 500, appium:mjpegServerPort: 6209, appium:nativeWebTap: true, appium:newCommandTimeout: 180, appium:noReset: true, appium:platformVersion: 18.0, appium:udid: 00008030-001D79403C84202E, appium:useJSONSource: false, appium:useNewWDA: false, appium:usePrebuiltWDA: true, appium:wdaLaunchTimeout: 120000, appium:wdaLocalPort: 6109, appium:wdaStartupRetries: 4, platformName: IOS}
```

### Further Information

appium@3.1.0
xcuitest@10.4.3

issue NOT reproducing

KazuCocoa:
Could you share the entire appium log? I couldn't see any logs with the below info level. I wondered what kind log was available on your env.

[...]

PR #21945, opened 2026-02-10
fix(appium): Fix CLI arguments normalization

mykola-mokhnach:
https://github.com/appium/appium/issues/21940#issuecomment-3876771876

The root cause was missing dest for schema-derived server args in the CLI parser, which made --log-level end up under log_level instead of loglevel. Setting dest for server args fixes the CLI behavior; normalization in main remains for programmatic callers.

### DIFF
--- a/packages/appium/lib/cli/parser.js
+++ b/packages/appium/lib/cli/parser.js
@@ -14,7 +14,7 @@ import {
   SERVER_SUBCOMMAND,
   SETUP_SUBCOMMAND
 } from '../constants';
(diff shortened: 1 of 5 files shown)


KazuCocoa (approved):
This change looks good, while I'm not sure this will fix the linked issue since the reported log format is `info HTTP -->` which might be a different format from our https://github.com/appium/appium/blob/master/packages/base-driver/lib/express/express-logging.js#L55

mykola-mokhnach:
@KazuCocoa it's easy to verify the fix by just running the server with changes from this PR and same STR as in the original bug report ;)

Issue #17791, opened 2022-11-17
when installing an extension, it appears installation succeeds even if validation fails


boneskull:
In `installViaNpm()`, we show a spinner while the installation happens, then report "success" (shows a green checkmark) once it's completed without error and the `package.json` is parseable.

Thereafter, we perform additional validation against the content of `package.json` (the `appium` field, mostly).  This step can fail.

Maybe it'd be better to wrap the entire install/validate operation in a spinner (or try something different) so that a user does not see installation has succeeded when, in fact, it has failed due to validation.

While it may _technically_ be true that "installation" succeeded, it does not matter to the end user, who did not experience success.

boneskull:
This might be a morass, which is why I'm making the issue instead of just going to fix it.

jlipps:
yeah i think we should wrap the entire operation, or have two spinner steps, one for install and one for validate

subhashvkotra:
I am in favour of two spinner steps, it gives clearer picture to anyone trying.

mykola-mokhnach:
Closed as fixed

PR #21947, opened 2026-02-10
chore(appium): Split package installation and validation

mykola-mokhnach:
This is related to https://github.com/appium/appium/issues/17791

I don't have any strong feelings about the current implementation, so if you think this issue itself does not make sense then we could just close it as "won't fix".

### DIFF
--- a/packages/appium/lib/cli/extension-command.js
+++ b/packages/appium/lib/cli/extension-command.js
@@ -601,20 +601,23 @@ class ExtensionCliCommand {
    * @returns {Promise<ExtInstallReceipt<ExtType>>}
    */
   async installViaNpm({installSpec, pkgName, pkgVer, installType}) {
-    const msg = `Installing '${installSpec}'`;
+    const installMsg = `Installing '${installSpec}'`;
+    const validateMsg = `Validating '${installSpec}'`;
 
     // the string used for installation is either <name>@<ver> in the case of a standard NPM
     // package, or whatever the user sent in otherwise.
     const installStr = installType === INSTALL_TYPE_NPM ? `${pkgName}${pkgVer ? `@${pkgVer}` : ''}` : installSpec;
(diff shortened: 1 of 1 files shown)


packages/appium/lib/cli/extension-command.js
```diff
@@ -601,20 +601,22 @@ class ExtensionCliCommand {
...
-          installType,
-        });
+      const {pkg, installPath} = await spinWith(
+        this.isJsonOutput,
+        installMsg,
+        async () => await npm.installPackage(appiumHome, installStr, {pkgName, installType})
+      );
+
+      await spinWith(this.isJsonOutput, validateMsg, async () => {
         this.validatePackageJson(pkg, installSpec);
```
eglitise, line 619: I'm not against splitting this, but `validatePackageJson` is a synchronous operation, so it's unlikely that the spin will even have any effect for it

mykola-mokhnach: yes, that makes sense.

shall we then just close the original issue?

eglitise: But maybe it's fine - we currently already have a spinner for the updatable check, which is also sync

packages/appium/lib/cli/extension-command.js
```diff
@@ -601,20 +601,22 @@ class ExtensionCliCommand {
...
-          pkgName,
-          installType,
-        });
+      const {pkg, installPath} = await spinWith(
+        this.isJsonOutput,
+        installMsg,
+        async () => await npm.installPackage(appiumHome, installStr, {pkgName, installType})
+      );
+
+      await spinWith(this.isJsonOutput, validateMsg, async () => {
```
eglitise, line 618: The `async` from the callback function can probably be removed

eglitise (approved):

KazuCocoa (approved):

PR #21956, opened 2026-02-12
test(base-driver): Migrate e2e tests to typescript

mykola-mokhnach:


packages/base-driver/test/unit/jsonwp-proxy/mock-request.js
```diff
@@ -1,4 +1,4 @@
-function resFixture(url, method, _json) {
+function resFixture(url, method, _json) { // eslint-disable-line @typescript-eslint/no-unused-vars
```
eglitise: It's probably cleaner to just not extract the `json` property on line 25

packages/base-driver/test/e2e/basedriver/helpers.e2e.spec.ts
```diff
@@ -1,39 +1,35 @@
 import path from 'node:path';
 import {fs} from '@appium/support';
+
+declare const __dirname: string;
```
eglitise: If we want to fix the issue with __dirname not being available in ESM, I'm not sure if this would work, but the `schema` package is using the recommended approach ([here](https://github.com/appium/appium/blob/master/packages/schema/scripts/generate-schema-json.mjs#L21)). Although there are still plenty of other files in the repo that just call __dirname directly.

packages/base-driver/test/unit/jsonwp-proxy/mock-request.js
mykola-mokhnach: removed it

packages/base-driver/test/e2e/basedriver/helpers.e2e.spec.ts
mykola-mokhnach: this ESM vs CJS is a mess. What've learned so far ts-node assumes the transpiled modules is ESM if it uses url.meta and CJS if it has mentions of __dirname or __filename.

Lets try to run it without this declare and see what happens. If it fails then I am going to  put it back.

eglitise: Nice to see that CI did pass after all. But I think in the future we can also explore whether `tsx` + `tsc` could offer any benefits over `ts-node` in this repo's usecases.

eglitise (approved):

PR #21960, opened 2026-02-13
chore(support): replace moment with native Date

eglitise:
This PR replaces `moment.js` with a native `Date()`-based equivalent. There are several reasons for this:
* `moment.js` is in maintenance mode
* it is a fairly large package (4.3MB unpacked)
* this repo only uses `moment.js` in one location

The last point is also the reason why I am proposing to use the native API instead of another smaller package like `day.js`, even though it is a bit less readable due to no built-in formatter method.

More information:
* https://momentjs.com/docs/#/-project-status/
* https://e18e.dev/docs/replacements/moment.html

### DIFF
--- a/packages/support/lib/logging.js
+++ b/packages/support/lib/logging.js
@@ -1,11 +1,9 @@
 import globalLog, { markSensitive as _markSensitive } from '@appium/logger';
 import _ from 'lodash';
-import moment from 'moment';
 
 /** @type {import('@appium/types').AppiumLoggerLevel[]} */
 export const LEVELS = ['silly', 'verbose', 'debug', 'info', 'http', 'warn', 'error'];
 const MAX_LOG_RECORDS_COUNT = 3000;
-const PREFIX_TIMESTAMP_FORMAT = 'HH-mm-ss:SSS';
 // mock log object is used in testing mode to silence the output
 const MOCK_LOG = {
   unwrap: () => ({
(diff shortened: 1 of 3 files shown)


packages/support/lib/logging.js
```diff
@@ -132,7 +130,15 @@ function getFinalPrefix(prefix, shouldLogTimestamp = false) {
   if (!shouldLogTimestamp) {
     return result;
   }
-  const formattedTimestamp = `[${moment().format(PREFIX_TIMESTAMP_FORMAT)}]`;
```
mykola-mokhnach, line 135: can we use Intl.DateTimeFormat ?

eglitise: Theoretically yes, but according to Copilot (Claude), it would be less readable, since it also doesn't support custom separators, and doesn't have a milliseconds field. This is the code block equivalent it returned:
```js
const now = new Date();
const formatter = new Intl.DateTimeFormat('en-US', {
  hour: '2-digit',
  minute: '2-digit',
  second: '2-digit',
  hour12: false
});
const parts = formatter.formatToParts(now);
const hours = parts.find(p => p.type === 'hour').value;
const minutes = parts.find(p => p.type === 'minute').value;
const seconds = parts.find(p => p.type === 'second').value;
const milliseconds = String(now.getMilliseconds()).padStart(3, '0');
const formattedTimestamp = `[${hours}-${minutes}-${seconds}:${milliseconds}]`;
```

mykola-mokhnach (approved):

packages/support/lib/logging.js
```diff
@@ -132,7 +130,15 @@ function getFinalPrefix(prefix, shouldLogTimestamp = false) {
...
+  // Format the timestamp as HH-mm-ss:SSS
+  const now = new Date();
+  const pad = (n, z = 2) => String(n).padStart(z, '0');
+  const formattedTimestamp = `[${
+    pad(now.getHours()) + '-' +
```
mykola-mokhnach: not sure why we should use catenation here since the string is anyway templated

packages/support/lib/logging.js
mykola-mokhnach: nah, this does not look better

packages/support/lib/logging.js
eglitise: Changed to only use template strings

PR #21942, opened 2026-02-10
chore(base-driver): update type of express-serve-static-core

KazuCocoa:
## Proposed changes

https://github.com/appium/appium/pull/21937 indicates that a newer version of `@types/express-serve-static-core` changed the type of `req.params.sessionId`. So, the latest `req.params.sessionId` can be an array of strings or `'/session/:sessionId(.*)'` which doesn't exist in Appium.

So, this change creates a method to raise an error for such an array case.

## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [Contributing Guide](https://appium.io/docs/en/latest/contributing/)
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #21962, opened 2026-02-14
fix(base-driver): fix session id handling more

KazuCocoa:
## Proposed changes

An error I observed https://github.com/appium/appium/pull/21942#discussion_r2791740734 was by remaining `req.params.sessionId`. So, the fix is to use `getSessionId` everywhere. But this case occurs only when users use `/session/*sessionId/something` syntax, so usually it is a wrong one - the driver/plugin author should fix.

So, this PR prevents repeating the log in a request, but prints a message to provide a hint to fix the wrong routing definition for authors.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/packages/base-driver/lib/protocol/protocol.ts
+++ b/packages/base-driver/lib/protocol/protocol.ts
@@ -35,12 +35,12 @@ export function determineProtocol(createSessionArgs: any[]): keyof typeof PROTOC
 }
 
 /**
- * Extract and validate sessionId from Express route parameter.
+ * Extract and validate the sessionId from the Express route parameter.
  * Express may return route params as string | string[] | undefined.
  * Appium uses standard routes (e.g., /session/:sessionId) which should always be strings.
  * Only `*` such as `/session/*sessionId` can return `string[]`.
  * Then, this method will return the first element as the session id.
- * It may break existing appium routing hanlding also, thus this method will log
+ * It may break existing appium routing handling also, thus this method will log
  * received parameters as well to help debugging.
  * @param driver Running driver
  * @param req The request in Express
@@ -50,10 +50,10 @@ export function getSessionId(driver: Core<any>, req: Request): string | undefine
   if (Array.isArray(req.params.sessionId)) {
     const sessionId = req.params.sessionId[0];
     getLogger(driver, sessionId).warn(
-      `Received malformed sessionId as array from route: ${req.originalUrl}. ` +
-      `This indicates a route definition issue. The route should start with '/session/:sessionId' (named parameter) ` +
+      `Received malformed sessionId as array from the route: ${req.originalUrl}. ` +
+      `This indicates the route definition issue. The route should start with '/session/:sessionId' (named parameter) ` +
       `instead of '/session/*sessionId' (wildcard). ` +
-      `Using first element as session id: ${sessionId}. ` +
+      `Using the first element as session id: ${sessionId}. ` +
       `Please fix the route definition to prevent this error.`
     );
     // This is to not log the message multiple times.
--- a/packages/base-driver/lib/protocol/protocol.ts
+++ b/packages/base-driver/lib/protocol/protocol.ts
@@ -56,7 +56,7 @@ export function getSessionId(driver: Core<any>, req: Request): string | undefine
       `Using first element as session id: ${sessionId}. ` +
       `Please fix the route definition to prevent this error.`
     );
-    // This is to not log the message twice.
+    // This is to not log the message multiple times.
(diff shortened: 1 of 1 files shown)


eglitise (approved):

packages/base-driver/lib/protocol/protocol.ts
```diff
@@ -49,9 +49,15 @@ export function determineProtocol(createSessionArgs: any[]): keyof typeof PROTOC
...
-      `Received sessionId as array: ${JSON.stringify(req.params.sessionId)}. Using first element as the session id: ${sessionId}`
+    getLogger(driver, sessionId).warn(
+      `Received malformed sessionId as array from route: ${req.originalUrl}. ` +
+      `This indicates a route definition issue. The route should start with '/session/:sessionId' (named parameter) ` +
+      `instead of '/session/*sessionId' (wildcard). ` +
+      `Using first element as session id: ${sessionId}. ` +
+      `Please fix the route definition to prevent this error.`
     );
+    // This is to not log the message multiple times.
+    req.params.sessionId = sessionId;
```
mykola-mokhnach, line 60: I don't think this is going to help much. Each request creates a new `req` instance

Consider persiting this flag outside of the async handler closure

mykola-mokhnach:
fix(appium) -> fix(base-driver)

packages/base-driver/lib/protocol/protocol.ts
mykola-mokhnach: Perhaps you can reuse the same idea as for deprecatedCommandsLogged

KazuCocoa: Yeah, this is on a request-by-request basis right now. One request calls the `getSessionId` method in multiple places, so this will prevent logging the message multiple times in a request, so the scope is per request as this implementation

KazuCocoa: I thought of `deprecatedCommandsLogged`, which is process-wide memory. The incoming request could be several patterns like `[<session id>,"path1"]`, `[<session id>,"path2"]` etc. If we use the session id as a key, it will need to be removed after the session ends to avoid a memory leak.  Not alway the method can get `spec` like `deprecatedCommandsLogged` in the current implementation so I chosen simpler way to handle per request basis

packages/base-driver/lib/protocol/protocol.ts
```diff
@@ -49,9 +49,15 @@ export function determineProtocol(createSessionArgs: any[]): keyof typeof PROTOC
 export function getSessionId(driver: Core<any>, req: Request): string | undefined {
   if (Array.isArray(req.params.sessionId)) {
     const sessionId = req.params.sessionId[0];
-    getLogger(driver, sessionId).debug(
-      `Received sessionId as array: ${JSON.stringify(req.params.sessionId)}. Using first element as the session id: ${sessionId}`
+    getLogger(driver, sessionId).warn(
+      `Received malformed sessionId as array from route: ${req.originalUrl}. ` +
```
mykola-mokhnach: from the route

packages/base-driver/lib/protocol/protocol.ts
```diff
@@ -49,9 +49,15 @@ export function determineProtocol(createSessionArgs: any[]): keyof typeof PROTOC
...
+      `This indicates a route definition issue. The route should start with '/session/:sessionId' (named parameter) ` +
+      `instead of '/session/*sessionId' (wildcard). ` +
+      `Using first element as session id: ${sessionId}. ` +
```
mykola-mokhnach: the first element

packages/base-driver/lib/protocol/protocol.ts
```diff
@@ -49,9 +49,15 @@ export function determineProtocol(createSessionArgs: any[]): keyof typeof PROTOC
 export function getSessionId(driver: Core<any>, req: Request): string | undefined {
   if (Array.isArray(req.params.sessionId)) {
     const sessionId = req.params.sessionId[0];
-    getLogger(driver, sessionId).debug(
-      `Received sessionId as array: ${JSON.stringify(req.params.sessionId)}. Using first element as the session id: ${sessionId}`
+    getLogger(driver, sessionId).warn(
+      `Received malformed sessionId as array from route: ${req.originalUrl}. ` +
+      `This indicates a route definition issue. The route should start with '/session/:sessionId' (named parameter) ` +
```
mykola-mokhnach: The route definition

mykola-mokhnach (approved):

PR #21969, opened 2026-02-15
chore(support): Migrate fs.js to typescript

mykola-mokhnach:


KazuCocoa (approved):

packages/support/lib/fs.ts
```diff
@@ -0,0 +1,458 @@
...
+        if (isErrnoException(err) && err.code === 'EXDEV') {
+          await this.copyFile(String(src), String(dst));
+          await this.rimraf(src);
+        } else {
+          throw err;
+        }
+      }
+    };
+
+    let fromStat: import('node:fs').Stats;
```
eglitise: This can be a top-level type import

packages/support/lib/fs.ts
```diff
@@ -0,0 +1,458 @@
...
+  async mkdirp(dir: PathLike): Promise<string | undefined> {
+    return await fs.mkdir(dir, {recursive: true});
+  },
+
+  /**
+   * Walks a directory; callback is invoked with path joined with dir.
+   * @param dir - Directory path to start walking
+   * @param recursive - If true, walk subdirectories
+   * @param callback - Called for each path; return true to stop
+   * @returns The found path or null if not found
```
eglitise, line 293: I assume this docstring isn't needed, given that it's been removed from all other methods

packages/support/lib/fs.ts
```diff
@@ -0,0 +1,458 @@
...
+    let directoryCount = 0;
+    const timer = new Timer().start();
+    return await new B<string | null>(function (resolve, reject) {
+      let lastFileProcessed: Promise<string | undefined> = Promise.resolve(undefined);
+      walker = klaw(dir, {
+        depthLimit: recursive ? -1 : 0,
+      });
+      walker
+        .on('data', function (item: klaw.Item) {
+          const w = walker;
```
eglitise: Not sure why another identical variable is needed

packages/support/lib/fs.ts
```diff
@@ -0,0 +1,458 @@
...
+                resolve(item.path);
+                return item.path;
+              }
+              if (w) {
+                w.resume();
+              }
+              return undefined;
+            } catch (err) {
+              reject(err);
+              return undefined;
```
eglitise: Are explicit returns needed here?

packages/support/lib/fs.ts
mykola-mokhnach: fixed

packages/support/lib/fs.ts
eglitise: I assume `return item.path;` also isn't needed

packages/support/lib/fs.ts
mykola-mokhnach: I still find it useful, only jsdoc where only param types are provided were erased

packages/support/lib/fs.ts
mykola-mokhnach: removed

packages/support/lib/fs.ts
mykola-mokhnach: removed

mykola-mokhnach: > I assume `return item.path;` also isn't needed

return is needed there

eglitise (approved):

PR #21975, opened 2026-02-17
chore: remove obsolete root-level files

eglitise:
This removes a few top-level files that appear to be obsolete at the moment:
* `.lintstagedrc.json` - the repo does not use `lint-staged` anymore
* `CONDUCT.md` - duplicated from the org-level equivalent
* `IDEAS.md` - most of these are either already done or not relevant anymore, and it's succeeded by the roadmap
* `triagers.json` - I believe this was needed by https://github.com/jlipps/triager, which is also not used anymore


KazuCocoa:
COUNDUCT.md still has a link from http://appium.io/docs/en/latest/contributing/, need to update there if we removed that.

KazuCocoa (approved):

PR #21990, opened 2026-02-19
feat(execute-driver-plugin)!: replace bluebird usage inside scripts with native promises

eglitise:
BREAKING CHANGE: The `Promise` object in the script context now matches the global `Promise` object instead of `bluebird`. Methods specific to `bluebird` are no longer supported.

* Replace `bluebird` with native `Promise`s in the module code
* The `Promise` object in the script context now matches the global `Promise` object instead of `bluebird`
* Add support for `setTimeout` and `clearTimeout` functions in the script context

### DIFF
--- a/packages/execute-driver-plugin/lib/execute-child.ts
+++ b/packages/execute-driver-plugin/lib/execute-child.ts
@@ -1,16 +1,9 @@
+import * as asyncbox from 'asyncbox';
 import _ from 'lodash';
 import vm from 'node:vm';
 import {promisify} from 'node:util';
 import {logger, util} from 'appium/support';
 import type {DriverScriptMessageEvent, ScriptResult, RunScriptResult} from './types';
-
-// Historically, scripts could use the Promise.delay function, which relied on bluebird.
-// The native promise library does not have a 1:1 equivalent, so we add it to maintain backwards compatibility.
-class PromiseWithDelay<T> extends Promise<T> {
-  static delay(ms: number): Promise<void> {
-    return new Promise<void>((resolve) => setTimeout(resolve, ms));
-  }
-}
 
 const log = logger.getLogger('ExecuteDriver Child');
 let send: (res: ScriptResult) => Promise<void>;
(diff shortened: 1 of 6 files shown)


packages/execute-driver-plugin/lib/execute-child.ts
```diff
@@ -1,9 +1,17 @@
 import _ from 'lodash';
-import B from 'bluebird';
 import vm from 'node:vm';
+import {promisify} from 'node:util';
 import {logger, util} from 'appium/support';
 import type {DriverScriptMessageEvent, ScriptResult, RunScriptResult} from './types';
 
+// Historically, scripts could use the Promise.delay function, which relied on bluebird.
```
mykola-mokhnach: is it the only diff? I assume there are more differences, so we could break user scripts if we merge it

eglitise: I assume the current plugin version supports all of `bluebird`'s methods. The issue here is that this was not documented anywhere, and we'd probably want to support native promises instead (which would also be more intuitive for users). I can turn this into a breaking change if that makes sense.

mykola-mokhnach: yep, we don't know what people actually use in their scripts. It would be safer to bump the major version of this plugin

eglitise: updated

packages/execute-driver-plugin/lib/execute-child.ts
```diff
@@ -46,11 +47,11 @@ async function runScript(eventParams: DriverScriptMessageEvent): Promise<RunScri
...
   log.info('Running driver script in Node vm');
 
-  // run the driver script, giving user access to the driver object, a fake
-  // console logger, and a promise library
+  // run the driver script, giving user access to the driver object, a fake console logger,
+  // the asyncbox module, and standard setTimeout/clearTimeout functions
   let result = await vm.runInNewContext(
     fullScript,
-    {driver, console: consoleFns, Promise: B},
+    {driver, console: consoleFns, asyncbox, setTimeout, clearTimeout},
```
mykola-mokhnach: I wouldn't add asyncbox as it's interface may change

eglitise: Hmm, well I can remove it, but I think it would still be fairly safe, since its interface is in our control anyway, and writing delay functions is much cleaner than in the native Promise implementation.

mykola-mokhnach: it's rather about clients not having control over module versioning. asyncbox only contains simple async wrappers, that are easy to recreate

eglitise: I don't think it would be a problem if we ensure that major asyncbox releases also bump the major version of this plugin, but fair enough, I'll remove it

mykola-mokhnach (approved):

PR #21991, opened 2026-02-20
fix(logger): Make sure we always have single logger instance per process

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/21940

Root cause analysis:

- If the main Appium server sets `global._global_npmlog` **before** any driver/remotexpc code runs, then when **both** copies of support run `_getLogger()`, they both see the global and both use that **one** Log. In that case you still have **two** `Log` instances in memory (one from each logger module load), but only **one** is actually used → no duplicate from “two loggers”.
- If `global._global_npmlog` is **not** set when the second copy of support loads (e.g. plugin/driver loads before the server sets it), then that second support uses **its** `globalLog` (Log B). Then you can have **two** Logs in use (driver’s and remotexpc’s), and depending on who logs where, you can get duplicate lines.

So: **different instances are still created** because **each time `@appium/logger` is loaded as a separate module instance, it unconditionally creates a new `Log()`**. The global is only used inside **support** to choose which logger to use; it does not stop the logger package from creating another instance.

### DIFF
--- a/packages/appium/lib/logsink.js
+++ b/packages/appium/lib/logsink.js
@@ -5,10 +5,6 @@ import _ from 'lodash';
 import { adler32 } from './utils';
 import { LRUCache } from 'lru-cache';
 
-// set up distributed logging before everything else
-global._global_npmlog = globalLog;
-// npmlog is used only for emitting, we use winston for output
-globalLog.level = 'info';
 const LEVELS_MAP = {
   debug: 4,
   info: 3,
@@ -42,6 +38,7 @@ const COLORS_CACHE = new LRUCache({
   updateAgeOnGet: true,
 });
 
+// npmlog is used only for emitting, we use winston for output (global is set by support)
 /** @type {import('winston').Logger?} */
 let log = null;
 
@@ -102,6 +99,8 @@ export async function init(args) {
       }
     }
   });
+  // Only Winston produces output; avoid duplicate lines from the logger's default stream
+  globalLog.stream = null;
 }
 
 /**
--- a/packages/logger/lib/log.ts
+++ b/packages/logger/lib/log.ts
@@ -44,7 +44,7 @@ export class Log extends EventEmitter implements Logger {
   prefixStyle: StyleObject;
   headingStyle: StyleObject;
   heading: string;
-  stream: Writable; // Defaults to process.stderr
+  stream: Writable | null; // Defaults to process.stderr; set to null when using custom output (e.g. Winston)
 
   _asyncStorage: AsyncLocalStorage<Record<string, any>>;
   _colorEnabled?: boolean;
--- a/packages/logger/lib/log.ts
+++ b/packages/logger/lib/log.ts
@@ -422,5 +422,18 @@ interface ArgumentFormatResult {
   stack: string | undefined,
 }
 
-export const GLOBAL_LOG = new Log();
+// Unique key for the process-wide logger on globalThis (avoids collisions with other globals).
+const GLOBAL_NPMLOG_KEY = 'appium-logger-global-8f4a1c2b-5e6d-4a9b-8c3f-7d2e1b0a9c6e';
+
+type GlobalWithLogger = typeof globalThis & { [K in typeof GLOBAL_NPMLOG_KEY]: Log | undefined };
+
+// Reuse process-wide logger so multiple loads of this module use the same Log instance.
+const g = globalThis as GlobalWithLogger;
+export const GLOBAL_LOG =
+  g[GLOBAL_NPMLOG_KEY] ??
+  (() => {
+    const log = new Log();
+    g[GLOBAL_NPMLOG_KEY] = log;
+    return log;
+  })();
 export default GLOBAL_LOG;
--- a/packages/logger/lib/types.ts
+++ b/packages/logger/lib/types.ts
@@ -1,5 +1,6 @@
 import type {EventEmitter} from 'node:events';
 import type {AsyncLocalStorage} from 'node:async_hooks';
+import type {Writable} from 'node:stream';
 
 export interface Logger extends EventEmitter {
   level: string;
@@ -8,7 +9,7 @@ export interface Logger extends EventEmitter {
   prefixStyle: StyleObject;
   headingStyle: StyleObject;
   heading: string;
-  stream: any; // Defaults to process.stderr
+  stream: Writable | null; // Defaults to process.stderr; set to null when using custom output (e.g. Winston)
 
   /**
    * Creates a log message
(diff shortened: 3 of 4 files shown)


KazuCocoa:
This didn't help. Quickly checked with below steps.

My env already has `.appium` dir and xcuitest driver `10.2.0`

1. built this branch and ran appium with `npx appium` in this repo
2. copy the logger built module in globally installed appium (So i updated `/Users/kazu/.nvm/versions/node/v22.19.0/lib/node_modules/appium/node_modules/@appium/logger/build/lib/log.js` with this change)

I also tried to copy the built log.js into xcuitest driver's `node_modules/@appium/logger/build/lib/log.js` as well, but it also didn't help.

mykola-mokhnach:
> This didn't help. Quickly checked with below steps.

:(

Could you please verify the following:

- Add console log of __filename to make sure each module is using a patched version
- Make sure the `@appium/support` package is patched  inside node_modules of the following places:
  - .appium
  - xcuitest driver
  - remotexpc (it should be inside xcuitest driver dependencies unless linked manually)

KazuCocoa:
- remotexpc itself didn't have `node_modules`, so I hope it referred to the xcuitest or appium's one... XCUITest driver itself is in `.appium` and the `@appium/logger` was this built log.js
- `@appium/support` didn't have `node_modules/@appium/logger`, btw

Yes, i have tested with code diff below:

```diff
diff --git a/packages/logger/lib/log.ts b/packages/logger/lib/log.ts
index f7edc3e54..259283042 100644
--- a/packages/logger/lib/log.ts
+++ b/packages/logger/lib/log.ts
@@ -252,7 +252,7 @@ export class Log extends EventEmitter implements Logger {
       id: this._id++,
       timestamp: Date.now(),
       level,
-      prefix: this._secureValuesPreprocessor.preprocess(unleakString(prefix || '')),
+      prefix: `[debug] ${__filename}: ${this._secureValuesPreprocessor.preprocess(unleakString(prefix || ''))}`,
       message: this._secureValuesPreprocessor.preprocess(unleakString(formattedMessage)),
     };
```

and the result is:

```
info [debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: AppiumDriver@0e3c New XCUITestDriver session created successfully, session 1fa5a6e6-3f01-4433-9a2c-2e5f373e1c68 added to master session list
[1fa5a6e6][[debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: AppiumDriver@0e3c] Event 'newSessionStarted' logged at 1771644135172 (19:22:15 GMT-0800 (Pacific Standard Time))
[1fa5a6e6][[debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: XCUITestDriver@9771] Cached the protocol value 'W3C' for the new session 1fa5a6e6-3f01-4433-9a2c-2e5f373e1c68
[1fa5a6e6][[debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: XCUITestDriver@9771] Responding to client with driver.createSession() result: {"capabilities":{"webStorageEnabled":false,"locationContextEnabled":false,"browserName":"","platform":"MAC","javascriptEnabled":true,"databaseEnabled":false,"takesScreenshot":true,"networkConnectionEnabled":false,"platformName":"ios","platformVersion":"26.2","deviceName":"iPhone 17","automationName":"xcuitest","webDriverAgentUrl":"http://192.168.4.23:8100","udid":"00008030-000C51E13C29802E"}}
[1fa5a6e6][[debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: HTTP] <-- POST /session 200 2101 ms - 456
info [debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: HTTP <-- POST /session 200 2101 ms - 456
[1fa5a6e6][[debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: HTTP] --> POST /session/1fa5a6e6-3f01-4433-9a2c-2e5f373e1c68/timeouts {"implicit":30000}
info [debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: HTTP --> POST /session/1fa5a6e6-3f01-4433-9a2c-2e5f373e1c68/timeouts {"implicit":30000}
[1fa5a6e6][[debug] /Users/kazu/github/appium/packages/logger/build/lib/log.js: XCUITestDriver@9771] Calling AppiumDriver.timeouts() with args: [null,null,null,null,30000,"1fa5a6e6-3f01-4433-9a2c-2e5f373e1c68"]
```

So, still it has `info` prefix etc... which is the issue's one...

mykola-mokhnach:
Thanks for checking it @KazuCocoa 

I've investigated the issue further and I assume the potential cause might be Winston/logsink logic.
Could you please try the updated PR again when you have time?
Feel free to add your own fixes if you get more insights in local env

packages/appium/lib/logsink.js
```diff
@@ -102,6 +99,8 @@ export async function init(args) {
       }
     }
   });
+  // Only Winston produces output; avoid duplicate lines from the logger's default stream
+  globalLog.stream = null;
```
KazuCocoa, line 103: It looks like this line fixed the issue..

Great finding...

KazuCocoa:
It looks like the minimal fix is only below on my local:

```
kazu $ git diff
diff --git a/packages/appium/lib/logsink.js b/packages/appium/lib/logsink.js
index 9cd77b221..6bdcc87e7 100644
--- a/packages/appium/lib/logsink.js
+++ b/packages/appium/lib/logsink.js
@@ -102,6 +102,7 @@ export async function init(args) {
       }
     }
   });
+  globalLog.stream = null;
 }

 /**
diff --git a/packages/logger/lib/log.ts b/packages/logger/lib/log.ts
index 3f433a601..5f000d428 100644
--- a/packages/logger/lib/log.ts
+++ b/packages/logger/lib/log.ts
@@ -44,7 +44,7 @@ export class Log extends EventEmitter implements Logger {
   prefixStyle: StyleObject;
   headingStyle: StyleObject;
   heading: string;
-  stream: Writable; // Defaults to process.stderr
+  stream: Writable | null; // Defaults to process.stderr; set to null when using custom output (e.g. Winston)

   _asyncStorage: AsyncLocalStorage<Record<string, any>>;
   _colorEnabled?: boolean;
```

KazuCocoa (approved):
lgtm based on the behavior on my local, while the minimal code change to fix this behavior is https://github.com/appium/appium/pull/21991#issuecomment-3938450575

I'm good for both, this pr itself or simplified to the minimal fixes

mykola-mokhnach:
> lgtm based on the behavior on my local, while the minimal code change to fix this behavior is [#21991 (comment)](https://github.com/appium/appium/pull/21991#issuecomment-3938450575)
> 
> I'm good for both, this pr itself or simplified to the minimal fixes

nice, thanks for confirming it.

PR #22010, opened 2026-02-25
chore(support): Migrate timing.js to typescript

mykola-mokhnach:


packages/support/lib/timing.ts
```diff
@@ -3,15 +3,16 @@ import _ from 'lodash';
 const NS_PER_S = 1e9;
 const NS_PER_MS = 1e6;
 
+/** High-resolution start time: tuple from process.hrtime() or bigint from process.hrtime.bigint() */
+export type HrTime = [number, number] | bigint;
```
eglitise: Do we want to export this? Reading the unchanged parts of the code, it looks like there's some leftover usecases for Node 10, and if we remove those in a future PR, this type won't be needed. If we do export it now, its future removal would technically be a breaking change.

mykola-mokhnach: let me remove the export

KazuCocoa (approved):

PR #22017, opened 2026-02-27
chore(support): migrate zip.js to typescript

mykola-mokhnach:


packages/support/lib/zip.ts
```diff
@@ -274,54 +107,60 @@ async function _extractEntryTo(zipFile, entry, destDir) {
...
-  const zipReadStreamPromise = new B((resolve, reject) => {
+  const zipReadStreamPromise = new B<void>((resolve, reject) => {
     zipReadStream.once('end', resolve);
     zipReadStream.once('error', reject);
   });
   zipReadStream.pipe(writeStream);
 
   // Wait for the zipReadStream and writeStream to end before returning
-  return await B.all([zipReadStreamPromise, writeStreamPromise]);
+  await B.all([zipReadStreamPromise, writeStreamPromise]);
```
eglitise, line 127: Should the `return` keyword not be preserved? I don't think we should change signatures for exported methods without a breaking change

packages/support/lib/zip.ts
```diff
@@ -332,46 +171,53 @@ async function readEntries(zipFilePath, onEntry) {
   zipfile.readEntry();
 
   // Wait for the entries to finish being iterated through
-  return await zipReadStreamPromise;
+  await zipReadStreamPromise;
```
eglitise, line 174: Same here

packages/support/lib/zip.ts
```diff
@@ -478,48 +326,51 @@ async function assertValidZip(filePath) {
...
+  src: ZipSourceOptions = {},
+  opts: ZipCompressionOptions = {}
+): Promise<void> {
   const {level = 9} = opts;
   const {pattern = '**/*', cwd = path.dirname(dstPath), ignore = []} = src;
   const archive = archiver('zip', {zlib: {level}});
-  const stream = fs.createWriteStream(dstPath);
-  return await new B((resolve, reject) => {
+  const outStream = fs.createWriteStream(dstPath);
+  await new B<void>((resolve, reject) => {
```
eglitise, line 363: Same here

packages/support/lib/zip.ts
mykola-mokhnach: this method was defined as return void, typechecks were not dealing properly with it

packages/support/lib/zip.ts
mykola-mokhnach: this promise does not return anything (void->undefined(

packages/support/lib/zip.ts
mykola-mokhnach: this promise does not return anything

eglitise (approved):

PR #20797, opened 2024-12-02
fix{support}: keep backward compatibility for https://github.com/appium/appium/pull/20788

KazuCocoa:
## Proposed changes

To fix https://github.com/appium/appium/issues/20796

https://github.com/appium/appium/pull/20788 modified `installPackage` interface in `@appium/support` package, thus `appium@2.12.1` and older Appium package was broken.

We should bump the `@appium/support`'s major version first, then should drop the argument to avoid this issue. Before that, let's add a temporary workaround for existing version's backward compatibility.

So, after this PR:

`appium@2.12.2` -> has proper fix of https://github.com/appium/appium/pull/20788
`appium@2.12.1` and lower plus `@appium/support` v5 -> still git/github installation has an issue but can fix https://github.com/appium/appium/issues/20796

After this PR, I'd like to bump the support lib's version with this PR's revert as v6, then I'd like to bump the `appium` package's support lib version to use the major version. 
https://github.com/appium/appium/blob/master/packages/appium/package.json#L68

With this:
```
% npx appium driver install xcuitest
✔ Checking if 'appium-xcuitest-driver' is compatible
✔ Installing 'xcuitest' using NPM install spec 'appium-xcuitest-driver'
ℹ Driver xcuitest@7.32.0 successfully installed
- automationName: XCUITest
- platformNames: ["iOS","tvOS"]
% npx appium -v
2.11.5
```

```
% npx appium driver install xcuitest
✔ Checking if 'appium-xcuitest-driver' is compatible
✔ Installing 'xcuitest'
ℹ Driver xcuitest@7.32.0 successfully installed
- automationName: XCUITest
- platformNames: ["iOS","tvOS"]
% npx appium --version
2.12.2
```

## Types of changes

What types of changes does your code introduce to Appium?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [Contributing Guide](https://appium.io/docs/en/latest/contributing/)
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
[...]

PR #22026, opened 2026-02-28
chore(support): drop non-breaking uses of bluebird

eglitise:
Same as previous replacements, but retaining any uses that would be breaking.
* The `fn.ReadFn` type has been marked as deprecated, as its signature matches the now-native implementation of `read.__promisify__`
* `B.reduce` was replaced with ~~a for loop, since the usage here was simply returning as soon as any promise rejected~~ `asyncbox`'s `asyncmap`
* The custom promisified `pipeline` was replaced with the native equivalent from `node:stream/promises`


packages/support/lib/fs.ts
```diff
@@ -425,22 +425,22 @@ export const fs = {
...
   readFile: fsPromises.readFile,
   readlink: fsPromises.readlink,
   realpath: fsPromises.realpath,
   rename: fsPromises.rename,
   stat: fsPromises.stat,
   symlink: fsPromises.symlink,
   unlink: fsPromises.unlink,
-  write: B.promisify(write),
+  write: promisify(write),
   writeFile: fsPromises.writeFile,
```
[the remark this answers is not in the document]
eglitise: While true, `B.promisify(read)` had this same issue, but it had an explicitly defined (and exported) type that incorrectly defined an object as the return value. Now that custom type is (more) correct, so it would make sense to also keep an object as the return value for `promisify(write)`.

packages/support/lib/mjpeg.ts
```diff
@@ -135,21 +134,34 @@ export class MJpegStream extends Writable {
...
-    );
-
-    (this.responseStream as Readable & {pipe<T extends Writable>(dest: T): T})
-      .once('close', onClose)
-      .on('error', onErr)
-      .pipe(this.consumer as unknown as Writable)
-      .pipe(this);
-
-    await startPromise;
+    let timeoutId: ReturnType<typeof setTimeout> | null = null;
```
mykola-mokhnach: I believe node has a dedicated type for the timer

packages/support/lib/util.ts
```diff
@@ -116,24 +117,24 @@ export function localIp(): string | undefined {
...
     resolve = _resolve;
     reject = _reject;
     timer = setTimeout(() => resolve(), ms);
   });
 
-  (delay as B<void> & {cancel: () => void}).cancel = function () {
+  (delay as Promise<void> & {cancel: () => void}).cancel = function () {
     clearTimeout(timer);
-    reject(new B.CancellationError());
+    reject(new Error('cancellation error'));
```
mykola-mokhnach: this change is breaking. Client code may rely on this particular exception class

packages/support/lib/util.ts
```diff
@@ -315,8 +316,10 @@ export async function isSameDestination(
   ...pathN: string[]
 ): Promise<boolean> {
   const allPaths = [path1, path2, ...pathN];
-  if (!(await B.reduce(allPaths, async (a, b) => a && (await fs.exists(b)), true))) {
-    return false;
+  for (const p of allPaths) {
```
mykola-mokhnach: It would probably be faster to use Promise.all or asyncmap

packages/support/lib/mjpeg.ts
```diff
@@ -135,21 +134,34 @@ export class MJpegStream extends Writable {
       this.lastChunk = null;
     };
 
-    const startPromise = new B<void>((res, rej) => {
-      this.registerStartSuccess = res;
-      this.registerStartFailure = rej;
-    }).timeout(
```
mykola-mokhnach, line 141: this might also be a breaking change since bluebird throws a custom TimeoutError that clients may expect

packages/support/lib/mjpeg.ts
eglitise: updated

packages/support/lib/util.ts
eglitise: added a custom error class

packages/support/lib/util.ts
eglitise: fixed

packages/support/lib/mjpeg.ts
eglitise: added a custom error class

mykola-mokhnach: I'm afraid it won't work if clients do something like `if (err instanceof B.TimeoutError)`.

instanceof does not check the class name, but rather the prototype chain of err. Since our TimeoutError (TimeoutError → Error → Object) does not inherit from B.TimeoutError  the client check still fails.

eglitise: Ok, then I guess there's no way around making this a breaking change. In that case I believe it's fine to drop the custom error classes.

mykola-mokhnach: The problem with support package is that we cannot make breaking changes in it unless we bump the major version of the server. Drivers import it directly as peer (e.g. `appium/support` vs `@appium/support`).

Same applies to several other packages that are imported as peers, like base driver, base plugin etc

eglitise: Versions 4, 5 and 6 of the support package did not come with a server bump, why would these changes be different?

eglitise: I could also keep the dependency and only apply the non-breaking changes, and then apply the remainder for Appium 4, if we want.

mykola-mokhnach: v4:

imageUtil: removed Jimp-based methods/constants; stack moves to sharp. -> none of these were used in drivers

v5:

Logger/secure-values APIs removed from support and moved to @appium/logger. -> the stuff was internal and was only used inside of the monorepo

v6:

Revert of backward-compat fix #20797 (behavior that was temporarily preserved is no longer guaranteed). -> was actually a fix

mykola-mokhnach: It makes sense to first check usages of affected functions under appium org modules to make sure we are not going to break anything there (don't forget that older versions of them would still load the most recent version of the `appium/support`). 

As for external drivers and plugins, I'm afraid it would be just a bet as there is no good way to check all of them.

To summarize:

if you feel lucky, verified existing appium packages and don't mind fixing possible issues due the change then go for it

if we want to do it "by the book" then it makes sense to only apply non-breaking changes and finish it as soon as we start working on Appium 4

eglitise: I've now removed any breaking changes and added comments that they should be applied with Appium 4

packages/support/lib/fs.ts
```diff
@@ -425,21 +425,18 @@ export const fs = {
...
-  open: B.promisify(open) as (
-    path: PathLike,
-    flags: OpenMode,
-    mode?: Mode | null
-  ) => B<number>,
+  open: promisify(open),
   openFile: fsPromises.open,
   readdir: fsPromises.readdir,
-  read: B.promisify(read) as unknown as ReadFn,
+  read: promisify(read),
```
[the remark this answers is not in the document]
eglitise: The custom `ReadFn` type (incorrectly) specified the resolved value as `{bytesRead, buffer}`, so now the behavior actually matches that. Another way to apply the fix would be to retain `B.promisify` and update the `ReadFn` signature - not sure which would be preferable.

packages/support/lib/fs.ts
```diff
@@ -65,6 +64,7 @@ export type WalkDirCallback = (
 /**
  * Promisified fs.read signature.
  * @template TBuffer - Buffer type (e.g. NodeJS.ArrayBufferView)
+ * @deprecated use `typeof read.__promisify__` instead
  */
 export type ReadFn<TBuffer extends NodeJS.ArrayBufferView = NodeJS.ArrayBufferView> = (
```
mykola-mokhnach, line 69: I would say it is ok to ditch ReadFn. This type has been introduced very recently while I was migrating support to typescript and has not even published yet.

eglitise: It was previously defined as JSDoc, and was still exported, so removing it would be a breaking change

mykola-mokhnach: hm, I have doubts it was used anywhere, but it's also ok to keep this typedef for now and mark it deprecated

mykola-mokhnach (approved):

KazuCocoa (approved):

PR #22034, opened 2026-03-04
chore(base-driver): Migrate websocket and express logging helpers to typescript

mykola-mokhnach:


KazuCocoa (approved):

packages/base-driver/lib/express/express-logging.ts
```diff
@@ -0,0 +1,84 @@
...
+export const startLogFormatter: RequestHandler = morgan(startLogFormatterHandler, {
+  immediate: true,
+});
+
+// #region Private types and helpers
+type MorganTokens = unknown;
+type FormatFn = (tokens: MorganTokens, req: Request, res: Response) => string;
+
+function endLogFormatterHandler(tokens: MorganTokens, req: Request, res: Response): void {
+  log.info(requestEndLoggingFormat(tokens, req, res));
```
eglitise, line 28: Should  the `(res.jsonResp || '').grey` part have been removed?

mykola-mokhnach: yes, it was a dead code. `jsonResp` property does not exist on `res`. And even if it existed it would not be not a good idea to log the full response

eglitise (approved):

PR #22037, opened 2026-03-04
chore(base-driver): Migrate capability helpers to typescript

mykola-mokhnach:


packages/base-driver/lib/helpers/capabilities.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (_.has(c, 'firstMatch') && _.has(c, 'alwaysMatch')) {
+    return isFirstMatchValid() && isAlwaysMatchValid();
+  }
+  if (_.has(c, 'firstMatch')) {
+    return isFirstMatchValid();
+  }
+  if (_.has(c, 'alwaysMatch')) {
+    return isAlwaysMatchValid();
+  }
+  return false;
```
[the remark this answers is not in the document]
mykola-mokhnach: changing the logic would mean breaking the backward compatibility

packages/base-driver/lib/helpers/capabilities.ts
```diff
@@ -0,0 +1,76 @@
...
+      value = value.trim();
+      let newValue = parseInt(value as string, 10);
+      if (value !== `${newValue}`) {
+        newValue = parseFloat(value as string);
+      }
+      log.warn(
+        `Capability '${cap}' changed from string ('${value}') to integer (${newValue}). ` +
+        `This may cause unexpected behavior`
+      );
+      caps[cap] = newValue;
```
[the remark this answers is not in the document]
mykola-mokhnach: fixed

packages/base-driver/lib/helpers/capabilities.ts
```diff
@@ -0,0 +1,93 @@
...
+  for (const cap of booleanCaps) {
+    const value = oldCaps[cap];
+    if (!_.isString(value)) {
+      continue;
+    }
+
+    if (!['true', 'false'].includes(value.toLowerCase())) {
+      log.warn(
+        `String capability '${cap}' ('${value}') cannot be converted to a boolean. ` +
+        `This may cause an unexpected behavior`
```
eglitise: It's more correct to leave `This may cause unexpected behavior`, same in all other occurrences

mykola-mokhnach: changed

packages/base-driver/lib/helpers/capabilities.ts
```diff
@@ -0,0 +1,83 @@
...
+  const intCaps = _.keys(_.pickBy(desiredCapConstraints, (k) => k.isNumber === true));
+  for (const cap of intCaps) {
+    const value = oldCaps[cap];
+    if (!_.isString(value)) {
+      continue;
+    }
+
+    const intValue = parseInt(value as string, 10);
+    const floatValue = parseFloat(value as string);
+    const newValue = floatValue !== intValue ? floatValue : intValue;
```
eglitise, line 71: Shouldn't this be more like
```ts
const intValue = parseInt(value as string, 10);
const newValue = value !== `${intValue}` ? parseFloat(value as string) : intValue;
```
?

mykola-mokhnach: I believe the updated logic is more precise

eglitise: Hmm, I had assumed `parseFloat` always returns a float, seems not. Then the logic does make sense

eglitise (approved):

packages/base-driver/lib/helpers/capabilities.ts
mykola-mokhnach: yes, it is sometimes surprising that javascript has no separate type for integers. everything is a number :)

KazuCocoa (approved):

PR #22042, opened 2026-03-06
fix(support): add path traversal check in `_extractEntryTo`

eglitise:
Related to #22014

### DIFF
--- a/packages/support/lib/zip.ts
+++ b/packages/support/lib/zip.ts
@@ -90,8 +90,7 @@ export async function _extractEntryTo(
   destDir: string
 ): Promise<void> {
   const dstPath = path.resolve(destDir, entry.fileName);
-  const normalizedDestDir = path.normalize(destDir).toLowerCase();
-  if (!dstPath.toLowerCase().startsWith(normalizedDestDir)) {
+  if (!dstPath.startsWith(path.normalize(destDir))) {
     throw new Error(
       `Out of bound path "${dstPath}" found while processing file ${entry.fileName}`
     );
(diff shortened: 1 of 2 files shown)


packages/support/lib/zip.ts
```diff
@@ -90,6 +90,11 @@ export async function _extractEntryTo(
   destDir: string
 ): Promise<void> {
   const dstPath = path.resolve(destDir, entry.fileName);
+  if (!dstPath.startsWith(path.normalize(destDir))) {
```
mykola-mokhnach: consider using isSubPath util instead

eglitise: added

KazuCocoa (approved):

packages/support/lib/zip.ts
```diff
@@ -90,6 +90,11 @@ export async function _extractEntryTo(
   destDir: string
 ): Promise<void> {
   const dstPath = path.resolve(destDir, entry.fileName);
+  if (!isSubPath(dstPath, destDir)) {
```
mykola-mokhnach, line 93: this could be a breaking change as we use the system unzip by default and it does not case about checking for subpaths

mykola-mokhnach (approved):

packages/support/lib/zip.ts
```diff
@@ -90,6 +90,11 @@ export async function _extractEntryTo(
   destDir: string
 ): Promise<void> {
```
mykola-mokhnach, line 91: maybe it would make sense to add a precrafted zip to assets and verify this behaviour in tests

eglitise: We already had a base64 zipfile in the tests, so I just added a new test to verify this particular fix

packages/support/lib/zip.ts
eglitise: This fix applies to the relative filepaths for individual entries inside the zipfile, which aren't expected to contain backwards path traversal (like in the existing tests for `readEntries()`)

PR #22043, opened 2026-03-06
ci: Add manual release workflow

mykola-mokhnach:
I don't know exactly what might be missing there, but it definitely makes sense to be able to trigger a regular server release from a GH pipeline

@KazuCocoa @eglitise Could you please help me to polish this workflow and add missing steps?

### DIFF
--- a/.github/workflows/manual-release.yml
+++ b/.github/workflows/manual-release.yml
@@ -18,7 +18,7 @@ permissions:
 jobs:
   release:
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/master'
+    if: github.ref == 'refs/heads/master' && contains(fromJSON('["mykola-mokhnach", "jlipps", "eglitise", "KazuCocoa"]'), github.actor)
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/manual-release.yml
+++ b/.github/workflows/manual-release.yml
@@ -1,0 +1,53 @@
+name: Appium Server Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Dry run (no publishing or tagging)'
+        required: true
+        type: boolean
+        default: true
+
+permissions:
+  contents: write
+  pull-requests: write
+  issues: write
+  id-token: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Configure git user
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-release.yml
```diff
@@ -0,0 +1,53 @@
...
+permissions:
+  contents: write
+  pull-requests: write
+  issues: write
+  id-token: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
eglitise: We could limit the pipeline to this repo as well (disabling it on any forks):

`if: github.repository == 'appium/appium' && github.ref == 'refs/heads/master'`

.github/workflows/manual-release.yml
```diff
@@ -0,0 +1,53 @@
...
+        env:
+          DRY_RUN: ${{ inputs.dry_run }}
+        run: |
+          set -euo pipefail
+          if [ "$DRY_RUN" = "true" ]; then
+            echo "Running lerna publish in dry-run mode"
+            npm run release -- --yes --dry-run
+          else
+            echo "Running lerna publish for real"
+            npm run release -- --yes
```
eglitise, line 51: We'll need to double check if the usual release process uses `lerna publish` or `lerna version`

.github/workflows/manual-release.yml
KazuCocoa: Just copied from Copilot suggestion. Perhaps, could we limit the persons who can proceed the job? By default, all users who have write permission can trigger this. Before confirming the behavior, maybe we can add user check as well. After testing, maybe can be relaxed

```
jobs:
  release:
    if: |
      github.ref == 'refs/heads/master' &&
      contains(fromJson('["KazuCocoa","someMaintainer","anotherMaintainer"]'), github.actor)
```

KazuCocoa:
Lg to me.
Publishing via npm won't occur til we configure a trusted publisher, so maybe we could try out with a dry run and see the logs?

mykola-mokhnach:
@jlipps Would you have some time to look at this PR? Feel free to edit it if necessary

.github/workflows/manual-release.yml
mykola-mokhnach: added

KazuCocoa (approved):

jlipps:
The idea looks nice to me, but I don't think it will work without setting up npm auth tokens?

KazuCocoa:
Yes, I'd like to set up npm side once we confirm this works as expected with a dry run. (to avoid accidental release)

mykola-mokhnach:
@KazuCocoa I've merged the PR. Would you have some time later to test the OIDC auth?

KazuCocoa:
Sure, will do it

PR #22064, opened 2026-03-12
docs: add reference pages for caps & insecure features

eglitise:
This PR adds new pages in the Reference section for capabilities and insecure features. Since the server/base driver/official plugins do have handling for some of these, it makes sense to add them in the proper location.

The capabilities and security guides have also been updated and now link to these pages.


packages/appium/docs/en/guides/caps.md
```diff
@@ -2,80 +2,75 @@
 title: Session Capabilities
 ---
 
-"Capabilities" is the name given to the set of parameters used to start an Appium session. The
-information in the set describes what sort of "capabilities" you want your session to have, for
-example, a certain mobile operating system or a certain version of a device. Capabilities are
-represented as key-value pairs, with values allowed to be any valid JSON type, including
-other objects.
+Capabilities are the core parameters used to start an Appium session. They describe various
```
mykola-mokhnach, line 5: it makes sense to mention capabilities are immutable and cannot change during session lifecycle

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,80 +2,75 @@
...
+your Appium client, the prefix may be added automatically or in conjunction with certain interfaces,
+but it is always a good practice to explicitly include it for clarity.
 
-!!! note
+Here are a few commonly used Appium-specific capabilities:
+
+| Capability Name         | Type     | Description                                    |
+|-------------------------|----------|------------------------------------------------|
+| `appium:automationName` | `string` | The name of the Appium driver to use           |
+| `appium:deviceName`     | `string` | The name of a particular device to automate    |
```
mykola-mokhnach: this might be confusing. deviceName is only respected by xcuitest for simulator tests. Real devices use udid. Same for android

packages/appium/docs/en/guides/caps.md
```diff
@@ -2,80 +2,75 @@
...
+[Capabilities Reference document](../reference/session/caps.md). While this common capability set
+is fairly small, it can be greatly extended by Appium drivers and plugins, who can (and should)
+define their own capabilities. Make sure to reference their documentation to learn more about the
+capabilities they support. You can find a list of known drivers in the [Ecosystem Drivers](../ecosystem/drivers.md) document.
 
-    Capabilities are like parameters used when starting a session. After the capabilities are sent
-    and the session is started, they cannot be changed. If a driver supports updating aspects of
-    its behaviour in the course of a session, it will provide a [Setting](./settings.md) for this
-    purpose instead of, or in addition to, a capability.
+Drivers are also able to place more complex constraints on capabilities as a group. For example,
```
mykola-mokhnach, line 46: > requires that at least one of `browserName`, `appium:app`, or `appium:bundleId`

What actually happens
If browserName = 'safari':
The driver sets bundleId to Safari’s bundle and runs Safari.

If appium:app or appium:bundleId is provided (and browserName is not Safari):
The driver configures/installs that app and uses its bundle id.

If none of browserName, appium:app, or appium:bundleId is provided:

validateDesiredCaps does not fail; it only logs that neither app nor bundleId is present and that WDA will start “without the default app”.
During start(), the if (browserName === 'safari') { … } else if (app || bundleId) { … } branches are skipped, so no AUT is configured/installed/launched.
WebDriverAgent is still started, and the session begins with the device on the Home screen, as documented in docs/reference/capabilities.md (for app/bundleId).
So: you can start a session without any of those three caps; it just won’t auto-install or auto-launch an app, and you’re expected to handle launching yourself (e.g. via mobile: launchApp or similar).

packages/appium/docs/en/guides/caps.md
eglitise: This is mentioned later in the document

packages/appium/docs/en/guides/caps.md
eglitise: That's why they are listed as commonly used instead of fully supported everywhere. deviceName is not mentioned in the reference document since the base driver doesn't do anything with it

packages/appium/docs/en/reference/cli/index.md
```diff
@@ -31,5 +31,5 @@ The executable has four main subcommands: `server`, `driver`, `plugin`, and `set
 
 </div>
 
-Appium also recognizes several [environment variables](./env-vars.md), which may be used for
-advanced configuration.
+Appium also recognizes several [environment variables](./env-vars.md) and [insecure feature names](./insecure-features.md),
```
mykola-mokhnach: Appium server

packages/appium/docs/en/reference/session/caps.md
```diff
@@ -0,0 +1,54 @@
...
+
+These capabilities are not directly used in the Appium server/base driver, but are still subject to
+validation checks due to their relevance in other drivers. The actual handling of these
+capabilities is optional and entirely driver-dependent.
+
+| Capability           | Validation                                                    |
+|----------------------|---------------------------------------------------------------|
+| `platformVersion`    | Must be `string`                                              |
+| `appium:app`         | Must be `string`; empty values are ignored                    |
+| `appium:autoLaunch`  | Must be `boolean`                                             |
```
mykola-mokhnach: this cap is deprecated

packages/appium/docs/en/guides/caps.md
eglitise: The part about the absence of all three caps simply opening the home screen is indeed there in the current docs - I simply omitted it because the goal here isn't to explain what exactly happens in the driver, but rather just the concept of additional constraints. But I could add it back if needed.

packages/appium/docs/en/guides/caps.md
eglitise: Though I could replace it with udid. I guess that might actually be more common.

packages/appium/docs/en/guides/caps.md
mykola-mokhnach: my point is that the statement itself is confusing. Neither of these caps are actually required

packages/appium/docs/en/guides/caps.md
mykola-mokhnach: I still assume some people might assume it must work like that in all driver. That is why it makes sense to be more expilcit about its purpose

mykola-mokhnach: perhaps, we could also decide to deprecate it and remove from base driver since Appium 4 as it's not abstract enough

eglitise: Added extra clarification.
`deviceName` is already not used in the base driver, and as for `udid,` I don't think we should remove its handling, since its purpose is quite universal

mykola-mokhnach (approved):

PR #22072, opened 2026-03-14
chore(appium): Migrate logsink to typescript

mykola-mokhnach:


KazuCocoa (approved):

packages/appium/lib/logsink.ts
```diff
@@ -104,37 +109,21 @@ export async function init(args) {
...
-/**
- *
- * @param {ParsedArgs} args
- * @param {string} logLvl
- * @returns {transports.ConsoleTransportInstance}
- */
-function createConsoleTransport(args, logLvl) {
-  /** @type {AppiumConsoleTransportOptions} */
-  const opt = {
-    name: 'console',
```
eglitise, line 137: Should the `name` property be removed here and in 2 other methods?

mykola-mokhnach: yes, name was not in the typedef

eglitise (approved):

PR #22090, opened 2026-03-20
feat: use exact version for dependencies in monorepo packages instead of ^

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/20215

I've decided to apply this change to all monorepo packages. LMK if it only makes change to do it for the server (e.g. `packages/appium`) one.


eglitise:
I would actually suggest using exact versions for packages within this monorepo. It doesn't really make a difference because we always publish all changed packages at once, and we don't expect people to separately update individual packages anyway.

Exact versions can be achieved without needing to add a new script, by just adding `--exact` to the `lerna publish` command: https://github.com/lerna/lerna/tree/main/libs/commands/version#--exact
(`publish` supports all of `version`'s flags, including `--exact`)

mykola-mokhnach:
> I would actually suggest using exact versions for packages within this monorepo. It doesn't really make a difference because we always publish all changed packages at once, and we don't expect people to separately update individual packages anyway.
> 
> Exact versions can be achieved without needing to add a new script, by just adding `--exact` to the `lerna publish` command: https://github.com/lerna/lerna/tree/main/libs/commands/version#--exact (`publish` supports all of `version`'s flags, including `--exact`)

makes sense to me. Changed the configuration to use `exact`.

cc @KazuCocoa @jlipps

package.json
```diff
@@ -57,7 +57,7 @@
     "rebuild": "run-s clean build",
     "reinstall": "run-s clean clean:deps ci",
     "reinstall:angry": "npm run clean:dangerously && npm ci",
-    "release": "lerna publish",
+    "release": "lerna publish --exact",
```
eglitise, line 60: I think the extra flag here and in the workflow file might not be needed, considering it's also added to `lerna.json`

mykola-mokhnach: I just added it to make sure this option is always applied as it has priority over the config one.

eglitise (approved):

KazuCocoa (approved):
lg. we can release more easily today

PR #22118, opened 2026-03-29
fix(docutils): do not silence errors when building without serve

eglitise:
The current behavior of `appium-docs build`, when running without the `--serve` option, silences the error messages from the original function (either `python -m mkdocs build` or `mike deploy`). This makes it impossible to identify the root cause if these functions return an error (for example in CI: https://github.com/appium/appium-xcuitest-driver/actions/runs/23710556037/job/69070407758).

This PR fixes the issue by reading the `err.stderr` field returned by `teen_process`, which is where the original error message is stored, and appending it to the full error message.

Here's a quick comparison when I changed `mkdocs build` to `mkdocs buil`:

Before:
```
Command '/usr/bin/python3 -m mkdocs buil -f /home/ee/Development/appium/packages/appium/docs/mkdocs-en.yml' exited with code 2
```

After:
```
Command '/usr/bin/python3 -m mkdocs buil -f /home/ee/Development/appium/packages/appium/docs/mkdocs-en.yml' exited with code 2
Command error:
Usage: python -m mkdocs [OPTIONS] COMMAND [ARGS]...
Try 'python -m mkdocs -h' for help.

Error: No such command 'buil'.
```

### DIFF
--- a/packages/docutils/lib/util.ts
+++ b/packages/docutils/lib/util.ts
@@ -111,7 +111,7 @@ export async function execWithErrorHandling(
     return await exec(cmd, args, opts);
   } catch (err) {
     const execErr = err as ExecError;
-    const execStderr = execErr.stderr ? `\nCommand error:\n${execErr.stderr}` : '';
-    throw new Error(`${execErr.message}${execStderr}`);
+    execErr.message = execErr.stderr ? `${execErr.message}\nCommand error:\n${execErr.stderr}` : execErr.message;
+    throw execErr;
   }
 }
--- a/packages/docutils/lib/util.ts
+++ b/packages/docutils/lib/util.ts
@@ -7,6 +7,8 @@ import _ from 'lodash';
 import type {SpawnOptions} from 'node:child_process';
 import {spawn} from 'node:child_process';
 import path from 'node:path';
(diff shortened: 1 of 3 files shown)


packages/docutils/lib/util.ts
```diff
@@ -96,3 +98,20 @@ export async function spawnBackgroundProcess(command: string, args: string[], op
...
+  cmd: string,
+  args?: string[],
+  opts?: TeenProcessExecOptions,
+) {
+  try {
+    return await exec(cmd, args, opts);
+  } catch (err) {
+    const execErr = err as ExecError;
+    const execStderr = execErr.stderr ? `\nCommand error:\n${execErr.stderr}` : '';
+    throw new Error(`${execErr.message}${execStderr}`);
```
mykola-mokhnach: not sure if it's an issue, but we are loosing the original stacktrace here.
We could pass the original error as `cause` to fix that or mutate the message of the original error instead of creating a new instance as a workaround

mykola-mokhnach (approved):

packages/docutils/lib/util.ts
eglitise: changed

PR #22125, opened 2026-03-31
chore: update Renovate config

eglitise:
Add 2 new Renovate config options:
* `security:minimumReleaseAgeNpm` for better supply chain security
* `:configMigration` for easier maintenance of Renovate itself

PR #22128, opened 2026-04-01
fix: adjust Renovate config

eglitise:
Follow-up to #22125.
I may have misunderstood the `matchRepositories` option - it seems that it checks against the name of the repo that is using Renovate, instead of the name of the repo of the dependency itself.
This PR changes the option to `matchSourceUrls` instead - this option is also used in Renovate's built-in monorepo presets, so it is likely to work the way we expect it to.

### DIFF
--- a/renovate/default.json
+++ b/renovate/default.json
@@ -21,7 +21,7 @@
     },
     {
       "description": "Disable minimum release age checks for Appium organization packages",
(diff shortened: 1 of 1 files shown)


mykola-mokhnach (approved):

mykola-mokhnach:
chore: -> fix:

PR #22183, opened 2026-04-16
fix: All the rest of linter warnings

mykola-mokhnach:


packages/eslint-config-appium-ts/index.mjs
```diff
@@ -53,6 +53,7 @@ export default defineConfig([
       'import-x/resolver-next': [
         createTypeScriptImportResolver({
           project: ['tsconfig.json', './packages/*/tsconfig.json'],
+          noWarnOnMultipleProjects: true,
```
eglitise: I don't think this option is supported by `eslint-import-resolver-typescript`

packages/eslint-config-appium-ts/index.mjs
```diff
@@ -214,6 +215,7 @@ export default defineConfig([
     languageOptions: {
       parserOptions: {
         projectService: true,
+        noWarnOnMultipleProjects: true,
```
eglitise: This isn't supported by `typescript-eslint` either

packages/eslint-config-appium-ts/index.mjs
mykola-mokhnach: strangely this still helped to get rid of the warning
can you try the same locally?

eglitise: My mistake, it seems the option is supported, but not properly documented in their readme.
The warning is certainly still there, but I'd prefer to actually address it rather than suppress it, or just leave it as is for now.

mykola-mokhnach: > The warning is certainly still there, but I'd prefer to actually address it rather than suppress it, or just leave it as is for now.

weird, I don't see the warning anymore on my local

could you please paste the output of `npm run lint`?

eglitise: Sorry, yes, the warning is there without this line, and adding the line makes it disappear.

eglitise: Can we then either address the warning or remove this line? Silencing it would mean we'll just forget about it, which doesn't solve anything.
It's not coming from any linting rule in the first place, but the resolver package itself.

mykola-mokhnach: > Multiple projects found, consider using a single `tsconfig` with `references` to speed up, or use `noWarnOnMultipleProjects` to suppress this warning

this is exactly what the message itself mentions. Would you like to investigate adding references to tsconfig instead?

eglitise: I would prefer if we could handle this in a separate PR, since it's not directly related to linting rule fixes.

mykola-mokhnach: ok, I can split that

eglitise (approved):

PR #22185, opened 2026-04-17
perf(base-driver): replace fastest-levenshtein with optimized-fastest-levenshtein (Rust, 6–9× faster)

dev-kjma:
## What this does

Replaces [`fastest-levenshtein`](https://github.com/nicolo-ribaudo/fastest-levenshtein) with [`optimized-fastest-levenshtein`](https://github.com/dev-kjma/faster-levenshtein) — a **full drop-in replacement** implemented in Rust via N-API.

The import is updated to use the aliased `distance`/`closest` exports so **zero call-site changes** are needed elsewhere.

## Benchmark results (Apple M2, Node.js v25.2.1, mitata)

| Input size | `fastest-levenshtein` (JS) | `optimized-fastest-levenshtein` (Rust) | Speedup |
|---|---|---|---|
| Medium strings (35–45 chars) | 1,750 ns | **1,320 ns** | **1.33× faster** |
| Long strings (~810 chars) | 342 µs | **55 µs** | **6.2× faster** |
| Very long strings (~700 chars) | 1,280 µs | **138 µs** | **9.3× faster** |

The algorithm complexity drops from **O(n × m)** to **O(n × ⌈m/64⌉)** — processing 64 DP columns per CPU instruction using bitwise arithmetic. The benefit grows with string length.

## Why it's safe

- **139 parity tests** verify bit-for-bit identical output vs `fast-levenshtein` (same underlying Myers algorithm)
- Exports `get(a,b)`, `distance(a,b)` (alias), and `closest(str, arr)` — complete `fastest-levenshtein` API parity
- Zero production dependencies — links only against Node's built-in N-API
- 8 releases since v1.0.0 (Aug 2024), MIT license

**Source:** https://github.com/dev-kjma/faster-levenshtein · https://www.npmjs.com/package/optimized-fastest-levenshtein

### DIFF
--- a/packages/base-driver/lib/basedriver/commands/execute.ts
+++ b/packages/base-driver/lib/basedriver/commands/execute.ts
@@ -10,7 +10,7 @@ import type {
 } from '@appium/types';
 import {mixin} from './mixin';
 import type {BaseDriver} from '../driver';
-import {distance} from 'fastest-levenshtein';
+import { get as distance } from 'optimized-fastest-levenshtein';
 
 declare module '../driver' {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
--- a/packages/base-driver/package.json
+++ b/packages/base-driver/package.json
@@ -54,7 +54,6 @@
     "bluebird": "3.7.2",
     "body-parser": "2.2.2",
     "express": "5.2.1",
-    "fastest-levenshtein": "1.0.16",
     "http-status-codes": "2.3.0",
     "lodash": "4.18.1",
     "lru-cache": "11.3.5",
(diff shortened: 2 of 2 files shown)


eglitise:
I'm all in support of using Rust-based replacements for performance improvements, but I do not think it is a good idea in migrating to a package that's been out for less than a day, from an author who joined GH this month. The fact that you've opened similar PRs in dozens of other repos also serves more as a spam indicator.

We can reevaluate this once you've shown you are willing to maintain the package in the long-term and it's received production uses elsewhere.

dev-kjma:
I understand, Thank you for taking the time to review this, i am trying to gain knowledge and experience in this industry and i genuienly engoy programming like this. In the future what should i do to come across better to Collaborators like you? 

Also, I totally understand why you didnt approve my pull request, this projects launch has been a rough one for me I havent been able to get many dependants or even npm downloads, again thank you for reviewing it

Thanks - KJ

mykola-mokhnach:
I have briefly observed the module implementation and since it uses Rust/NAPI users who install it would need to have build tools and rust installed on their machines, which would potentially fail the whole server package installation as this module is listed in required dependencies.
Consider supplying your node module with prebuilt binaries for different platforms (at least the most popular ones).

Beside of that, the implementation lacks any automated CI/CD process.

PR #22190, opened 2026-04-17
fix(eslint-config-appium-ts): Suppress lint warnings

mykola-mokhnach:
https://github.com/appium/appium/pull/22183#discussion_r3095501257

### DIFF
--- a/packages/eslint-config-appium-ts/index.mjs
+++ b/packages/eslint-config-appium-ts/index.mjs
@@ -53,6 +53,7 @@ export default defineConfig([
       'import-x/resolver-next': [
         createTypeScriptImportResolver({
           project: ['tsconfig.json', './packages/*/tsconfig.json'],
+          noWarnOnMultipleProjects: true,
         })
       ],
       jsdoc: {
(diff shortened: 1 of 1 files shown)


eglitise:
I fail to see the point of this change. All it does is hide a harmless warning, causing it to not actually be resolved, which means it will just be forgotten as a result.
At the same time, I also don't know how complex it is to properly add the references field. If it's non-trivial, I'd rather the warning remains as it is, so it's not forgotten.

mykola-mokhnach:
The warning was always there for some years and nobody was able to resolve it. Thus it would make sense to just mute it. Eventually I am happy with the actual linter performance in the monorepo.

You can also keep a dedicated issue if a reminder is needed.

eglitise:
That's a fair point. In that case I guess we can add a TODO comment for that line, or something similar.

packages/eslint-config-appium-ts/index.mjs
```diff
@@ -214,6 +215,7 @@ export default defineConfig([
     languageOptions: {
       parserOptions: {
         projectService: true,
+        noWarnOnMultipleProjects: true,
```
eglitise: This line isn't needed, the warning is already suppressed by the above one.

eglitise (approved):

packages/eslint-config-appium-ts/index.mjs
mykola-mokhnach: removed

PR #22213, opened 2026-04-27
feat(support): Add various helper utils

mykola-mokhnach:
While removing lodash dependencies from repo modules I can observe that there are some particular lodash constructs that we use more often than others. So it makes sense to put them into shared utils

### DIFF
--- a/packages/support/lib/util.ts
+++ b/packages/support/lib/util.ts
@@ -63,18 +63,7 @@ export function memoize<Fn extends (...args: any[]) => any, Key = unknown>(
   fn: Fn,
   resolver?: (...args: Parameters<Fn>) => Key
 ): Fn & {cache: Map<Key, ReturnType<Fn>>} {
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):

eglitise:
...should these new utils still remain based on Lodash?

packages/support/lib/util.ts
```diff
@@ -374,6 +474,16 @@ export interface LockFileOptions {
   tryRecovery?: boolean;
 }
 
+/** Options for truncateString(). */
+export interface TruncateStringOptions {
+  /** Maximum length of the resulting string. Default 30. */
```
eglitise, line 458: This default isn't used in the method itself

eglitise: I guess the limit comes from lodash itself

mykola-mokhnach: yes, it does

eglitise (approved):

PR #22230, opened 2026-05-02
chore: fix new linting errors in ESLint 10 

eglitise:
The ESLint 10 upgrade is finally unblocked after #22222 - however, the new version [adds three new rules in the recommended preset](https://eslint.org/docs/latest/use/migrate-to-10.0.0#eslint-recommended). This PR addresses errors caused by these new rules, separating these changes from the actual ESLint 10 upgrade PR.

### DIFF
--- a/packages/appium/lib/utils.ts
+++ b/packages/appium/lib/utils.ts
@@ -22,9 +22,9 @@ export const getAppiumModuleRoot = _.memoize(function getAppiumModuleRoot(): str
 export function adler32(str: string, seed: number | null = null): number {
   let a = 1,
     b = 0,
-    M = 0,
-    c = 0,
-    d = 0;
+    M: number,
+    c: number,
+    d: number;
   const L = str.length;
   if (typeof seed === 'number') {
     a = seed & 0xffff;
--- a/packages/appium/lib/utils.ts
+++ b/packages/appium/lib/utils.ts
@@ -21,10 +21,8 @@ export const getAppiumModuleRoot = _.memoize(function getAppiumModuleRoot(): str
  */
 export function adler32(str: string, seed: number | null = null): number {
   let a = 1,
-    b = 0,
-    M: number,
-    c: number,
-    d: number;
+    b = 0;
+  let M: number, c: number, d: number;
   const L = str.length;
   if (typeof seed === 'number') {
     a = seed & 0xffff;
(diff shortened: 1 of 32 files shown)


KazuCocoa (approved):

packages/appium/lib/utils.ts
```diff
@@ -22,9 +22,9 @@ export const getAppiumModuleRoot = _.memoize(function getAppiumModuleRoot(): str
 export function adler32(str: string, seed: number | null = null): number {
   let a = 1,
     b = 0,
-    M = 0,
-    c = 0,
-    d = 0;
+    M: number,
```
mykola-mokhnach: default values are missing

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -130,7 +130,6 @@ export class BaseDriver<
           unexpectedShutdownRejecter = null;
           // @ts-ignore typescript cannot understand this
           unexpectedShutdownResolver?.();
-          unexpectedShutdownResolver = null;
```
mykola-mokhnach, line 133: why it was removed?

packages/support/lib/plist.ts
```diff
@@ -37,7 +37,7 @@ export async function parsePlistFile(
     return {};
   }
 
-  let obj: object = {};
+  let obj: object;
```
mykola-mokhnach, line 40: why the default value is removed?

eglitise:
For all 3 comments - the default values were removed because of the [`no-useless-assignment` rule](https://eslint.org/docs/latest/rules/no-useless-assignment). In all cases, the assigned value was either changed later without being used, or the variable itself was not used anymore afterwards.

mykola-mokhnach:
> For all 3 comments - the default values were removed because of the [`no-useless-assignment` rule](https://eslint.org/docs/latest/rules/no-useless-assignment). In all cases, the assigned value was either changed later without being used, or the variable itself was not used anymore afterwards.

would it make sense then to refactor these variable definitions and move them to corresponding blocks to make the above fact more visible?

packages/appium/lib/utils.ts
```diff
@@ -21,10 +21,8 @@ export const getAppiumModuleRoot = _.memoize(function getAppiumModuleRoot(): str
  */
 export function adler32(str: string, seed: number | null = null): number {
   let a = 1,
-    b = 0,
-    M = 0,
-    c = 0,
-    d = 0;
+    b = 0;
+  let M: number, c: number, d: number;
```
mykola-mokhnach: this works, but my idea was a bit different, e.g. instead of 

```js
let x: number;
while (true) {
  x = 1;
}
```

have something like

```js
while (true) {
  const x = 1;
}
```

eglitise: changed

mykola-mokhnach (approved):

PR #22252, opened 2026-05-06
ci: tune release workflows

eglitise:
* ~~Move the release job's dry-run option into a separate workflow~~
  * ~~This is to prevent dry-runs from triggering a docs build~~
* Remove extra workaround scripts in the release job
* Clarify the purpose of the version job


.github/workflows/manual-release-dry.yml
```diff
@@ -0,0 +1,25 @@
...
+          fetch-depth: 0
+
+      - name: Setup Node and install deps (LTS)
+        uses: ./.github/actions/setup-build
+        with:
+          node-version: lts/*
+          install-command: npm ci
+          run-build: 'true'
+
+      - name: Run Lerna version (dry run)
```
mykola-mokhnach: missing GH_TOKEN

eglitise: I can add it, I just thought it's not required since nothing is being pushed anywhere

.github/workflows/manual-release.yml
```diff
@@ -43,96 +37,4 @@ jobs:
           # for lerna createRelease
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DRY_RUN: ${{ inputs.dry_run }}
```
mykola-mokhnach, line 45: this was removed

.github/workflows/manual-release-dry.yml
```diff
@@ -0,0 +1,25 @@
+name: Appium Server Release (Dry Run)
```
mykola-mokhnach: it is more convenient to have a dedicated workflow for the dry run?

eglitise: Not really, but it's specifically to prevent dry-runs from triggering the docs being built. If there's a way to modify the trigger for the docs build instead, I'd be fine with that as well.

mykola-mokhnach: you may still execute workflow steps based in the input flag value, for example:

```
- name: Run only on main branch
   if: github.ref == 'refs/heads/main'
   run: echo "This runs only on main branch"

- name: Skip if commit message contains [skip-ci]
   if: "!contains(github.event.head_commit.message, '[skip-ci]')"
   run: echo "This runs unless commit message has [skip-ci]"
```

eglitise: The `github.event` API doesn't expose the inputs of the original workflow.
I think I'd prefer to tackle this in a separate PR and only leave the other two changes in this PR.

mykola-mokhnach (approved):

PR #22278, opened 2026-05-15
feat: make assignIpc internal

mykola-mokhnach:


jlipps (approved):
ok, seems like it's worth it!

PR #22280, opened 2026-05-15
chore(base-driver): Simplify async generator logic in IpcSubscription

mykola-mokhnach:


KazuCocoa (approved):

jlipps (approved):
yes, I felt something like this should be possible 😆

PR #22318, opened 2026-05-28
chore(support): Deprecate sharp-related primitives

mykola-mokhnach:
sharp is heavy and there is no need to pull it into every module that depends on support.
The next step would be to move deprecated entities into corresponding drivers and to fetch sharp only where it is needed by the driver.

### DIFF
--- a/packages/support/lib/image-util.ts
+++ b/packages/support/lib/image-util.ts
@@ -4,6 +4,7 @@ let _sharp: typeof sharp | undefined;
 
 /**
  * @deprecated Sharp-backed image helpers are deprecated and will be removed in a future major version.
+ * Consumers are expected to implement this functionality on their side.
  * @returns The sharp module for image processing
  */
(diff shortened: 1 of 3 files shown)


packages/support/lib/image-util.ts
```diff
@@ -3,6 +3,7 @@ import type sharp from 'sharp';
 let _sharp: typeof sharp | undefined;
 
 /**
+ * @deprecated Sharp-backed image helpers are deprecated and will be removed in a future major version.
```
KazuCocoa, line 6: It might need to tell users to implement these by themselves with the deprecated message if we will not prepare an alternative.

KazuCocoa (approved):

packages/support/lib/image-util.ts
mykola-mokhnach: makes sense. Updated

PR #22332, opened 2026-06-01
chore(support): Optimize zip reading performance

mykola-mokhnach:
## Summary

Optimizes the **yauzl-based** JS ZIP read path in `@appium/support` without changing the public API or adding dependencies. Evaluated `node-stream-zip` on real IPAs (including VodQA ~200 MB); full-extract times were similar, so we stayed on yauzl and fixed client-side usage instead.

## Changes

- **Reuse `openReadStream`** — `createOpenReadStream()` binds `promisify` once per archive; used by `ZipExtractor`, `readEntries`, and `_extractEntryTo`.
- **Sequential entry iteration** — new `processYauzlEntriesSequentially()` drives `lazyEntries` with a promise queue (no re-entrant `async` `entry` handlers).
- **`readEntries` early exit** — calls `zipfile.close()` when the callback returns `false` (replaces `zipfile.emit('end')`).
- **`_extractEntryTo`** — uses `pipeline()` instead of manual `pipe()` + dual promises; normalizes names via `toEntryFileName()` when `entry.fileName` is a `Buffer`.
- **`extractAllTo` (JS path)** — zip-slip checks use `isSubPath()` against the once-`realpath`'d dest dir (drops per-entry `realpath`).

**Unchanged:** system `unzip` / PowerShell path, `archiver` write path (`toInMemoryZip`, `toArchive`), `ZipEntry` / `ExtractAllOptions` contracts.

## Benchmark notes (informal)

On `VodQAReactNative.ipa` (~200 MB), naive vs optimized yauzl were within noise for full extract (~230 ms) and Info.plist early-exit (~1 ms). Gains are mainly correctness and lower overhead on hot paths, not orders-of-magnitude speedups on that IPA.

### DIFF
--- a/packages/support/lib/zip.ts
+++ b/packages/support/lib/zip.ts
@@ -25,8 +25,6 @@ const IFMT = 0b1111000000000000;
 const IFDIR = 0b0100000000000000;
 const IFLNK = 0b1010000000000000;
 
-// Internal extraction helpers are defined near the end of the file.
-
 export interface ExtractAllOptions {
   /**
    * The encoding to use for extracted file names.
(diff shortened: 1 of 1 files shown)


packages/support/lib/zip.ts
```diff
@@ -110,77 +111,54 @@ class ZipExtractor {
   }
 
   async extract(): Promise<void> {
-    const {dir, fileNamesEncoding} = this.opts;
+    const {fileNamesEncoding} = this.opts;
```
KazuCocoa, line 114: Is this (`dir` move) to move the dir close to the actual usage?

KazuCocoa (approved):

packages/support/lib/zip.ts
mykola-mokhnach: yes

PR #22348, opened 2026-06-04
chore(support): use native `randomUUID` where possible

gameroman:


packages/support/package.json
```diff
@@ -70,7 +70,6 @@
     "supports-color": "10.2.2",
     "teen_process": "4.1.3",
     "type-fest": "5.7.0",
-    "uuid": "14.0.0",
```
mykola-mokhnach, line 73: unfortunately we cannot drop this dependency yet because uuid APIs are used in `util`

gameroman: ah sorry, missed that

mykola-mokhnach:
@gameroman Could you please address the linter failure?

packages/support/lib/node.ts
```diff
@@ -1,9 +1,9 @@
 import {isWindows} from './system';
 import log from './logger';
 import {exec} from 'teen_process';
+import {randomUUID} from 'node:crypto'
```
gameroman:
```suggestion
import {randomUUID} from 'node:crypto';
```

mykola-mokhnach (approved):

PR #22374, opened 2026-06-10
feat(docutils): add option to only use major version when deploying

eglitise:
This PR adds the optional `use-prefixed-major-deploy-version` flag to the `appium-docs build` command, which causes the docs to be built for the current major version (plus the `v` prefix), instead of the major+minor version. This results in simpler documentation maintenance for projects that commonly receive minor versions (such as drivers).

### DIFF
--- a/packages/docutils/lib/builder/deploy.ts
+++ b/packages/docutils/lib/builder/deploy.ts
@@ -74,9 +74,16 @@ export interface DeployOpts {
    */
   message?: string;
   /**
-   * Version (dir) to deploy build to
+   * Version (dir) to deploy build to. If omitted, the version number is extracted from the
+   * `package.json` file.
    */
   deployVersion?: string;
+  /**
+   * If `true`, when extracting the deployment version from `package.json`, always use a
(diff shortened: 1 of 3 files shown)


mykola-mokhnach (approved):

packages/docutils/lib/builder/deploy.ts
```diff
@@ -234,9 +242,14 @@ async function doDeploy(mikePath: string, args: string[] = [], opts: TeenProcess
 /**
  * Derives a deployment version from `package.json`
  * @param packageJsonPath Path to `package.json` if known
+ * @param usePrefixedMajorVersion Whether to extract a v-prefixed major version
  * @param cwd Current working directory
  */
-async function findDeployVersion(packageJsonPath?: string, cwd = process.cwd()): Promise<string> {
+async function findDeployVersion(
```
mykola-mokhnach: do we need to cover this feature with a unit test?

eglitise: Added. I didn't add any tests if a non-semver version is used, because that results in `readPackageJson` throwing an error.

KazuCocoa (approved):

PR #22414, opened 2026-06-21
chore: Ditch unmaintained midnight-smoker dependency

mykola-mokhnach:
> glob  10.2.0 - 10.4.5
> Severity: high
> glob CLI: Command injection via -c/--cmd executes matches with shell:true - https://github.com/advisories/GHSA-5j98-mcp5-4vw2
> fix available via `npm audit fix --force`
> Will install midnight-smoker@2.0.4, which is a breaking change
> node_modules/midnight-smoker/node_modules/glob
>   midnight-smoker  >=6.0.0
>   Depends on vulnerable versions of glob
>   node_modules/midnight-smoker


eglitise (approved):
Even though this dependency has previously helped in catching some issues, I agree that it is currently effectively unmaintained.

PR #22421, opened 2026-06-23
chore: Update pluginE2EHarness to drop implicit mocha deps

mykola-mokhnach:
This does breaking changes to `plugin-test-support`, although changing it breaks building of other plugins, so I had to bump the major version manually and  also added the corresponding changelog entry, but the CI still passes.

Split of https://github.com/appium/appium/pull/22415 (one of many)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18944,7 +18944,7 @@
     },
     "packages/plugin-test-support": {
       "name": "@appium/plugin-test-support",
-      "version": "1.2.4",
+      "version": "2.0.0",
       "license": "Apache-2.0",
       "dependencies": {
         "async-lock": "1.4.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -18947,6 +18947,7 @@
       "version": "1.2.4",
       "license": "Apache-2.0",
       "dependencies": {
+        "async-lock": "1.4.1",
         "teen_process": "4.1.5"
       },
       "devDependencies": {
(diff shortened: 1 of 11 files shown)


packages/plugin-test-support/tsconfig.json
```diff
@@ -9,7 +9,7 @@
       "appium/support": ["../support"],
       "appium": ["../appium"]
     },
-    "types": ["mocha", "chai", "sinon", "chai-as-promised", "node"]
+    "types": ["chai", "sinon", "chai-as-promised", "node"]
```
eglitise: Are the `chai` and `sinon` types still needed here if the module does not use them?

package-lock.json
```diff
@@ -18947,6 +18947,7 @@
       "version": "1.2.4",
```
eglitise: I guess this version also needs to be bumped (`npm i` might also automatically bump it).
Generally I'd prefer if the breaking change could be extracted separately (despite failing plugin tests), but if Lerna can properly handle a combined PR, I suppose it's fine.

packages/plugin-test-support/tsconfig.json
mykola-mokhnach: removed

package-lock.json
mykola-mokhnach: updated the lock

eglitise (approved):

PR #22450, opened 2026-06-30
chore: Migrate from prettier to oxfmt

mykola-mokhnach:
oxfmt is rust-based and is much faster than prettier, which is highly relevant to our current large codebase


mykola-mokhnach:
Most of the changes are format-only, except of 
- package.json
- .oxfmtrc.json

KazuCocoa (approved):
lg. Will you apply this for all repos, or basically this codebase only?

.oxfmtrc.json
```diff
@@ -0,0 +1,35 @@
...
+  "tabWidth": 2,
+  "useTabs": false,
+  "endOfLine": "lf",
+
+  "semi": true,
+  "singleQuote": true,
+  "quoteProps": "as-needed",
+  "trailingComma": "all",
+  "bracketSpacing": false,
+  "sortImports": true,
```
eglitise: Most of these options are set to the same as their default value in both Prettier and Oxfmt, and can therefore be removed: `tabWidth`, `useTabs`, `endOfLine`, `semi`, `quoteProps` and `trailingComma`

mykola-mokhnach: I prefer when these are set to explicit values, so we have more visibility over how formatter does its work

.oxfmtrc.jsonc
```diff
@@ -0,0 +1,35 @@
...
+    "**/.vscode/**",
+    "**/*.html",
+    "**/generated/**",
+    "packages/schema/lib/appium-config.schema.json",
+    "packages/types/lib/appium-config.ts",
+    "**/*.hbs",
+    ".nx/**",
+    "packages/docutils/**/*.{yml,yaml}"
+  ]
+}
```
eglitise, line 28: * `oxfmt` ignores `node_modules` by default
* `oxfmt` ignores lockfiles by default, so both `**/*-lock.json` and `package-lock.json` can be removed
* `package.json` is also worth removing, because `oxfmt` supports sorting of its fields, which is also enabled by default. Could also come in a follow-up PR if needed.
* `packages/docutils/**/*.{yml,yaml}` is an odd syntax, since the package only has one YAML file (`base-mkdocs.yml`). If we don't want to apply formatting to it, maybe a more obvious pattern would be `**/*mkdocs.{yml,yaml}`?

package.json
```diff
@@ -47,8 +47,8 @@
     "docs:publish": "npm run -w packages/appium publish:docs",
     "doctor": "node ./scripts/doctor-stub.mjs",
     "install-docs-deps": "npm run -w packages/appium install-docs-deps",
-    "format": "prettier -w ./packages ./scripts",
-    "format:check": "prettier --check ./packages ./scripts",
+    "format": "oxfmt \"./packages/**/*\" \"./scripts/**/*\"",
+    "format:check": "oxfmt check \"./packages/**/*\" \"./scripts/**/*\"",
```
eglitise: The correct syntax is `oxfmt --check`, not `oxfmt check`.
Also I wonder if it's fine to just remove the file patterns and also check files in the root directory (such as `package.json`)

.oxfmtrc.json
eglitise: What makes these specific options special? Both formatters have many other options with default values, but they are not explicitly added here.

.oxfmtrc.jsonc
mykola-mokhnach: > oxfmt ignores node_modules by default
> oxfmt ignores lockfiles by default, so both **/*-lock.json and package-lock.json can be removed

removed from ignores

> package.json is also worth removing, because oxfmt supports sorting of its fields, which is also enabled by default. Could also come in a follow-up PR if needed.

removed from ignores and pointed formatter to the root of the monorepo

> packages/docutils/**/*.{yml,yaml} is an odd syntax, since the package only has one YAML file (base-mkdocs.yml). If we don't want to apply formatting to it, maybe a more obvious pattern would be **/*mkdocs.{yml,yaml}

updated the pattern

package.json
mykola-mokhnach: changed to --check

.oxfmtrc.json
mykola-mokhnach: they are not special, just reflect the same structure as editorconfig does

eglitise: Could we clearly separate/distinguish the default values from the modified values for clarity? If needed, oxfmt supports the `.jsonc` configuration file, which would allow adding comments.

.oxfmtrc.jsonc
eglitise: `package-lock.json` can also be removed

.oxfmtrc.json
mykola-mokhnach: changed to .jsonc and added a comment these options are synced from .editorconfig

eglitise (approved):

PR #22458, opened 2026-07-01
fix: Typescript references

mykola-mokhnach:
Fixes https://github.com/appium/appium/actions/runs/28525587302/job/84561015184

### DIFF
--- a/packages/fake-driver/package.json
+++ b/packages/fake-driver/package.json
@@ -50,6 +50,9 @@
     "asyncbox": "6.3.0",
     "xpath": "0.0.34"
   },
+  "devDependencies": {
+    "@appium/driver-test-support": "2.0.0"
+  },
   "peerDependencies": {
     "appium": "^3.0.0-beta.0"
   },
--- a/packages/fake-plugin/tsconfig.json
+++ b/packages/fake-plugin/tsconfig.json
@@ -7,12 +7,5 @@
   },
   "extends": "@appium/tsconfig/tsconfig.plugin.json",
   "include": ["lib", "test"],
-  "references": [
-    {
-      "path": "../base-plugin"
-    },
-    {"path": "../logger"},
-    {"path": "../support"},
-    {"path": "../types"}
-  ]
+  "references": [{"path": "../base-plugin"}, {"path": "../logger"}, {"path": "../support"}, {"path": "../types"}]
 }
--- a/packages/fake-plugin/tsconfig.json
+++ b/packages/fake-plugin/tsconfig.json
@@ -7,5 +7,12 @@
   },
   "extends": "@appium/tsconfig/tsconfig.plugin.json",
   "include": ["lib", "test"],
-  "references": [{"path": "../base-plugin"}]
+  "references": [
+    {
+      "path": "../base-plugin"
+    },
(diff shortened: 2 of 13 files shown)


packages/fake-driver/package.json
```diff
@@ -50,6 +50,9 @@
     "asyncbox": "6.3.0",
     "xpath": "0.0.34"
   },
+  "devDependencies": {
+    "@appium/driver-test-support": "2.0.0"
+  },
```
eglitise, line 55: Is this needed? `base-driver` also uses `fake-driver` for tests, and references it in its `tsconfig`, but it doesn't have it in `devDependencies`

packages/support/tsconfig.json
```diff
@@ -8,5 +8,5 @@
     "types": ["node", "chai", "chai-as-promised"]
   },
   "include": ["lib", "test"],
-  "references": [{"path": "../types"}]
+  "references": [{"path": "../types"}, {"path": "../logger"}, {"path": "../types"}]
```
eglitise: Duplicate entry for `types`

packages/universal-xml-plugin/tsconfig.json
```diff
@@ -7,5 +7,6 @@
     "types": ["node"]
   },
   "extends": "@appium/tsconfig/tsconfig.plugin.json",
-  "include": ["lib", "test"]
+  "include": ["lib", "test"],
+  "references": [{"path": "../base-plugin"}, {"path": "../support"}, {"path": "../types"}]
```
eglitise, line 11: I assume this plugin also needs the `logger` reference, since all other plugins do

packages/fake-driver/package.json
mykola-mokhnach: not really, base driver has it's own (minimized) class for fake driver.
It cannot be true anyway because fake driver already imports base driver, so the reversed import would create a circular reference

packages/universal-xml-plugin/tsconfig.json
mykola-mokhnach: it does not have any logging

packages/support/tsconfig.json
mykola-mokhnach: good catch, fixed

packages/fake-plugin/tsconfig.json
```diff
@@ -7,5 +7,12 @@
   },
   "extends": "@appium/tsconfig/tsconfig.plugin.json",
   "include": ["lib", "test"],
-  "references": [{"path": "../base-plugin"}]
+  "references": [
+    {
+      "path": "../base-plugin"
```
eglitise: Formatting here is inconsistent, strange how oxfmt considers it fine

mykola-mokhnach: 🤷

packages/fake-driver/package.json
eglitise: Sorry, my mistake - what I meant to say was that `base-driver`, like `fake-driver`, has `driver-test-support` in its `tsconfig`, but it does not include `driver-test-support` in its `devDependencies`

PR #22462, opened 2026-07-02
chore(fake-plugin): Drop mocha and chai

mykola-mokhnach:


packages/fake-plugin/test/unit/plugin.spec.ts
```diff
@@ -1,11 +1,8 @@
-import chai from 'chai';
-import chaiAsPromised from 'chai-as-promised';
+import assert from 'node:assert';
```
eglitise: Should it be `node:assert/strict`, like for other already-migrated packages?

mykola-mokhnach: changed to strict

eglitise (approved):

PR #22469, opened 2026-07-03
chore(appium): Drop mocha

mykola-mokhnach:


packages/appium/test/unit/extension/manifest.spec.ts
```diff
@@ -493,32 +491,19 @@ describe('Manifest', function () {
...
+          MockAppiumSupport.fs.glob.rejects(new Error('bogus'));
         });
-        it('should reject', function () {
-          expect(manifest.syncWithInstalledExtensions()).to.be.rejectedWith(Error, 'bogus');
+        it('should reject', async function () {
+          await expect(manifest.syncWithInstalledExtensions()).to.be.rejectedWith(Error, 'bogus');
         });
       });
 
       describe('when the underlying implementation completes with an error', function () {
```
eglitise: This test is no longer any different from the previous test `when the underlying implementation emits "error"`

mykola-mokhnach: yep, deleted

eglitise:
Since this is the last package with mocha, should we also delete `@types/mocha` from `package.json`, the root-level `.mocharc` file, and the references to it in the `build.yml` workflow?

mykola-mokhnach:
> Since this is the last package with mocha, should we also delete `@types/mocha` from `package.json`, the root-level `.mocharc` file, and the references to it in the `build.yml` workflow?

good idea, done

eglitise (approved):

PR #22477, opened 2026-07-06
feat(appium): add --allow-unknown-args to tolerate unrecognized server CLI arguments

AsijitM:
### Problem
`ArgParser#parseArgs` already parses with argparse's `parse_known_args`, and only throws when
unrecognized arguments remain (the `run` subcommand is the exception — it forwards them as
`extraArgs`):

```ts
} else if (unknownArgs?.length) {
  throw new Error(`[ERROR] Unrecognized arguments: ${unknownArgs.join(' ')}`);
}
```

When the Appium CLI is wrapped by external tooling (device farms, CI harnesses, launchers) that
appends extra flags the server doesn't recognize, this hard-exit is difficult to work around
without patching Appium. Today the only options are to strip those args before invocation or to
patch the parser/argparse.

### Fix
Add an opt-in boolean server argument **`--allow-unknown-args`** (default `false`). When set,
unrecognized server arguments are logged (stderr) and ignored instead of exiting. Default behavior
is unchanged — unknown args are still an error.

### Changes
- `packages/schema/lib/appium-config-schema.ts`: add the `allow-unknown-args` boolean server property.
- `packages/schema/lib/appium-config.schema.json`: mirror the generated entry (hand-added to match
  the source; can be regenerated via the schema build if preferred).
- `packages/appium/lib/cli/parser.ts`: when `allowUnknownArgs` is set, log + ignore leftover args
  instead of throwing.

### Notes
- Opt-in and default-off, so it can't mask typos for existing users.
- I hand-mirrored the generated `appium-config.schema.json` entry to keep source/artifact in sync;
  happy to regenerate it via the repo's schema generator and to add unit tests + docs if the
  direction looks good to maintainers.

### DIFF
--- a/packages/appium/lib/cli/parser.ts
+++ b/packages/appium/lib/cli/parser.ts
@@ -323,6 +323,13 @@ export class ArgParser {
       if (unknownArgs?.length && (knownArgs.driverCommand === 'run' || knownArgs.pluginCommand === 'run')) {
         return ArgParser._transformParsedArgs(knownArgs, unknownArgs);
       } else if (unknownArgs?.length) {
+        if (knownArgs.allowUnknownArgs) {
+          // eslint-disable-next-line no-console
+          console.error(
+            `[WARN] Ignoring unrecognized arguments because '--allow-unknown-args' is set: ${unknownArgs.join(' ')}`,
+          );
+          return ArgParser._transformParsedArgs(knownArgs);
+        }
         throw new Error(`[ERROR] Unrecognized arguments: ${unknownArgs.join(' ')}`);
       }
       return ArgParser._transformParsedArgs(knownArgs);
--- a/packages/appium/docs/en/reference/cli/server.md
+++ b/packages/appium/docs/en/reference/cli/server.md
@@ -29,6 +29,7 @@ appium
 |`--address`, `-a`|IPv4/IPv6 address to listen on|string|`0.0.0.0`|
 |`--allow-cors`|Allow web browser connections from any host|boolean|`false`|
 |`--allow-insecure`|List of [insecure features](../../guides/security.md) that should be allowed in this server's sessions. Individual features can be overridden by `--deny-insecure`. Has no effect in combination with `--relaxed-security`.|array<string>|`[]`|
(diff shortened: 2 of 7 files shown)


eglitise:
This seems like a reasonable feature. I'm guessing the tests are failing due to the types and test fixtures not being updated. Could you try adjusting them, similarly to #22279?

AsijitM:
Thanks for the review @eglitise — you're right, I'd updated the schema but not the generated type or the fixtures. Now done, mirroring #22279:

- regenerated `packages/types/lib/appium-config.ts` (adds `AllowUnknownArgsConfig` + the `allow-unknown-args` server property)
- added `allowUnknownArgs: false` to `packages/appium/test/fixtures/default-args.ts`
- documented `--allow-unknown-args` in `docs/en/reference/cli/server.md`

That should settle the failing unit tests (types + fixtures were out of sync with the schema change). PTAL when you get a chance.

eglitise:
Thanks, looks good. Could you also add a simple test for this flag? It should go in `appium/test/e2e/args.e2e.spec.ts`.

packages/appium/lib/cli/parser.ts
```diff
@@ -323,6 +323,13 @@ export class ArgParser {
       if (unknownArgs?.length && (knownArgs.driverCommand === 'run' || knownArgs.pluginCommand === 'run')) {
         return ArgParser._transformParsedArgs(knownArgs, unknownArgs);
       } else if (unknownArgs?.length) {
+        if (knownArgs.allowUnknownArgs) {
+          // eslint-disable-next-line no-console
+          console.error(
```
mykola-mokhnach, line 328: why it's logged as an error if this is a warning?

mykola-mokhnach (approved):

mykola-mokhnach:
@AsijitM Would you have some time to address the above comment about having tests covering the new arg?

eglitise (approved):

AsijitM:
@eglitise Could you please share which version this change will be included in?

eglitise:
@AsijitM it will be in the next release, so Appium 3.6.

PR #22489, opened 2026-07-11
feat: Add centralised Oxc lint and format config

mykola-mokhnach:
## Summary

Introduces `@appium/oxc-config`, a shared [Oxc](https://oxc.rs/) toolchain package that centralises Oxlint and Oxfmt settings for the monorepo and Appium-adjacent projects. The package migrates rules from `@appium/eslint-config-appium-ts` to Oxlint and consolidates the existing Oxfmt settings from `.oxfmtrc.jsonc`, pinning `oxlint`, `oxlint-tsgolint`, and `oxfmt` as dependencies so consumers only need one package.

- Adds `packages/oxc-config` with separate subpath exports: `@appium/oxc-config/oxlint` and `@appium/oxc-config/oxfmt`
- Switches monorepo `lint` / `lint:ci` / `lint:fix` scripts from ESLint to Oxlint
- Replaces root `.oxfmtrc.jsonc` with `oxfmt.config.mjs` wired to the shared format config
- Removes direct `oxlint`, `oxlint-tsgolint`, and `oxfmt` devDependencies from the root `package.json`

## Motivation

Mirror the Prettier → Oxfmt migration with ESLint → Oxlint, using the fast Oxc toolchain while keeping rule coverage as close as possible to the legacy config. Centralising versions and settings in one publishable package makes Appium and extension projects consistent.

## Migration approach

Oxlint config was generated from `packages/eslint-config-appium-ts/index.mjs` via [`@oxlint/migrate`](https://oxc.rs/docs/guide/usage/linter/migrate-from-eslint.html) (`--type-aware --js-plugins=false --with-nursery`) and hand-tuned for severity/options parity.

### Intentionally omitted

| Area | Reason |
| --- | --- |
| `@stylistic/*` rules | Handled by Oxfmt |
| Mocha rules | No longer used in Appium projects |
| `perfectionist/sort-modules` | No native Oxlint equivalent |

### No Oxlint equivalent yet

These rules have no native Oxlint equivalent. They could be restored via [ESLint-compatible JS plugins](https://oxc.rs/docs/guide/usage/linter/js-plugins.html) in a follow-up, at a performance cost:

- `@typescript-eslint/member-ordering`
- `n/no-deprecated-api`
- `jsdoc/require-jsdoc`

### Oxlint findings deferred to follow-up PRs

Initial Oxlint runs surfaced a few issues in existing packages (duplicate imports, non-null assertions, etc.). Those fixes were intentionally **not** included here to avoid unrelated feature-level version bumps in affected packages. They can land in separate PRs per package.

## Package design

```
@appium/oxc-config
├── oxlint.mjs              → rules, defineConfig, ignorePatterns
├── oxfmt.mjs               → format options, defineConfig, ignorePatterns
└── scripts/test-smoke.mjs  → export/config smoke test

dependencies:
  oxlint@1.73.0
  oxlint-tsgolint@0.24.0
  oxfmt@0.59.0
```

Both subpaths export `defineConfig` for type checking and autocomplete. Root configs extend the shared package and add monorepo-specific `ignorePatterns` only.

### Oxfmt and `.editorconfig`

Oxfmt reads [`.editorconfig` properties](https://oxc.rs/docs/guide/usage/formatter/config#editorconfig) for unset formatting options. The shared Oxfmt config:

- Always sets Appium-specific options (`semi`, `singleQuote`, `quoteProps`, `trailingComma`, `bracketSpacing`, `sortImports`)
- Applies Appium fallbacks for `.editorconfig`-mapped options (`printWidth`, `tabWidth`, `useTabs`, `endOfLine`, `insertFinalNewline`) only when the nearest `.editorconfig` does not define the corresponding property
- Leaves defined options unset so Oxfmt can still apply section-specific values (including glob overrides) at format time

This keeps the monorepo aligned with its root `.editorconfig` while giving other org repos sensible defaults when they have no `.editorconfig`, or only a partial one.

**Note:** ESLint dependencies and `eslint.config.mjs` remain for now — only the `lint` scripts were switched. A follow-up can remove the ESLint ecosystem once confidence is high.

## Test plan

- [x] `npm run -w packages/oxc-config test:smoke`
- [x] `npm run lint:ci`
- [x] `npm run format:check`


eglitise:
I suggest first applying the Oxlint findings in a separate PR, otherwise the affected packages will also receive a feature level version bump

packages/oxtools-config/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "exports": {
+    "./oxlint": "./oxlint.mjs",
+    "./oxfmt": "./oxfmt.mjs"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "test:smoke": "node --input-type=module -e \"import oxlintConfig, {defineConfig, ignorePatterns} from './oxlint.mjs'; import oxfmtConfig, {fmtIgnorePatterns} from './oxfmt.mjs'; if (!oxlintConfig?.overrides?.length) throw new Error('invalid oxlint config'); if (!ignorePatterns?.length) throw new Error('invalid ignorePatterns'); if (typeof defineConfig !== 'function') throw new Error('invalid defineConfig'); if (!oxfmtConfig?.singleQuote) throw new Error('invalid oxfmt config'); if (!fmtIgnorePatterns?.length) throw new Error('invalid fmtIgnorePatterns')\""
+  },
```
eglitise: This is a bit hard to read, how about extracting this into a separate script file?

packages/oxtools-config/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+Create `oxfmt.config.mjs` in your project root:
+
+```js
+import appiumFmtConfig, {fmtIgnorePatterns} from '@appium/oxtools-config/oxfmt';
+
+export default {
+  ...appiumFmtConfig,
+  ignorePatterns: [...fmtIgnorePatterns],
+};
```
eglitise: oxfmt also supports `defineConfig` for type checking and autocomplete

eglitise:
> No Oxlint equivalent yet

While true, Oxlint also supports ESLint-compatible JS plugins, so we can continue using these rules with Oxlint: https://oxc.rs/docs/guide/usage/linter/js-plugins.html

Of course, this would affect performance.

packages/oxtools-config/README.md
```diff
@@ -0,0 +1,77 @@
...
+```bash
+npx oxlint -c oxlint.config.mjs .
+```
+
+### Oxfmt
+
+Create `oxfmt.config.mjs` in your project root:
+
+```js
+import appiumFmtConfig, {fmtIgnorePatterns} from '@appium/oxtools-config/oxfmt';
```
eglitise: For consistency I would suggest the same names as for `/oxlint`: `appiumConfig` and `ignorePatterns`

mykola-mokhnach: done

packages/oxtools-config/README.md
mykola-mokhnach: added defineConfig

packages/oxtools-config/package.json
mykola-mokhnach: moved to scripts

packages/oxtools-config/oxfmt.mjs
```diff
@@ -0,0 +1,34 @@
...
+  '**/*.html',
+  '**/generated/**',
+  '**/*.hbs',
+  '**/*mkdocs.{yml,yaml}',
+];
+
+/**
+ * Shared Oxfmt configuration for Appium projects.
+ *
+ * Format options are synced with the monorepo `.editorconfig`.
```
eglitise: This is fine in the context of the monorepo, but if we want to reuse this config across the org, most repos don't have an `.editorconfig` file to pull from, so its rules won't be applied. It might be best to redefine those rules here as well.

eglitise:
I think the name could be simplified to `@appium/oxc-config`, since `oxc` is the official name for this collection of tools

packages/oxtools-config/oxfmt.mjs
mykola-mokhnach: updated the logic to set appropriate defaults if no editorconfig is present

mykola-mokhnach:
> I think the name could be simplified to `@appium/oxc-config`, since `oxc` is the official name for this collection of tools

Renamed

KazuCocoa (approved):
lgtm

packages/oxc-config/oxfmt.mjs
```diff
@@ -0,0 +1,168 @@
...
+  tabWidth: 2,
+  useTabs: false,
+  endOfLine: 'lf',
+  insertFinalNewline: true,
+};
+
+/** @type {Record<keyof typeof editorConfigFallbacks, string[]>} */
+const editorConfigOptionKeys = {
+  printWidth: ['max_line_length'],
+  tabWidth: ['indent_size', 'tab_width'],
```
eglitise: oxfmt does not read the `tab_width` property, so `editorConfigOptionKeys` can be simplified to a string->string mapping.

packages/oxc-config/oxfmt.mjs
```diff
@@ -0,0 +1,168 @@
...
+    sortImports: true,
+    ...resolveEditorConfigFallbacks(cwd),
+  };
+}
+
+/**
+ * Shared Oxfmt configuration for Appium projects.
+ */
+/** @type {Record<string, unknown>} */
+const config = {
```
eglitise, line 163: It would be good to also preserve the schema entry (`"$schema": "./node_modules/oxfmt/configuration_schema.json"`), although I'm not sure if it should go here or in the consumer config file.

packages/oxc-config/oxfmt.mjs
mykola-mokhnach: simplified

packages/oxc-config/oxfmt.mjs
eglitise: Correction, I missed that we're now using the JS file format instead of the JSONC one.

mykola-mokhnach: For our setup I don't think we need to add it back. We moved to oxfmt.config.mjs with defineConfig from oxfmt, which already gives editor autocomplete and type checking via TypeScript types. $schema is mainly useful for JSON/JSONC configs (.oxfmtrc.json, .oxfmtrc.jsonc), where editors rely on it for validation.

packages/oxc-config/oxlint.mjs
```diff
@@ -0,0 +1,255 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+export {defineConfig} from 'oxlint';
+
+/**
+ * Default ignore patterns for Appium Oxlint projects.
+ *
+ * Oxlint does not inherit `ignorePatterns` via `extends`, so consumers should
+ * use `resolveIgnorePatterns()` in their root config. See README for usage.
```
eglitise: The README has no mention of `resolveIgnorePatterns`, and simply uses the same spread syntax as for oxfmt. I suppose the function is not needed then?
It's also worth noting that oxlint automatically ignores files matched by `.gitignore`, so perhaps `readGitignorePatterns` isn't needed either.

packages/oxc-config/oxlint.mjs
```diff
@@ -0,0 +1,255 @@
...
+ */
+export const ignorePatterns = [
+  '**/.*',
+  '**/*-d.ts',
+  '**/*.min.js',
```
eglitise: oxlint also automatically ignores files with `.min.` in the name, so this line can be omitted

packages/oxc-config/oxlint.mjs
```diff
@@ -0,0 +1,255 @@
...
+  },
+  options: {
+    typeAware: true,
+  },
+  env: {
+    builtin: true,
+  },
+  overrides: [
+    {
+      files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'],
```
eglitise, line 178: I wonder if this block can be moved up to the config root? Oxlint only supports JS/TS files plus frameworks like React/Vue/Svelte etc, and this file pattern already covers all the supported file extensions we use in the organization anyway.

packages/oxc-config/oxlint.mjs
```diff
@@ -0,0 +1,255 @@
...
+      plugins: ['promise', 'import', 'typescript', 'unicorn'],
```
eglitise, line 179: Could probably also add the `eslint` and `oxc` plugins, although that might also be done in the future, and this PR can then aim to only match the existing ESLint config.

packages/oxc-config/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "scripts": {
+    "test:smoke": "node scripts/test-smoke.mjs"
+  },
+  "dependencies": {
+    "oxfmt": "0.59.0",
+    "oxlint": "1.73.0",
+    "oxlint-tsgolint": "0.24.0"
+  },
+  "engines": {
+    "node": "^20.19.0 || ^22.12.0 || >=24.0.0",
```
eglitise, line 44: Worth noting that `.ts` config files technically require Node `^22.18.0 || >=24.0.0`, though I understand that Lerna would attempt to synchronize this field with the rest of the monorepo

packages/oxc-config/oxlint.mjs
mykola-mokhnach: yes, this could be done in a separate PR

packages/oxc-config/package.json
mykola-mokhnach: true story

packages/oxc-config/oxlint.mjs
mykola-mokhnach: yep, removed it completely

packages/oxc-config/oxlint.mjs
mykola-mokhnach: removed it from linter ignores

packages/oxc-config/oxlint.mjs
mykola-mokhnach: Oxlint's extends only merges rules, plugins, and overrides — not root-level env or globals ([docs](https://oxc.rs/docs/guide/usage/linter/nested-config#extending-configuration-files)). Those must stay in an override for extends: [appiumConfig] to work.

Also, plugins must stay in the file-scoped override: moving them to root changed which rules ran and surfaced 52 new errors. Keeping plugins: [] at root plus plugin activation in the override preserves existing behavior.

Overrides now only covers environment/globals setup, TS-specific rules, and test relaxations — not the full rule set. Smoke test and lint:ci both pass with 0 errors.

packages/oxc-config/oxfmt.mjs
```diff
@@ -0,0 +1,168 @@
...
+  '**/.*',
+  '**/build/**',
+  '**/fixtures/**',
+  '**/*.min.*',
+  '**/*.md',
+  '**/*.html',
+  '**/generated/**',
+  '**/*.hbs',
+  '**/*mkdocs.{yml,yaml}',
+];
```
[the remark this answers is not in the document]
mykola-mokhnach: I don't want to include these to common patterns

packages/oxc-config/oxlint.mjs
```diff
@@ -0,0 +1,210 @@
...
+    },
+    {
+      files: ['**/*.{ts,tsx,mtsx}'],
+      plugins: ['typescript'],
+      rules: {
+        'typescript/no-floating-promises': 'warn',
+      },
+    },
+    {
+      files: ['**/test/**', '*.spec.*js', '-specs.*js', '*.spec.ts'],
```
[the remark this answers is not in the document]
mykola-mokhnach: only kept `'**/test/**'` pattern

packages/oxc-config/scripts/test-smoke.mjs
```diff
@@ -0,0 +1,79 @@
...
+  throw new Error('invalid editorConfigFallbacks');
+}
+
+const optionsWithoutEditorConfig = createFormatOptions('/tmp/no-editorconfig-here');
+if (optionsWithoutEditorConfig.printWidth !== 120) {
+  throw new Error('expected editorConfigFallbacks when .editorconfig is absent');
+}
+
+const monorepoRoot = path.resolve(import.meta.dirname, '../../..');
+const optionsWithEditorConfig = createFormatOptions(monorepoRoot);
```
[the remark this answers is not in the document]
mykola-mokhnach: updated

packages/oxc-config/scripts/test-smoke.mjs
```diff
@@ -0,0 +1,79 @@
...
+
+if (typeof defineOxfmtConfig !== 'function') {
+  throw new Error('invalid oxfmt defineConfig');
+}
+
+if (editorConfigFallbacks.printWidth !== 120 || editorConfigFallbacks.endOfLine !== 'lf') {
+  throw new Error('invalid editorConfigFallbacks');
+}
+
+const optionsWithoutEditorConfig = createFormatOptions('/tmp/no-editorconfig-here');
```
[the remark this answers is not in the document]
mykola-mokhnach: updated

packages/oxc-config/oxfmt.mjs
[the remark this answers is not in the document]
eglitise: Maybe `**/.*/**` would work? It would then also include files inside directories like `.github`, however.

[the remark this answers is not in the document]
mykola-mokhnach: I'm inclined to keep it as is for now. We can always tune these settings if any issues pop up

eglitise (approved):

PR #22532, opened 2026-07-22
chore: Linter warnings

mykola-mokhnach:
## Summary

- Fix oxlint warnings: drop non-null assertions in `createAppiumURL` by giving `session`/`pathname` empty-string defaults, and prefix the unused inferred type param in `system.spec.ts` with `_`.
- Rename the CI step from `ESLint` to `Lint` and swap the stale `.eslintignore` path trigger in `build.yml` for `oxlint.config.mjs`/`oxfmt.config.mjs`, reflecting the earlier oxlint migration.
- Bump the `test:e2e` timeout from 90s to 200s in `packages/appium/package.json`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,8 @@ on:
       - 'lerna.json'
       - '.npmrc'
       - '.github/workflows/build.yml' # this file
-      - '.eslintignore'
+      - 'oxlint.config.mjs'
+      - 'oxfmt.config.mjs'
       - '!**/sample-code/**'
       - '!packages/*/docs/**'
   pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -191,7 +193,7 @@ jobs:
           node-version: lts/*
           install-command: npm ci
           run-build: 'false'
-      - name: ESLint
+      - name: Lint
         run: npm run lint:ci
       - name: Prettier
```
eglitise: I guess this step can also be renamed

eglitise:
Also I guess `chore` is more appropriate

.github/workflows/build.yml
mykola-mokhnach: renamed to 'Format check'

mykola-mokhnach:
> Also I guess `chore` is more appropriate

The PR fixes linter warnings, that is why I categorized it as a fix

eglitise:
I tend to use `fix` only for changes that are visible to the end user. Linting/formatting/refactoring/etc is purely internal.

eglitise (approved):

mykola-mokhnach:
> I tend to use `fix` only for changes that are visible to the end user. Linting/formatting/refactoring/etc is purely internal.

changed to `chore`

PR #22556, opened 2026-07-29
chore(base-driver): Group route definitions

mykola-mokhnach:
## Summary

- Split packages/base-driver/lib/protocol/routes.ts (633 lines, one flat object) into a routes/ directory of small, per-spec files, assembled via nested barrel index.ts files. No route content, types, or behavior changed — METHOD_MAP, ALL_COMMANDS, NO_SESSION_ID_COMMANDS, and routeToCommandName all still export from ./routes with identical output.
- routes/w3c/ — one file per W3C WebDriver spec section (session, navigation, window, element, document, cookie, actions, alert, capture) plus the smaller W3C specs WebDriver extends automation to (permissions, device-posture, sensor, webauthn, payment, fedcm, pressure, privacy, reporting), merged by w3c/index.ts.
- routes/appium/ — Appium-specific routes split into session, settings, commands, device, merged by appium/index.ts.
- routes/jsonwp.ts — legacy JSONWP routes, now also absorbing the two previously-unknown.ts routes (receive_async_response, element/:elementId) whose origin was unclear but which fit naturally alongside other legacy/deprecated routes.
- routes/mjsonwp.ts — unchanged, mobile JSONWP routes.
- routes/other.ts (renamed from other-protocols.ts) — now holds only the genuinely non-W3C-track routes: Selenium/Chromium logging, Chromium CDP execute, WHATWG custom handlers, and the Privacy CG storage-access draft.
- Top-level routes/index.ts merges all groups into METHOD_MAP and keeps the existing routeToCommandName/cache logic verbatim.

## Why

The route map had grown into a single 600+ line object that was hard to scan or diff. Grouping by spec/section makes it easy to find where a given route lives, review changes to one area in isolation, and cross-reference against the actual W3C/Appium spec sections.


eglitise:
The idea of grouping the routes makes sense, but I'm not sure about the proposed groups.
I find it useful to clearly separate standard W3C WebDriver routes from extended routes, which is difficult to achieve in the current proposal. It is also worth noting that all the supported extended spec routes are W3C regardless of their website host URL, so it should not be used to classify any of them as 'non-W3C'.
In addition, the webdriver spec doesn't specify any grouping for the standard routes, so the subgroups here may seem somewhat arbitrary, making it more difficult to find a specific route definition.

I would propose the following groups instead:
* JSONWP (same as proposed)
* MJSONWP (same as proposed)
* W3C standard (all standard routes in one file)
* Appium (either all routes in one file, or separate the `/device` routes into a second file)
* Other extension spec protocols (one file per extension spec, for 13 files total. Similar to proposed, but also includes the specs currently labeled as 'non-W3C')

mykola-mokhnach:
Good points, thanks! Pushed a restructure along the lines you suggested:

* `w3c.ts` — all standard WebDriver routes back in a single file (agreed the spec doesn't define subgroups, so splitting them was arbitrary)
* `appium.ts` + `appium-device.ts` — kept device routes split out since that group is fairly large on its own
* `jsonwp.ts` / `mjsonwp.ts` — unchanged
* `extensions/` — one file per extension spec (13 total), with the ones I'd previously filed as "non-W3C" (Selenium log routes, Chromium CDP, custom handlers, storage access) folded in alongside the rest — you're right that spec host URL isn't a meaningful way to split these

Let me know if this looks better.

packages/base-driver/test/unit/protocol/routes.spec.ts
```diff
@@ -43,7 +43,11 @@ describe('Routes', function () {
         }
       }
       const hash = shasum.digest('hex').substring(0, 8);
-      expect(hash).to.equal('ae870415');
+      // Hash intentionally updated: the route definitions were reorganized into
+      // multiple files (see git history), changing iteration order but not the
+      // actual set of routes/commands/params. Update this value again only when
+      // an intentional route/command/param change is made.
```
eglitise: Considering the hash is changed every time the routes are updated, I think it's fine to only leave the last sentence of this comment

eglitise (approved):

packages/base-driver/test/unit/protocol/routes.spec.ts
mykola-mokhnach: Done, trimmed to just the last sentence.

PR #22557, opened 2026-07-29
feat(universal-xml-plugin)!: Migrate package to ESM

mykola-mokhnach:
## Summary

Migrate @appium/universal-xml-plugin from CommonJS to native ESM:

- Add "type": "module" and an exports map (., ./package.json) to package.json.
- Set module/moduleResolution to NodeNext in tsconfig.json.
- Add explicit .js extensions to all relative imports, as required under NodeNext resolution.
- Replace CJS-only globals with their ESM equivalents: require.main === module → import.meta.url check in the CLI entrypoint; __filename → fileURLToPath(import.meta.url) in test fixtures.
- Add a local tsconfig.json paths override for appium/driver.js / appium/plugin.js, since the appium core package has no exports map and its bare-specifier subpaths don't resolve under strict ESM module resolution the way they did under classic CommonJS resolution.

BREAKING CHANGE: @appium/universal-xml-plugin is now an ESM-only package. It can no longer be loaded via CommonJS require(); consumers must use import or dynamic import(). Appium's own extension loader already loads plugins via dynamic import(), so appium plugin install/use universal-xml is unaffected. Deep imports into the package's internals are no longer possible — only the public entry point is exposed via exports.

### DIFF
--- a/packages/universal-xml-plugin/lib/index.ts
+++ b/packages/universal-xml-plugin/lib/index.ts
@@ -1,9 +1,10 @@
-export {UniversalXMLPlugin} from './plugin';
-export {transformSourceXml} from './source';
+export {UniversalXMLPlugin} from './plugin.js';
+export {transformSourceXml} from './source.js';
 import fs from 'node:fs/promises';
+import {fileURLToPath} from 'node:url';
 
-import {UniversalXMLPlugin} from './plugin';
-import {transformSourceXml} from './source';
+import {UniversalXMLPlugin} from './plugin.js';
+import {transformSourceXml} from './source.js';
 
 export default UniversalXMLPlugin;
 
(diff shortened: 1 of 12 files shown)


packages/universal-xml-plugin/lib/index.ts
```diff
@@ -33,6 +34,6 @@ export async function main(): Promise<void> {
   }
 }
 
-if (require.main === module) {
+if (process.argv[1] && fileURLToPath(import.meta.url) === process.argv[1]) {
```
eglitise, line 37: I probably don't quite understand the purpose of this guard - only run if the filename is passed as an argument? Is it still needed with ESM?

mykola-mokhnach: this is used in the smoke test

eglitise (approved):

Issue #22051, opened 2026-03-08
Appium 4


mykola-mokhnach:
We are going to use this ticket as the main place to collect all ideas we'd like to implement or change in the next major server release.

gioboa:
Hi 👋 
Do you mind if I take "Change default CLI logging level to info"? Happy to open a PR against `appium4` today if nobody's on it.

eglitise:
> Do you mind if I take "Change default CLI logging level to info"?

Thanks for the offer, but we are handling all of these items internally.

PR #22558, opened 2026-07-29
feat(storage-plugin): Add route aliases

mykola-mokhnach:
## Summary
Addresses the storage-plugin item from #22051 ("Update storage plugin endpoints to follow W3C format: `/storage/*` -> `/appium/storage/*`").

- Adds new primary endpoints under `/appium/storage/*` (`add`, `list`, `reset`, `delete`), matching the `appium/`-namespaced convention used elsewhere in the W3C protocol.
- Keeps the existing `/storage/*` endpoints working as deprecated aliases for backward compatibility — each route is registered under both prefixes and dispatches to the same handlers.
- Logs a one-time warning per deprecated route path when a client hits `/storage/*`, pointing it to the `/appium/storage/*` replacement.
- The WebSocket `stream`/`events` paths returned by `add` now honor whichever prefix the client actually called through, so both old and new clients get correctly-prefixed WS URLs.
- Updates the plugin README (example + a note about the deprecated alias) and the e2e test (now exercises `/appium/storage/*` as the primary path, with an added case verifying `/storage/*` still works).

### DIFF
--- a/packages/storage-plugin/README.md
+++ b/packages/storage-plugin/README.md
@@ -64,7 +64,7 @@ If a folder with the same name already exists in the storage, an error will be t
 [Refer to the Appium documentation](https://appium.io/docs/en/latest/reference/api/plugins/#storage-plugin).
 
 > [!NOTE]
-> Prior to plugin version 1.1.8, all endpoints were mounted under the `/storage` prefix (e.g. `/storage/add`)
+> Prior to plugin version 1.2.0, all endpoints were mounted under the `/storage` prefix (e.g. `/storage/add`)
 > instead of `/appium/storage`. These legacy routes are still available for backward compatibility,
 > but are deprecated and will be removed in a future version of the plugin.
 
--- a/packages/storage-plugin/README.md
+++ b/packages/storage-plugin/README.md
@@ -39,7 +39,7 @@ The procedure for storing a local file on the Appium server is as follows:
 
 - Calculate the [SHA1](https://en.wikipedia.org/wiki/SHA-1) hash of the source file
 - Decide the name of the destination file in the server storage (it can be the same as the original file name)
-- Send a `POST` request to the `/storage/add` endpoint, which will return the `events` and `stream` websocket paths
+- Send a `POST` request to the `/appium/storage/add` endpoint, which will return the `events` and `stream` websocket paths
 - Connect to both web sockets
 - Start listening for messages on the `events` web socket. Each message there is a JSON object wrapped
   to a string. The message must be either `{"value": {"success": true, "name":"app.ipa","sha1":"ccc963411b2621335657963322890305ebe96186"}}` to notify about a successful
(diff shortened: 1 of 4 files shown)


packages/storage-plugin/lib/plugin.ts
```diff
@@ -14,9 +14,16 @@ import type {AddRequestResult, ItemOptions, StorageItem} from './types';
 const log = logger.getLogger('StoragePlugin');
 
 let SHARED_STORAGE: Storage | null = null;
-const STORAGE_PREFIX = '/storage';
+const STORAGE_PREFIX = '/appium/storage';
+/**
+ * @deprecated Use `STORAGE_PREFIX` (`/appium/storage`) instead. This alias only exists for
```
KazuCocoa, line 19: 👍

KazuCocoa (approved):
lg entirely

packages/storage-plugin/README.md
```diff
@@ -63,6 +63,11 @@ If a folder with the same name already exists in the storage, an error will be t
 
 [Refer to the Appium documentation](https://appium.io/docs/en/latest/reference/api/plugins/#storage-plugin).
 
+> [!NOTE]
+> Prior to plugin version 1.1.8, all endpoints were mounted under the `/storage` prefix (e.g. `/storage/add`)
```
eglitise: I guess this should be version 1.2.0, since the current latest version is 1.1.8, and this PR will bump it to 1.2.0.

eglitise:
It would be good to also update the Reference docs: https://github.com/appium/appium/blob/master/packages/appium/docs/en/reference/api/plugins.md

packages/storage-plugin/lib/plugin.ts
```diff
@@ -25,40 +32,58 @@ const STORAGE_ADDITIONS_CACHE: LRUCache<string, () => any> = new LRUCache({
...
-        logger.markSensitive(util.truncateString(JSON.stringify(body.value), {length: 200})),
+    const buildHandler =
+      (methodName: string, basePath: string, isDeprecated: boolean) => async (req: Request, res: Response) => {
+        if (isDeprecated && !deprecatedRoutesLogged.has(req.path)) {
+          deprecatedRoutesLogged.add(req.path);
+          log.warn(
+            `The '${req.path}' endpoint has been deprecated and will be removed in a future version ` +
+              `of Appium. Please use '${req.path.replace(DEPRECATED_STORAGE_PREFIX, STORAGE_PREFIX)}' instead`,
+          );
+        }
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch — fixed in 71fd0ce4a. The warning is now keyed off the statically registered route path instead of `req.path`, so it can't be logged twice for the same endpoint due to a trailing slash. Also reworded to say "future version of the storage plugin" instead of "Appium", to match the README.

packages/storage-plugin/README.md
mykola-mokhnach: Fixed in 71fd0ce4a — now says 1.2.0.

mykola-mokhnach:
Updated in 71fd0ce4a — `packages/appium/docs/en/reference/api/plugins.md` now shows the `/appium/storage/*` paths and notes the deprecated `/storage` alias.

eglitise (approved):

PR #22585, opened 2026-08-02
feat(docutils)!: Migrate to ESM

mykola-mokhnach:
## Summary

- Migrates @appium/docutils to a native ESM package ("type": "module"), following the same pattern already applied to @appium/universal-xml-plugin (#22557), @appium/strongbox (#22563), @appium/relaxed-caps-plugin (#22573), @appium/images-plugin (#22574), and @appium/execute-driver-plugin (#22576).
- Adds "type": "module" and an explicit exports map (., ./package.json) to package.json, and sets module/moduleResolution to "NodeNext" in tsconfig.json.
- Adds .js extensions to all relative imports across lib/, bin/, and test/, including directory-index imports that resolved implicitly under CommonJS (./utils → ./utils/index.js, ./builder → ./builder/index.js, ./command → ./command/index.js).
- Replaces `require('consola')` in lib/constants.ts and lib/logger.ts with proper `import`/`import type` statements. The original code destructured `LogLevel` and `ConsolaInstance` off a `require()` call purely to reference their types (since `require()` types as `any`); these are now imported as types directly from `consola`.
- Replaces `__dirname` in lib/constants.ts (used to compute `PKG_ROOT_DIR`) with `path.dirname(fileURLToPath(import.meta.url))`.
- Replaces `require.main === module` in lib/cli/index.ts with `process.argv[1] && fileURLToPath(import.meta.url) === process.argv[1]`, matching the check used in the "stawhere in this migration series.
- Converts bin/appium-docs.js to ESM import syntax with explicit .js extensions.
- No test-mocking rewrite was needed here (unlike #22563/#22574) — none of docutils' unit/e2e tests stub internal modules, so the existing test:unit/test:e2e scripts are unchanged.

BREAKING CHANGE: @appium/docutils is now an ESM-only package. It can no longer be loaded via CommonJS require(); consumers must use import or dynamic import(). Deep imports into the package's internals are no longer possible — only the public entry point is exposed via exports.

### DIFF
--- a/packages/docutils/bin/appium-docs.js
+++ b/packages/docutils/bin/appium-docs.js
@@ -1,10 +1,8 @@
 #!/usr/bin/env node
 // @ts-check
 
-'use strict';
-
-const {main} = require('../build/lib/cli');
-const {getLogger} = require('../build/lib/logger');
+import {main} from '../build/lib/cli/index.js';
+import {getLogger} from '../build/lib/logger.js';
 
 const log = getLogger('cli');
 
--- a/packages/docutils/lib/builder/deploy.ts
+++ b/packages/docutils/lib/builder/deploy.ts
@@ -17,12 +17,12 @@ import {
   NAME_BIN,
   NAME_MIKE,
   NAME_MKDOCS_YML,
-} from '../constants';
-import {DocutilsError} from '../error';
-import {findMike, findMkDocsYml, isMkDocsInstalled, readPackageJson, requirePython} from '../fs';
-import {getLogger} from '../logger';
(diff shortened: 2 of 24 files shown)


eglitise:
I recently found out that `path.dirname(fileURLToPath(import.meta.url))` can be simplified to just `import.meta.dirname` for files: https://nodejs.org/api/esm.html#importmetadirname
Similarly, `fileURLToPath(import.meta.url)` can be simplified to `import.meta.filename`.

eglitise (approved):

KazuCocoa (approved):

Issue #17060, opened 2022-06-14
refactor task: send in only one parameter to createSession


jlipps:
Currently it takes 3 caps parameters. We should only accept and deal with one.

This will potentially be a breaking change for drivers, so would want to coordinate PRs. not sure if it will break existing driver installs if a user upgrades to an Appium with this change. it might.

PR #22587, opened 2026-08-03
chore(base-driver): Deprecate the legacy createSession API shape

mykola-mokhnach:
## Summary

Addresses #17060 by giving `createSession` a proper single-argument
overload without breaking any currently-published drivers.

- `AppiumDriver.createSession` (the client-facing umbrella method) gets a single-argument
  overload as the primary signature, with the legacy 3-position shape marked
  `@deprecated`. This is type-only, same as the driver-side change below: the `/session`
  route's `payloadParams` still sends `capabilities` three times, because that array is
  spread directly into any plugin hooking `createSession` (`AppiumDriver#wrapCommandWithPlugins`),
  not just into `AppiumDriver.createSession` itself. I confirmed `relaxed-caps-plugin`
  (shipped in this repo) reads `caps1`/`caps2`/`caps3` positionally, so shrinking the
  route's payload would be a wire-level breaking change for that and any similarly-written
  third-party plugin, even though `AppiumDriver.createSession`'s own caps-resolution
  logic happens to tolerate it. Deferred to a coordinated future breaking change, same as
  the driver-side `driverData` positional shape below.
- `BaseDriver`/`ExternalDriver.createSession` (the contract driver packages implement)
  gets the same single-argument overload added at the type level, with the legacy
  `(w3cCaps1, w3cCaps2, w3cCaps3, driverData)` shape marked `@deprecated`. This is
  type-only — no runtime behavior changes here. I checked the installed sources for
  `appium-xcuitest-driver` and `appium-uiautomator2-driver`, and both override
  `createSession` with the old 4-positional-arg signature, forwarding `driverData`
  positionally to `super()`. Appium's internal call to `driverInstance.createSession(...)`
  therefore still sends the legacy 4-arg form, so those drivers keep working unchanged.
  Driver authors can now migrate to the single-arg form at their own pace.
- `driverData` itself is deprecated end-to-end (the `DriverData` type, the
  `Core.driverData` getter, and the `driverData` params on `createSession` and
  `deleteSession`), pointing at `IAppiumIpc` as the replacement. I confirmed this is safe
  to deprecate: every real driver I checked (`xcuitest-driver`, `uiautomator2-driver`,
  `espresso-driver`) stubs `get driverData()` to return `{}`, and several others
  (`mac2-driver`, `windows-driver`, `safari-driver`) don't override it at all. The only
  place it's meaningfully exercised is Appium's own `fake-driver` test double.
- Updated the driver-authoring guide (`build-drivers.md`) to show the new single-arg
  `createSession` pattern and flag the old multi-arg/`driverData` pattern as deprecated.

## Non-goals / follow-ups

- Appium's internal call to the inner driver's `createSession` still uses the legacy
  4-arg convention on purpose, to stay compatible with drivers built against the old
  contract. Switching that call site to the new single-arg form is a breaking change for
  any driver still reading `driverData` positionally, and should be coordinated
  separately as part of the future Appium 4 major version bump.
- Actually removing `driverData`/the legacy overloads is left for Appium 4
  breaking-change pass.

### DIFF
--- a/packages/appium/lib/appium.ts
+++ b/packages/appium/lib/appium.ts
@@ -76,6 +76,8 @@ type SessionHandlerDeleteResult = SessionHandlerResult<void>;
  * Tuple shape of the deprecated multi-argument overload of {@link AppiumDriver.createSession}.
  * Shared between that overload's declaration and its implementation signature so the parameter
  * list only needs to be written out once.
+ *
+ * @deprecated Use the single-argument overload of {@link AppiumDriver.createSession} instead.
  */
 type LegacyCreateSessionArgs = [
   w3cCapabilities1: W3CAppiumDriverCaps,
--- a/packages/appium/lib/appium.ts
+++ b/packages/appium/lib/appium.ts
@@ -282,8 +282,9 @@ export class AppiumDriver extends DriverCore<AppiumDriverConstraints> {
    */
   async createSession(w3cCapabilities: W3CAppiumDriverCaps): Promise<SessionHandlerCreateResult>;
   /**
-   * @deprecated Legacy call sites may pass the same W3C caps in up to three positions; the first
-   * W3C-shaped value wins. Use the single-argument overload of {@linkcode createSession} instead.
+   * @deprecated Legacy call sites may pass the same W3C caps in up to three positions. These
+   * positions are intended to carry the same value; if they differ, which one wins is
+   * unspecified. Use the single-argument overload of {@linkcode createSession} instead.
    */
   async createSession(...legacyArgs: LegacyCreateSessionArgs): Promise<SessionHandlerCreateResult>;
   async createSession(...legacyArgs: LegacyCreateSessionArgs): Promise<SessionHandlerCreateResult> {
--- a/packages/appium/lib/appium.ts
+++ b/packages/appium/lib/appium.ts
@@ -72,6 +72,17 @@ type SessionHandlerCreateResult = SessionHandlerResult<
 
 type SessionHandlerDeleteResult = SessionHandlerResult<void>;
 
+/**
+ * Tuple shape of the deprecated multi-argument overload of {@link AppiumDriver.createSession}.
+ * Shared between that overload's declaration and its implementation signature so the parameter
+ * list only needs to be written out once.
+ */
+type LegacyCreateSessionArgs = [
+  w3cCapabilities1: W3CAppiumDriverCaps,
+  w3cCapabilities2?: W3CAppiumDriverCaps,
+  w3cCapabilities3?: W3CAppiumDriverCaps,
+];
+
 /** @internal Not part of {@link ExternalDriver}; used only when wiring session IPC. */
 type IpcAssignable = {
   assignIpc?: (ipc: IAppiumIpc) => Promise<void>;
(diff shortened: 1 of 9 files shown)


packages/appium/lib/appium.ts
```diff
@@ -263,9 +263,20 @@ export class AppiumDriver extends DriverCore<AppiumDriverConstraints> {
...
+   * envelope with either `[sessionId, caps, protocol]` or an error.
+   *
+   * @param w3cCapabilities - the new session capabilities in W3C format
    */
+  async createSession(w3cCapabilities: W3CAppiumDriverCaps): Promise<SessionHandlerCreateResult>;
+  /**
+   * @deprecated Legacy call sites may pass the same W3C caps in up to three positions; the first
+   * W3C-shaped value wins. Use the single-argument overload of {@linkcode createSession} instead.
+   */
+  async createSession(
```
eglitise: Isn't this legacy shape duplicated?
Same in `base-driver/lib/basedriver/driver.ts`

packages/base-driver/lib/protocol/routes/w3c.ts
```diff
@@ -10,7 +10,7 @@ export const W3C_ROUTES = {
     POST: {
       command: 'createSession',
       payloadParams: {
-        optional: ['capabilities', 'capabilities', 'capabilities'],
```
eglitise, line 13: Not sure if this is a safe change. The PR description only mentions the lack of `driverData` as the reason for simplifying it, but I assume it should still match the currently supported shape of 3 arguments

packages/appium/lib/appium.ts
mykola-mokhnach: Good catch, and you're right that it's the same shape in both files. I dug into why: TypeScript requires the deprecated overload to be declared as a distinct signature from the single-arg one for external callers to be able to pass 3 arguments at all (verified with a minimal repro — without it, a 3-arg call becomes a hard compile error, since the implementation signature alone isn't visible to callers unless a matching overload precedes it). So some duplication between the overload declaration and the implementation is inherent to this pattern.

That said, I've pushed a fix that removes the *textual* duplication: the legacy parameter list is now a shared `LegacyCreateSessionArgs` tuple type (exported from `@appium/types`, reused by `BaseDriver.createSession`; a local copy in `appium.ts` since the umbrella method has no `driverData` param), and both the overload declaration and the implementation signature just spread it as `...legacyArgs`. The parameter list itself is now written out exactly once per method.

packages/base-driver/lib/protocol/routes/w3c.ts
mykola-mokhnach: You're right, thanks for catching this — I traced how these args actually flow and this wasn't safe. The route's `payloadParams` array is spread directly into any plugin hooking 'createSession' (see `AppiumDriver#wrapCommandWithPlugins`, which does `cmdHandler.call(plugin, next, driver, ...args)`), not just into `AppiumDriver.createSession` itself. `relaxed-caps-plugin` (shipped in this repo) is one such consumer that declares `createSession(next, driver, caps1, caps2, caps3, ...restArgs)` positionally. I actually ran it both ways: with 1 arg it still produces the same result today, but only because `AppiumDriver.createSession`'s `.find(isW3cCaps)` fallback papers over the now-`undefined` 2nd/3rd positions — a third-party plugin that isn't written that defensively (e.g. one that checks `args.length`, or reads position 3 specifically per the pre-#16408 convention) could break.

I've reverted this route back to sending 'capabilities' three times, with a comment explaining why. `AppiumDriver.createSession`'s single-arg overload is now type-level only, same conservative approach as the `BaseDriver`/`ExternalDriver` side — no wire-level shape changes in this PR. Pushed.

packages/appium/lib/appium.ts
```diff
@@ -263,14 +276,18 @@ export class AppiumDriver extends DriverCore<AppiumDriverConstraints> {
...
-   * same W3C caps in up to three positions; the first W3C-shaped value wins.
+   * envelope with either `[sessionId, caps, protocol]` or an error.
+   *
+   * @param w3cCapabilities - the new session capabilities in W3C format
+   */
+  async createSession(w3cCapabilities: W3CAppiumDriverCaps): Promise<SessionHandlerCreateResult>;
+  /**
+   * @deprecated Legacy call sites may pass the same W3C caps in up to three positions; the first
+   * W3C-shaped value wins. Use the single-argument overload of {@linkcode createSession} instead.
    */
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch. That claim was pre-existing text I carried over verbatim from the code I moved into this overload's JSDoc, and it doesn't hold up — the .find() precedence here is really an unspecified implementation detail (and, worse, it's not even the same order in `BaseDriver.createSession`: appium.ts checks position 3, then 2, then 1, while driver.ts checks 3, then 1, then 2). Rather than document an order that isn't a real contract and could drift, I've updated both JSDoc blocks to say these positions are intended to carry the same value and which one wins if they differ is unspecified. Pushed.

packages/base-driver/lib/protocol/routes/w3c.ts
```diff
@@ -10,6 +10,10 @@ export const W3C_ROUTES = {
     POST: {
       command: 'createSession',
       payloadParams: {
+        // Deliberately sent 3 times: this array is spread directly into any plugin hooking
+        // 'createSession' (see AppiumDriver#wrapCommandWithPlugins), so changing its shape here
+        // would be a wire-level breaking change for third-party plugins, not just drivers. Keep
+        // this in sync with the deprecated multi-argument overload of `createSession`.
         optional: ['capabilities', 'capabilities', 'capabilities'],
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch — the PR description was stale after I reverted this route to keep the triple-`capabilities` shape for plugin compatibility (per the earlier thread on this same file). Updated the PR description to reflect that the route intentionally still sends `capabilities` three times, and that `AppiumDriver.createSession`'s single-argument overload is type-level only, same as the driver-side contract.

eglitise (approved):

PR #22590, opened 2026-08-04
feat(storage-plugin)!: Migrate to ESM

mykola-mokhnach:
## Summary
- Migrate `@appium/storage-plugin` to a pure ESM package (`"type": "module"`), following the same pattern used for `@appium/docutils` (#22585), `@appium/relaxed-caps-plugin`, and `@appium/images-plugin`.
- Add `.js` extensions to relative and `appium/*` subpath imports and switch `tsconfig.json` to `module`/`moduleResolution: "NodeNext"`, as required for ESM.
- Fix two latent bugs only caught by strict ESM module resolution:
  - `EventEmitter` was imported from `node:stream`, which only worked because Node's CJS loader tolerates it as a legacy compat property; ESM has no such export, so it now imports from `node:events`.
  - `WebSocket.Server` doesn't exist on `ws`'s ESM entry point (it's a CJS-only static alias); switched to the named `WebSocketServer` export.
- Add a narrow, documented type cast where `@appium/types` (still CommonJS) and this now-ESM package resolve `ws`'s `Server` type through different conditional-export branches, producing structurally distinct (but runtime-identical) types — a known `@types/ws` dual-package hazard.

BREAKING CHANGE: `@appium/storage-plugin` is now an ESM-only package. It can no longer be loaded via CommonJS `require()`; consumers must use `import` or dynamic `import()`. Deep imports into the package's internals are no longer possible — only the public entry point is exposed via `exports`.

### DIFF
--- a/packages/storage-plugin/lib/plugin.ts
+++ b/packages/storage-plugin/lib/plugin.ts
@@ -17,6 +17,7 @@ const log = logger.getLogger('StoragePlugin');
 // @appium/types is still CommonJS, so its `ws` Server type resolves through the "require"
 // condition, while this ESM package resolves the same class through "import" — TypeScript
 // treats them as structurally distinct even though they are identical at runtime.
+// TODO: Remove this workaround once @appium/types is migrated to ESM.
 type WSHandlerServer = Parameters<AppiumServer['addWebSocketHandler']>[1];
 
 let SHARED_STORAGE: Storage | null = null;
--- a/packages/storage-plugin/lib/plugin.ts
+++ b/packages/storage-plugin/lib/plugin.ts
@@ -1,17 +1,23 @@
-import {EventEmitter} from 'node:stream';
+import {EventEmitter} from 'node:events';
 
 import {fs, logger, tempDir, util} from '@appium/support';
 import type {AppiumServer} from '@appium/types';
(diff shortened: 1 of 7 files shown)


packages/storage-plugin/lib/plugin.ts
```diff
@@ -1,18 +1,24 @@
...
+import type WebSocket from 'ws';
 
-import {requireValidItemOptions, Storage, StorageArgumentError, validateStorageItemName} from './storage';
-import type {AddRequestResult, ItemOptions, StorageItem} from './types';
+import {requireValidItemOptions, Storage, StorageArgumentError, validateStorageItemName} from './storage.js';
+import type {AddRequestResult, ItemOptions, StorageItem} from './types.js';
 
 const log = logger.getLogger('StoragePlugin');
 
+// @appium/types is still CommonJS, so its `ws` Server type resolves through the "require"
```
eglitise, line 17: It might be worth adding a TODO here to remove this workaround once `@appium/types` is migrated to ESM

eglitise (approved):

packages/storage-plugin/lib/plugin.ts
mykola-mokhnach: Good idea, added a TODO. Thanks!

PR #22591, opened 2026-08-04
feat(driver-test-support)!: Migrate to ESM

mykola-mokhnach:
## Summary

- Convert @appium/driver-test-support to a native ESM package: add "type": "module" and an exports map in package.json.
- Switch tsconfig.json to module/moduleResolution: NodeNext and add explicit .js extensions to relative imports (lib/index.ts, test/unit/index.spec.ts), as required under NodeNext resolution.
- The exports map uses a default condition (rather than import-only) so existing CommonJS consumers can still require() the package — Node's require(esm) support (available on the Node versions this package already targets) loads it synchronously without any consumer changes.

BREAKING CHANGE: @appium/driver-test-support is now an ESM-only package. It can no longer be loaded via CommonJS require(); consumers must use import or dynamic import(). Deep imports into the package's internals are no longer possible — only the public entry point is exposed via exports.

### DIFF
--- a/packages/driver-test-support/lib/index.ts
+++ b/packages/driver-test-support/lib/index.ts
@@ -1,1 +1,1 @@
-export {createAppiumURL, getTestPort, TEST_HOST} from './helpers';
+export {createAppiumURL, getTestPort, TEST_HOST} from './helpers.js';
--- a/packages/driver-test-support/package.json
+++ b/packages/driver-test-support/package.json
@@ -32,8 +32,16 @@
     "lib",
     "tsconfig.json"
   ],
+  "type": "module",
   "main": "./build/lib/index.js",
   "types": "./build/lib/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./build/lib/index.d.ts",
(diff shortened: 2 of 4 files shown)


eglitise (approved):

eglitise:
> The exports map uses a default condition (rather than import-only) so existing CommonJS consumers can still require() the package — Node's require(esm) support (available on the Node versions this package already targets) loads it synchronously without any consumer changes.

Can't CJS customers also `require(esm)` the other already-migrated drivers/plugins? Not sure what makes this package different

mykola-mokhnach:
> > The exports map uses a default condition (rather than import-only) so existing CommonJS consumers can still require() the package — Node's require(esm) support (available on the Node versions this package already targets) loads it synchronously without any consumer changes.
> 
> Can't CJS customers also `require(esm)` the other already-migrated drivers/plugins? Not sure what makes this package different

changed it to the common sentence

eglitise:
I was actually wondering why this package needs the `default` entry under `exports` instead of `import`, which the description was not clear about

PR #22593, opened 2026-08-04
test(strongbox): replace chai with node:assert/strict

mykola-mokhnach:


packages/strongbox/test/e2e/strongbox.e2e.spec.ts
```diff
@@ -109,8 +105,11 @@ describe('@appium/strongbox', function () {
         for await (const item of box) {
           iterated.push(item);
         }
-        expect(iterated.map((i) => i.name)).to.eql(listed.map((i) => i.name));
-        expect(iterated).to.eql(listed);
+        assert.deepStrictEqual(
+          iterated.map((i) => i.name),
+          listed.map((i) => i.name),
+        );
+        assert.deepStrictEqual(iterated, listed);
```
eglitise: I wonder if the second assertion already covers the first one

packages/strongbox/test/unit/strongbox.spec.ts
```diff
@@ -193,41 +197,41 @@ describe('Strongbox', function () {
...
-          expect(box.getItem('test')).to.be.undefined;
+          assert.strictEqual(box.getItem('test'), undefined);
         });
       });
 
       describe('when there is a known Item with the given id', function () {
         it('should return the Item', async function () {
           const item = await box.createItem('test');
-          expect(box.getItem(item.id)).to.equal(item);
+          assert.strictEqual(box.getItem(item.id), item);
```
eglitise, line 234: I assume this should be `deepStrictEqual`

packages/strongbox/test/unit/strongbox.spec.ts
```diff
@@ -306,8 +316,11 @@ describe('Strongbox', function () {
         for await (const item of box) {
           fromIter.push(item);
         }
-        expect(fromIter.map((i) => i.name)).to.eql(fromList.map((i) => i.name));
-        expect(fromIter).to.eql(fromList);
+        assert.deepStrictEqual(
+          fromIter.map((i) => i.name),
+          fromList.map((i) => i.name),
+        );
+        assert.deepStrictEqual(fromIter, fromList);
```
eglitise: Same as earlier, the second assert appears to cover the first one

packages/strongbox/test/unit/strongbox.spec.ts
mykola-mokhnach: this is correct, we want to compare object references

packages/strongbox/test/e2e/strongbox.e2e.spec.ts
mykola-mokhnach: Good catch, thanks! Removed the redundant assertion in 314b7fb13.

eglitise: There's a second similar case around line 320 (I left a comment but resolved it by mistake)

mykola-mokhnach: Found it — same fix applied to packages/strongbox/test/unit/strongbox.spec.ts:320 in d80953687.

eglitise (approved):

PR #22594, opened 2026-08-04
test(support): replace chai with node:assert/strict

mykola-mokhnach:


packages/support/test/unit/logger/logger-normal.spec.ts
```diff
@@ -33,18 +32,18 @@ describe('normal logger', function () {
...
       throw log.errorWithException('msg1');
-    }).to.throw('msg1');
-    expect(() => {
+    }, /msg1/);
+    assert.throws(() => {
       throw log.errorWithException(new Error('msg2'));
-    }).to.throw('msg2');
+    }, /msg2/);
     assertOutputContains(writers, 'msg1');
     assertOutputContains(writers, 'msg2');
```
eglitise, line 42: I wonder if it's worth wrapping these `assertOutputContains` and `assertOutputDoesntContain` helpers in `assert.doesNotThrow()`

packages/support/test/unit/node.spec.ts
```diff
@@ -1,63 +1,63 @@
...
       const obj2 = node.deepFreeze({a: {b: 'c'}});
-      expect(() => ((obj2 as any).a.b = 'd')).to.throw();
-      expect(node.deepFreeze(1)).to.eql(1);
-      expect(node.deepFreeze(null)).to.equal(null);
+      assert.throws(() => ((obj2 as any).a.b = 'd'));
+      assert.strictEqual(node.deepFreeze(1), 1);
+      assert.strictEqual(node.deepFreeze(null), null);
       const obj3 = [1, {}, 3];
-      expect(node.deepFreeze(obj3)).to.equal(obj3);
+      assert.strictEqual(node.deepFreeze(obj3), obj3);
```
eglitise, line 60: `deepStrictEqual`?

mykola-mokhnach: comparing by reference is ok here

packages/support/test/unit/logger/logger-normal.spec.ts
mykola-mokhnach: if they would throw an exception the test is not going to pass anyway

packages/support/test/unit/util.spec.ts
```diff
@@ -200,68 +197,68 @@ describe('util', function () {
...
-      await expect(delay).to.eventually.be.rejectedWith(/cancellation error/);
+      await assert.rejects(delay, /cancellation error/);
     });
   });
 
   describe('safeJsonParse', function () {
     it('should pass object through', function () {
       const obj = {a: 'a', b: 'b'};
-      expect(util.safeJsonParse(obj)).to.equal(obj);
+      assert.strictEqual(util.safeJsonParse(obj), obj);
```
eglitise, line 207: `deepStrictEqual` might be better here, if only for consistency with the following tests

eglitise (approved):

PR #22598, opened 2026-08-04
test(logger): replace chai with node:assert/strict

mykola-mokhnach:


packages/logger/test/unit/basic.spec.ts
```diff
@@ -112,7 +111,7 @@ describe('basic', function () {
     });
 
     it('should work', function () {
-      expect(log.stream).to.equal(s);
+      assert.strictEqual(log.stream, s);
```
eglitise, line 114: `deepStrictEqual`?

mykola-mokhnach: this is correct, we want compare object references rather than their payload

eglitise (approved):

PR #22600, opened 2026-08-04
test(fake-driver): replace chai with node:assert/strict

mykola-mokhnach:


packages/fake-driver/test/e2e/general.e2e.spec.ts
```diff
@@ -26,57 +23,57 @@ export function generalTests(context: {port: number}) {
     });
     it('should get geolocation', async function () {
       const geo = await driver.getGeoLocation();
-      expect(geo.latitude).to.exist;
-      expect(geo.longitude).to.exist;
+      assert.ok(geo.latitude !== undefined && geo.latitude !== null);
+      assert.ok(geo.longitude !== undefined && geo.longitude !== null);
```
eglitise, line 27: Just `assert.ok(geo.x)` might be fine

packages/fake-driver/test/unit/driver.spec.ts
```diff
@@ -37,39 +34,39 @@ describe('FakeDriver unit suite', function () {
...
+    assert.strictEqual(d.sessionId, null);
     const sessionId2 = await d.createSession(structuredClone(w3cCaps));
-    expect(sessionId1).to.not.eql(sessionId2);
+    assert.notDeepStrictEqual(sessionId1, sessionId2);
   });
 
   it('should get the current session', async function () {
     const [, caps] = await d.createSession(w3cCaps);
-    expect(caps).to.equal(await d.getSession());
+    assert.strictEqual(caps, await d.getSession());
```
eglitise, line 69: `deepStrictEqual`?

packages/fake-driver/test/unit/driver.spec.ts
```diff
@@ -318,96 +315,96 @@ describe('FakeDriver unit suite', function () {
...
       it('should return false', function () {
-        expect(d.proxyActive(sessId)).to.be.false;
+        assert.strictEqual(d.proxyActive(sessId), false);
       });
       it('should throw an error when sessionId is wrong', function () {
-        expect(() => {
+        assert.doesNotThrow(() => {
           d.proxyActive('aaa');
-        }).to.throw;
+        });
```
eglitise: Is this the same case as https://github.com/appium/appium-uiautomator2-driver/pull/1031#discussion_r3677434662 ?
If yes, might be worth removing this test, along with equivalents for `getProxyAvoidList` and `canProxy`

packages/fake-driver/test/unit/driver.spec.ts
mykola-mokhnach: comparing references is more precise

packages/fake-driver/test/unit/driver.spec.ts
mykola-mokhnach: I don't have any particular preferences on this one. Such check also works for me

packages/fake-driver/test/e2e/general.e2e.spec.ts
mykola-mokhnach: we want to explicitly check the value is not null or undefined. the other option may contain 0, which is ok, but would fail the test

packages/fake-driver/test/unit/driver.spec.ts
eglitise: I'd probably remove the tests since the function output doesn't depend on session ID anyway

mykola-mokhnach: Agreed — removed all three (proxyActive, getProxyAvoidList, canProxy) in ada43feee.

eglitise (approved):

PR #22603, opened 2026-08-04
test(base-driver): replace chai with node:assert/strict

mykola-mokhnach:


packages/base-driver/test/unit/basedriver/capability.spec.ts
```diff
@@ -35,32 +32,34 @@ describe('Desired Capabilities', function () {
...
 
   it('should require platformName and deviceName', async function () {
-    await expect(
+    await assert.rejects(
       d.createSession({
         alwaysMatch: {},
         firstMatch: [{}],
       }),
-    ).to.be.rejectedWith(errors.SessionNotCreatedError, /platformName/);
+      (err: any) => err instanceof errors.SessionNotCreatedError && /platformName/.test(err.message),
```
eglitise: Can this and other similar assertions use the object form already used in other PRs?
```
{
  name: 'errors.SessionNotCreatedError',
  message: /platformName/
}
```

packages/base-driver/test/e2e/protocol/protocol.e2e.spec.ts
```diff
@@ -350,14 +353,14 @@ describe('Protocol', function () {
...
-          expect(message).to.match(/Method has not yet been implemented/);
-          expect(stacktrace).to.match(/protocol\.(js|ts)/);
-          expect(w3cError).to.be.a('string');
-          expect(w3cError).to.equal(errors.NotYetImplementedError.error());
-          expect(message).to.match(/Method has not yet been implemented/);
+          assert.match(message, /Method has not yet been implemented/);
+          assert.match(stacktrace, /protocol\.(js|ts)/);
+          assert.strictEqual(typeof w3cError, 'string');
+          assert.strictEqual(w3cError, errors.NotYetImplementedError.error());
+          assert.match(message, /Method has not yet been implemented/);
```
eglitise: Looks like this assertion was originally duplicated (on line 359), so it can be removed

packages/base-driver/test/unit/basedriver/capability.spec.ts
mykola-mokhnach: Good call, updated to use the {name, message} object form (matches the pattern from support/env.spec.ts) in both this file and device-settings.spec.ts, which had the same instanceof+regex predicate.

eglitise (approved):

packages/base-driver/test/e2e/protocol/protocol.e2e.spec.ts
mykola-mokhnach: Good catch, removed the duplicate assertion.

PR #22604, opened 2026-08-04
test(appium): replace chai with node:assert/strict

mykola-mokhnach:


eglitise:
Looks like CI isn't happy

mykola-mokhnach:
> Looks like CI isn't happy

now it is. There is a random nodejs bug that makes native node e2e tests to fail while piping spawned process output. It should be already fixed in node24, but the fix has not been ported back to v22

packages/appium/test/e2e/args.e2e.spec.ts
```diff
@@ -52,7 +48,8 @@ describe('argument parsing', function () {
               timeout: 5000,
             },
           ),
-        ).to.be.rejectedWith(Error, /timed out/);
+          (err: unknown) => err instanceof Error && /timed out/.test(err.message),
```
eglitise: This and other rejection assertions can use the object form (https://github.com/appium/appium/pull/22603#discussion_r3720700356)

packages/appium/test/e2e/cli-driver.e2e.spec.ts
```diff
@@ -338,8 +357,8 @@ describe('Driver CLI', {timeout: 90000}, function () {
 
     it('should uninstall a driver based on its driver name', async function () {
       const uninstall = await runUninstall(['fake']);
-      expect(uninstall).to.not.have.key('fake');
-      await expect(fs.exists(path.join(appiumHome, 'node_modules', '@appium', 'fake-driver'))).to.eventually.be.false;
+      assert.notDeepStrictEqual(Object.keys(uninstall), ['fake']);
```
eglitise: This line has a slightly different assertion. If it's not intentional, the following would be more equivalent:
```
assert.ok(!('fake' in Object.keys(uninstall)));
```

Or perhaps even
```
assert.ok(!Object.hasOwn(uninstall, 'fake'));
```

packages/appium/test/e2e/cli-driver.e2e.spec.ts
```diff
@@ -37,6 +33,26 @@ interface ExtensionListResult {
   };
 }
 
+/**
+ * Asserts that every key/value pair in `expected` is present (via deep equality) on `actual`,
+ * mirroring chai's `expect(actual).to.deep.include(expected)` semantics for objects.
+ */
+function assertDeepInclude(actual: unknown, expected: unknown): void {
```
eglitise, line 40: Since other files also used `expect(actual).to.deep.include(expected)`, maybe it's worth moving these to `../helpers`, so they can be reused?

packages/appium/test/e2e/config-file.e2e.spec.ts
```diff
@@ -181,19 +179,28 @@ describe('config file behavior', function () {
...
+                    additionalProperty: 'appium-home',
+                  },
+                  message: 'must NOT have additional properties',
+                  isIdentifierLocation: true,
+                });
+                return true;
+              } catch {
+                return false;
+              }
+            }),
```
eglitise: Since all of this is already wrapped in `assert.ok`, and this file already imports `util` from `@appium/support`, maybe a simpler format could be
```js
result.errors?.some((error) => util.isEqual(error, {instancePath: '', ... }));
```

packages/appium/test/e2e/local.e2e.spec.ts
```diff
@@ -110,12 +109,12 @@ describe('when Appium is a dependency of the current project', function () {
...
 
         it('should list the driver', function () {
-          expect(res).to.have.property('fake');
+          assert.ok(Object.hasOwn(res, 'fake'));
         });
 
         it('should update the manifest', async function () {
           const manifestParsed = await readManifest();
-          expect(manifestParsed).to.have.nested.property('drivers.fake');
+          assert.ok(manifestParsed.drivers?.fake !== undefined);
```
eglitise: Perhaps just `assert.ok(manifestParsed.drivers?.fake)` is ok
Same for the following tests

packages/appium/test/unit/bootstrap/config-file.spec.ts
```diff
@@ -192,7 +189,7 @@ describe('bootstrap/config-file', function () {
             });
 
             it('should resolve with an object having a nonempty array of errors', function () {
-              expect(result).to.have.property('errors').that.is.not.empty;
+              assert.ok(Array.isArray(result.errors) && result.errors.length > 0);
```
eglitise: This should probably be split into two separate assertions, following other similar examples for objects (e.g. line 140 in this file)

packages/appium/test/unit/bootstrap/config-file.spec.ts
```diff
@@ -285,7 +283,7 @@ describe('bootstrap/config-file', function () {
...
```
eglitise: Same here

packages/appium/test/unit/extension/driver-config.spec.ts
```diff
@@ -56,12 +53,12 @@ describe('DriverConfig', function () {
       describe('when the DriverConfig is not yet associated with a Manifest', function () {
         it('should return a new DriverConfig', function () {
           const config = DriverConfig.create(manifest);
-          expect(config).to.be.an.instanceof(DriverConfig);
+          assert.ok(config instanceof (DriverConfig as unknown as new (...args: any[]) => unknown));
```
eglitise: Maybe I'm misunderstanding this line but the type conversions make it seem like `config` is compared against a generic class signature

packages/appium/test/e2e/cli-driver.e2e.spec.ts
eglitise: After looking closer at the other files, it seems the usage was different after all. Here it's used to check whether all keys from `expected` are in `actual`, whereas other files check whether the `actual` array includes the `expected` object.

packages/appium/test/unit/extension/driver-config.spec.ts
```diff
@@ -119,64 +117,78 @@ describe('DriverConfig', function () {
...
-              err: 'Missing or incorrect supported platformNames list.',
-              val: undefined,
-            });
+            assert.ok(
+              driverConfig
+                .getConfigProblems({})
+                .some((p: unknown) =>
+                  isDeepStrictEqual(p, {err: 'Missing or incorrect supported platformNames list.', val: undefined}),
+                ),
+            );
```
eglitise: Since multiple tests in this file use such checks, it might be worth creating a helper method, similarly to `assertDeepInclude` in `cli-driver.e2e.spec.ts`

packages/appium/test/unit/extension/driver-config.spec.ts
```diff
@@ -270,17 +289,17 @@ describe('DriverConfig', function () {
       describe('when the extension data is missing `schema`', function () {
         it('should throw', async function () {
           delete (extData as {schema?: string}).schema;
-          await expect(driverConfig.readExtensionSchema(extName, extData)).to.be.rejectedWith(
-            TypeError,
-            /why is this function being called/i,
+          await assert.rejects(
+            driverConfig.readExtensionSchema(extName, extData),
+            (err: unknown) => err instanceof TypeError && /why is this function being called/i.test(err.message),
```
eglitise: This can use the object form for the error name & message

packages/appium/test/unit/extension/manifest.spec.ts
```diff
@@ -48,15 +45,15 @@ describe('Manifest', function () {
         it('should return the same object both times', function () {
           const firstInstance = Manifest.getInstance('/some/path');
           const secondInstance = Manifest.getInstance('/some/path');
-          expect(firstInstance).to.equal(secondInstance);
+          assert.strictEqual(firstInstance, secondInstance);
```
eglitise, line 48: Not sure if this and the next test should use `deepStrictEqual` instead

packages/appium/test/unit/extension/manifest.spec.ts
```diff
@@ -136,9 +133,10 @@ describe('Manifest', function () {
           MockAppiumSupport.fs.readFile.resolves('{');
         });
         it('should reject', async function () {
-          await expect(manifest.read()).to.be.eventually.rejectedWith(
-            Error,
-            /trouble loading the extension installation cache file/i,
+          await assert.rejects(
+            manifest.read(),
+            (err: unknown) =>
+              err instanceof Error && /trouble loading the extension installation cache file/i.test(err.message),
```
eglitise: This and other similar reject asserts in this file can use the error object form

packages/appium/test/unit/extension/manifest.spec.ts
```diff
@@ -279,7 +284,8 @@ describe('Manifest', function () {
 
       it('should add a clone of the extension manifest to the internal data object', function () {
         manifest.setExtension(DRIVER_TYPE, 'foo', extData);
-        expect(manifest.getExtensionData(DRIVER_TYPE).foo).to.eql(extData).and.not.to.equal(extData);
+        assert.deepStrictEqual(manifest.getExtensionData(DRIVER_TYPE).foo, extData);
+        assert.notStrictEqual(manifest.getExtensionData(DRIVER_TYPE).foo, extData);
```
eglitise, line 280: I'm a little confused how this passes both `deepStrictEqual` and `notStrictEqual`

packages/appium/test/unit/extension/package-changed.spec.ts
```diff
@@ -36,32 +32,37 @@ describe('package-changed', function () {
...
+      );
     });
 
     describe('when it cannot create the parent dir', function () {
       it('should reject', async function () {
         MockAppiumSupport.fs.mkdirp.rejects(new Error('some error'));
-        await expect(packageDidChange('/some/path')).to.be.rejectedWith(Error, /could not create the directory/i);
+        await assert.rejects(
+          packageDidChange('/some/path'),
+          (err: unknown) => err instanceof Error && /could not create the directory/i.test(err.message),
```
eglitise: This and line 106 can also use the object form like in my previous comments

packages/appium/test/unit/extension/plugin-config.spec.ts
```diff
@@ -56,12 +53,12 @@ describe('PluginConfig', function () {
       describe('when the PluginConfig is not yet associated with a Manifest', function () {
         it('should return a new PluginConfig', function () {
           const config = PluginConfig.create(manifest);
-          expect(config).to.be.an.instanceof(PluginConfig);
+          assert.ok(config instanceof (PluginConfig as unknown as new (...args: any[]) => unknown));
```
eglitise: Same question here as the one I left in `driver-config.spec.ts`

packages/appium/test/unit/extension/driver-config.spec.ts
eglitise: Same applies to `plugin-config.spec.ts`

packages/appium/test/unit/helpers/capability.spec.ts
```diff
@@ -13,20 +12,16 @@ import {
...
-  beforeEach(async function () {
-    use(chaiAsPromised);
-  });
-
   describe('parseCapsForInnerDriver()', function () {
     it('should return an error if only JSONWP provided', function () {
       const res = parseCapsForInnerDriver(BASE_CAPS as unknown as W3CCapabilities<Constraints>);
-      expect('error' in res && res.error).to.be.ok;
-      expect((res as {error: {message: string}}).error.message).to.match(/W3C/);
+      assert.ok('error' in res && res.error);
```
eglitise: Probably can be simplified to `assert.ok(res.error)`

packages/appium/test/unit/schema/cli-args.spec.ts
```diff
@@ -141,18 +138,21 @@ describe('cli-args', function () {
...
           });
         });
 
         describe('null', function () {
           it('should throw', async function () {
             const schema = {properties: {foo: {type: 'null'}}, type: 'object'};
-            await expect(getArgs({extType, extName, schema})).to.be.rejectedWith(TypeError, /unknown or disallowed/);
+            await assert.rejects(
+              getArgs({extType, extName, schema}),
+              (err: Error) => err instanceof TypeError && /unknown or disallowed/.test(err.message),
```
eglitise: Same as my previous comments about using the error object form, for this and 2 other tests in this file

packages/appium/test/unit/schema/format-errors.spec.ts
```diff
@@ -40,43 +36,51 @@ describe('schema/format-errors', function () {
...
       });
     });
 
     describe('when provided `errors` as `undefined`', function () {
       it('should throw', function () {
-        expect(() => formatErrors()).to.throw(TypeError, 'Array of errors must be non-empty');
+        assert.throws(
+          () => formatErrors(),
+          (err: Error) => err instanceof TypeError && /Array of errors must be non-empty/.test(err.message),
+        );
```
eglitise: Same as above

packages/appium/test/unit/schema/schema.spec.ts
```diff
@@ -65,56 +62,62 @@ describe('schema', function () {
     describe('error conditions', function () {
       describe('when provided no parameters', function () {
         it('should throw a TypeError', async function () {
-          await expect((registerSchema as (...args: unknown[]) => unknown)()).to.be.rejectedWith(
-            TypeError,
-            /expected extension type/i,
+          await assert.rejects(
+            (registerSchema as (...args: unknown[]) => Promise<unknown>)(),
+            (err: Error) => err instanceof TypeError && /expected extension type/i.test(err.message),
```
eglitise: Same as above for various tests in this file

packages/appium/test/e2e/args.e2e.spec.ts
mykola-mokhnach: Fixed, updated to the {name, message} object form here and in the other files you flagged (driver-config.spec.ts, package-changed.spec.ts, manifest.spec.ts, cli-args.spec.ts, schema.spec.ts, format-errors.spec.ts).

packages/appium/test/e2e/cli-driver.e2e.spec.ts
mykola-mokhnach: Good catch — went with your second suggestion (`Object.hasOwn`), since the first one (`'fake' in Object.keys(uninstall)`) checks array indices rather than values and wouldn't actually detect the key.

packages/appium/test/e2e/cli-driver.e2e.spec.ts
mykola-mokhnach: Sounds good, no action needed here given your follow-up.

packages/appium/test/e2e/config-file.e2e.spec.ts
mykola-mokhnach: Nice simplification, done.

packages/appium/test/e2e/local.e2e.spec.ts
mykola-mokhnach: Done, simplified this and the other `!== undefined` checks later in the file to plain truthy checks.

packages/appium/test/unit/bootstrap/config-file.spec.ts
mykola-mokhnach: Split into two assertions to match the pattern at line 140.

packages/appium/test/unit/bootstrap/config-file.spec.ts
mykola-mokhnach: Same here, split into two assertions.

packages/appium/test/unit/extension/driver-config.spec.ts
mykola-mokhnach: Good question — the cast was there because the local `DriverConfig` variable is typed with a hand-written `DriverConfigConstructor` interface (it's obtained dynamically via rewiremock, not the real class), and that interface only declared the two static methods, not a construct signature, so TS wouldn't accept `instanceof` without a cast. Added `new (...args: never[]): DriverConfig` to the interface instead, so `config instanceof DriverConfig` now just works directly.

packages/appium/test/e2e/cli-driver.e2e.spec.ts
mykola-mokhnach: Makes sense, thanks for double-checking — left the two array-based .some() usages here as-is.

packages/appium/test/unit/extension/driver-config.spec.ts
mykola-mokhnach: Extracted an `assertArrayIncludesDeep()` helper into test/helpers.ts (alongside the existing `assertDeepInclude`/`assertInclude` pattern) and used it here and in plugin-config.spec.ts.

packages/appium/test/unit/extension/driver-config.spec.ts
mykola-mokhnach: Fixed, updated to the {name, message} object form.

packages/appium/test/unit/extension/manifest.spec.ts
mykola-mokhnach: This one's intentionally strictEqual rather than deepStrictEqual — the test title is 'should return the same object both times', i.e. it's asserting singleton identity (same reference), not just structural equality. deepStrictEqual would incorrectly pass even if getInstance() returned two separate-but-equal objects, which defeats the point of the test. Left as-is.

packages/appium/test/unit/extension/manifest.spec.ts
mykola-mokhnach: Fixed, updated to the {name, message} object form.

packages/appium/test/unit/extension/manifest.spec.ts
mykola-mokhnach: These aren't contradictory — deepStrictEqual checks the stored value has the same *content* as extData, while notStrictEqual checks it's a *different reference* (i.e. a clone, not the same object). Both need to hold for 'should add a clone' to be true: same data, different identity. This mirrors the original chai `.to.eql(extData).and.not.to.equal(extData)`. Left as-is.

packages/appium/test/unit/extension/package-changed.spec.ts
mykola-mokhnach: Fixed, updated to the {name, message} object form.

packages/appium/test/unit/extension/plugin-config.spec.ts
mykola-mokhnach: Same fix as driver-config.spec.ts — added a construct signature to `PluginConfigConstructor` so `instanceof` no longer needs the double-cast.

packages/appium/test/unit/extension/driver-config.spec.ts
mykola-mokhnach: Done, extracted the shared assertArrayIncludesDeep() helper and used it in both files.

packages/appium/test/unit/helpers/capability.spec.ts
mykola-mokhnach: Simplified — turns out `res.error` alone doesn't quite type-check since `parseCapsForInnerDriver`'s return type is a union where only one branch has `error`, so I added an `as InvalidCaps` cast on the assignment instead (also removes the duplicate cast that was previously on the next line).

packages/appium/test/unit/schema/cli-args.spec.ts
mykola-mokhnach: Fixed, updated to the {name, message} object form for this and the other two tests in the file.

packages/appium/test/unit/schema/format-errors.spec.ts
mykola-mokhnach: Fixed, updated to the {name, message} object form.

packages/appium/test/unit/schema/schema.spec.ts
mykola-mokhnach: Fixed, updated to the {name, message} object form for the tests in this file. Note: SchemaUnsupportedSchemaError extends TypeError without overriding `.name`, so its runtime `.name` is actually 'TypeError' — used that as the matcher value for the two tests checking that specific error class.

eglitise (approved):

PR #22606, opened 2026-08-04
fix(base-driver): return a W3C error for a malformed JSON body

Mochxd:
﻿## Proposed changes

A request body that is not valid JSON currently gets an HTML error page back instead of a WebDriver error:

```
$ curl -i -X POST http://127.0.0.1:4723/session -H 'content-type: application/json' -d '{"capabilities": not json}'

HTTP/1.1 400 Bad Request
Content-Type: text/html; charset=utf-8

<!DOCTYPE html><html lang="en"><head><meta charset="utf-8"><title>Error</title></head>
<body><pre>SyntaxError: Unexpected token &#39;o&#39; ... </pre></body></html>
```

The spec says a POST body that fails to parse should come back as `invalid argument`, so clients try to read `value.error` and get a parser error of their own instead of a useful message.

Two things are going on:

`catchAllHandler` is registered before `bodyParser.json()` and before the routes. Express only hands an error to an error handler that sits after the middleware which raised it, so the one meant to be the last handler never ran for body parse failures and Express fell back to its own HTML page. It is now registered last.

That alone would return `unknown error` / 500, since a body-parser `SyntaxError` means nothing to `getResponseForW3CError`. Parse failures are now translated to `InvalidArgumentError` first, which keeps the 400 and gives the response the error code the spec asks for:

```json
{"value":{"error":"invalid argument","message":"Unexpected token 'o' ...","stacktrace":"..."}}
```

Extension routes added through `serverUpdaters` still register after this handler, so their uncaught errors keep the old behaviour. That looked like a separate change to me, happy to pull it in here if you would rather have it in one go.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

The new test posts a broken body to a running server and checks the response. On master it fails with `expected 'text/html; charset=utf-8' to include 'application/json'`.

### DIFF
--- a/packages/base-driver/lib/express/middleware.ts
+++ b/packages/base-driver/lib/express/middleware.ts
@@ -155,8 +155,18 @@ export function catchAllHandler(err: Error, _req: Request, res: Response, next: 
   }
 
   log.error(`Uncaught error: ${err.message}`);
-  const [status, body] = getResponseForW3CError(err);
+  const [status, body] = getResponseForW3CError(
+    isBodyParseError(err) ? new errors.InvalidArgumentError(err.message) : err,
+  );
   res.status(status).json(body);
+}
+
+/**
+ * Detects the error body-parser raises for a request body that is not valid JSON.
+ * The protocol expects such a request to be rejected as an invalid argument.
+ */
+function isBodyParseError(err: Error): boolean {
+  return (err as Error & {type?: string}).type === 'entity.parse.failed';
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/base-driver/lib/express/middleware.ts
```diff
@@ -155,10 +155,20 @@ export function catchAllHandler(err: Error, _req: Request, res: Response, next:
...
+  );
   res.status(status).json(body);
 }
 
+/**
+ * Detects the error body-parser raises for a request body that is not valid JSON.
+ * The protocol expects such a request to be rejected as an invalid argument.
+ */
+function isBodyParseError(err: Error): boolean {
+  return (err as Error & {type?: string}).type === 'entity.parse.failed';
```
mykola-mokhnach: We should not assume error is always an object (`.type -> ?.type`)

Mochxd: Good point, switched to ?.type

mykola-mokhnach (approved):

PR #22607, opened 2026-08-04
fix(execute-driver-plugin): fail fast if the script process dies

Mochxd:
﻿## Proposed changes

`executeDriverScript` waits for the child process through IPC only:

```js
const res = await new Promise((resolve) => {
  scriptProc.once('message', resolve); // this is node IPC
});
```

Nothing watches the process itself, so a child that dies before it can send anything leaves that promise pending and the only thing left to settle the race is the timeout, which defaults to an hour. The script errors you would normally expect are caught in `execute-child.js` and reported over IPC, so this is about the cases where the child never gets that far: killed by the OOM killer, a crash on startup, or the unhandled rejection you get when `process.send` itself fails.

The wait now also listens for `exit` and `error` and rejects. A child that answers first wins the race, so the `exit` that follows a normal run lands on an already settled promise and changes nothing. Instead of hanging, the client now gets:

```
Could not execute driver script. Original error was: Error: The driver script process ended without returning a result (exit code: 1, signal: null)
```

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

The new unit test stubs the fork so the child exits without sending a result. It rejects in about 8ms with this change; on master it sits there until the 5s test timeout kills it. The e2e suite still passes, so real scripts that do return a result are unaffected.

### DIFF
--- a/packages/execute-driver-plugin/lib/plugin.ts
+++ b/packages/execute-driver-plugin/lib/plugin.ts
@@ -95,8 +95,19 @@ export class ExecuteDriverPlugin extends BasePlugin {
 
         // promise that deals with the result from the child process
         const waitForResult = async () => {
-          const res = await new Promise<{error?: {message: string}; success?: any}>((resolve) => {
+          const res = await new Promise<{error?: {message: string}; success?: any}>((resolve, reject) => {
             scriptProc.once('message', resolve); // this is node IPC
+            // the child reports script errors over IPC as well, so reaching any of the
+            // handlers below means it died before it could tell us anything
+            scriptProc.once('error', reject);
+            scriptProc.once('exit', (code, signal) => {
+              reject(
+                new Error(
+                  `The driver script process ended without returning a result ` +
+                    `(exit code: ${code}, signal: ${signal})`,
+                ),
+              );
+            });
           });
 
           this.log.info('Received execute driver script result from child process, shutting it down');
--- a/packages/execute-driver-plugin/lib/plugin.ts
+++ b/packages/execute-driver-plugin/lib/plugin.ts
@@ -101,6 +101,11 @@ export class ExecuteDriverPlugin extends BasePlugin {
             // handlers below means it died before it could tell us anything
             scriptProc.once('error', reject);
             scriptProc.once('exit', (code, signal) => {
+              // a clean exit is left to the script timeout, since only an abnormal
+              // one tells us for sure that no result is coming
+              if (code === 0) {
+                return;
+              }
               reject(
                 new Error(
                   `The driver script process ended without returning a result ` +
--- a/packages/execute-driver-plugin/lib/plugin.ts
+++ b/packages/execute-driver-plugin/lib/plugin.ts
@@ -101,10 +101,11 @@ export class ExecuteDriverPlugin extends BasePlugin {
             // handlers below means it died before it could tell us anything
             scriptProc.once('error', reject);
             scriptProc.once('exit', (code, signal) => {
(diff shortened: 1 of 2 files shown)


packages/execute-driver-plugin/lib/plugin.ts
```diff
@@ -95,8 +95,19 @@ export class ExecuteDriverPlugin extends BasePlugin {
...
         // promise that deals with the result from the child process
         const waitForResult = async () => {
-          const res = await new Promise<{error?: {message: string}; success?: any}>((resolve) => {
+          const res = await new Promise<{error?: {message: string}; success?: any}>((resolve, reject) => {
             scriptProc.once('message', resolve); // this is node IPC
+            // the child reports script errors over IPC as well, so reaching any of the
+            // handlers below means it died before it could tell us anything
+            scriptProc.once('error', reject);
+            scriptProc.once('exit', (code, signal) => {
+              reject(
```
mykola-mokhnach, line 110: should we still reject if exit code is zero?

Mochxd: I think it should....reaching that handler at all means the child never sent a result, so there is nothing to return to the client no matter which code it exited with, and resolving instead would report a null result for a script that never ran.

The risk worth checking is the other direction, losing a result that was already on its way. I forked a child that calls `process.send()` and `process.exit(0)` in the same tick, and the parent saw `message` before `exit` 50 times out of 50, since node drains the IPC channel before it emits `exit`. The child here also does not exit on its own once it has sent something, it waits for the `finally` block to disconnect or kill it, so a normal run is settled well before any exit arrives.

Happy to limit it to non-zero codes if you would rather be careful about it.

Mochxd:
Rebased on master to pick up #22601, the new test now uses `node:assert/strict` like the rest of the file.

packages/execute-driver-plugin/lib/plugin.ts
mykola-mokhnach: yes, lest limit it to non-zero codes only (or no codes if killed by a signal)

Mochxd: Done in 58e1efe. A zero exit code now falls through to the script timeout, and only a non-zero code or a signal kill rejects.

Added tests for the three endings: an error code and a signal both reject, and a clean exit is left to the timeout.

KazuCocoa (approved):

packages/execute-driver-plugin/lib/plugin.ts
```diff
@@ -95,8 +95,24 @@ export class ExecuteDriverPlugin extends BasePlugin {
...
+          const res = await new Promise<{error?: {message: string}; success?: any}>((resolve, reject) => {
             scriptProc.once('message', resolve); // this is node IPC
+            // the child reports script errors over IPC as well, so reaching any of the
+            // handlers below means it died before it could tell us anything
+            scriptProc.once('error', reject);
+            scriptProc.once('exit', (code, signal) => {
+              // a clean exit is left to the script timeout, since only an abnormal
+              // one tells us for sure that no result is coming
+              if (code === 0) {
+                return;
```
mykola-mokhnach: timeout might be very long. I would rather just log the exit code and pass an undefined value as `message`: `return resolve({})`

Mochxd: Updated in the latest commit. A zero exit with no IPC message now logs and resolves with an empty result (`resolve({})` → `undefined` to the client) instead of waiting on the script timeout. Non-zero codes and signal kills still reject.

PR #22610, opened 2026-08-05
chore(deps): remove obsolete devDependencies

mykola-mokhnach:
Drop @types/lodash, @types/mv, and @types/xmldom (no longer used by any source or dependency), plus the root-level eslint plugin packages that are redundant now that eslint.config.mjs only needs eslint itself and packages/eslint-config-appium-ts, which already declares them.


package.json
```diff
@@ -117,12 +112,6 @@
     "cpy-cli": "7.0.0",
     "cross-env": "10.1.0",
     "eslint": "10.8.0",
```
eglitise: I assume this one can also be removed

eglitise (approved):

package.json
mykola-mokhnach: done

PR #22625, opened 2026-08-10
fix(base-driver): reject a non-string url without throwing TypeError

Mochxd:
﻿## Proposed changes

`POST /session/:id/url` validates the url by calling `.match()` on it. That only works for strings, so a body like `{"url": 123}` blows up with:

```
TypeError: url.match is not a function
```

which comes back as an `unknown error` instead of the intended validation message.

The check now rejects anything that is not a string first, so those requests get the same `Url or Uri must start with <scheme>://` error as a missing or schemeless value.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have signed the CLA

## Further comments

Added a small unit test around `validators.setUrl`. On master it fails with `actual: 'TypeError'`; with this change it passes.

### DIFF
--- a/packages/base-driver/lib/protocol/validators.ts
+++ b/packages/base-driver/lib/protocol/validators.ts
@@ -1,7 +1,11 @@
+import {util} from '@appium/support';
+
+const MAX_URL_ERROR_LENGTH = 100;
+
 export const validators = {
   setUrl: (url: any) => {
     if (!URL.canParse(url)) {
-      throw new Error('Url or Uri must be a valid URL');
+      throw new Error(`'${util.truncateString(String(url), {length: MAX_URL_ERROR_LENGTH})}' must be a valid URL`);
     }
   },
   setNetworkConnection: (type: any) => {
--- a/packages/base-driver/lib/protocol/validators.ts
+++ b/packages/base-driver/lib/protocol/validators.ts
@@ -1,8 +1,7 @@
 export const validators = {
   setUrl: (url: any) => {
-    // either an `xyz://`, `about:`, or `data:` scheme is allowed
-    if (typeof url !== 'string' || !url.match(/^([a-zA-Z0-9_+.-]+:\/\/)|(about:)|(data:)/)) {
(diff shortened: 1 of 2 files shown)


packages/base-driver/lib/protocol/validators.ts
```diff
@@ -1,7 +1,7 @@
 export const validators = {
   setUrl: (url: any) => {
     // either an `xyz://`, `about:`, or `data:` scheme is allowed
-    if (!url || !url.match(/^([a-zA-Z0-9_+.-]+:\/\/)|(about:)|(data:)/)) {
+    if (typeof url !== 'string' || !url.match(/^([a-zA-Z0-9_+.-]+:\/\/)|(about:)|(data:)/)) {
```
mykola-mokhnach: I think it would be more reliable to use `URL.canParse` . The check for particular schemas must be removed

Mochxd: Switched to ``URL.canParse`` and dropped the scheme-specific regex. Non-strings still fail cleanly (``canParse`` returns false) instead of throwing TypeError.

packages/base-driver/lib/protocol/validators.ts
```diff
@@ -1,8 +1,7 @@
 export const validators = {
   setUrl: (url: any) => {
-    // either an `xyz://`, `about:`, or `data:` scheme is allowed
-    if (!url || !url.match(/^([a-zA-Z0-9_+.-]+:\/\/)|(about:)|(data:)/)) {
-      throw new Error('Url or Uri must start with <scheme>://');
+    if (!URL.canParse(url)) {
+      throw new Error('Url or Uri must be a valid URL');
```
mykola-mokhnach: consider putting the actual url value into the error message instead of `Url or Uri` (converted to string and truncated)

Mochxd: Updated. The message now includes the rejected value, stringified and truncated with ``util.truncateString``.

mykola-mokhnach (approved):

Issue #17411, opened 2022-08-24
Unable to register global express middleware from appium plugin 


sudharsan-selvaraj:
## The problem

I'm currently building an appium plugin and I need to add an express middleware to handle all incoming API requests before it is handled by the appium server. When i tried adding the middleware inside `updateSever` methods as below,
```
public static async updateServer(expressApp: any, httpServer: any, cliArgs: any): Promise<void> {
  expressApp.use("/", (requuest, repsonse, next)=>{
    console.log(request.url)
  })
}
```
The middleware is not being called because appium routes are registered before the middleware and all requests are severd automatically by the matching appium route.

One hacky way to solve this issue is to manually update the express route stack and inject the middleware before the appium routes are registered

```
public static async updateServer(expressApp: any, httpServer: any, cliArgs: any): Promise<void> {
  let indexToPushMiddleware = expressApp._router.stack.findIndex((s: any) => s.route);
  expressApp.use("/", (requuest, repsonse, next)=>{
    console.log(request.url)
  });
 expressApp._router.stack.splice(indexToPushMiddleware, 0, expressApp._router.stack.pop());
}
```

Sample output of `console.log(expressApp._router.stack)` is https://gist.github.com/sudharsan-selvaraj/50729d447356dcb6563bbf8652ea6a94

saikrishna321:
@jlipps @boneskull What do you think about think? We are also doing the same in appium-device-farm for remote proxy the command. If anything from `expressApp` changes we need to change this hack as well.

jlipps:
yeah. i'm not sure we can cover all use cases with the express app in a clean way. at the moment i think using the "hack" is probably the right way. unless @boneskull has any ideas.

in the future we could add first class support for adding middleware to the express server, seems like it would need to be complex because there might be various places an extension could want the middleware to live.

boneskull:
yeah other than this workaround, that's what we can do right now. to support this exact use case, we'd need to split the server initialization into an "instantiate express" and "create routes" portions, then run `updateServer` between the two... but you'd also want to run `updateServer` _after_ Appium creates its own routes so that they can be modified.  I'm not sure there's anything to be done there, as your use-case is probably the more unusual of the two.  

another way to approach this would be some sort of `preServer` method might be useful to monkey w/ options provided to the server factory function.  you could then replace the `routeConfiguringFunction` option (which creates all the routes) and replace it with smth like `_.flow(myRouteConfiguringFunction, routeConfiguringFunction)`.

[...]

PR #22632, opened 2026-08-13
feat(base-driver): add httpServer.frontRouter for global extension middleware

mykola-mokhnach:
Drivers/plugins previously had no way to register Express middleware that observes every incoming request: expressApp.use() calls made in updateServer run after Appium's own routes are already registered, so they never see requests matching those routes (appium/appium#17411). configureHttp now mounts an express.Router() before any route is added and exposes it as httpServer.frontRouter, so extensions can attach middleware to it — even from within their existing updateServer hook — and have it observe every request, since Express walks a mounted router's stack per-request rather than at mount time.

### DIFF
--- a/packages/appium/docs/en/developing/build-plugins.md
+++ b/packages/appium/docs/en/developing/build-plugins.md
@@ -496,7 +496,7 @@ other parts of the Appium server!
 won't see requests to paths Appium already owns — Express matches middleware and routes in
 registration order. For middleware that must see *every* request (e.g. logging or auth), use
 `httpServer.frontRouter` instead: Appium mounts this
-[Router](https://expressjs.com/en/4x/api.html#router) before any route is registered, so anything
+[Router](https://expressjs.com/en/5x/api/router/) before any route is registered, so anything
 attached to it — even from inside `updateServer` — still runs ahead of route matching:
 
 ```js
--- a/packages/appium/docs/en/developing/build-plugins.md
+++ b/packages/appium/docs/en/developing/build-plugins.md
@@ -494,10 +494,13 @@ other parts of the Appium server!
 
 `updateServer` runs *after* Appium's own routes are registered, so `expressApp.use(...)` here
 won't see requests to paths Appium already owns — Express matches middleware and routes in
-registration order. For middleware that must see *every* request (e.g. logging or auth), use
-`httpServer.frontRouter` instead: Appium mounts this
+registration order. For middleware that must see requests to those routes too (e.g. logging or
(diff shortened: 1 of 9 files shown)


packages/appium/docs/en/developing/build-plugins.md
```diff
@@ -492,6 +492,22 @@ results you'll need to test! Warning: this should be considered an advanced feat
 knowledge of Express, as well as the care not to do anything that could affect the operation of
 other parts of the Appium server!
 
+`updateServer` runs *after* Appium's own routes are registered, so `expressApp.use(...)` here
+won't see requests to paths Appium already owns — Express matches middleware and routes in
+registration order. For middleware that must see *every* request (e.g. logging or auth), use
+`httpServer.frontRouter` instead: Appium mounts this
+[Router](https://expressjs.com/en/4x/api.html#router) before any route is registered, so anything
```
eglitise: We currently use Express 5:
```suggestion
[Router](https://expressjs.com/en/5x/api/router/) before any route is registered, so anything
```

mykola-mokhnach: Fixed in 7523d19e4 — updated the link to point at the Express 5 Router docs.

packages/appium/docs/en/developing/build-plugins.md
```diff
@@ -492,6 +492,22 @@ results you'll need to test! Warning: this should be considered an advanced feat
 knowledge of Express, as well as the care not to do anything that could affect the operation of
 other parts of the Appium server!
 
+`updateServer` runs *after* Appium's own routes are registered, so `expressApp.use(...)` here
+won't see requests to paths Appium already owns — Express matches middleware and routes in
+registration order. For middleware that must see *every* request (e.g. logging or auth), use
+`httpServer.frontRouter` instead: Appium mounts this
```
KazuCocoa: The docs might address already in another place, but I think it might be helpful to address this will be applied after security boundary in appium such as CORS, websocket or requests rejected during body parsing

KazuCocoa (commented):
lgtm entirely. Left one docs related comment

KazuCocoa (approved):

saikrishna321 (approved):

packages/appium/docs/en/developing/build-plugins.md
mykola-mokhnach: Good point — clarified in f2602d69e that frontRouter still runs after Appium's own baseline middleware (CORS, WebSocket upgrade, body parsing), so it won't see requests already rejected there.

PR #22635, opened 2026-08-15
fix(appium): parse --webhook as a URI instead of host:port

Mochxd:
## Proposed changes

`--webhook` / `server.webhook` is documented and schema-validated as a URI (`format: "uri"`). The fixtures even use values like `http://0.0.0.0/hook`. The Http transport setup still treated it as `host:port`:

```js
const hostAndPort = args.webhook.split(':');
host = hostAndPort[0];
port = parseInt(hostAndPort[1], 10);
```

So `http://0.0.0.0/hook` became `host: "http"`, `port: NaN`, and the path was always `/`. Log shipping never reached the listener people actually configured.

It now uses `URL` and passes `host`, `port`, `path`, and `ssl` into Winston.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have signed the CLA

## Further comments

One thing worth a second opinion: `localhost:9003` passes the `uri` format check, and `new URL()` reads it as scheme `localhost:` with path `9003`, so it would silently produce an empty host. That is also the one shape the old code handled correctly, so I made it throw with a message pointing at the expected form instead of quietly shipping logs nowhere. The error is caught where the transport is created, so the server still starts and the reason ends up in the log. Happy to keep accepting `host:port` for compatibility if you would rather not change that, though the `$comment` in the schema suggests restricting to http/https was the intent.

Added unit coverage around `parseWebhookUri` for the fixture-style URL, https defaults, an explicit port + query string, and the two rejected forms.

### DIFF
--- a/packages/appium/lib/logsink.ts
+++ b/packages/appium/lib/logsink.ts
@@ -185,7 +185,12 @@ export function parseWebhookUri(
     };
   }
   const [host, port] = webhook.split(':');
-  return {host, port: parseInt(port, 10), path: '/', ssl: false};
+  return {
+    host: host || DEFAULT_LEGACY_WEBHOOK_HOST,
+    port: parseInt(port, 10) || undefined,
+    path: '/',
+    ssl: false,
+  };
 }
 
 function createHttpTransport(args: ParsedArgs, logLvl: string): transports.HttpTransportInstance {
--- a/packages/appium/lib/logsink.ts
+++ b/packages/appium/lib/logsink.ts
@@ -153,20 +153,23 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
 /**
  * Parse a `--webhook` / `server.webhook` value into Winston Http transport options.
  *
- * The schema documents this option as a URI (e.g. `http://0.0.0.0/hook`), and such a
- * value has its host, port, path and ssl flag taken from `URL`. A bare `host:port`
- * keeps the meaning it had before, where log records are always posted to the root
- * path of a plain http listener.
+ * The schema documents it as a URI (e.g. `http://0.0.0.0/hook`). A bare `host:port`
+ * is still accepted and always posts to the root path over plain http.
  *
  * @internal
  */
 export function parseWebhookUri(
   webhook: string,
 ): Pick<transports.HttpTransportOptions, 'host' | 'port' | 'path' | 'ssl'> {
-  // `host:port` is parseable as well, but as a URI whose scheme is the host name,
-  // hence the protocol must be verified before the parsing result can be used
+  // `host:port` parses too, into a URI whose scheme is the host name and no host at all
   const url = URL.parse(webhook);
-  if (url && SUPPORTED_WEBHOOK_PROTOCOLS.includes(url.protocol)) {
+  if (url?.hostname) {
+    if (!SUPPORTED_WEBHOOK_PROTOCOLS.includes(url.protocol)) {
+      throw new Error(
+        `The logging webhook must be an http(s) URL, for example 'http://localhost:9003/logs'. ` +
+          `'${webhook}' is given instead`,
+      );
+    }
     const isSsl = url.protocol === HTTPS_PROTOCOL;
     const defaultPort = isSsl ? DEFAULT_HTTPS_WEBHOOK_PORT : DEFAULT_HTTP_WEBHOOK_PORT;
     return {
@@ -176,18 +179,10 @@ export function parseWebhookUri(
       ssl: isSsl,
     };
   }
-  if (!webhook.includes(':')) {
-    return {
-      host: DEFAULT_LEGACY_WEBHOOK_HOST,
-      port: DEFAULT_LEGACY_WEBHOOK_PORT,
-      path: '/',
-      ssl: false,
-    };
-  }
-  const [host, port] = webhook.split(':');
+  const [host, port] = webhook.includes(':') ? webhook.split(':') : ['', ''];
   return {
     host: host || DEFAULT_LEGACY_WEBHOOK_HOST,
-    port: parseInt(port, 10) || undefined,
+    port: parseInt(port, 10) || DEFAULT_LEGACY_WEBHOOK_PORT,
     path: '/',
     ssl: false,
   };
--- a/packages/appium/lib/logsink.ts
+++ b/packages/appium/lib/logsink.ts
@@ -34,8 +34,10 @@ const COLOR_CODE_PATTERN = /\u001b\[(\d+(;\d+)*)?m/g; // eslint-disable-line no-
 
 const HTTPS_PROTOCOL = 'https:';
 const SUPPORTED_WEBHOOK_PROTOCOLS = ['http:', HTTPS_PROTOCOL];
-const DEFAULT_HTTP_PORT = 80;
-const DEFAULT_HTTPS_PORT = 443;
+const DEFAULT_HTTP_WEBHOOK_PORT = 80;
+const DEFAULT_HTTPS_WEBHOOK_PORT = 443;
+const DEFAULT_LEGACY_WEBHOOK_HOST = '127.0.0.1';
+const DEFAULT_LEGACY_WEBHOOK_PORT = 9003;
 
 // https://www.ditig.com/publications/256-colors-cheat-sheet
 const MIN_COLOR = 17;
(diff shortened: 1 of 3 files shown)


packages/appium/lib/logsink.ts
```diff
@@ -143,20 +143,32 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
-    port = parseInt(hostAndPort[1], 10);
+/**
+ * Parse a `--webhook` / `server.webhook` URI into Winston Http transport options.
+ * The schema requires a URI (e.g. `http://0.0.0.0/hook`), so host/port/path/ssl
+ * are taken from `URL` rather than a naive `host:port` split.
+ */
+export function parseWebhookUri(
+  webhook: string,
+): Pick<transports.HttpTransportOptions, 'host' | 'port' | 'path' | 'ssl'> {
+  const url = new URL(webhook);
```
mykola-mokhnach: may this throw an error? should we first check canParse?

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +143,32 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
+  // 'host:port' parses into a URL whose scheme is the host name, which would
+  // otherwise leave us with an empty host and the port as the path
+  if (url.protocol !== 'http:' && url.protocol !== 'https:') {
+    throw new Error(`Expected an http(s) URL, for example 'http://localhost:9003/logs'`);
   }
+  const isSsl = url.protocol === 'https:';
```
mykola-mokhnach: ssl, so we could use a shortcut below?

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +143,32 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
+ * are taken from `URL` rather than a naive `host:port` split.
+ */
+export function parseWebhookUri(
+  webhook: string,
+): Pick<transports.HttpTransportOptions, 'host' | 'port' | 'path' | 'ssl'> {
+  const url = new URL(webhook);
+  // 'host:port' parses into a URL whose scheme is the host name, which would
+  // otherwise leave us with an empty host and the port as the path
+  if (url.protocol !== 'http:' && url.protocol !== 'https:') {
+    throw new Error(`Expected an http(s) URL, for example 'http://localhost:9003/logs'`);
```
mykola-mokhnach: Please also include the information about we expect this url in the logging webhook

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +143,32 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
   }
+  const isSsl = url.protocol === 'https:';
+  return {
+    host: url.hostname,
+    port: url.port ? parseInt(url.port, 10) : isSsl ? 443 : 80,
```
mykola-mokhnach: Please avoid using nested ternaries

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +143,32 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
-  let port = 9003;
-
-  if (args.webhook?.match(':')) {
-    const hostAndPort = args.webhook.split(':');
-    host = hostAndPort[0];
-    port = parseInt(hostAndPort[1], 10);
+/**
+ * Parse a `--webhook` / `server.webhook` URI into Winston Http transport options.
+ * The schema requires a URI (e.g. `http://0.0.0.0/hook`), so host/port/path/ssl
+ * are taken from `URL` rather than a naive `host:port` split.
```
mykola-mokhnach: nit: mark it as `@internal` if only exported for testing

packages/appium/lib/logsink.ts
Mochxd: whoever hits this in the log has no idea which URL we're complaining about. It now reads:
The logging webhook must be an http(s) URL, for example http://localhost:9003/logs. 'localhost:9003' is given instead

packages/appium/lib/logsink.ts
Mochxd: yes i agree it was getting hard to read. Pulled the inner one out into a `defaultPort` variable

packages/appium/lib/logsink.ts
Mochxd: Yeah, it throws on anything it can't parse. I went with `URL.parse` rather than `canParse` so we're not parsing the string twice - it hands back null instead of throwing, which also let me fold the bad protocol case into the same error. It's been there since Node 20.18, so it's fine for the engines range here

packages/appium/lib/logsink.ts
Mochxd: Yeah, that reads better... `isSsl` is worked out once now and the default port just falls out of it

packages/appium/lib/logsink.ts
Mochxd: Good catch, added

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +148,38 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
   return new transports.File(opt);
 }
 
-function createHttpTransport(args: ParsedArgs, logLvl: string): transports.HttpTransportInstance {
-  let host = '127.0.0.1';
-  let port = 9003;
```
mykola-mokhnach, line 148: It looks like the default port value has been lost on the way

mykola-mokhnach: @Mochxd I know AI is nice and helpful, but please check what it generates, and/or ask it to review its own creation. Sometimes the thing hallucinates and it's useful to have additional judge in between iterations

Mochxd: It did go, though I think it was only ever reachable by accident. Those two lines applied when the value contained no colon at all, and in that case they threw away whatever was passed and shipped the logs to 127.0.0.1:9003 instead. From the CLI or a config file you could never get there, since `format: uri` rejects a value with no scheme, so the only way in was the programmatic API, where args skip the parser and are just normalized.

It also is not documented as a default anywhere: the `--webhook` row in the CLI reference has an empty default column, and 9003 does not appear anywhere else in the repo.

So I let the port follow the URL, which means `http://example.com/hook` goes to 80 and the https example to 443. I am happy to put 9003 back as the fallback when the port is omitted if you would rather keep it around, I just found it more surprising for a URL written without a port to quietly end up on :9003. Your call.

packages/appium/lib/logsink.ts
```diff
@@ -32,6 +32,11 @@ const TO_WINSTON_LEVELS_MAP = {
 } as const;
 const COLOR_CODE_PATTERN = /\u001b\[(\d+(;\d+)*)?m/g; // eslint-disable-line no-control-regex
 
+const HTTPS_PROTOCOL = 'https:';
+const SUPPORTED_WEBHOOK_PROTOCOLS = ['http:', HTTPS_PROTOCOL];
+const DEFAULT_HTTP_PORT = 80;
```
mykola-mokhnach: please be specific about this constant purpose, e. g. DEFAULT_HTTP_WEBHOOK_PORT

packages/appium/lib/logsink.ts
mykola-mokhnach: I think we should document this behavior instead of changing how it actually works. Doing that might break existing people's scripts, which we don't want to happen

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +148,38 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
+ *
+ * @internal
+ */
+export function parseWebhookUri(
+  webhook: string,
+): Pick<transports.HttpTransportOptions, 'host' | 'port' | 'path' | 'ssl'> {
+  // 'host:port' is a parseable URI whose scheme is the host name, so the protocol
+  // must be verified as well, otherwise the host would be empty and the port a path
+  const url = URL.parse(webhook);
+  if (!url || !SUPPORTED_WEBHOOK_PROTOCOLS.includes(url.protocol)) {
```
mykola-mokhnach: I assume this change would also be breaking. The previous implementation allowed values like `host:port` or just `host` and now it requires mandatory protocol value

packages/appium/lib/logsink.ts
Mochxd: Sorry about the mess above - I missed reviewing a few of those replies before they went out and some of them posted mangled. That was my mistake, and I have fixed them since.

On the default port you are right, and my earlier reply understated it. `--webhook localhost:9003` passes the `uri` format check, so that form works from the CLI today, and rejecting it would break exactly the scripts you are talking about. Your other comment crossed with this one, and it is the same answer: the mandatory protocol is gone in what I have locally, `host:port` and a bare `host` both keep working.

Here is what I have locally, let me know if it looks right before I push:

- if the value parses as an http(s) URL, host/port/path/ssl come from `URL`, so `http://0.0.0.0/hook` finally lands where it says it does
- anything else falls through to the old `host:port` split untouched, including the `127.0.0.1:9003` fallback when there is no colon
- I ran the old and new implementations over the same inputs: every value that worked before returns exactly the same options, only the ones that produced host `http` with a `NaN` port changed
- constants renamed to `DEFAULT_HTTP_WEBHOOK_PORT` / `DEFAULT_HTTPS_WEBHOOK_PORT`, and the two legacy values got names as well
- the `--webhook` row in the CLI reference now documents the `host:port` form and the `127.0.0.1:9003` fallback

One detail on the bare `host` case, since it surprised me too: the current implementation does not really honour it. With no colon in the value it drops whatever host you passed and posts to `127.0.0.1:9003` anyway, so `--webhook myhost` never reached `myhost`. I kept that as it is, quirk included, because changing it would be a behaviour change rather than the compatibility you are asking for. If you would rather a bare `host` meant that host on port 9003, that is a one line change and I am happy to make it, I just did not want to slip it in as a side effect of this fix.

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +150,47 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
+  if (!webhook.includes(':')) {
+    return {
+      host: DEFAULT_LEGACY_WEBHOOK_HOST,
+      port: DEFAULT_LEGACY_WEBHOOK_PORT,
+      path: '/',
+      ssl: false,
+    };
   }
+  const [host, port] = webhook.split(':');
+  return {host, port: parseInt(port, 10), path: '/', ssl: false};
```
mykola-mokhnach: `port: parseInt(port, 10) || undefined`

mykola-mokhnach: host: host || DEFAULT_LEGACY_WEBHOOK_HOST

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +150,52 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
+ *
+ * @internal
+ */
+export function parseWebhookUri(
+  webhook: string,
+): Pick<transports.HttpTransportOptions, 'host' | 'port' | 'path' | 'ssl'> {
+  // `host:port` is parseable as well, but as a URI whose scheme is the host name,
+  // hence the protocol must be verified before the parsing result can be used
+  const url = URL.parse(webhook);
+  if (url && SUPPORTED_WEBHOOK_PROTOCOLS.includes(url.protocol)) {
```
mykola-mokhnach: for this check if url is still parseable, but contains an unsupported protocol we need to throw an error instead of falling back

packages/appium/lib/logsink.ts
```diff
@@ -143,20 +150,52 @@ function createFileTransport(args: ParsedArgs, logLvl: string): transports.FileT
...
+      host: DEFAULT_LEGACY_WEBHOOK_HOST,
+      port: DEFAULT_LEGACY_WEBHOOK_PORT,
+      path: '/',
+      ssl: false,
+    };
+  }
+  const [host, port] = webhook.split(':');
+  return {
+    host: host || DEFAULT_LEGACY_WEBHOOK_HOST,
+    port: parseInt(port, 10) || undefined,
```
mykola-mokhnach: I think it would still make sense to fallback to the default port here to make it consistent: `port: parseInt(port, 10) || DEFAULT_LEGACY_WEBHOOK_PORT`

mykola-mokhnach: the we could properly combine it with the above expression to avoid unnecessary duplication:

```
const [host, port] = webhook.includes(':') ? webhook.split(':') : [undefined, undefined];
return {
  host: host || DEFAULT_LEGACY_WEBHOOK_HOST,
  port: parseInt(port, 10) || DEFAULT_LEGACY_WEBHOOK_PORT,
  path: '/',
  ssl: false,
};
```

mykola-mokhnach (approved):

PR #22642, opened 2026-08-20
feat(base-driver): add /appium-prefixed routes for JSONWP/MJSONWP end points

mykola-mokhnach:
Adds /session/:sessionId/appium/{rotation,context,contexts,orientation} routes as non-deprecated aliases, and marks the legacy JSONWP/MJSONWP routes they replace as deprecated (routes already deprecated are untouched).

### DIFF
--- a/packages/base-driver/lib/protocol/routes/appium.ts
+++ b/packages/base-driver/lib/protocol/routes/appium.ts
@@ -8,6 +8,24 @@ export const APPIUM_ROUTES = {
   '/appium/sessions': {
     GET: {command: 'getAppiumSessions'},
   },
+  '/session/:sessionId/appium/rotation': {
+    GET: {command: 'getRotation'},
+    POST: {command: 'setRotation', payloadParams: {required: ['x', 'y', 'z']}},
+  },
+  '/session/:sessionId/appium/context': {
+    GET: {command: 'getCurrentContext'},
+    POST: {command: 'setContext', payloadParams: {required: ['name']}},
+  },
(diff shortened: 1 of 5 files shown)


KazuCocoa (approved):

eglitise:
Thanks. The reason I wasn't sure about adding these was that they would have the same method name as the current endpoints - would that not cause issues in drivers?

packages/base-driver/lib/protocol/routes/appium.ts
```diff
@@ -8,6 +8,24 @@ export const APPIUM_ROUTES = {
   '/appium/sessions': {
     GET: {command: 'getAppiumSessions'},
   },
+  '/session/:sessionId/appium/rotation': {
```
eglitise: I think the rotation and orientation endpoints are better suited to have the `/appium/device` prefix. Context should still stay under `/appium`.

mykola-mokhnach: Good call — moved `rotation`/`orientation` under `/appium/device`, kept `context`/`contexts` under `/appium`. Pushed in d39f97914.

mykola-mokhnach:
> Thanks. The reason I wasn't sure about adding these was that they would have the same method name as the current endpoints - would that not cause issues in drivers?

not sure why it should cause any issue. This name is the name of the js function in the driver implementation that handles the given route. For aliases it's expected that they must point to the same handler method.

eglitise (approved):
I'll update the docs in a follow-up PR.

PR #22644, opened 2026-08-20
fix(universal-xml-plugin): iOS findElements drops matches after the first one

Mochxd:
Found this while testing the universal-xml plugin on iOS.

If you call findElements with an xpath that matches more than one node, the plugin turns it into a union query like \/*[1]/*[1]/*[1] | /*[1]/*[1]/*[2]\. Before sending it to WDA we strip the leading \/*[1]\ because AppiumAUT shows up in our transformed page source but not in the source WDA actually queries.

The problem is that strip only happened on the first branch. The second path still had the extra segment, so WDA could not match it. You end up with fewer elements than the xpath should return and no error to tell you something went wrong.

The fix just runs the same strip on each branch of the union. Added a unit test with the existing iOS fixture (two Window nodes) that fails on master and passes with the fix.

## Tests
- [x] \
pm run test:unit\ in packages/universal-xml-plugin (22 tests pass)
- [x] new test fails without the fix (returns 1 instead of 2)

### DIFF
--- a/packages/universal-xml-plugin/lib/plugin.ts
+++ b/packages/universal-xml-plugin/lib/plugin.ts
@@ -109,8 +109,12 @@ export class UniversalXMLPlugin extends BasePlugin {
     if (platformName.toLowerCase() === 'ios') {
       // with the XCUITest driver, the <AppiumAUT> wrapper element is present in the source but is
       // not present in the source considered by WDA, so our index path based xpath queries will
-      // not work with WDA as-is. We need to remove the first path segment.
-      newSelector = newSelector.replace(/^\/\*\[1\]/, '');
+      // not work with WDA as-is. We need to remove the first path segment. A query matching
+      // multiple nodes is a union of paths, so every branch needs the segment removed.
+      newSelector = newSelector
+        .split('|')
+        .map((branch) => branch.trim().replace(/^\/\*\[1\]/, ''))
+        .join(' | ');
     }
     this.log.info(`Selector was translated to: ${newSelector}`);
 
--- a/packages/universal-xml-plugin/lib/plugin.ts
+++ b/packages/universal-xml-plugin/lib/plugin.ts
@@ -82,7 +82,6 @@ export class UniversalXMLPlugin extends BasePlugin {
     strategy: string,
     selector: string,
   ): Promise<Element | Element[]> {
-    const platformName = getPlatformName(driver);
     if (
       strategy.toLowerCase() !== 'xpath' ||
       !driver.getCurrentContext ||
@@ -91,7 +90,7 @@ export class UniversalXMLPlugin extends BasePlugin {
       return (await next()) as Element | Element[];
     }
     const xml = await this.getPageSource(null, driver, null, true);
-    let newSelector = transformQuery(selector, xml, multiple);
+    const newSelector = transformQuery(selector, xml, multiple);
 
     // if the selector was not able to be transformed, that means no elements were found that
     // matched, so do the appropriate thing based on element vs elements
@@ -106,16 +105,6 @@ export class UniversalXMLPlugin extends BasePlugin {
       throw new errors.NoSuchElementError();
     }
 
-    if (platformName.toLowerCase() === 'ios') {
-      // with the XCUITest driver, the <AppiumAUT> wrapper element is present in the source but is
-      // not present in the source considered by WDA, so our index path based xpath queries will
-      // not work with WDA as-is. We need to remove the first path segment. A query matching
-      // multiple nodes is a union of paths, so every branch needs the segment removed.
-      newSelector = newSelector
-        .split('|')
-        .map((branch) => branch.trim().replace(/^\/\*\[1\]/, ''))
-        .join(' | ');
-    }
     this.log.info(`Selector was translated to: ${newSelector}`);
 
     // otherwise just run the transformed query!
--- a/packages/universal-xml-plugin/lib/xpath.ts
+++ b/packages/universal-xml-plugin/lib/xpath.ts
@@ -28,22 +28,26 @@ export function transformQuery(query: string, xmlStr: string, multiple: boolean)
     return null;
   }
 
-  const newQueries = nodes.map((node) => {
-    const indexPath = getNodeAttrVal(node, 'indexPath');
-    // at this point indexPath will look like /0/0/1/1/0/1/0/2
-    const newQuery = indexPath
-      .substring(1) // remove leading / so we can split
-      .split('/') // split into indexes
-      .map((indexStr) => {
-        // map to xpath node indexes (1-based)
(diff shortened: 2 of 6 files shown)


packages/universal-xml-plugin/lib/plugin.ts
```diff
@@ -109,8 +109,12 @@ export class UniversalXMLPlugin extends BasePlugin {
     if (platformName.toLowerCase() === 'ios') {
       // with the XCUITest driver, the <AppiumAUT> wrapper element is present in the source but is
       // not present in the source considered by WDA, so our index path based xpath queries will
-      // not work with WDA as-is. We need to remove the first path segment.
-      newSelector = newSelector.replace(/^\/\*\[1\]/, '');
+      // not work with WDA as-is. We need to remove the first path segment. A query matching
+      // multiple nodes is a union of paths, so every branch needs the segment removed.
+      newSelector = newSelector
+        .split('|')
+        .map((branch) => branch.trim().replace(/^\/\*\[1\]/, ''))
```
mykola-mokhnach: This logic seems weird to me. If we don't need the root node then why it is still preserved in the transformSourceXml helper? Cannot we just cut it off there, so when we build a locator here then it's always built against "clean" xml?

Mochxd: the string surgery did look off. I moved it into the source helper.

The one catch with cutting the root node itself is that `AppiumAUT` is what becomes the universal `UI` root, and Android's `hierarchy` maps to the same thing:

```js
UI: {
  ios: 'AppiumAUT',
  android: 'hierarchy',
},
```

So removing the node would leave iOS with `App` as the root while Android keeps `UI`, which changes `getPageSource` output and breaks the shared shape the plugin exists to provide.

What I did instead is stop giving that root an index path at all. `transformSourceXml` was injecting `index="0"` into `<AppiumAUT>` purely so it would get counted, and I dropped that. Index paths now start at the node below it, which is exactly the hierarchy WDA queries against, so the locator comes out clean and `plugin.ts` no longer touches it.

One related thing that falls out of this: nodes with no index path are now skipped in `transformQuery`. That covers `//UI` and `//*`, which on iOS used to build a selector for the root and then strip it down to an empty string.

Generated selectors are byte for byte identical to before for everything that already worked. I regenerated the iOS index path fixture and updated the expected values in the xpath tests to match.

packages/universal-xml-plugin/lib/xpath.ts
```diff
@@ -28,22 +28,26 @@ export function transformQuery(query: string, xmlStr: string, multiple: boolean)
...
-      .split('/') // split into indexes
-      .map((indexStr) => {
-        // map to xpath node indexes (1-based)
-        const xpathIndex = parseInt(indexStr, 10) + 1;
-        return `*[${xpathIndex}]`;
-      })
-      .join('/'); // reapply /
+  const newQueries = nodes
+    // the ios root node has no index path because the driver does not include it in the hierarchy
+    // it queries, so there is no locator we can build for it
```
mykola-mokhnach: this explanation is confusing. We don't build locator for this node without index because we need to deliberately skip it. It's not part of xpath queries, but just an internal node.

packages/universal-xml-plugin/lib/xpath.ts
```diff
@@ -66,3 +70,14 @@ export function getNodeAttrVal(node: any, attr: string): string {
...
+
+/**
+ * Checks whether a node has a given attribute
+ *
+ * @param node - The XML node
+ * @param attr - The attribute name
+ * @returns True if the node has the attribute
+ */
+export function hasNodeAttr(node: any, attr: string): boolean {
+  return Object.values(node.attributes || {}).some((obj: any) => obj.name === attr);
```
mykola-mokhnach: can we have a better type here instead of `any`?

packages/universal-xml-plugin/lib/xpath.ts
Mochxd: yeah, that comment was worded badly....It described how it happens instead of why we do it, and it read like a general rule when it's really only iOS

on Android the raw `hierarchy` root has `index="0"`, so universal `UI` does get an indexPath and is still findable (`/*[1]` matches `hierarchy` in the source UIA2 queries). iOS is the odd one out because WDA doesn't have `AppiumAUT` in the tree it queries, so we leave that node without an index path on purpose in `transformSourceXml`

Reworded to:

```ts
// nodes with no indexPath cannot be located on the device, so skip them deliberately.
// the only such node is the ios AppiumAUT root (see transformSourceXml)
.filter((node) => node.hasAttribute('indexPath'))
```

The comment in `transformSourceXml` now mentions the Android side too, so the two read together instead of each half explaining itself

packages/universal-xml-plugin/lib/xpath.ts
Mochxd: i am thinking about it the helper didn't need to exist at all....xmldom ships the normal DOM interfaces, so `hasAttribute()` already does this and the `Object.values(node.attributes || {})` scan wasn't needed. Dropped `hasNodeAttr` and the filter just calls `node.hasAttribute('indexPath')`

`getNodeAttrVal` now takes `Element` and uses `getAttribute()`, and `runQuery` returns `Element[]`, so there's no `any` left in the file:

```ts
export function getNodeAttrVal(node: Element, attr: string): string {
  const value = node.getAttribute(attr);
  if (value === null) {
    throw new Error(`Tried to retrieve a node attribute '${attr}' but the node didn't have it`);
  }
  return value;
}
```

one thing worth flagging: `runQuery` needs `as unknown as Element[]`. The `xpath` typings describe lib.dom nodes, which don't overlap with xmldom's, so tsc rejects a direct conversion. I kept that in the one spot next to the existing `@ts-expect-error` for xmldom/xmldom#724 rather than letting `any` back into the rest of the module, but lets go for another one if you'd prefer

packages/universal-xml-plugin/lib/xpath.ts
```diff
@@ -6,12 +7,14 @@ import {select as xpathQuery} from 'xpath';
...
-export function runQuery(query: string, xmlStr: string): any[] {
+export function runQuery(query: string, xmlStr: string): Element[] {
   const dom = new DOMParser().parseFromString(xmlStr, MIME_TYPE.XML_TEXT);
   // @ts-expect-error Missing Node properties are not needed.
   // https://github.com/xmldom/xmldom/issues/724
   const nodes = xpathQuery(query, dom);
-  return nodes as any[];
+  // the xpath typings describe lib.dom nodes rather than xmldom ones, so this is the one place
+  // where the two type worlds have to be bridged
+  return nodes as unknown as Element[];
```
mykola-mokhnach, line 17: 👍

mykola-mokhnach (approved):

PR #22649, opened 2026-08-21
ci: adjust publish docs workflow trigger condition

eglitise:
This should prevent the publish docs workflow from triggering for a dry-run version of the release workflow.

### DIFF
--- a/.github/workflows/publish-doc.yml
+++ b/.github/workflows/publish-doc.yml
@@ -3,10 +3,10 @@ name: Publish docs
 on:
   workflow_dispatch:
   push:
-    # Runs when any new tag is pushed = there is a new release
-    # Pattern matched against refs/tags, so does not trigger for tags with a slash, e.g. 1.0/beta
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-doc.yml
```diff
@@ -2,10 +2,11 @@ name: Publish docs
...
   workflow_dispatch:
-  workflow_run:
-    workflows: ['Appium Server Release']
-    types:
-      - completed
+  push:
+    # Runs when any new tag is pushed = there is a new release
+    # Pattern matched against refs/tags, so does not trigger for tags with a slash, e.g. 1.0/beta
+    tags:        
+      - '*'
```
KazuCocoa: Would it be triggered every tags?
Shall we restrict more with `appium@*`, so it will be triggered with the primary appium tag?

eglitise: Good idea, changed

mykola-mokhnach (approved):

KazuCocoa (approved):

PR #22663, opened 2026-08-23
feat(release): add next-major beta-branch support to release tooling

mykola-mokhnach:
## Summary

- Add opt-in `betaBranch`/`betaChannel` options to `@appium/semantic-release-config` so driver/plugin repos can later publish a long-lived next-major branch as `-beta.N` prereleases, anchored at a fixed target major instead of escalating further on every breaking change.
- Extend this monorepo's own Lerna-based release workflows (`manual-release.yml`, `manual-version.yml`) to allow that beta branch alongside `master`, passing `--conventional-prerelease --preid beta` / `--dist-tag beta` when running on it.
- Add a one-time `bootstrap_packages` workflow input to force a package straight to its first `<major>.0.0-beta.0`, so the anchor doesn't depend on remembering to mark a commit as breaking.
- Update `lerna.json`'s `allowBranch` to match.

### DIFF
--- a/packages/semantic-release-config/index.mjs
+++ b/packages/semantic-release-config/index.mjs
@@ -62,7 +62,13 @@ const APP_GIT_ASSETS = ['package.json', 'package-lock.json', 'CHANGELOG.md'];
  * flavor (`conventionalcommits` commit-analyzer preset, `npmPublish: false`, no GitHub release
  * assets by default, PR/issue comments disabled).
  * @param {string[]} [opts.branches] - Passed through as the top-level `branches` option; omit to
- * use semantic-release's own default branches.
+ * default to `['master']` (or semantic-release's own defaults if `betaBranch` is also omitted).
+ * @param {string} [opts.betaBranch] - Name of a long-lived next-major prerelease branch (e.g.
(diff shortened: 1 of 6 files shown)


packages/semantic-release-config/index.mjs
```diff
@@ -131,8 +139,11 @@ export default function semanticReleaseConfig(opts = {}) {
     plugins: [commitAnalyzerPlugin, releaseNotesGeneratorPlugin, changelogPlugin, npmPlugin, gitPlugin, githubPlugin],
   };
 
-  if (branches) {
-    config.branches = branches;
+  if (branches || betaBranch) {
+    config.branches = [
+      ...(branches ?? ['master']),
```
eglitise, line 144: Note: not all Appium repos have `master` as the default branch, e.g. `appium-chromium-driver`, so those may need to be explicitly adjusted

eglitise (approved):

packages/semantic-release-config/index.mjs
mykola-mokhnach: yes, it's mentioned in the documentation that we use `master` as the default value

PR #22664, opened 2026-08-23
chore(eslint-config-appium-ts): deprecate in favour of oxc-config

mykola-mokhnach:
Consumers should migrate to @appium/oxc-config.

### DIFF
--- a/packages/eslint-config-appium-ts/package.json
+++ b/packages/eslint-config-appium-ts/package.json
@@ -3,6 +3,7 @@
   "version": "3.2.4",
   "gitHead": "f7b20335eab4022e5cbbb627ec86866a994444f8",
   "description": "Shared ESLint config for Appium projects (TypeScript version)",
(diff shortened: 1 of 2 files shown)


KazuCocoa (approved):

packages/eslint-config-appium-ts/package.json
```diff
@@ -53,5 +53,6 @@
   "engines": {
     "node": "^20.19.0 || ^22.12.0 || >=24.0.0",
     "npm": ">=10"
-  }
+  },
+  "deprecated": "This package is deprecated and will be removed from the Appium monorepo. Please migrate to @appium/oxc-config."
```
eglitise: This field isn't recognized by `npm`, so it can be omitted. From what I could find, deprecation on npm is only done through the npm website or CLI.

mykola-mokhnach: Good catch, thanks — you're right, npm doesn't read that field from package.json, deprecation on the registry has to go through `npm deprecate` after publish instead. Dropped it and kept just the README notice.

eglitise (approved):

PR #22665, opened 2026-08-23
fix(strongbox): custom container paths are always rejected on Windows

Mochxd:
On Windows every custom container path is rejected, so `strongbox(name, {container})` cannot be used there at all.

`checkOptions` slugifies the container one path segment at a time, and on Windows the drive letter is just another segment. `slugify` drops everything that is not alphanumeric, so `C:` turns into `C`, the path stops being absolute, and the very next line throws:

```
input   : C:\Users\me\appium-data
slugged : C\Users\me\appium-data   (isAbsolute=false)
result  : TypeError: container slug C\Users\me\appium-data must be an absolute path
```

The existing `should use the provided container path` test already covers this. It builds its path with `path.resolve(path.sep, 'somewhere')`, which is `/somewhere` on Linux but `C:\somewhere` on Windows, so it fails there today. It just never runs on Windows, since unit tests only run on `ubuntu-latest`.

The fix checks that the path is absolute first, then slugifies only the segments below the path root, leaving the root alone. Relative paths still throw, and the error message now shows what was actually passed in rather than the slugified version.

## Tests
- [x] `npm run test:unit` in packages/strongbox on Windows, 34 pass (2 fail on master, including the pre-existing container path test)
- [x] `npm run test:e2e` in packages/strongbox, 12 pass
- [x] default container path behaviour is unchanged, only the custom container branch is touched

### DIFF
--- a/packages/strongbox/test/unit/strongbox.spec.ts
+++ b/packages/strongbox/test/unit/strongbox.spec.ts
@@ -63,6 +63,16 @@ describe('Strongbox', function () {
           const container = path.resolve(path.sep, 'somewhere');
           assert.strictEqual(strongbox('test', {container}).container, container);
         });
+
+        it('should slugify the path segments but keep the path absolute', function () {
+          const container = path.resolve(path.sep, 'some dir', 'another one');
+          const {container: actual} = strongbox('test', {container});
+
+          assert.ok(path.isAbsolute(actual));
+          // the root is what makes the path absolute, so it must survive untouched
+          assert.strictEqual(path.parse(actual).root, path.parse(container).root);
+          assert.strictEqual(actual, path.join(path.parse(container).root, 'some-dir', 'another-one'));
+        });
       });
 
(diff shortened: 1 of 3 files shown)


packages/strongbox/test/unit/strongbox.spec.ts
```diff
@@ -63,6 +63,23 @@ describe('Strongbox', function () {
           const container = path.resolve(path.sep, 'somewhere');
           assert.strictEqual(strongbox('test', {container}).container, container);
         });
+
+        it('should slugify the path segments but keep the path absolute', function () {
```
mykola-mokhnach, line 67: can we add a test to verify a windows and a UNC path slugification?

Mochxd: added a windows drive path and a UNC path case....they run through `path.win32` so they actually execute on the linux unit tests

mykola-mokhnach (approved):

PR #22674, opened 2026-08-28
feat!: convert remaining monorepo packages to ESM-only

mykola-mokhnach:
## Proposed changes

Converts the remaining CJS packages in the monorepo to be ESM-only ahead of the next major (breaking-change) release line tracked on `appium4` (see #22663). This now covers the whole monorepo except the already-deprecated `@appium/eslint-config-appium-ts`:

- `@appium/support`
- `@appium/base-driver`
- `@appium/base-plugin`
- `appium`
- `@appium/fake-plugin`
- `@appium/fake-driver`
- `@appium/types`
- `@appium/logger`
- `@appium/schema`
- `@appium/tsconfig`

Each package gets `"type": "module"` added to its `package.json`, and every remaining CJS idiom (`require()`, `__dirname`/`__filename`, `module.exports`, `require.main`) is converted to its ESM equivalent (`import.meta.dirname`/`import.meta.filename`, `createRequire()` for genuine optional CJS deps, dynamic `import()` for extension/schema loading, etc.). `@appium/tsconfig` has no JS/TS source of its own — it's just `tsconfig*.json` consumed via TypeScript's `extends`, unaffected by the module type — but is included for consistency, along with a `test:smoke` fix (see below).

Test suites for the original six packages (`support`, `base-driver`, `base-plugin`, `appium`, `fake-plugin`, `fake-driver`) are rewritten from `rewiremock` (which only worked under CJS) to Node's native `node:test` `mock.module()` API, gated behind `--experimental-test-module-mocks`. `types`, `logger`, `schema`, and `tsconfig` didn't use `rewiremock`, so no test-mocking rewrite was needed for those.

Also fixed along the way:
- `packages/appium/scripts/autoinstall-extensions.js` had a broken import path (`../build/lib/extension.js` instead of `../build/lib/extension/index.js`) left over from the ESM conversion, which made the `npm install -g appium --drivers=...`/`--plugins=...` postinstall auto-install silently fail on every real (non-monorepo) install.
- `driver.js`/`plugin.js`/`support.js`'s `export *` re-exports didn't forward each package's default export (ES `export *` never does); added explicit `export {default}` alongside.
- `APPIUM_RELOAD_EXTENSIONS`'s `require.cache` eviction was a silent no-op for genuinely ESM extensions (no public cache-eviction API exists for real ESM); replaced with a cache-busting import specifier that works for both CJS and ESM extensions.
- A Node 22–specific `mock.module()` bug: spreading `@appium/support`'s namespace (which has a `default` export) into `namedExports` crashed with `SyntaxError: Unexpected token 'default'` on Node 22 (though not Node 24); fixed by excluding `default` from the spread at each affected call site.

BREAKING CHANGE: `require('@appium/support')` no longer works; consumers must `import` it instead.
BREAKING CHANGE: `require('@appium/base-driver')` no longer works; drivers extending `BaseDriver` via `require` must switch to `import`.
BREAKING CHANGE: `require('@appium/base-plugin')` no longer works; plugins extending `BasePlugin` via `require` must switch to `import`.
BREAKING CHANGE: `require('appium')` no longer works; plugins extending the package via `require` must switch to `import`.
BREAKING CHANGE: `require('@appium/fake-plugin')` no longer works; consumers must `import` it instead.
BREAKING CHANGE: `require('@appium/fake-driver')` no longer works; consumers must `import` it instead.
BREAKING CHANGE: `require('@appium/types')` no longer works; consumers must `import` it instead.
BREAKING CHANGE: `require('@appium/logger')` no longer works; consumers must `import` it instead.
BREAKING CHANGE: `require('@appium/schema')` no longer works; consumers must `import` it instead.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)


packages/appium/lib/extension/extension-config.ts
```diff
@@ -579,9 +589,12 @@ export abstract class ExtensionConfig<ExtType extends ExtensionType> {
       );
     }
     // note: this will only reload the entry point
-    if (process.env.APPIUM_RELOAD_EXTENSIONS && require.cache[entryPointFullPath]) {
-      log.debug(`Removing ${entryPointFullPath} from require cache`);
-      delete require.cache[entryPointFullPath];
+    if (process.env.APPIUM_RELOAD_EXTENSIONS) {
+      const req = createRequire(import.meta.url);
+      if (req.cache[entryPointFullPath]) {
```
eglitise: Maybe there's an ESM equivalent for this?

packages/appium/lib/logsink.ts
```diff
@@ -1,13 +1,17 @@
...
+import type {ParsedArgs} from 'appium/types/index.js';
 import {LRUCache} from 'lru-cache';
 import type {Logform, Logger} from 'winston';
 import {createLogger, format, transports} from 'winston';
 import type Transport from 'winston-transport';
 
-import {adler32} from './utils';
+import {adler32} from './utils/index.js';
+
+// `@appium/logger` is still CJS; under real ESM interop a default import binds to the
```
eglitise: Maybe logger (and I guess schema and tsconfig) should also be converted to ESM in this PR?

packages/appium/scripts/autoinstall-extensions.js
```diff
@@ -73,12 +74,18 @@ async function init() {
     return false;
   }
   try {
-    ({env, util, logger} = require('@appium/support'));
-    // @ts-ignore This is OK
-    ({runExtensionCommand} = require('../build/lib/cli/extension'));
-    ({DRIVER_TYPE, PLUGIN_TYPE} = require('../build/lib/constants'));
-    // @ts-ignore This is OK
-    ({loadExtensions} = require('../build/lib/extension'));
+    const [supportMod, extensionCliMod, constantsMod, extensionMod] = await Promise.all([
```
eglitise, line 77: Note: fixing and optimizing this script isn't that important, since it's slated for removal anyway

mykola-mokhnach: I will 🔪 it in following PR, just did not want to do everything at once

packages/appium/lib/logsink.ts
mykola-mokhnach: I can add it, but this is going to make the PR even larger

packages/appium/lib/extension/extension-config.ts
mykola-mokhnach: Yes — pushed a fix in b40a683d2. There's no public API to evict a module from ESM's cache (unlike `require.cache`), so `require.cache` eviction was actually a silent no-op for any extension that's genuinely ESM (only worked for CJS extensions, since Node's ESM loader delegates to the CJS loader for those). Replaced it with a cache-busting query string on the `requireAsync()` import specifier instead, which works uniformly for both. Same scope as before — only the entry point itself reloads, not files it imports internally.

packages/appium/scripts/autoinstall-extensions.js
eglitise: Sure, it's just a note that it's not worth fixing bugs in it, like what was done in an earlier commit

packages/appium/lib/logsink.ts
eglitise: Usually I'd prefer smaller PRs, but in this case those modules are fairly small compared to the ones already changed here, and it would finish the ESM migration (other than the deprecated eslint-plugin)

packages/appium/test/unit/bootstrap/node-helpers.spec.ts
```diff
@@ -112,9 +112,13 @@ describe('bootstrap/node-helpers', function () {
       }
     });
 
-    it('should adjust NODE_PATH', async function () {
+    it('should no-op under ESM, leaving NODE_PATH unset', function () {
+      // `adjustNodePath()` relies on `require('node:module').Module._initPaths()` to make an
+      // adjusted `NODE_PATH` actually take effect; under ESM there's no `require` global, so it
+      // intentionally leaves `NODE_PATH` unset rather than setting a value Node won't honor.
       adjustNodePath();
```
eglitise, line 116: Note: I assume this method will be removed in a later PR?

mykola-mokhnach: yes

packages/appium/lib/logsink.ts
mykola-mokhnach: ok, migrated the rest of packages

packages/appium/driver.d.ts
```diff
@@ -1 +1,2 @@
 export * from '@appium/base-driver';
+export {default} from '@appium/base-driver';
```
eglitise, line 2: Perhaps it's worth getting rid of the default import altogether?
e.g. consumers will need to use `import {BaseDriver} from '@appium/driver'` etc.
I guess this can also come in a follow-up PR

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -444,6 +458,11 @@ export class BaseDriver<
   }
 }
 
-export * from './commands';
```
eglitise, line 447: Is this export ok to remove?

packages/base-driver/package.json
```diff
@@ -33,16 +33,24 @@
...
   "publishConfig": {
     "access": "public"
   },
   "scripts": {
     "test": "run-p test:unit test:types",
-    "test:e2e": "node --test --enable-source-maps --test-force-exit --test-concurrency=1 --test-timeout=20000 \"./build/test/e2e/**/*.spec.js\"",
+    "test:e2e": "node --test --experimental-test-module-mocks --enable-source-maps --test-force-exit --test-concurrency=1 --test-timeout=20000 \"./build/test/e2e/**/*.spec.js\"",
     "test:smoke": "node ./build/lib/index.js",
-    "test:unit": "node --test --enable-source-maps --test-timeout=10000 \"./build/test/unit/**/*.spec.js\"",
+    "test:unit": "node --test --experimental-test-module-mocks --test-concurrency=1 --enable-source-maps --test-timeout=10000 \"./build/test/unit/**/*.spec.js\"",
```
eglitise, line 53: Is the concurrency needed?
Same for `appium`, `@appium/base-plugin`, `@appium/support`

packages/appium/package.json
```diff
@@ -42,6 +42,7 @@
     "types",
     "tsconfig.json"
   ],
+  "type": "module",
```
eglitise, line 45: Does this package not need the `exports` field like in others?
Same for fake-driver and fake-plugin

packages/support/test/unit/process.spec.ts
```diff
@@ -79,16 +97,20 @@ describe('process', function () {
...
+      });
+      await assert.rejects(mockedProcess.killProcess('tail'), /Oops/);
     });
-    it('should throw an error if pkill fails', async function () {
-      const innerExecStub = sandbox.stub();
-      innerExecStub.returns({stdout: '42\n'});
-      innerExecStub.throws({message: 'Oops', code: 2});
-      (sandbox.stub(teenProcess, 'exec') as any).get(() => innerExecStub);
-      await assert.rejects(process.killProcess('tail'), /Oops/);
+    it('should throw an error if pkill fails', async function (t) {
```
eglitise, line 106: This test is now identical to the one on line 100

packages/base-driver/package.json
```diff
@@ -66,6 +74,10 @@
     "spdy": "4.0.2"
   },
   "tsd": {
+    "compilerOptions": {
```
eglitise, line 77: What's the purpose of these?
Same for `@appium/types`

packages/base-driver/lib/basedriver/driver.ts
mykola-mokhnach: yes, it's not used and shouldn't be there

packages/appium/driver.d.ts
mykola-mokhnach: yes, I wanted to do that long ago. in ESM world there default exports are a pain

packages/base-driver/package.json
mykola-mokhnach: I keep it for now, will take care about tests later

packages/base-driver/package.json
mykola-mokhnach: this for tsd - types validation

KazuCocoa:
Three concrete runtime regressions remain in the ESM migration:

1. **The installed CLI can exit without calling `main()`.** In `packages/appium/index.js`, `process.argv[1] === fileURLToPath(import.meta.url)` is not safe for npm bin execution on Unix: `argv[1]` can retain the invoked symlink path while `import.meta.url` identifies the resolved target. Comparing resolved realpaths (or an equivalent check supported by the minimum Node version) and adding a smoke test that invokes an npm-style bin symlink would cover the real installation path.

2. **`APPIUM_RELOAD_EXTENSIONS` still does not reload CommonJS extensions.** The query-string cache bust in `packages/appium/lib/extension/extension-config.ts` works for ESM, but CJS is cached by resolved filename, so importing the same CJS entry point with a different query still returns the previously evaluated module. Please preserve query busting for ESM and also use `createRequire(import.meta.url)` to resolve and clear the CJS `require.cache` entry, with regression tests proving that both ESM and CJS entry points are reevaluated.

3. **`adjustNodePath()` no longer supports CommonJS extensions.** The ESM no-op in `packages/appium/lib/bootstrap/node-helpers.ts` prevents CJS drivers/plugins from picking up peer dependencies added through `NODE_PATH`. ESM can preserve the previous behavior by using `createRequire(import.meta.url)` to load `node:module` and call `Module._initPaths()`; a test should cover a CJS extension resolving a dependency through the adjusted `NODE_PATH`.

These are runtime compatibility regressions, so I think they should be addressed before merging.

packages/appium/index.js
```diff
@@ -1,9 +1,11 @@
 #!/usr/bin/env node
 
-const appium = require('./build/lib/main.js');
+import {fileURLToPath} from 'node:url';
 
-if (require.main === module) {
+import * as appium from './build/lib/main.js';
+
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
```
KazuCocoa: `process.argv[1]` retains the invoked npm-bin symlink path on Unix, while `fileURLToPath(import.meta.url)` identifies the resolved target, so this predicate is false when the installed `appium` command is launched normally and `main()` is never called. Please compare resolved realpaths (or use an equivalent entry-point check supported by the minimum Node version) and add a smoke test that invokes an npm-style bin symlink; `node ./index.js` will not catch this.

packages/appium/lib/extension/extension-config.ts
```diff
@@ -296,7 +308,15 @@ export abstract class ExtensionConfig<ExtType extends ExtensionType> {
...
     log.debug(`Requiring ${this.extensionType} at ${reqPath}`);
     // https://github.com/nodejs/node/issues/31710
-    const importPath = system.isWindows() ? pathToFileURL(reqPath).href : reqPath;
+    let importPath = system.isWindows() ? pathToFileURL(reqPath).href : reqPath;
+    // note: this will only reload the entry point, same as the old require.cache-eviction
+    // approach — files the entry point itself imports are still served from their own cache
+    if (process.env.APPIUM_RELOAD_EXTENSIONS) {
+      // ESM has no public API to evict a module from its cache (unlike CJS's `require.cache`),
+      // so force a fresh copy by giving this import a unique specifier instead.
+      importPath += `?reload=${reloadCounter++}`;
```
KazuCocoa, line 326: This cache-bust does not reload a CommonJS extension. When `import()` loads CJS, Node caches it by resolved filename in the CJS cache, so changing only the query string still returns the same evaluated CJS module. Please keep query-string busting for ESM, but also use `createRequire(import.meta.url)` to resolve the CJS entry point and clear its `require.cache` entry before importing. Add regression tests proving that both CJS and ESM extension entry points are reevaluated with `APPIUM_RELOAD_EXTENSIONS`.

packages/appium/lib/bootstrap/node-helpers.ts
```diff
@@ -38,8 +38,15 @@ export function adjustNodePath(): void {
 
   const refreshRequirePaths = (): boolean => {
     try {
+      // `require` is not a global under ESM, so this intentionally no-ops here (see
+      // the doc comment above) rather than using `createRequire`, which would make it
+      // actually succeed and contradict that documented, intentional behavior.
+      const req = (globalThis as {require?: NodeRequire}).require;
```
KazuCocoa: No-oping here under ESM breaks the purpose of `adjustNodePath()`: CommonJS drivers/plugins still use `require()` internally and need `Module._initPaths()` to pick up the updated `NODE_PATH` for peer-dependency resolution. ESM can still preserve the old behavior with `createRequire(import.meta.url)('node:module').Module._initPaths()`. Please retain that behavior and cover a CJS extension resolving a dependency through the adjusted `NODE_PATH`.

packages/appium/package.json
mykola-mokhnach: Added one in 290a78e48. fake-driver and fake-plugin deliberately still don't get one though: the extension loader resolves each installed driver/plugin's `appium.schema`/`appium.scripts` manifest paths via Node's package-subpath resolution (mirroring `require.resolve`), against whatever relative path a third-party extension author declares. A restrictive `exports` map blocks anything not explicitly listed — I actually tried adding one to both and it broke driver/plugin installation in the e2e suite (`Package subpath '...' is not defined by "exports"`). `appium` itself is never resolved that way, so it's safe there.

packages/support/test/unit/process.spec.ts
mykola-mokhnach: Good catch, thanks — fixed in 290a78e48. Both mocks unconditionally threw on the first exec() call, so killProcess() always failed at the pgrep step and never reached pkill; made the mock command-aware (succeeds on pgrep, throws on pkill) so this test genuinely exercises the pkill failure path now.

packages/appium/package.json
mykola-mokhnach: Actually had to revert this in 84bf73615 — turns out 6 downstream packages (execute-driver-plugin, images-plugin, plugin-test-support, relaxed-caps-plugin, storage-plugin, universal-xml-plugin) redirect the bare `appium` specifier to a directory via tsconfig `paths`, and once `exports` was added this broke on a genuinely cold `tsc -b` build (fresh clone + npm ci matches CI; a warm/incremental rebuild masked it locally, which is how it slipped through). Repointing those `paths` entries at appium's built output or source didn't survive a cold build either — the CI failure at https://github.com/appium/appium/actions/runs/33235889192 is this. So `appium` stays without an exports map for now; `fake-driver`/`fake-plugin` already don't have one for the schema/scripts resolution reason mentioned above.

packages/appium/index.js
mykola-mokhnach: Confirmed and fixed in f2e793c7d — realpath() both sides now. Verified the failure mode directly: a bare `node.js`-symlink invocation gave `process.argv[1] !== fileURLToPath(import.meta.url)`, so main() never ran. Added test/e2e/bin-symlink.e2e.spec.ts, which symlinks index.js into a temp dir and invokes it that way; confirmed it fails against the old comparison and passes against the fix. Applied the same realpath fix everywhere else this pattern was duplicated (main.ts, base-plugin/fake-plugin/schema's index.ts, gen-cli-args-docs.js).

packages/appium/lib/extension/extension-config.ts
mykola-mokhnach: Good catch, confirmed and fixed in f2e793c7d. Verified empirically: import(cjsPath + '?query') returns the same cached module for a genuine CJS file (Node's CJS loader keys by resolved filename, ignoring the query string). Now clears the resolved entry point from require.cache via createRequire *and* keeps the query-string bust, covering both formats. Added two regression tests against temp-dir CJS/ESM fixtures — there's no longer a real CJS extension anywhere in this repo to reuse for testing, since every package here is ESM now — and confirmed the CJS one fails without the require.cache eviction.

packages/appium/lib/bootstrap/node-helpers.ts
mykola-mokhnach: You're right, fixed in f2e793c7d. My earlier assumption that this needed `require` in scope was wrong — `import {Module} from 'node:module'` gives the same class either way, so `Module._initPaths()` works fine from ESM too. Restored the original 'should adjust NODE_PATH' test and added one that resolves a real dependency through the adjusted NODE_PATH via a fresh createRequire(), confirmed it fails against a simulated no-op. (This also means the 'will this be removed' thread below is moot now — replied there separately.)

packages/appium/test/unit/bootstrap/node-helpers.spec.ts
mykola-mokhnach: Correction: KazuCocoa caught a bug in my earlier fix — I'd made this a no-op based on a wrong assumption (thought Module._initPaths() needed `require` in scope; it doesn't, `import {Module} from 'node:module'` works fine). Restored it to actually function in f2e793c7d, so this method is genuinely useful again and won't be removed after all — sorry for the earlier confusion.

KazuCocoa (approved):
lgtm... 🤞

eglitise (approved):
I don't quite understand the mocking changes in `appium/test/unit/extension`, but the rest seems fine

Issue #21835, opened 2025-12-24
💡 feat: remove ability to autoinstall extensions alongside Appium server


eglitise:
### Is your feature request related to a problem?

Appium has a feature where users can install the server component along with any extensions using just one command. This can be achieved by adding a flag to the regular server installation command as follows:
```sh
npm install -g appium --drivers=xcuitest,uiautomator2 --plugins=images
```
The problem with this feature is that, while undoubtedly useful, it relies on `npm`'s postinstall hook feature, where packages are able to define code that should be executed upon their installation. This feature can be dangerous from a security perspective, and is already being discouraged:
* Supply chain attacks commonly rely on such hooks
* `npm` `v10` suppresses any output from postinstall scripts, meaning that since Appium 3, the autoinstallation script is already partially broken
* `npm` `v12` will throw errors for non-`npm` arguments like `--drivers` and `--plugins`, so the script will break completely
* Other package managers like `pnpm` disable postinstall scripts by default

### Describe the solution you'd like.

Remove the postinstall script completely, and redirect users to other ways for installing extensions

### Describe alternatives you've considered.

Users will still be able to install extensions using the approaches listed [in the Managing Extensions guide](https://appium.io/docs/en/latest/guides/managing-exts/):
* Using Appium's CLI
* Using `package.json` like any other package

eglitise:
Fixed in #22675

PR #22675, opened 2026-08-29
feat(appium)!: remove postinstall autoinstall-extensions script

mykola-mokhnach:
npm's postinstall hooks are a supply-chain security risk, npm@10 already suppresses their output, and npm@12 will reject the non-npm --drivers/ --plugins arguments outright, breaking this feature entirely.

BREAKING CHANGE: `npm install -g appium --drivers=... --plugins=...` no longer installs extensions automatically. Use `appium driver/plugin install` or add extensions to package.json instead.

Fixes #21835


packages/appium/docs/en/guides/migrating-1-to-2.md
```diff
@@ -31,10 +31,6 @@ installing it only installs the core Appium server, without any drivers.
 
 When it comes to installing Appium 2 drivers, there are several approaches you can take:
 
-* Add the `--drivers` flag when installing Appium, for example:
-```bash
-npm i -g appium --drivers=xcuitest,uiautomator2
```
eglitise, line 36: I was unsure about removing this, since Appium 2 users should still be using npm < 10 which doesn't discourage postinstall scripts. But this command would install the latest drivers anyway, which would require the latest major server version, and in turn a newer Node and npm version. So, it's best to remove this after all.

eglitise (approved):

PR #22677, opened 2026-08-29
feat(appium)!: remove Selenium Grid 3 support

mykola-mokhnach:
## Summary

Removes support for the legacy Selenium Grid 3 node registration feature, per the "Remove support for Selenium Grid 3" item in the Appium 4 tracking issue (#22051).

- Deletes the Grid 3 hub-registration client (`grid-v3-register.ts`) and its call site in the main server runner.
- Drops the `--nodeconfig` CLI/config argument from the config schema and generated types.
- Removes the "Using Selenium Grid 3" section from the Grid docs (the Grid 4 relay guide is unaffected).
- Updates fixtures, sample configs, and tests that referenced `nodeconfig`.

BREAKING CHANGE: The `--nodeconfig` server argument and the `server.nodeconfig` config file property no longer exist. Users who need to register with a Selenium Grid should use the Grid 4 relay feature instead (see the Grid guide).


eglitise:
There's one more remaining `nodeconfig` key in `appium/sample-code/appium.config.sample.yaml`

mykola-mokhnach:
> There's one more remaining `nodeconfig` key in `appium/sample-code/appium.config.sample.yaml`

removed

eglitise (approved):

PR #22681, opened 2026-08-30
feat(base-driver)!: remove legacy (M)JSONWP protocol support

mykola-mokhnach:
Removes legacy (M)JSONWP protocol support from `@appium/base-driver`. Appium and every actively maintained driver already speak W3C WebDriver only, so this cleanup drops the machinery that existed purely to interoperate with the old Selenium JSON Wire Protocol / Mobile JSON Wire Protocol.

## What changed

- Removed JSONWP status-code table and `getSummaryByCode` (`jsonwp-status`).
- Removed the downstream MJSONWP↔W3C proxy request transformer (`ProtocolConverter` / `protocol-converter.ts`); `WebDriverProxy` now assumes downstream servers speak W3C only.
- Removed `jsonwpCode` from `ProtocolError` and the `errorFromMJSONWPStatusCode`/`errorFromCode` mapping function.
- Removed the `ELEMENT`/`element-6066-...` dual element-key duplication (`MJSONWP_ELEMENT_KEY`) from response formatting.
- Removed `determineProtocol` and the legacy `{sessionId, status, value}` response-wrapping branch; responses are now always W3C-shaped.
- Removed `setProtocolMJSONWP()` from `DriverCore`.
- Renamed `lib/jsonwp-proxy` → `lib/wd-proxy`.
- Renamed the `JWProxy` class to `WebDriverProxy` and dropped the deprecated `JWProxy` alias entirely.
- Removed the orphaned, unlinked `docs/mjsonwp/` reference docs.

The JSONWP/MJSONWP **route tables** themselves (`JSONWP_ROUTES`, `MJSONWP_ROUTES`) are untouched and still dispatch to their commands — only the legacy-protocol-specific plumbing around them was removed.

BREAKING CHANGE: `@appium/base-driver` no longer exports `statusCodes` or `getSummaryByCode` (the `jsonwp-status` module was removed).
BREAKING CHANGE: `errorFromMJSONWPStatusCode` (aliased as `errorFromCode`) has been removed; use `errorFromW3CJsonCode` for W3C error-signature mapping.
BREAKING CHANGE: `ProtocolError` instances no longer have a `jsonwpCode` property.
BREAKING CHANGE: `WebDriverProxy` (formerly `JWProxy`) no longer proxies to or interprets responses from downstream servers speaking the legacy MJSONWP protocol; the `ProtocolConverter` class, `COMMAND_URLS_CONFLICTS`, and the `downstreamProtocol` getter/setter have been removed. Downstream automation servers must speak W3C WebDriver.
BREAKING CHANGE: Responses no longer duplicate element references under the legacy `ELEMENT` key; only the W3C `element-6066-11e4-a52e-4f735466cecf` key is present. The `MJSONWP_ELEMENT_KEY` constant has been removed.
BREAKING CHANGE: `determineProtocol` has been removed, and drivers can no longer produce the legacy `{sessionId, status, value}` response shape; all responses are now W3C-shaped regardless of how the client formatted its `createSession` capabilities.
BREAKING CHANGE: `DriverCore#setProtocolMJSONWP()` has been removed.
BREAKING CHANGE: The deprecated `JWProxy` export has been removed; use `WebDriverProxy` instead. Its source also moved from `lib/jsonwp-proxy/proxy.js` to `lib/wd-proxy/proxy.js`.
BREAKING CHANGE: `DriverCore#isMjsonwpProtocol()` has been removed.


packages/base-driver/lib/wd-proxy/proxy.ts
```diff
@@ -244,17 +213,9 @@ export class JWProxy {
...
-          const raw = data.sessionId ?? value?.sessionId;
-          this.sessionId = typeof raw === 'string' ? raw : raw != null ? String(raw) : null;
-        }
-        this.downstreamProtocol = this.getProtocolFromResBody(data) ?? this.downstreamProtocol;
-        this.log.info(`Determined the downstream protocol as '${this.downstreamProtocol}'`);
-      }
-      if (Object.hasOwn(data, 'status') && parseInt(data.status as string, 10) !== 0) {
-        throwProxyError(data);
+      if (isSessionCreationRequest && status === 200) {
+        const raw = (data.value as Record<string, unknown> | undefined)?.sessionId;
```
eglitise, line 217: This removes the check for `data.sessionId`, is that okay?
I guess it's JSONWP-only?

packages/base-driver/lib/wd-proxy/proxy.ts
```diff
@@ -418,6 +348,16 @@ export class JWProxy {
     res.status(statusCode).json(ensureW3cResponse(resBody));
   }
 
+  /** Reshapes a legacy `/timeouts` `{type, ms}` body into the single-key W3C form downstream servers expect. */
+  private getTimeoutsRequestBody(body: HTTPBody): HTTPBody {
```
eglitise, line 334: Is this needed?

packages/base-driver/test/unit/protocol/errors.spec.ts
```diff
@@ -203,13 +185,6 @@ const errorsList: ErrorListItem[] = [
     errorName: 'NotYetImplementedError',
     errorMsg: 'Method has not yet been implemented',
     error: 'unknown method',
-    errorCode: 405,
-  },
-  {
-    errorName: 'UnknownCommandError',
-    errorMsg:
-      'The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource.',
```
eglitise, line 211: Should this one be removed?

packages/base-driver/test/unit/wd-proxy/proxy.spec.ts
```diff
@@ -178,14 +177,6 @@ describe('proxy', function () {
       const j = mockProxy({sessionId: '123'});
       await assert.rejects(j.proxy('/badurl', 'GET'), /Could not proxy/);
     });
-    it('should proxy error responses and codes', async function () {
```
eglitise, line 181: Should this and the following 2 tests be removed?

packages/base-driver/lib/wd-proxy/proxy.ts
mykola-mokhnach: yes, W3C protocol does not require session id in responses except for the session creation one

packages/base-driver/lib/wd-proxy/proxy.ts
mykola-mokhnach: I would keep for now, just to be on the safe side. Would you prefer to only keep w3c params?

packages/base-driver/test/unit/protocol/errors.spec.ts
mykola-mokhnach: Yes — that was a stray duplicate of the `UnknownCommandError` entry already listed earlier in the array (same `errorMsg`/`error`, just missing `errorCode`). Removed as part of consolidating the list to one entry per error class; no coverage lost.

packages/base-driver/test/unit/wd-proxy/proxy.spec.ts
mykola-mokhnach: Yes to all 3, but with a nuance worth calling out:
- `should proxy error responses and codes` — `proxy()` no longer inspects/throws on error bodies itself (that moved entirely into `command()`), already covered by `command proxy > should throw when a command fails`.
- `should throw when a command fails with a 200 because the status is not 0` — tested the legacy JSONWP body-embedded-status-despite-200-HTTP override, which we intentionally no longer support; a real HTTP 200 is now trusted as success.
- `should throw when a command fails with a 100` — this one I've actually restored (as `should throw when a command returns a non-2xx, non-error-shaped response`) in 479a6c2b8, since it now exercises real, still-relevant behavior: my `command()` rewrite originally only checked `statusCode < 300` (so 1xx wrongly counted as success) until a review comment on this PR caught it; it's now `>= 200 && < 300`, so a downstream 100 correctly falls through to `UnknownError` again.

packages/base-driver/lib/wd-proxy/proxy.ts
eglitise: I'm just unsure what makes the timeouts endpoints special, compared to all the other ones in ProtocolHandler that are not retained.

packages/base-driver/test/unit/wd-proxy/mock-request.ts
```diff
@@ -28,22 +28,19 @@ export async function request(opts: MockRequestOpts): Promise<MockRequestRespons
...
   if (/\/element\/bad\/text$/.test(url)) {
-    return [500, {status: 11, value: {message: 'Invisible element'}}];
-  }
-  if (/\/element\/200\/text$/.test(url)) {
-    return [200, {status: 11, value: {message: 'Invisible element'}}];
+    return [500, {value: {error: 'element not visible', message: 'Invisible element'}}];
   }
   if (/\/element\/200\/value$/.test(url)) {
-    return [200, {status: 0, sessionId: 'innersessionid', value: 'foobar'}];
+    return [200, {sessionId: 'innersessionid', value: 'foobar'}];
```
eglitise: The object response isn't wrapped in `{value: {..}` like all the other mocks, is that okay?
(since `sessionId` is only expected to be included in the first session response)

packages/base-driver/lib/wd-proxy/proxy.ts
mykola-mokhnach: afaik some drivers were still abusing these params to carry additional timeout info, I need to check for that separately

packages/base-driver/test/unit/wd-proxy/mock-request.ts
mykola-mokhnach: Good catch — that fixture's shape was actually masking something bigger: `ensureW3cResponse()` unconditionally strips `status`/`sessionId` from every proxied response right after this block runs, so the substitution above it never had any observable effect on the client response — it just mutated a field that got deleted moments later. It only existed to normalize a top-level `sessionId`, a shape a legacy JSONWP-style downstream would send on *every* response; a real W3C-only downstream never includes one outside the nested `value.sessionId` on session creation.

Removed the whole substitution block (and the now-unused `getSessionIdFromUrl()`) in 2e3708630, and repurposed the fixture/test to just verify `ensureW3cResponse`'s stray-field stripping, which is the part that actually matters.

packages/base-driver/lib/wd-proxy/proxy.ts
mykola-mokhnach: I've verified it and this value is used to configure newCommandTimeout - an Appium-specific extension with no W3C equivalent (script/pageLoad/implicit only cover the three standard timeout types). That's the only runtime API for changing a session's command timeout after creation; there's no updateSettings-based alternative.

eglitise (approved):
Note: as discussed, in a future PR we should tune the timeouts endpoints to align more closely with the W3C standard

PR #22685, opened 2026-08-31
feat!: bump minimum supported Node.js engine to ^22.22.2 || ^24.15.0 || >=26.0.0

mykola-mokhnach:
BREAKING CHANGE: the minimum supported Node.js engine is set to `^22.22.2 || ^24.15.0 || >=26.0.0`


eglitise:
Why are we dropping 22? It's still supported until May 2027.
I would suggest using the `^22.22.2 || ^24.15.0 || >=26.0.0` range instead, as this matches the requirements of some of the Node-related dependency bumps currently on hold.

mykola-mokhnach:
> Why are we dropping 22? It's still supported until May 2027. I would suggest using the `^22.22.2 || ^24.15.0 || >=26.0.0` range instead, as this matches the requirements of some of the Node-related dependency bumps currently on hold.

Yes, according to https://endoflife.date/nodejs version 22 is still supported by the above date, hovewer it will be only approximately 6 months after we release Appium 4. This means we will again have to delay merging multiple packages because we cannot bump the engine requirements in the server without bumping its major version. So in order to avoid it I would rather prefer to start with 24, which we know is going to be relevant at least until spring 2028.

eglitise:
We can always release Appium 5 closer to the EOL date of Node 22.
I would expect the Node 24 requirement to cause great difficulties for end-users running Node 22 and wanting to upgrade. The fact that this simplifies maintenance _for us_ should be much less important.

eglitise:
Also, I very much doubt that any other packages integrating Appium (such as WDIO) would be willing to bump their minimum requirements to Node 24, until Node 22 EOL.

KazuCocoa:
I agree with leaving 22 yet, relevant tools' version such as WDIO. Companies would be slower to move, so dropping 22 early might not have so much benefit for users.

mykola-mokhnach:
Updated the limit to `^22.22.2 || ^24.15.0 || >=26.0.0` per agreement

eglitise:
:+1: 
Should we also bump npm? 11 might be safe, while for 12 we have some adjustments to make

mykola-mokhnach:
> 👍 Should we also bump npm? 11 might be safe, while for 12 we have some adjustments to make

I thought about it, but there is no practical necessity in doing that. npm 10 still supports all the features we need

eglitise (approved):

PR #22705, opened 2026-09-02
fix(support): replace bplist-creator/bplist-parser with plist v5

mykola-mokhnach:
## Summary

- Drop the unmaintained `bplist-creator`/`bplist-parser` dependencies from `@appium/support` — `plist@5` now ships native `buildBinary()`/`parseBinary()` that cover the same binary plist (`bplist00`) functionality.
- Bump `plist` from `4.0.0` to `5.0.0`.
- Fix a data-corruption bug this migration would otherwise introduce: `plist@5`'s `parseBinary`/`parse` return embedded `<data>` payloads as plain `Uint8Array` instead of `Buffer`. Downstream consumers (`appium-ios-device`, `appium-remote-debugger`) call Buffer-only APIs (`Buffer.isBuffer(...)`, `.toString('utf8')`) on decoded WebInspector data keys (`WIRMessageDataKey`, `WIRSocketDataKey`, `WIRDestinationKey`) — left unpatched, these silently produce garbage (a comma-joined byte string) instead of throwing. `parsePlist`/`parseBinaryPlist` now recursively upgrade any `Uint8Array` back to `Buffer`.
- Document (via characterization tests) a real upstream breaking change in `plist@5`: xml `build()` now silently drops `null` dict values and `null` array elements (previously serialized as `<null/>`), to match Apple's plist DTD. `createBinaryPlist`/`buildBinary` are unaffected and still preserve `null`. No current consumer relies on the old `<null/>` output, so this is left as-is and just locked in with tests.

## Verification

Audited plist/bplist usage across `appium-xcuitest-driver`, `appium-ios-device`, `appium-ios-simulator`, and `appium-remote-debugger`:
- `appium-xcuitest-driver` / `appium-ios-simulator`: only use `parsePlistFile`/`updatePlistFile` on simple Info.plist-style data — unaffected.
- `appium-ios-device`: the heaviest consumer (`usbmux`, `webinspector`, `plist-service`, `mcinstall` transformers) — verified `parsePlist`/`createPlist`/`createBinaryPlist` signatures and behavior are preserved. Its own direct `bplist-creator`/`bplist-parser` dependency (used separately for NSKeyedArchiver's `UID` type in `nskeyed.js`) is untouched by this change.
- `appium-remote-debugger`: doesn't depend on `plist` directly but consumes binary-plist-decoded WebInspector messages from `appium-ios-device` — this is where the Buffer/Uint8Array regression would have surfaced.
- Checked all xml-build call sites (`certificate.ts`'s `toMobileConfig`, `usbmux-encoder.js`'s protocol payloads) for `null` values — none found, so the documented `null`-drop change has no live impact today.


eglitise:
Since plist v5 ships with types, we can also remove `@types/plist`

packages/support/lib/plist.ts
```diff
@@ -128,14 +126,14 @@ export function parsePlist(data: string | Buffer | Uint8Array | ArrayBuffer): ob
   if (typeof data === 'string') {
     return data.startsWith(BPLIST_IDENTIFIER.TEXT)
       ? parseBinaryPlistRoot(Buffer.from(data))
-      : (plistParse(data) as object);
+      : (toBufferDeep(plistParse(data)) as object);
```
eglitise, line 129: Does this line and line 136 need `toBufferDeep`? Both are passed string parameters, and therefore `toBufferDeep` will just return them verbatim.

mykola-mokhnach: Good question, but I believe both are still needed. `toBufferDeep` isn't operating on `data` (the string) itself — it wraps `plistParse(data)`'s *return value*, which is the parsed plist object/tree, not the raw string.

Verified this is still an issue even when the input is a plain string rather than a Buffer:

```js
import {build, parse} from 'plist';
const xml = build({payload: Buffer.from('hello')}); // xml is a string
parse(xml).payload.constructor.name // => 'Uint8Array', not preserved as a string
```

So an XML plist parsed from a string can still carry embedded `<data>` fields that plist@5 decodes to `Uint8Array`, same as the Buffer/Uint8Array/ArrayBuffer branches below. Dropping `toBufferDeep` on lines 129/136 would reintroduce the Buffer-vs-Uint8Array regression for any XML plist that happens to arrive as a string (e.g. `.plist` files read as UTF-8 text) rather than a Buffer.

mykola-mokhnach:
Good catch — removed `@types/plist` in 6a48f5560, since `plist@5` ships its own `.d.ts` and `tsc` builds clean without it.

eglitise (approved):

PR #22708, opened 2026-09-02
feat(support)!: move env/APPIUM_HOME resolution into appium package

mykola-mokhnach:
`env` (resolveAppiumHome/hasAppiumDependency/resolveManifestPath) and the `normalize-package-data`-based package.json reader it relied on were only ever consumed by the `appium` package (directly, or via `npm.ts`'s `installPackage`, itself only called from `appium`). Move them there so `@appium/support` no longer needs `normalize-package-data` at all.

BREAKING CHANGE: `npm.installPackage` now takes a `hasAppiumDependency` boolean computed by the caller instead of importing the check itself, avoiding a support -> appium dependency. 
BREAKING CHANGE: The `env` export is removed from `@appium/support`.


mykola-mokhnach:
> ### 🔵 Needs a closer look
> `findAppiumDependencyPackage` is memoized without a resolver, so differing `acceptableVersionRange` values can return incorrect cached results for the same `cwd`.
> 
> Pull request overview
> File summaries
> Review details
> 💡 [Add a `code-review` agent skill](/appium/appium/new/appium4?filename=.github/skills/code-review/SKILL.md) or configure MCP servers for context-aware, tailored reviews. [Learn more in the docs.](https://docs.github.com/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review?tool=webui#mcp-servers-and-agent-skills)

- It's a real, technically-correct observation about the code's behavior.
- It's not a regression from this PR — it's pre-existing behavior that just happens to show up as "new" lines in the diff because the whole function moved.
- It's latent, not active — no current caller in the codebase passes a custom acceptableVersionRange (every call site uses just cwd), so it can't actually misfire today.

KazuCocoa (approved):

PR #22709, opened 2026-09-02
fix(appium): do not treat a directory as a CLI argument file

Mochxd:
Starting the server from a folder that happens to contain a directory named after the argument value makes the server refuse to start:

```
$ mkdir adb_shell && appium --allow-insecure adb_shell
appium server: error: argument --allow-insecure: Could not read file 'adb_shell': EISDIR: illegal operation on a directory, read
```

The `csv` and `json` transformers accept either a literal value or a path to a file holding that value, and they tell the two apart with `existsSync`. That returns `true` for a directory, so `readFileSync` runs against the folder and fails with `EISDIR`. The value is never given a chance to be parsed as what the user actually meant.

This affects every argument that goes through a transformer: `--allow-insecure` and `--deny-insecure` (`appiumCliTransformer: 'csv'`), `--log-filters` (`appiumCliTransformer: 'json'`), `--default-capabilities` and any `object`-typed extension argument (the `json` transformer is applied automatically). `--use-drivers` and `--use-plugins` are not affected, since array-typed properties without an explicit transformer only ever get `parseCsvLine`.

The fix replaces `existsSync` with a `statSync().isFile()` check, so a directory falls through and gets parsed as a literal value like any other non-path string.

One detail worth calling out: the helper has to treat *every* stat failure as "not a file", not just `ENOENT`. `existsSync` already swallowed everything, and these arguments are usually not paths at all. On Linux a single path component over 255 bytes fails with `ENAMETOOLONG`, and `--default-capabilities` with a realistic caps blob is well past that, so an `ENOENT`-only check would break passing capabilities inline.

## Tests

- [x] `csv` and `json` are covered directly in a new `test/unit/schema/cli-transformers.spec.ts`: a directory stays a literal value, a real file is still read, and a JSON string longer than the max filename length still parses
- [x] `--allow-insecure` with a same-named directory in the cwd is covered in `parser.spec.ts`, alongside the existing `--default-capabilities` cases
- [x] the three new directory/length tests fail on `master` (two with `EISDIR`, one on the parser) and pass with this change
- [x] `npm run test:unit` in packages/appium, 67 pass in the two touched spec files; the remaining suite failures on Windows are pre-existing and unrelated (ESM loader in `extension-config.spec.ts`)
- [x] `npm run lint` clean

### DIFF
--- a/packages/appium/lib/schema/cli-transformers.ts
+++ b/packages/appium/lib/schema/cli-transformers.ts
@@ -1,4 +1,4 @@
-import {existsSync, readFileSync} from 'node:fs';
+import {readFileSync, statSync} from 'node:fs';
 
 import {ArgumentTypeError} from 'argparse';
 
@@ -28,7 +28,7 @@ export const transformers = {
     let csv = csvOrPath;
     let loadedFromFile = false;
     // Value could be a single CSV token or a filepath; attempt file first.
-    if (existsSync(csvOrPath)) {
+    if (isRegularFile(csvOrPath)) {
       try {
         csv = readFileSync(csvOrPath, 'utf8');
       } catch (err) {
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
it looks like node22 unit tests are not very happy

Mochxd:
@mykola-mokhnach let's try again i think it is related to timeout not an issue in the PR

mykola-mokhnach (approved):

PR #22710, opened 2026-09-02
fix(appium): match insecure feature prefixes by automationName, case-insensitively

Mochxd:
The security guide says the prefix on `--allow-insecure` / `--deny-insecure` is the driver's `automationName`:

> The prefix can be either the driver's `automationName`, or the wildcard (`*`) symbol

But when a session starts, `filterInsecureFeatures` compares that prefix against the name the driver is installed under, and it does so case-sensitively. Those are two different strings: UiAutomator2 is installed as `uiautomator2` and its automationName is `UiAutomator2`.

So this:

```bash
appium --allow-insecure=UiAutomator2:adb_shell
```

is accepted at startup, the session starts fine, and `adb_shell` is never enabled. The entry is dropped while being copied onto the driver, and nothing is logged about it.

`isFeatureEnabled` on the driver side already lowercases the prefix before comparing, so the two ends of this disagreed.

This makes the filter accept either the install name or the automationName, and compares case-insensitively so it lines up with `isFeatureEnabled`. The automationName is read from the manifest via `findMatchingDriver`, not from the session capabilities, so nothing the client sends can influence which features get switched on.

Worth being precise about the scope: for every officially maintained driver the install name is just the lowercased automationName (`xcuitest`/`XCUITest`, `espresso`/`Espresso`), so in practice it is the case-sensitivity that bites people following the docs. The automationName lookup matters for third-party drivers that pick a `driverName` unrelated to their `automationName`, which the manifest allows.

Two tests added, both failing before the change:

- a driver installed as `my-driver` with automationName `MyAutomation`, allowed via `MyAutomation:foo`
- `FAKE:foo` and `Fake:bar` against a driver whose names are `fake` / `Fake`

### DIFF
--- a/packages/appium/lib/insecure-features.ts
+++ b/packages/appium/lib/insecure-features.ts
@@ -48,6 +48,9 @@ export function configureGlobalFeatures(this: AppiumDriver) {
  * @param driverName
  */
 export function configureDriverFeatures(this: AppiumDriver, driver: ExternalDriver, driverName: string) {
+  // the docs tell users to scope a feature with the driver's automationName, which is also what
+  // the driver itself compares against in `isFeatureEnabled`
+  const automationName = this.driverConfig?.installedExtensions?.[driverName]?.automationName;
   if (this.relaxedSecurityEnabled) {
     this.log.info(
       `Enabling relaxed security for this session as per the server configuration. ` +
@@ -55,13 +58,13 @@ export function configureDriverFeatures(this: AppiumDriver, driver: ExternalDriv
     );
     driver.relaxedSecurityEnabled = true;
   }
-  const allowedDriverFeatures = filterInsecureFeatures(this.allowInsecure, driverName);
+  const allowedDriverFeatures = filterInsecureFeatures(this.allowInsecure, driverName, automationName);
   if (!util.isEmpty(allowedDriverFeatures)) {
     this.log.info('Explicitly enabling insecure features for this session ' + 'as per the server configuration:');
     allowedDriverFeatures.forEach((a) => this.log.info(`    ${a}`));
     driver.allowInsecure = allowedDriverFeatures;
   }
-  const deniedDriverFeatures = filterInsecureFeatures(this.denyInsecure, driverName);
+  const deniedDriverFeatures = filterInsecureFeatures(this.denyInsecure, driverName, automationName);
   if (util.isEmpty(deniedDriverFeatures)) {
     return;
   }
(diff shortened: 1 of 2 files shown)


packages/appium/lib/insecure-features.ts
```diff
@@ -45,23 +45,29 @@ export function configureGlobalFeatures(this: AppiumDriver) {
...
- * @param driverName
+ * @param driverName the name the driver is installed under, e.g. `uiautomator2`
+ * @param automationName the driver's automationName, e.g. `UiAutomator2`
  */
-export function configureDriverFeatures(this: AppiumDriver, driver: ExternalDriver, driverName: string) {
+export function configureDriverFeatures(
+  this: AppiumDriver,
+  driver: ExternalDriver,
+  driverName: string,
+  automationName: string,
```
mykola-mokhnach: this is a breaking change to a public method

packages/appium/lib/extension/driver-config.ts
```diff
@@ -64,7 +65,12 @@ export class DriverConfig extends ExtensionConfig<DriverType> {
     );
 
     try {
-      const {driverName, mainClass, version} = this._getDriverBySupport(automationName, platformName);
+      const {
+        driverName,
+        mainClass,
+        version,
+        automationName: manifestAutomationName,
```
mykola-mokhnach: is automationName not the same as manifestAutomationName ?

mykola-mokhnach:
I think it makes more sense to align documentation there instead of making the breaking change to the implementation.

Mochxd:
@mykola-mokhnach dropped the signature change. `configureDriverFeatures` keeps its two arguments and `driver-config.ts` is reverted, so `MatchedDriver` and the `manifestAutomationName` rename are both gone. The automationName is looked up from the manifest inside the function instead, since `this` is already the `AppiumDriver`:

```ts
const automationName = this.driverConfig?.installedExtensions?.[driverName]?.automationName;
```

Nothing public changes now. The diff is `insecure-features.ts` and its tests.

On documenting the install name instead of the automationName: I think that would leave the driver side broken. `isFeatureEnabled` compares the prefix against `opts.automationName`, not the name the driver is installed under:

https://github.com/appium/appium/blob/master/packages/base-driver/lib/basedriver/core.ts#L264-L267

For the official drivers it makes no difference, since `uiautomator2` is just `UiAutomator2` lowercased. But for a driver installed as `my-driver` with `automationName: MyAutomation`, `my-driver:foo` would pass the server-side filter and get copied onto `driver.allowInsecure`, then fail in `isFeatureEnabled` because that compares against `myautomation`. The feature stays off even though the user followed the docs.

The case-insensitivity part is separate and holds either way: `isFeatureEnabled` lowercases the prefix and the server-side filter did not, so `Fake:foo` and `fake:foo` behaved differently at the two ends.

Happy to cut this down to just the case fix if you would rather not accept the automationName as a prefix at all.

packages/appium/lib/insecure-features.ts
```diff
@@ -97,20 +100,29 @@ function validateFeatures(features: string[]): string[] {
...
+ * The prefix may be the driver's install name or its automationName, which is the one the
+ * security guide documents. Both are compared case-insensitively, like the prefix check
+ * `isFeatureEnabled` does on the driver side.
+ *
  * @param features
  * @param driverName
+ * @param automationName
  */
-function filterInsecureFeatures(features: string[], driverName: string = ALL_DRIVERS_MATCH): string[] {
+function filterInsecureFeatures(features: string[], driverName?: string, automationName?: string): string[] {
```
mykola-mokhnach, line 114: is it a breaking change that the default value of `driverName` has been removed?

Mochxd: It's not exported, only `appium.ts` imports this module, so nothing outside can call it.

Behaviour is the same for the no-arg calls in `configureGlobalFeatures` too. It used to default to `*` and compare `['*', '*'].includes(prefix)`, now the set is just `{'*'}`. Either way only `*:` entries get through.

I dropped the default because `*` ends up in the set anyway, so it wasn't doing anything

mykola-mokhnach: I observe there are still other places where filterInsecureFeatures is called with single argument - these will be broken

mykola-mokhnach (approved):

PR #22715, opened 2026-09-03
test: use native WebSocket client in e2e tests instead of ws

mykola-mokhnach:
Node 22+ ships a native WebSocket client global, so the outbound test sockets in these e2e specs no longer need the `ws` import. Server-side WebSocketServer/handleUpgrade usage is unaffected since Node has no native WebSocket server.

### DIFF
--- a/packages/storage-plugin/test/e2e/storage.e2e.spec.ts
+++ b/packages/storage-plugin/test/e2e/storage.e2e.spec.ts
@@ -139,6 +139,8 @@ describe('StoragePlugin', function () {
           'message',
           async (event) => {
             const data = event.data;
+            // Native WebSocket.data for a text frame is always a plain string (never a Buffer),
+            // unlike 'ws', so only the string case needs handling here.
             if (typeof data !== 'string') {
               return;
             }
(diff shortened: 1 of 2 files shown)


KazuCocoa (approved):

packages/storage-plugin/test/e2e/storage.e2e.spec.ts
```diff
@@ -134,46 +133,50 @@ describe('StoragePlugin', function () {
...
-              resolve();
-            } else {
-              reject(new Error(JSON.stringify(value)));
+        streamWs.addEventListener('error', () => reject(new Error('streamWs connection error')), {once: true});
+        eventsWs.addEventListener('error', () => reject(new Error('eventsWs connection error')), {once: true});
+        eventsWs.addEventListener(
+          'message',
+          async (event) => {
+            const data = event.data;
+            if (typeof data !== 'string') {
```
eglitise, line 144: The existing implementation also checks if `data` is a buffer, is it okay to omit here?

mykola-mokhnach: Good catch to double-check. Yes, it's safe to omit: with the native `WebSocket`, `event.data` for a text frame is always a plain JS string — never a Node `Buffer` (that's a `ws`-specific data type for its EventEmitter-style `message` callback). Since the server side always sends events as `wsUpstream.send(JSON.stringify(value))` (a string, i.e. a text frame), the client here will never see anything other than a string, or the non-string cases (Blob/ArrayBuffer) that the original code also just ignored via its `else { return; }` branch. Added a short comment in 3b512398c to make that explicit.

eglitise (approved):

PR #22716, opened 2026-09-03
refactor!: remove deprecated driverData mechanism

mykola-mokhnach:
## Summary

- Removes the deprecated `driverData` mechanism (`Core.driverData`, the `driverData` param on `createSession`/`deleteSession`, and the `DriverData` type from `@appium/types`), which was superseded by `IAppiumIpc` for cross-session coordination.
- Drops `FakeDriver`'s `uniqueApp` capability and exclusivity check, the only real consumer of `driverData` — its cross-session coordination has no working replacement in the current `IAppiumIpc`, since IPC channels are scoped per-session and only wired up *after* `createSession` resolves.
- Updates the fake-driver unit test and appium e2e test that exercised `uniqueApp`, and the appium unit test mocks that asserted on the old `driverData` args.

BREAKING CHANGE: `Core.driverData`, `DriverData`, and the `driverData` parameter on `ISessionHandler.createSession`/`deleteSession` are gone. Any driver overriding `get driverData()` or relying on the umbrella driver passing sibling-session data into `createSession`/`deleteSession` needs to migrate off this mechanism.
BREAKING CHANGE: `FakeDriver` no longer supports the `uniqueApp` capability.

### DIFF
--- a/packages/appium/docs/en/developing/build-drivers.md
+++ b/packages/appium/docs/en/developing/build-drivers.md
@@ -231,13 +231,10 @@ async createSession(w3cCaps) {
 
 !!! warning "Deprecated"
 
-    Older drivers may still define `createSession` with up to four parameters, e.g.
-    `createSession(jwpCaps, reqCaps, w3cCaps, otherDriverData)`. This shape is a holdover from the
-    retired JSON Wire Protocol, which required desired and required caps as the first two
-    arguments; only the first W3C-shaped argument was ever used. The `otherDriverData` parameter is
-    likewise deprecated in favor of [IPC](#send-messages-to-plugins-running-on-the-same-session) for
-    coordinating with other concurrently-running sessions. New drivers should use the single-argument
-    form shown above.
+    Older drivers may still define `createSession` with up to three parameters, e.g.
+    `createSession(jwpCaps, reqCaps, w3cCaps)`. This shape is a holdover from the retired JSON Wire
+    Protocol, which required desired and required caps as the first two arguments; only the first
+    W3C-shaped argument was ever used. New drivers should use the single-argument form shown above.
 
 You'll want to make sure to call `super.createSession` in order to get the session ID as well as
 the processed capabilities (note that capabilities are also set on `this.caps`; modifying `caps`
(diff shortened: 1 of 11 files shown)


packages/appium/lib/appium.ts
```diff
@@ -365,37 +358,22 @@ export class AppiumDriver extends DriverCore<AppiumDriverConstraints> {
...
-          // TODO remove this existence guard as a breaking change in Appium 3
-          await extDriver.assignIpc(this.sessionIpcs[innerSessionId]);
-        }
-      } finally {
-        pull(this.pendingDrivers[InnerDriver.name], driverInstance);
+      [innerSessionId, dCaps] = (await driverInstance.createSession(
+        processedW3CCapabilities as never,
+        processedW3CCapabilities,
+        processedW3CCapabilities,
+      )) as [string, DriverCaps<AppiumDriverConstraints> & {webSocketUrl?: string | boolean}];
```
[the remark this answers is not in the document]
mykola-mokhnach: Fixed in c012da798 — removed the driverData-based resource-coordination section and the deprecated 4th createSession argument mention from packages/appium/docs/en/developing/build-drivers.md.

packages/appium/docs/en/developing/build-drivers.md
```diff
@@ -798,50 +795,19 @@ that multiple simultaneous sessions don't use the same resources:
...
-determine which ones you want to use for this particular session.
 
 !!! warning
 
-    Be careful here, since `driverData` is only passed between sessions of a single running Appium
-    server. There's nothing to stop a user from running multiple Appium servers and requesting your
-    driver simultaneously on each of them. In this case, you won't be able to ensure independence
-    of resources via `driverData`, so you might consider using file-based locking mechanisms or
-    something similar.
+    Appium doesn't provide any built-in mechanism for a driver to discover what resources other
```
eglitise: I'm not very familiar with how the IPC implementation works - according to these docs, it's only between sessions of different drivers/plugins, whereas `driverData` was only between sessions of the same driver?

mykola-mokhnach: ipc can be used for both

eglitise: This says 'Appium doesn't provide any built-in mechanism for a driver to discover what resources other concurrently-running sessions of the same driver are using', implying it only works between different driver/plugin types

mykola-mokhnach: although only in scope of a single process. If multiple server instances are used to start different sessions then ipc won't work between these

mykola-mokhnach: Good catch — you're right that they're orthogonal. IPC channels are per-session, shared only by the driver and plugins active in that one session (see the "IPC Notes" section); driverData was cross-session, shared between concurrent sessions of the same driver. IPC never actually replaced driverData's use case. Clarified this explicitly in d01b26208, both in the resources warning and in the IPC Notes bullets.

mykola-mokhnach: You're right, I was wrong — thanks for the pointer. IPC *can* coordinate concurrent sessions of the same driver, just not via the per-session channel Appium auto-wires through onIpcInit. xcuitest-driver's session-claim-handler.ts works by holding its own standalone AppiumIpc instance at module scope (memoized), so every session of that driver loaded in the same Appium server process shares it. The real limit is only that it can't cross Node process/Appium server boundaries. Corrected in 8eac4f692, and now references xcuitest-driver's session-claim-handler.ts as a worked example.

PR #22719, opened 2026-09-04
refactor(logger)!: remove default exports and dead logging APIs

mykola-mokhnach:
Cleans up the @appium/logger package and its downstream consumers (@appium/support, @appium/types, appium):

- Removed the default exports from @appium/logger (log) and @appium/support's internal logging.ts module — use the named log export instead.
- Removed the deprecated AppiumLogger#errorAndThrow method (superseded by errorWithException).
- Removed the long-disabled, no-op progress/unicode API (enableProgress, disableProgress, progressEnabled, enableUnicode, disableUnicode) from the Logger interface and implementation, along with the internal clearProgress/showProgress stubs that backed them.
- Made several internal Log fields private, since nothing outside the class read or wrote them.

BREAKING CHANGE: @appium/logger and @appium/support no longer export a default log — import the named log export instead.
BREAKING CHANGE: AppiumLogger#errorAndThrow has been removed; use errorWithException instead.
BREAKING CHANGE: Logger#enableProgress, disableProgress, progressEnabled, enableUnicode, disableUnicode have been removed; they were already permanently disabled no-ops.

### DIFF
--- a/packages/appium/lib/logsink.ts
+++ b/packages/appium/lib/logsink.ts
@@ -1,4 +1,4 @@
-import globalLog, {type MessageObject} from '@appium/logger';
+import {log as globalLog, type MessageObject} from '@appium/logger';
 import {fs, util} from '@appium/support';
 import type {ParsedArgs} from 'appium/types/index.js';
 import {LRUCache} from 'lru-cache';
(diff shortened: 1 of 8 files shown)


eglitise (approved):

eglitise:
One small thing I missed - would be nice to also update the `@appium/logger` README for the correct import.

mykola-mokhnach:
> One small thing I missed - would be nice to also update the `@appium/logger` README for the correct import.

Updated

PR #22720, opened 2026-09-04
fix(types): declare express and ws as direct dependencies

mykola-mokhnach:
@appium/types publicly exports AppiumServer, AppiumServerExtension and UpdateServerCallback which reference Express/Router and ws.Server types, but express/ws (and their @types packages) were only present as root-level devDependencies. Consumers installing @appium/types or @appium/base-driver standalone couldn't resolve these types.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19061,7 +19061,6 @@
         "@appium/support": "7.2.7",
         "@appium/types": "1.7.0",
         "@types/express": "5.0.6",
-        "@types/ws": "8.18.1",
         "async-lock": "1.4.1",
         "asyncbox": "6.4.2",
         "axios": "1.19.0",
@@ -19073,8 +19072,7 @@
         "method-override": "3.0.0",
         "morgan": "1.11.0",
         "path-to-regexp": "8.4.2",
-        "type-fest": "5.8.0",
-        "ws": "8.21.3"
+        "type-fest": "5.8.0"
       },
       "engines": {
         "node": "^22.22.2 || ^24.15.0 || >=26.0.0",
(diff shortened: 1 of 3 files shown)


eglitise:
Does base-driver need the `@types/express` dependency, if it already depends on `@appium/types` which itself now depends on `@types/express`?

mykola-mokhnach:
Yes — `base-driver` needs its own `@types/express` dependency independently of `@appium/types`.

The reason `@appium/types`'s `@types/express` dependency doesn't cover it: TypeScript resolves a type-only import relative to the file that contains it, walking up `node_modules` from *that file's* location. `@appium/types`'s dependency on `@types/express` only guarantees resolution for imports written inside `@appium/types`'s own `.d.ts` files (e.g. `AppiumServer` referencing `Router`).

`base-driver` has its own direct `import type {...} from 'express'` in several files — `express/express-logging.ts`, `express/idempotency.ts`, `express/middleware.ts`, `express/server.ts`, `protocol/protocol.ts`, `wd-proxy/proxy.ts` — none of which go through `@appium/types` at all. Those resolve starting from `base-driver`'s own directory, so it needs `@types/express` listed directly; relying on it being hoisted transitively from `@appium/types` isn't guaranteed (e.g. under stricter/non-hoisting installs).

By contrast, I did drop `@types/ws` from `base-driver` in the second commit — it has no direct `'ws'` imports anywhere, it only consumes the `AppiumServer` type re-exported from `@appium/types`, so that one *does* resolve transitively through `@appium/types`'s own dependency tree.

eglitise (approved):

PR #22722, opened 2026-09-05
fix(appium): discover hoisted monorepo extensions from search root

### LINKED ISSUE #19819: bug: `appium driver list --installed` is unable to find drivers in monorepo
### Do I have the most recent component updates?

- [X] I use the most recent available driver/plugin and server versions

### Is the component officially supported by the Appium team?

- [X] I have verified the component repository is present under the Appium organization in GitHub

### Current Behavior

`appium driver list --installed` lists no drivers in a monorepo when drivers are hoisted to the root workspace. Example:

```
monorepo
├── node_modules
│   ├── appium
│   ├── appium-uiautomator2-driver  <-- drivers were installed here
│   └── appium-xcuitest-driver      <-- drivers were installed here
├── package.json
└── packages
    └── my-app
        ├── node_modules  <-- no drivers here; they were hoisted to root
        ├── package.json  <-- drivers were requested here
        └── src
```

### Expected Behavior

`appium driver list --installed` should list installed drivers

If you have drivers installed globally, move them elsewhere before running the following commands:

```
git clone https://github.com/microsoft/react-native-test-app.git
cd react-native-test-app
git reset --hard 329f07c5e580c792b549b9a19c90f7e79f4f062c
yarn
cd example
appium driver list --installed
```

### Environment

- Operating system: macOS 14.3.1
- Appium server version (output of `appium --version`): 2.5.1
  - appium-uiautomator2-driver 3.0.0
  - appium-xcuitest-driver 7.1.0
- Appium plugin(s) and their version(s): n/a
- Node.js version (output of `node --version`): v18.19.1
- `npm` version (output of `npm --version`): 10.2.4
- Last component(s) version which did _not_ exhibit the problem: n/a
- Platform and version under test: ?
- Real device or emulator/simulator: emulator/simulator

### Link to Appium Logs

n/a

### Further Information

We've worked around this issue with the following patch:

```diff
diff --git a/build/lib/extension/manifest.js b/build/lib/extension/manifest.js
index 5b650a6ff4dbe737ffdf8821e255e9e4e3d89da3..6179965231285c3ae1d3cddee0c45c57ab3fba03 100644
--- a/build/lib/extension/manifest.js
+++ b/build/lib/extension/manifest.js
@@ -195,12 +195,14 @@ class Manifest {
             onMatch(path_1.default.join(__classPrivateFieldGet(this, _Manifest_appiumHome, "f"), 'package.json'), true),
         ];
         // add dependencies to the queue
-        const filepaths = await support_1.fs.glob('node_modules/{*,@*/*}/package.json', {
-            cwd: __classPrivateFieldGet(this, _Manifest_appiumHome, "f"),
-            absolute: true,
-        });
-        for (const filepath of filepaths) {
-            queue.push(onMatch(filepath));
+        const appPkgPath = path_1.default.join(process.cwd(), 'package.json');
+        const appPkg = JSON.parse(await support_1.fs.readFile(appPkgPath, 'utf8'));
+        for (const name of Object.keys(appPkg.devDependencies)) {
+          try {
+            queue.push(onMatch(require.resolve(name + "/package.json")));
+          } catch (_) {
+            // Ignore packages that don't export `./package.json`
+          }
         }
         // wait for everything to finish
         await bluebird_1.default.all(queue);
```

shubmittal:
## Proposed changes

Add `--ext-search-root` to the Appium server and extension list commands so a workspace can explicitly identify the `package.json` whose declared dependencies contain Appium extensions. Dependencies are resolved using Node module resolution, allowing drivers and plugins hoisted to a monorepo root to be discovered without broadly globbing unrelated packages.

Search-root discoveries are transient and isolated from the persisted `APPIUM_HOME` manifest. The existing discovery behavior remains unchanged when the option is omitted.

Fixes #19819.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist

- [x] I have read the Contributing Guide
- [ ] I have signed the CLA

## Further comments

The repository dependency restore could not complete in the local Windows environment because the configured package proxy returned a 404 for a locked package and direct npm registry access repeatedly failed TLS negotiation. Formatting, non-type-aware linting, strict type-checking of every changed production file, diff checks, and substantive code review completed locally; CI should run the full unit and E2E suites in a clean dependency environment.

### DIFF
--- a/packages/appium/lib/bootstrap/appium-initializer.ts
+++ b/packages/appium/lib/bootstrap/appium-initializer.ts
@@ -12,7 +12,7 @@ import type {
 import {AppiumDriver, type AppiumDriverConstraints} from '../appium';
 import {runExtensionCommand} from '../cli/extension';
 import {injectAppiumSymlinks} from '../cli/extension-command';
-import {ArgParser, getParser} from '../cli/parser';
+import {ArgParser, getExtensionSearchRoot, getParser} from '../cli/parser';
 import {runSetupCommand} from '../cli/setup-command';
 import {SERVER_SUBCOMMAND} from '../constants';
 import {type ExtensionConfigs, loadExtensions} from '../extension';
@@ -57,7 +57,8 @@ export class AppiumInitializer {
 
     adjustNodePath();
 
-    const {driverConfig, pluginConfig} = await loadExtensions(appiumHome);
+    const extensionSearchRoot = args ? args.extSearchRoot : getExtensionSearchRoot();
+    const {driverConfig, pluginConfig} = await loadExtensions(appiumHome, extensionSearchRoot);
 
     const {preConfigArgs, throwInsteadOfExit} = await this.parsePreConfigArgs(args);
     this.throwInsteadOfExit = throwInsteadOfExit;
--- a/packages/appium/lib/cli/parser.ts
+++ b/packages/appium/lib/cli/parser.ts
@@ -359,3 +359,38 @@ export async function getParser(debug = false): Promise<ArgParser> {
 
   return new ArgParser(debug);
 }
+
+/**
+ * Extracts `--ext-search-root` before extension schemas are loaded and the full CLI can be parsed.
+ */
+export function getExtensionSearchRoot(args: string[] = process.argv.slice(2)): string | undefined {
+  if ((args[0] === DRIVER_TYPE || args[0] === PLUGIN_TYPE) && args[1] !== EXT_SUBCOMMAND_LIST && args[1] !== 'ls') {
+    return undefined;
+  }
+  let result: string | undefined;
+  for (let index = 0; index < args.length; index++) {
+    const arg = args[index];
+    if (arg === '--') {
+      break;
+    }
+    if (arg.startsWith('--') && '--ext-search-root'.startsWith(arg)) {
+      const value = args[++index];
+      if (!value || value.startsWith('-')) {
+        throw new Error('[ERROR] argument --ext-search-root: expected one argument');
+      }
+      result = value;
+    } else if (arg.includes('=')) {
+      const separatorIndex = arg.indexOf('=');
+      const name = arg.slice(0, separatorIndex);
+      const value = arg.slice(separatorIndex + 1);
+      if (!name.startsWith('--') || !'--ext-search-root'.startsWith(name)) {
+        continue;
+      }
+      if (!value) {
+        throw new Error('[ERROR] argument --ext-search-root: expected one argument');
+      }
+      result = value;
+    }
+  }
+  return result;
+}
--- a/packages/appium/lib/cli/parser.ts
+++ b/packages/appium/lib/cli/parser.ts
@@ -7,9 +7,11 @@ import type {SubArgumentParserOptions, SubparsersAction} from 'argparse';
 
 import {
   DRIVER_TYPE,
+  EXT_SEARCH_ROOT_ARG,
   EXT_SUBCOMMAND_DOCTOR,
   EXT_SUBCOMMAND_INSTALL,
   EXT_SUBCOMMAND_LIST,
+  EXT_SUBCOMMAND_LIST_ALIASES,
   EXT_SUBCOMMAND_RUN,
   EXT_SUBCOMMAND_UNINSTALL,
   EXT_SUBCOMMAND_UPDATE,
@@ -203,7 +205,7 @@ export class ArgParser {
           command: EXT_SUBCOMMAND_LIST,
           args: extensionArgs[type].list,
           help: `List available and installed ${type}s`,
-          aliases: ['ls'],
+          aliases: [...EXT_SUBCOMMAND_LIST_ALIASES],
         },
         {
           command: EXT_SUBCOMMAND_INSTALL,
@@ -315,9 +317,9 @@ export class ArgParser {
       const [knownArgs, unknownArgs] = parsed;
       // XXX: you'd think that argparse, when given an alias for a subcommand,
       // would set this value to the original subcommand name, but it doesn't.
-      if (knownArgs?.driverCommand === 'ls') {
+      if (EXT_SUBCOMMAND_LIST_ALIASES.includes(knownArgs?.driverCommand)) {
         knownArgs.driverCommand = 'list';
-      } else if (knownArgs?.pluginCommand === 'ls') {
+      } else if (EXT_SUBCOMMAND_LIST_ALIASES.includes(knownArgs?.pluginCommand)) {
         knownArgs.pluginCommand = 'list';
       }
       if (unknownArgs?.length && (knownArgs.driverCommand === 'run' || knownArgs.pluginCommand === 'run')) {
@@ -364,7 +366,11 @@ export async function getParser(debug = false): Promise<ArgParser> {
  * Extracts `--ext-search-root` before extension schemas are loaded and the full CLI can be parsed.
  */
 export function getExtensionSearchRoot(args: string[] = process.argv.slice(2)): string | undefined {
-  if ((args[0] === DRIVER_TYPE || args[0] === PLUGIN_TYPE) && args[1] !== EXT_SUBCOMMAND_LIST && args[1] !== 'ls') {
+  const extensionType = args[0];
+  if (
+    (extensionType === DRIVER_TYPE || extensionType === PLUGIN_TYPE) &&
+    !extensionCommandSupportsArg(extensionType, args[1], EXT_SEARCH_ROOT_ARG)
+  ) {
     return undefined;
   }
   let result: string | undefined;
@@ -373,24 +379,34 @@ export function getExtensionSearchRoot(args: string[] = process.argv.slice(2)): 
     if (arg === '--') {
       break;
     }
-    if (arg.startsWith('--') && '--ext-search-root'.startsWith(arg)) {
+    if (arg.startsWith('--') && EXT_SEARCH_ROOT_ARG.startsWith(arg)) {
       const value = args[++index];
       if (!value || value.startsWith('-')) {
-        throw new Error('[ERROR] argument --ext-search-root: expected one argument');
+        throw new Error(`[ERROR] argument ${EXT_SEARCH_ROOT_ARG}: expected one argument`);
       }
       result = value;
     } else if (arg.includes('=')) {
       const separatorIndex = arg.indexOf('=');
       const name = arg.slice(0, separatorIndex);
       const value = arg.slice(separatorIndex + 1);
-      if (!name.startsWith('--') || !'--ext-search-root'.startsWith(name)) {
+      if (!name.startsWith('--') || !EXT_SEARCH_ROOT_ARG.startsWith(name)) {
         continue;
       }
       if (!value) {
-        throw new Error('[ERROR] argument --ext-search-root: expected one argument');
+        throw new Error(`[ERROR] argument ${EXT_SEARCH_ROOT_ARG}: expected one argument`);
       }
       result = value;
     }
   }
   return result;
+}
+
+function extensionCommandSupportsArg(
+  extensionType: DriverType | PluginType,
+  command: string | undefined,
+  arg: string,
+): boolean {
+  const normalizedCommand = EXT_SUBCOMMAND_LIST_ALIASES.includes(command as 'ls') ? EXT_SUBCOMMAND_LIST : command;
+  const definitions = getExtensionArgs()[extensionType][normalizedCommand as CliExtensionSubcommand];
+  return Boolean(definitions && [...definitions.keys()].some((names) => names.includes(arg)));
 }
--- a/packages/appium/lib/extension/extension-config.ts
+++ b/packages/appium/lib/extension/extension-config.ts
@@ -16,7 +16,7 @@ import type {
 import {APPIUM_VER} from '../helpers/build';
 import {log} from '../logger';
 import {ALLOWED_SCHEMA_EXTENSIONS, isAllowedSchemaFileExtension, registerSchema} from '../schema/schema';
-import {capitalize, resolveFrom} from '../utils';
+import {capitalize} from '../utils';
 import type {Manifest} from './manifest';
 
 const DEFAULT_ENTRY_POINT = 'index.js';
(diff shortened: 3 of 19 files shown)


packages/appium/lib/extension/extension-config.ts
```diff
@@ -118,7 +118,7 @@ export abstract class ExtensionConfig<ExtType extends ExtensionType> {
     }
     let moduleObject: any;
     if (typeof argSchemaPath === 'string') {
-      const schemaPath = await resolveFrom(appiumHome, path.join(pkgName, argSchemaPath));
+      const schemaPath = path.resolve(extManifest.installPath ?? appiumHome, argSchemaPath);
```
mykola-mokhnach, line 121: this change breaks the previous logic, it does not validate if the target module could be loaded

packages/appium/lib/extension/manifest.ts
```diff
@@ -336,6 +390,39 @@ export class Manifest {
...
+        }
+      }
+    }
+    return data;
+  }
+}
+
+async function resolvePackageJson(searchRoot: string, dependencyName: string): Promise<string> {
+  try {
+    return await resolveFrom(searchRoot, `${dependencyName}/package.json`);
```
mykola-mokhnach, line 418: always use path.join, do not hardcode separators

shubmittal: Replaced the hardcoded separator with `path.join(dependencyName, 'package.json')` and updated the manifest tests to use the same platform-native path construction.

packages/appium/lib/extension/extension-config.ts
shubmittal: Restored `resolveFrom` before requiring string schemas. Search-root entries still resolve relative to their discovered `installPath`, while legacy entries retain package-relative resolution from `APPIUM_HOME`; focused tests now assert both resolution paths.

packages/appium/lib/bootstrap/appium-initializer.ts
```diff
@@ -57,7 +57,8 @@ export class AppiumInitializer {
 
     adjustNodePath();
 
-    const {driverConfig, pluginConfig} = await loadExtensions(appiumHome);
+    const extensionSearchRoot = args ? args.extSearchRoot : getExtensionSearchRoot();
```
mykola-mokhnach, line 60: I thought `args.extSearchRoot` is an addition rather than replacement. Do we really want to avoid resolving the "standard" root if `--ext-search-root` is provided?

packages/appium/lib/cli/parser.ts
```diff
@@ -359,3 +359,38 @@ export async function getParser(debug = false): Promise<ArgParser> {
 
   return new ArgParser(debug);
 }
+
+/**
+ * Extracts `--ext-search-root` before extension schemas are loaded and the full CLI can be parsed.
+ */
+export function getExtensionSearchRoot(args: string[] = process.argv.slice(2)): string | undefined {
+  if ((args[0] === DRIVER_TYPE || args[0] === PLUGIN_TYPE) && args[1] !== EXT_SUBCOMMAND_LIST && args[1] !== 'ls') {
```
mykola-mokhnach: this condition seems quite fragile to me, especially if new subcommands are added. Is there any possibility to make it more flexible or hardcoding subcommand names is the only way?

packages/appium/lib/cli/parser.ts
```diff
@@ -359,3 +359,38 @@ export async function getParser(debug = false): Promise<ArgParser> {
...
+    return undefined;
+  }
+  let result: string | undefined;
+  for (let index = 0; index < args.length; index++) {
+    const arg = args[index];
+    if (arg === '--') {
+      break;
+    }
+    if (arg.startsWith('--') && '--ext-search-root'.startsWith(arg)) {
```
mykola-mokhnach: Please avoid using `--ext-search-root` magic everywhere and prefer  a constant instead

packages/appium/lib/cli/parser.ts
```diff
@@ -359,3 +359,38 @@ export async function getParser(debug = false): Promise<ArgParser> {
...
+      const value = args[++index];
+      if (!value || value.startsWith('-')) {
+        throw new Error('[ERROR] argument --ext-search-root: expected one argument');
```
mykola-mokhnach: do we only want to allow a single search root or multiple?

packages/appium/lib/extension/extension-config.ts
```diff
@@ -118,7 +118,11 @@ export abstract class ExtensionConfig<ExtType extends ExtensionType> {
     }
     let moduleObject: any;
     if (typeof argSchemaPath === 'string') {
-      const schemaPath = await resolveFrom(appiumHome, path.join(pkgName, argSchemaPath));
+      const fromDirectory = extManifest.installPath ?? appiumHome;
```
mykola-mokhnach: could you please explain what this change does and why it's needed?

packages/appium/lib/extension/manifest.ts
```diff
@@ -22,24 +24,33 @@ const INITIAL_MANIFEST_DATA: Readonly<ManifestData> = Object.freeze({
...
-   * Returns the memoized manifest for an `APPIUM_HOME` directory (one instance per home).
+   * Returns the memoized manifest for an `APPIUM_HOME` and extension search root pair.
    *
    * @param appiumHome - `APPIUM_HOME` path used as the cache key
+   * @param extensionSearchRoot - Optional project directory whose dependencies contain extensions
    */
-  static getInstance = util.memoize((appiumHome: string): Manifest => new Manifest(appiumHome));
+  static getInstance = util.memoize(
+    (appiumHome: string, extensionSearchRoot?: string): Manifest =>
+      new Manifest(appiumHome, extensionSearchRoot && path.resolve(extensionSearchRoot)),
```
mykola-mokhnach: why do we need to resolve extensionSearchRoot here, shouldn't it be coming already resolved?

packages/appium/lib/extension/manifest.ts
```diff
@@ -210,12 +228,48 @@ export class Manifest {
 
     const queue: Promise<void>[] = [onMatch(path.join(this.#appiumHome, 'package.json'), true)];
 
-    const filepaths = await fs.glob('node_modules/{*,@*/*}/package.json', {
-      cwd: this.#appiumHome,
-      absolute: true,
-    });
-    for (const filepath of filepaths) {
-      queue.push(onMatch(filepath));
+    if (extensionSearchRoot) {
```
mykola-mokhnach: this addition makes the method too big and complicated. Consider splitting/simplifying it

packages/appium/lib/extension/extension-config.ts
```diff
@@ -118,7 +118,11 @@ export abstract class ExtensionConfig<ExtType extends ExtensionType> {
...
+      const moduleId = extManifest.installPath
+        ? path.resolve(extManifest.installPath, argSchemaPath)
```
KazuCocoa: This also could change schema loading for existing manifest entries with an `installPath`, even when `--ext-search-root` is omitted. For example, a driver with `appium.schema: "schema.json"` and `exports: {"./schema.json": "./build/schema.json"}` previously loaded through `resolveFrom(appiumHome, "driver-package/schema.json")`. The absolute path here bypasses the export mapping and looks for a nonexistent file at the package root. I reproduced `MODULE_NOT_FOUND` with the new schema reader; schema validation then excludes the extension. Could we preserve package-subpath resolution while resolving from the correct installed package, and add a regression test for an export-mapped schema?

packages/appium/lib/extension/manifest.ts
```diff
@@ -336,6 +390,39 @@ export class Manifest {
...
+}
+
+async function resolvePackageJson(searchRoot: string, dependencyName: string): Promise<string> {
+  try {
+    return await resolveFrom(searchRoot, path.join(dependencyName, 'package.json'));
+  } catch (err) {
+    if (!(err instanceof Error && 'code' in err && err.code === 'ERR_PACKAGE_PATH_NOT_EXPORTED')) {
+      throw err;
+    }
+    const entryPoint = await resolveFrom(searchRoot, dependencyName);
```
KazuCocoa: If a hoisted ESM extension hides `./package.json` and declares `exports: {".": {"import": "./index.js"}}`, the first resolution throws `ERR_PACKAGE_PATH_NOT_EXPORTED`, but this fallback throws the same error because `resolveFrom` uses CommonJS resolution and does not select the `import` condition. The caller catches the error and skips the dependency, so the driver/plugin remains undiscoverable despite being loadable by Appium's ESM extension loader.
Could the fallback locate the package manifest without requiring a CommonJS-resolvable entry point, with a real-filesystem test for import-only exports?

packages/appium/lib/extension/manifest.ts
```diff
@@ -336,6 +390,39 @@ export class Manifest {
...
+    return path.join(fs.findRoot(path.dirname(entryPoint)), 'package.json');
```
KazuCocoa: For an extension that hides `./package.json`, exports `./build/index.js`, and includes `build/package.json` containing only `{"type": "commonjs"}`, `fs.findRoot(path.dirname(entryPoint))` stops at `build/`. The resolver consequently returns `build/package.json` instead of the package-root manifest containing the extension metadata. `onMatch` then silently ignores it because it has no `appium` metadata, leaving the installed driver/plugin undiscovered. Could we identify the owning package root rather than stopping at an arbitrary nested manifest, and cover this layout with a real-filesystem test?

packages/appium/lib/extension/extension-config.ts
shubmittal: Fixed this by resolving schema strings as package subpaths from the known installed package. If the package declares `exports`, its own export map is honored (including `"./schema.json": "./build/schema.json"`) and excluded subpaths remain excluded; packages without `exports` resolve from the known root to preserve linked/aliased installs. Added real-filesystem coverage for the mapped schema and competing-package cases, plus driver schema validation coverage.

packages/appium/lib/bootstrap/appium-initializer.ts
shubmittal: Agreed. I split standard and explicit-root discovery so the additive behavior is clear: `read()` keeps the normal `APPIUM_HOME` refresh, snapshots that persistent data, and then merges declared dependencies from `--ext-search-root` transiently. I added regression coverage asserting that drivers from both roots are visible, the standard-root entry is written, and the explicit-root entry is not persisted.

packages/appium/lib/cli/parser.ts
shubmittal: Added `EXT_SEARCH_ROOT_ARG` and replaced the repeated production magic strings with that constant.

packages/appium/lib/extension/extension-config.ts
shubmittal: Extensions discovered through the explicit project root may be hoisted outside `APPIUM_HOME`, so schema resolution must start from the recorded installation. The final implementation preserves package-subpath semantics: packages with `exports` honor their mapping and exclusions from their known installation, while linked or aliased packages without `exports` resolve relative to that known package root. Legacy entries without `installPath` keep package-relative resolution from `APPIUM_HOME`.

packages/appium/lib/extension/manifest.ts
shubmittal: Split the method into standard installed-extension discovery and declared-dependency discovery. Package loading, declared-manifest parsing, and per-dependency resolution now live in focused helpers, while `read()` makes the persistent-then-transient sequence explicit.

packages/appium/lib/extension/manifest.ts
shubmittal: Moved normalization to the `loadExtensions` boundary. `Manifest.getInstance` now receives the absolute root and no longer calls `path.resolve` in both the memoized factory and cache-key resolver.

packages/appium/lib/extension/manifest.ts
shubmittal: Replaced the entry-point fallback with package-manifest lookup through Node’s module search paths. It reads the owning dependency’s `package.json` directly, so discovery no longer requires a CommonJS-resolvable root export. Added a real-filesystem import-only export case and verified the root Appium metadata is read.

packages/appium/lib/cli/parser.ts
shubmittal: I kept this as a single root. Issue #19819, the proposed `--ext-search-root` API, current docs, and `extSearchRoot?: string` all describe one project root; repeated options retain argparse’s last-value behavior. Supporting multiple roots would need an explicit precedence/merge contract, so I did not expand the API speculatively in this fix.

packages/appium/lib/cli/parser.ts
shubmittal: Reworked this to derive support from the selected extension command’s registered argument definitions instead of hardcoding `list`/`ls`. The list alias is centralized too, with focused coverage that adds the option to another command definition and checks bootstrap detection follows it.

packages/appium/lib/extension/manifest.ts
shubmittal: The resolver now constructs the declared dependency’s package-root manifest candidate from Node’s module search paths and canonicalizes that path; it no longer walks upward from an exported entry point. Added a real-filesystem package containing `build/package.json` and verified resolution still returns the root manifest with Appium metadata.

packages/appium/lib/extension/manifest.ts
```diff
@@ -336,6 +352,81 @@ export class Manifest {
...
+    pkg = JSON.parse(await fs.readFile(packageJsonPath, 'utf8')) as unknown;
+  } catch (err) {
+    throw new Error(`Could not read extension search root package manifest at ${packageJsonPath}`, {cause: err});
+  }
+  if (!util.isPlainObject(pkg)) {
+    throw new TypeError(`Extension search root package manifest at ${packageJsonPath} must contain an object`);
+  }
+
+  const dependencyNames = new Set<string>();
+  for (const field of ['dependencies', 'devDependencies'] as const) {
```
KazuCocoa, line 420: Is it intentional to not include `optionalDependencies` here?

shubmittal: No, that omission was not intentional. I added `optionalDependencies` alongside `dependencies` and `devDependencies`, with coverage for installed and missing optional packages and deduplication. Missing optional packages remain non-fatal. `peerDependencies` and `overrides` remain excluded from this discovery pass.

KazuCocoa:
I wondered if appium4 branch is appropriate if you don't have strong motivation to v3 specifically

shubmittal:
I don’t have a v3-specific requirement. I’ve prepared and validated a focused Appium 4 port locally. Because the branches and manifest architecture have diverged, directly retargeting #22722 would include unrelated commits; a clean Appium 4 branch keeps the change scoped. If this direction looks right, I can publish it as a replacement PR against `appium4`.

PR #22729, opened 2026-09-05
test(base-driver): fix invalid import

eglitise:
Not sure how TS did not catch this before.
Fixes failing CI for #22728

### DIFF
--- a/packages/base-driver/test/types/basedriver/driver.test-d.ts
+++ b/packages/base-driver/test/types/basedriver/driver.test-d.ts
@@ -1,7 +1,7 @@
 import {BaseDriverCapConstraints, Driver, DriverOpts, ExternalDriver} from '@appium/types';
 import {expectAssignable} from 'tsd';
 
-import {BaseDriver} from '.../../../lib/basedriver/driver';
+import {BaseDriver} from '../../../lib/basedriver/driver';
 
 expectAssignable<Driver<BaseDriverCapConstraints>>(
   new BaseDriver<BaseDriverCapConstraints>({} as DriverOpts<BaseDriverCapConstraints>),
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
Pushed a follow-up commit to get CI green.

The invalid-import fix was correct, but it uncovered a second issue: `tsd` builds its type-check program from an explicit file list rather than `tsconfig`'s `include` globs, and `@appium/tsconfig` sets `composite: true`, which requires every reachable file to be declared explicitly. Once the import actually resolved to the real `lib/basedriver/driver.ts`, `tsd` pulled in its whole transitive dependency tree (`constants.ts`, `express/server.ts`, `protocol/*`, etc.) and failed with `TS6059: File is not listed within the file list of project`.

Fixed by importing from the package entry point instead of raw source, matching the pattern already used in `packages/strongbox/test/types/strongbox.test-d.ts`:

```diff
-import {BaseDriver} from '../../../lib/basedriver/driver';
+import {BaseDriver} from '@appium/base-driver';
```

Verified locally with `npm run test:types` in `packages/base-driver`.

mykola-mokhnach:
Update: the `@appium/base-driver` self-import fix broke the main build instead — `Setup, install and build` failed across the board with a big cascade of unrelated-looking type errors in downstream packages (`fake-plugin`, `images-plugin`, etc.).

Root cause: `packages/base-driver/tsconfig.json` has `"include": ["lib", "test"]`, so `test/types` is part of the main composite `tsc -b` build too, not just the separate `tsd` step. Once `driver.test-d.ts` imports `@appium/base-driver` by package name, that's a self-reference — during the main build, TypeScript needs `build/lib/index.d.ts` to resolve it, but that file doesn't exist yet since it's mid-generation in that very build. Everything downstream of base-driver then compiled against a broken/missing base-driver output, hence the unrelated-looking cascade.

`packages/strongbox` already sidesteps this: its tsconfig only includes `["lib", "test/unit", "test/e2e"]`, deliberately excluding `test/types` from the main build — `tsd` targets `test/types` independently via its own `"directory"` config in `package.json`, so it's unaffected by the exclude.

Applied the same pattern to base-driver's tsconfig. Verified from a fully clean `npm ci --foreground-scripts` (matching CI) that the full build now succeeds with zero errors, and `npm run test:types` in `base-driver` still passes.

PR #22731, opened 2026-09-05
feat(appium)!: replace --allow-cors flag with --allow-insecure=*:cors

mykola-mokhnach:
CORS is now gated through the existing insecure-feature mechanism (allow-insecure/deny-insecure/relaxed-security) instead of a dedicated flag, consistent with how other server-level features like session_discovery work.

BREAKING CHANGE: The --allow-cors CLI flag and allow-cors config file property have been removed. Use --allow-insecure=*:cors (or --relaxed-security, optionally overridden with --deny-insecure=*:cors) instead.


eglitise:
Should the test `should read paths with hyphenated property segments` at `appium/test/unit/utils/object.spec.ts` also be adjusted here? I understand it's a unit test and all, so the exact value isn't relevant, but still.

mykola-mokhnach:
Good catch — updated in cf7764198: the test now uses `allow-insecure` instead of the removed `allow-cors` as its example key.

eglitise (approved):

Issue #15916, opened 2021-09-29
handle more validation via ajv


boneskull:
After #15664 (hopefully) lands, we will have the opportunity to validate various bits of user-provided configuration via `ajv`.

There are a few places (off the top of my head) where this would make sense. In relative order of impact:

1.  Extension manifest.  That means stuff that lives in the `appium` property of an extension's `package.json`.  
2. Less important, but related, would be to define the shape of `$APPIUM_HOME/extensions.yaml`.   This file is not intended to be end-user-editable (afaik) but it would be good to check in the case that an end-user monkeys with or otherwise corrupts it. 
3. The contents of the value of `--default-capabilities` arg (`server.defaultCapabilities` object).  This is not validated by the code in #15664 other than "it should be an object".  Note that the schema varies based on drivers used.  Unclear if _any_ validation happens here.
4. Common and extension capabilities during runtime.  wdio provides a very thorough set of types for capabilities for reference.  Ideally, schemas for this exist already (e.g., a schema for W3C-style capabilities), but if they don't, working backwards from the types may be helpful.  This either _depends on_ or _is depended on_ by the previous list item.

The goal of this task should result in a lot of existing handrolled validation code being removed, and avoiding it in the future.

Should this be labeled `enhancement`? It's more of a refactor, though it'll likely positively impact UX and DX.

jlipps:
> Note that the schema varies based on drivers used. Unclear if any validation happens here.

These just get put into caps and I believe pass through the normal driver cap validation flow.

> Ideally, schemas for this exist already (e.g., a schema for W3C-style capabilities)

There are only a handful of official W3C capabilities. The rest are invented by Appium. Drivers should really be the ones defining capabilities and exposing validators/schemas.

boneskull:
Fair enough.  Hopefully drivers won't need to actually _implement_ any validation functions, though.

jlipps:
Right, they just define the minimal schema like they currently do.

mykola-mokhnach:
Added more validation in v4

PR #22732, opened 2026-09-05
feat(appium): validate extension manifests and extensions.yaml via ajv

mykola-mokhnach:
## Proposed changes

Progresses [#15916](https://github.com/appium/appium/issues/15916) ("handle more validation via ajv"), a long-standing epic proposing that Appium replace hand-rolled config/manifest validation with `ajv`/JSON Schema. Two of that issue's items were still unaddressed: validation of the `appium` field in an extension's `package.json`, and validation of the overall shape of `$APPIUM_HOME/extensions.yaml` (which today has none at all beyond catching YAML parse errors).

This PR:

- Adds `packages/appium/lib/extension/manifest-schema.ts` and `manifest-validator.ts`, which replace the hand-rolled `typeof`/`Array.isArray` checks in `ExtensionConfig.getGenericConfigProblems()` and `DriverConfig.getConfigProblems()` with `ajv` schema validation. User-facing problem messages are unchanged, so no behavior changes for existing valid/invalid manifests.
- Adds a shallow `ajv` envelope check for `extensions.yaml` itself, run in `Manifest.read()` right after parsing. A corrupted-but-syntactically-valid file (e.g. `drivers` not being an object) now resets to the initial empty manifest and logs a warning instead of silently loading bad data that crashes later downstream with confusing errors (e.g. `automationName.toLowerCase is not a function`).
- Keeps the stateful cross-driver duplicate-`automationName` check as plain code, since it depends on state across multiple extensions and can't be expressed as a single-object JSON schema.
- Extracts the `ajv` CJS/ESM interop type-cast (previously only in `lib/schema/schema.ts`) into a shared `lib/schema/ajv-ctor.ts`, since the new validator needs its own dedicated `Ajv` instance (manifest loading happens before the `AppiumSchema` singleton used for config-file/CLI validation is ever finalized).

## Further comments

The new envelope/manifest schemas are deliberately narrow — scoped to exactly what the existing hand-rolled checks covered (plus the new top-level envelope shape) — rather than a full `ManifestData` schema, so that old-but-valid manifest revisions (v2/v3, before `installPath`/`dev` install type existed) keep parsing without changes. A dedicated `Ajv` instance is used instead of reusing the `AppiumSchema` singleton from `lib/schema/schema.ts`, since manifest loading happens during `AppiumInitializer.init()` well before that singleton's `finalize()` is ever called.

### DIFF
--- a/packages/appium/lib/extension/driver-config.ts
+++ b/packages/appium/lib/extension/driver-config.ts
@@ -1,12 +1,11 @@
-import {util} from '@appium/support';
 import type {DriverClass, DriverType, StringRecord} from '@appium/types';
 import type {ExtManifest, ExtName, ExtRecord} from 'appium/types/index.js';
 
 import {DRIVER_TYPE} from '../constants.js';
 import {log} from '../logger.js';
-import type {ExtManifestProblem} from './extension-config.js';
 import {ExtensionConfig} from './extension-config.js';
-import type {Manifest} from './manifest.js';
+import {manifestValidator} from './manifest/index.js';
+import type {ExtManifestProblem, Manifest} from './manifest/index.js';
 
 export type MatchedDriver = {
   driver: DriverClass;
@@ -87,45 +86,20 @@ export class DriverConfig extends ExtensionConfig<DriverType> {
 
   protected override getConfigProblems(extManifest: ExtManifest<DriverType>, extName: string): ExtManifestProblem[] {
     void extName;
-    const problems: ExtManifestProblem[] = [];
-    const {platformNames, automationName} = extManifest;
+    const problems = manifestValidator.getDriverManifestProblems(extManifest);
+    // `?.`, not destructuring: a corrupted entry may be null/non-object (already reported above).
+    const automationName = (extManifest as unknown as Record<string, unknown> | null | undefined)?.automationName;
 
-    if (!Array.isArray(platformNames)) {
-      problems.push({
-        err: 'Missing or incorrect supported platformNames list.',
-        val: platformNames,
-      });
-    } else if (util.isEmpty(platformNames)) {
-      problems.push({
-        err: 'Empty platformNames list.',
-        val: platformNames,
-      });
-    } else {
-      for (const pName of platformNames) {
-        if (typeof pName !== 'string') {
-          problems.push({
-            err: 'Incorrectly formatted platformName.',
-            val: pName,
-          });
-        }
+    // Only track actual strings — `undefined` would make every automationName-less entry a "duplicate".
+    if (typeof automationName === 'string') {
+      if (this.knownAutomationNames.has(automationName)) {
+        problems.push({
+          err: 'Multiple drivers claim support for the same automationName',
+          val: automationName,
+        });
       }
+      this.knownAutomationNames.add(automationName);
     }
-
-    if (typeof automationName !== 'string') {
-      problems.push({
-        err: 'Missing or incorrect automationName',
-        val: automationName,
-      });
-    }
-
-    if (this.knownAutomationNames.has(automationName as string)) {
-      problems.push({
-        err: 'Multiple drivers claim support for the same automationName',
-        val: automationName,
-      });
-    }
-
-    this.knownAutomationNames.add(automationName as string);
 
     return problems;
   }
--- a/packages/appium/lib/extension/manifest/validator.ts
+++ b/packages/appium/lib/extension/manifest/validator.ts
@@ -1,0 +1,156 @@
+import {Ajv, type ErrorObject, type ValidateFunction} from '../../schema/ajv.js';
+import {commonExtManifestProblemsSchema, driverExtManifestProblemsSchema, manifestEnvelopeSchema} from './schema.js';
+
+export interface EnvelopeValidationResult {
+  valid: boolean;
+  errors: ErrorObject[];
+}
+
+export type ExtManifestProblem = {err: string; val: unknown};
+
+function topLevelProp(error: ErrorObject): string | undefined {
+  if (error.keyword === 'required') {
+    return error.params.missingProperty as string;
+  }
+  return error.instancePath.split('/').filter(Boolean)[0];
+}
+
+/** Whether `errors` includes a failure of the root value itself (e.g. a string/array/null instead of an object). */
+function hasRootTypeError(errors: ErrorObject[]): boolean {
+  return errors.some((error) => error.instancePath === '' && error.keyword === 'type');
+}
+
+/** Validates parsed `extensions.yaml` and individual extension manifest entries via a dedicated Ajv instance. */
+class ManifestValidator {
+  private readonly validateEnvelope: ValidateFunction;
+  private readonly validateCommonProblems: ValidateFunction;
+  private readonly validateDriverProblems: ValidateFunction;
+
+  constructor() {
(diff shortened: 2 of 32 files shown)


packages/appium/lib/schema/schema.ts
```diff
@@ -3,23 +3,20 @@ import path from 'node:path';
 import {AppiumConfigJsonSchema} from '@appium/schema';
 import {util} from '@appium/support';
 import type {ExtensionType} from '@appium/types';
-import AjvImport, {type ErrorObject, type SchemaObject, type ValidateFunction} from 'ajv';
+import {type ErrorObject, type SchemaObject, type ValidateFunction} from 'ajv';
```
eglitise: This line can be changed to type-only

mykola-mokhnach: Fixed in db8ba9c1f — switched to a type-only `import type {ErrorObject, SchemaObject, ValidateFunction} from 'ajv';`.

packages/appium/lib/extension/driver-config.ts
```diff
@@ -87,36 +87,8 @@ export class DriverConfig extends ExtensionConfig<DriverType> {
...
-    }
-
-    if (typeof automationName !== 'string') {
-      problems.push({
-        err: 'Missing or incorrect automationName',
-        val: automationName,
-      });
-    }
+    const problems = getDriverManifestProblems(extManifest);
+    const {automationName} = extManifest;
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch, fixed in db8ba9c1f. Read `automationName` via `(extManifest as unknown as Record<string, unknown> | null | undefined)?.automationName` instead of destructuring, so a null/non-object entry can't throw — `getDriverManifestProblems()` already reports it as a problem above without throwing. Also only add to `knownAutomationNames` when it's actually a string, so two unrelated entries that both lack `automationName` no longer get flagged as duplicates of each other. Added regression tests for both cases.

packages/appium/lib/extension/manifest.ts
```diff
@@ -126,6 +128,18 @@ export class Manifest {
         }
       }
 
+      if (!shouldWrite) {
+        const {valid, errors} = validateManifestEnvelope(data);
+        if (!valid) {
+          log.warn(
+            `Appium had trouble validating the extension installation cache file (${manifestPathResolved}); ` +
+              `it will be reset. Specific error(s): ${errors.map((err) => err.message).join('; ')}`,
+          );
```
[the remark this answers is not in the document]
mykola-mokhnach: Fixed in db8ba9c1f — added a describeValidationErrors() helper in manifest-validator.ts that falls back to `${error.keyword} error at "${error.instancePath || '/'}"` when `message` is unset, and the warning now uses that instead of mapping `errors` to `.message` directly.

packages/appium/test/unit/cli/cli.spec.ts
```diff
@@ -6,7 +6,7 @@ import {createSandbox} from 'sinon';
 
 import DriverCommand from '../../../lib/cli/driver-command.js';
 import {loadExtensions} from '../../../lib/extension/index.js';
-import {Manifest} from '../../../lib/extension/manifest.js';
+import {Manifest} from '../../../lib/extension/manifest/manifest.js';
```
eglitise, line 9: Should this and other test files import from `manifest/index.js` instead of `manifest/manifest.js`, to follow non-test files?

packages/appium/lib/schema/schema.ts
```diff
@@ -3,25 +3,13 @@ import path from 'node:path';
 import {AppiumConfigJsonSchema} from '@appium/schema';
 import {util} from '@appium/support';
 import type {ExtensionType} from '@appium/types';
-import AjvImport, {type ErrorObject, type SchemaObject, type ValidateFunction} from 'ajv';
-import addFormatsImport from 'ajv-formats';
 
 import {DRIVER_TYPE, PLUGIN_TYPE} from '../constants.js';
 import {bindAll, kebabCase, omitKeys, setPath} from '../utils/index.js';
+import {addFormats, Ajv, type AjvInstance, type ErrorObject, type SchemaObject, type ValidateFunction} from './ajv.js';
```
eglitise: This would look cleaner if type imports were split into their own line

packages/appium/test/unit/cli/cli.spec.ts
mykola-mokhnach: in tests it's ok to use direct imports

packages/appium/lib/schema/schema.ts
mykola-mokhnach: Fixed in c5f02d132 — split into a separate `import type {...}` line, and applied the same split to the other three imports added in this PR with the same mixed value/type pattern (manifest/validator.ts, keywords.ts, cli/args.ts).

packages/appium/lib/extension/manifest/validator.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+export type ExtManifestProblem = {err: string; val: unknown};
+
+function topLevelProp(error: ErrorObject): string | undefined {
+  if (error.keyword === 'required') {
+    return error.params.missingProperty as string;
+  }
+  return error.instancePath.split('/').filter(Boolean)[0];
+}
```
eglitise, line 24: Not sure I fully understand this - I assume the first branch would return the name of the field if it's missing, but how does the `instancePath` branch work? Is it still the name of the field, but because it wasn't `required`, it must be extracted from a different location in the `error` object?

packages/appium/lib/extension/manifest/validator.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    const errors = this.validateDriverProblems.errors ?? [];
+    // See the comment in `getCommonManifestProblems` about casting once and reading via `?.`.
+    const manifest = extManifest as Record<string, unknown> | null | undefined;
+
+    // A root type error means both fields are missing; no per-property error to attribute either to.
+    if (hasRootTypeError(errors)) {
+      return [
+        {err: 'Missing or incorrect supported platformNames list.', val: manifest?.platformNames},
```
eglitise: There's some duplication here for both of these error messages and the value.
A later line defines `platformNames`, so since `manifest?.automationName` is also requested twice, for consistency it's worth defining that too.

packages/appium/lib/extension/manifest/validator.ts
mykola-mokhnach: Right — same field name, just fetched from a different spot depending on the error keyword: for `required` errors, ajv's `instancePath` points at the object that's *missing* the field (not the field itself), so the name only shows up in `error.params.missingProperty`. For every other keyword (e.g. `type`, `minItems`), `instancePath` points directly at the offending field, so its first segment (after the leading `/`) is the field name — the array-index case (`/platformNames/0`) just means a second segment tags along, which is why only the first one is taken. Added a doc comment explaining this in 1461753d1.

packages/appium/lib/extension/manifest/validator.ts
mykola-mokhnach: Good catch, fixed in 1461753d1 — hoisted `platformNames`/`automationName` above the root-type-error branch (matching how `platformNames` was already read once further down) and extracted both error message strings into named constants so they're no longer repeated.

eglitise (approved):

PR #22733, opened 2026-09-06
fix(base-driver): clean up aborted session creation

### LINKED ISSUE #13492: Appium Server Doesn't Handle Aborted Requests
## The problem

It appears the Appium server doesn't handle aborted requests, particularly on long running requests such as [Create Session](http://appium.io/docs/en/commands/session/create/).

I would expect that it either:
- Aborts immediately (or where possible as I appreciate all the orchestration going on).
- Effectively calls [End Session](http://appium.io/docs/en/commands/session/delete/) internally at the end.

This would help to avoid sessions continuing long after the client has disconnected. A common case is when the user sends a signal interruption (e.g. Ctrl+C).

## Environment

* Appium version that exhibits the issue: `1.15.1`.
* Desktop OS/version used to run Appium: `macOS Mojave 10.14.6`.
* Node.js version: `10.13.0`.
* Yarn package manager: `1.19.1`.
* Mobile platform/version under test: `iOS 12.4 (via Xcode 10.3)`.
* Real device or emulator/simulator: `Simulator`.
* Appium CLI or Appium.app: `Appium CLI`.

## Steps To Reproduce

1. Launch the Appium server (`1.15.1`).
2. Make a `POST` request to `http://localhost:4723/wd/hub/session` using a HTTP client such as [Postman](https://www.getpostman.com/) with the following payload:
```
{
    "desiredCapabilities": {
        "platformName": "iOS",
        "platformVersion": "12.4",
        "deviceName": "iPhone X",
        "automationName": "XCUITest",
        "app": "<path to .app file>"
    }
}
```
3. Cancel the request while it's pending. Observe that it continues to boot the simulator and install WebDriverAgent before installing the app.

JustasMonkev:
## Proposed changes

  Fixes https://github.com/appium/appium/issues/13492.

  Appium now deletes a new session when the client disconnects before receiving the session response.

  Retries with the same idempotency key wait for cleanup to finish. They then share one fresh session response.

  ## Types of changes

  ## Checklist

  - [x] I have read the Contributing Guide (https://appium.io/docs/en/latest/contributing/)
  - [x] I have signed the CLA


packages/base-driver/lib/protocol/protocol.ts
```diff
@@ -592,13 +596,51 @@ function buildHandler(
       }
       res.status(httpStatus).json(ensureW3cResponse(httpResBody));
     }
+    return newSessionId;
+  };
+
+  const newSessionHandler = async (req: Request, res: Response) => {
```
mykola-mokhnach, line 599: this method is too big and should be split into smaller chunks for better readibility

packages/appium/docs/en/intro/clients.md
```diff
@@ -39,6 +39,14 @@ So, for example, the `Find Element` command corresponds to an HTTP `POST` reques
 endpoint `/session/:sessionid/element` (where in this case, `:sessionid` is a placeholder for the
 unique session ID generated by the server in a previous call to `Create Session`).
 
+If the client disconnects before Appium finishes sending a `Create Session` response, Appium
+deletes the new session once driver setup completes. Setup is not interrupted: the simulator may
```
mykola-mokhnach: this doc must not mention simulators and keep the the descriptions abstract

packages/appium/docs/en/intro/clients.md
```diff
@@ -39,6 +39,14 @@ So, for example, the `Find Element` command corresponds to an HTTP `POST` reques
...
+still boot or the app may still install before cleanup runs. Retrying with the same
```
mykola-mokhnach: Not sure about this one. let say I am creating a new session with idempotency key X, then disconnect. 
Then send the same session creation request with the same key X. Question: would be the session still deleted after the second request returns?

packages/base-driver/lib/express/idempotency.ts
mykola-mokhnach, line 113: this function is too complicated and should be simplifed

packages/appium/docs/en/intro/clients.md
JustasMonkev: No. The abandoned session is deleted before the retry starts a replacement with a new session ID

mykola-mokhnach: I assume we still need to pick up the old one instead of killing it. this is the whole purpose of idempotency id

mykola-mokhnach:
@JustasMonkev I would appreciate it if you could write short summaries to my replies/requests describing what has been changed/fixed instead of just resolving them. This way it would be easier to do the next review iteration for me.

packages/base-driver/lib/protocol/protocol.ts
JustasMonkev: Moved deletion and error logging into `deleteAbandonedSession` Response closure uses one non-rejecting promise, and newSessionHandler releases retries only after cleanup finishes

packages/appium/docs/en/intro/clients.md
JustasMonkev: Droped those

packages/base-driver/lib/express/idempotency.ts
JustasMonkev: Separated socket capture from cache completion and replaced recursive completion callbacks, the response size limit error is left

packages/appium/docs/en/intro/clients.md
JustasMonkev: Got it, same key should return the original session. I'll keep that session for retries and only clean up disconnected requests without a key. Will update it tommorow

JustasMonkev: yeah, updated it. with an idempotency key, we now keep the original session and cache its response even if the client disconnects

KazuCocoa (commented):
Found one cleanup edge case. TypeScript compilation and all 22 added tests pass, but a focused test for a disconnect during upstream middleware reproduces the issue below.

packages/base-driver/lib/protocol/protocol.ts
```diff
@@ -592,16 +593,48 @@ function buildHandler(
...
     }
+    return newSessionId;
   };
+
+  const newSessionHandler = async (req: Request, res: Response) => {
+    if (preserveIdempotentSessionResponse(res)) {
+      return await asyncHandler(req, res);
+    }
+    // Sending after a disconnect can still set writableFinished, so snapshot it at close.
+    const responseClosed = new Promise<boolean>((resolve) => res.once('close', () => resolve(res.writableFinished)));
```
KazuCocoa, line 604: **[P2] Handle responses that closed before the session handler starts**

If the client disconnects while asynchronous `frontRouter` middleware is running, `close` has already fired when this listener is registered. Once the middleware calls `next()`, session creation still proceeds, but `responseClosed` never resolves and `deleteAbandonedSession` is never called. I reproduced this by pausing an unkeyed `/session` request in `frontRouter`, destroying the client and awaiting the server response's `close` event, then releasing the middleware: the session remained active with zero deletion calls. Please account for an already-destroyed response before waiting for a future `close` event, and add a regression test for this ordering.

packages/base-driver/lib/express/idempotency.ts
```diff
@@ -33,61 +48,65 @@ const MAX_CACHED_PAYLOAD_SIZE_BYTES = 1 * 1024 * 1024; // 1 MiB
...
+      replayResponse(res, cachedResponse);
+    };
+    responseStateListener.once('ready', onReady);
+    res.once('close', onClose);
+  }
+}
+
+function replayResponse(res: Response, response: ReplayResponse): void {
+  if (Buffer.isBuffer(response)) {
+    res.socket?.write(response);
```
mykola-mokhnach, line 107: we must always check if the socket is writeable. These callbacks must never throw

JustasMonkev: Restored the comment explaining why writableFinished must be captured at close, and removed comments that just restated the code.

packages/base-driver/lib/protocol/protocol.ts
JustasMonkev: Removed isResponseFinished() and the redundant removeListener. trackResponseClose() now reads res.writableFinished directly.

KazuCocoa (approved):

mykola-mokhnach:
@JustasMonkev this is a big addition and since we are currently working on appium 4 it makes more sense to retarget it to the `appium4` branch. This will allow also to avoid possible conflicts

JustasMonkev:
> @JustasMonkev this is a big addition and since we are currently working on appium 4 it makes more sense to retarget it to the `appium4` branch. This will allow also to avoid possible conflicts

Changed the branch, seeing a lot of conflicts today or tomorrow will try to fix them

JustasMonkev:
fixed @mykola-mokhnach

mykola-mokhnach:
I think it still makes sense to delete the session for the scenario where a request keyed with idempotency id is terminated mid-flight and no second request with the same request arrives before session creation finishes (same idea as for non-keyed requests).

JustasMonkev:
@mykola-mokhnach could you check it please?

mykola-mokhnach (approved):

PR #22734, opened 2026-09-06
fix(storage-plugin): map StorageArgumentError to InvalidArgumentError

Mochxd:
`StorageArgumentError` is a plain `Error`, so `add` with a name that cannot be a file, or a sha1 of the wrong length, comes back as `500 unknown error`. The same happens on the upload web socket status event when the hashes do not match.

This makes `StorageArgumentError` extend `InvalidArgumentError`, so `getResponseForW3CError` maps those cases to `400 invalid argument`. Delete is unchanged: a name that is missing, unusable, or simply not in the storage still returns `false`.

## Tests

- [x] `add` with `foo/bar` returns `400 invalid argument`
- [x] `validateStorageItemName` maps through `getResponseForW3CError` to `400`
- [x] both cases fail on master with `500` / a plain `Error`
- [x] `npm run test:unit` in packages/storage-plugin, 11 pass
- [x] lint clean

### DIFF
--- a/packages/storage-plugin/lib/plugin.ts
+++ b/packages/storage-plugin/lib/plugin.ts
@@ -116,14 +116,8 @@ STORAGE_HANDLERS.listStorageItems = async function listStorageItems(): Promise<S
 };
 
 STORAGE_HANDLERS.deleteStorageItem = async function deleteStorageItem(req: Request): Promise<boolean> {
-  let name: string;
-  try {
-    name = parseRequestArgs(req, ['name']).name;
-    validateStorageItemName(name);
-  } catch (e) {
-    log.error(`Failed to parse the request body for deleting a storage item: ${(e as Error).message}`);
-    return false;
-  }
+  const name = parseRequestArgs(req, ['name']).name;
+  validateStorageItemName(name);
   return await executeStorageMethod(async (storage: Storage) => await storage.delete(name));
 };
 
--- a/packages/storage-plugin/lib/plugin.ts
+++ b/packages/storage-plugin/lib/plugin.ts
@@ -116,8 +116,14 @@ STORAGE_HANDLERS.listStorageItems = async function listStorageItems(): Promise<S
 };
 
 STORAGE_HANDLERS.deleteStorageItem = async function deleteStorageItem(req: Request): Promise<boolean> {
-  const name = parseRequestArgs(req, ['name']).name;
-  validateStorageItemName(name);
+  let name: string;
+  try {
+    name = parseRequestArgs(req, ['name']).name;
+    validateStorageItemName(name);
+  } catch (e) {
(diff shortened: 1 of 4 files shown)


Mochxd:
I hate those tests so much :""D, as usual i think the failure related to the timeout not the PR
Appium Build / Unit Tests (22) (pull_request)

mykola-mokhnach:
> I hate those tests so much :""D, as usual i think the failure related to the timeout not the PR Appium Build / Unit Tests (22) (pull_request)

Tests are good, they just need some love. Do not hesitate to push a fix that makes them more stable

mykola-mokhnach:
This would be a breaking change. The current behavior is expected, we want delete to be safe, there is no point in it throwing errors when we could just return false

Mochxd:
Fair enough on the item itself. If the name is valid but not there, or not usable as a file name, returning false is the right call and I'll put that back.

The part I'd push on is that the same catch also swallows `parseRequestArgs`. That one fires when the body is not a JSON object at all, or when `name` is missing entirely. We return false without ever looking anything up, so a client that sends `{"fileName": "x"}` gets a clean 200 forever and never finds out why its deletes do nothing. That is not a safe delete, we just could not read the request.

So I would keep the catch around `validateStorageItemName` and let only `parseRequestArgs` through. Nothing changes for a name that is simply absent or unusable.

Separate from delete, `add` answers 500 `unknown error` for the same invalid name today, because `StorageArgumentError` is a plain `Error`. Making it an `InvalidArgumentError` turns that into a 400, same as #22679. That half does not touch delete at all, so I am happy to keep only that if you would rather leave delete alone entirely.

Mochxd:
On the Unit Tests (22) failure, it is not this PR. storage-plugin ran 14/14 @mykola-mokhnach 

The red job is `packages/images-plugin/build/test/unit/plugin.spec.js`. Every assertion in that file passed (`# pass 77`, `# fail 0`), then the file itself was killed at 5001ms against `--test-timeout=5000`. Node treats the whole file as a test, so TAP reports a timeout with no failing assertion.

`compareImages` already has `{timeout: 6000}` because those OpenCV cases are slow, but that only lifts the suite. The parent file stays on 5000ms, so the override never helps on a loaded runner.

I would bump images-plugin's unit timeout to 10000, same as base-driver and fake-driver. Happy to do that as a separate PR rather than mix it in here.

mykola-mokhnach:
I am ok to inherit StorageArgumentError from errors.InvalidArgumentError. As for other things this PR tries to change I am still not convinced.

Mochxd:
Dropped the delete changes. `deleteStorageItem` is back to returning false, and the PR now only makes `StorageArgumentError` extend `InvalidArgumentError`.

mykola-mokhnach (approved):

PR #22735, opened 2026-09-06
refactor(support)!: replace glob package with native node:fs glob

mykola-mokhnach:
## Summary

Replaces the `glob` npm dependency in `@appium/support` with Node's native `fs.promises.glob`, which has been stable since the monorepo's minimum supported engine (`^22.22.2 || ^24.15.0 || >=26.0.0`). Addresses the "Investigate replacing `glob` with native `node:fs` version" item on the Appium 4 tracking issue (#22051).

- Removed the `glob` dependency (and its transitive `path-scurry`/`minimatch`/etc.) from `packages/support/package.json`.
- `fs.glob` now delegates to `fs.promises.glob` internally, but exposes its own narrow `GlobOptions` (`cwd`, `withFileTypes`, `absolute`, `lazy`) instead of leaking `node:fs`'s own option shape — `absolute` has no native equivalent, so it's resolved manually via `path.join`.
- Surveyed real-world `fs.glob()` usage across the driver ecosystem (appium-adb, appium-ios-simulator, appium-chromium-driver, appium-chromedriver, appium-xcuitest-driver, appium-espresso-driver, appium-mac2-driver) — every caller awaits it and treats the result as a plain array, none use `for await`. Kept that as the default behavior (`Promise<string[] | Dirent[]>`), and added an opt-in `lazy: true` option for callers that want to stream matches via an async generator instead (used internally by `calculateFolderIntegrity` in `base-driver` to avoid materializing a full directory listing just to count it).
- `storage-plugin` now types glob's `withFileTypes` results as native `fs.Dirent` instead of `path-scurry`'s `Path` (which was only ever a transitive dependency, never declared directly).

BREAKING CHANGE: `@appium/support`'s `fs.glob` option shape changed from the `glob` package's `GlobOptions` to a smaller, `@appium/support`-owned `GlobOptions` interface. Supported fields: `cwd`, `withFileTypes`, `absolute`, `lazy`. Options like `nodir`/`ignore`/`nocase` are no longer supported directly — callers needing directory filtering can pass `withFileTypes: true` and filter on `.isFile()`/`.isDirectory()` (as `storage-plugin` already does internally); callers needing path exclusion can filter the returned array by pattern.

### DIFF
--- a/packages/appium/lib/extension/manifest/manifest.ts
+++ b/packages/appium/lib/extension/manifest/manifest.ts
@@ -373,7 +373,7 @@ async function findNodeModulesPackageJsons(nodeModulesDir: string): Promise<stri
   }
 
   const pkgJsonPathsByEntry = await asyncmap(
-    entries,
+    entries.filter((entry) => !entry.startsWith('.')),
     async (entry): Promise<string[]> => {
       const entryPath = path.join(nodeModulesDir, entry);
       if (!entry.startsWith('@')) {
@@ -388,7 +388,7 @@ async function findNodeModulesPackageJsons(nodeModulesDir: string): Promise<stri
         return [];
       }
       const scopedPkgJsonPaths = await asyncmap(
-        scopedEntries,
+        scopedEntries.filter((scopedEntry) => !scopedEntry.startsWith('.')),
         async (scopedEntry) => {
           const pkgJsonPath = path.join(entryPath, scopedEntry, 'package.json');
           return (await fs.exists(pkgJsonPath)) ? pkgJsonPath : null;
--- a/packages/support/lib/fs.ts
+++ b/packages/support/lib/fs.ts
@@ -80,7 +80,10 @@ export interface GlobOptions {
 
 /** Overloaded call signature for {@linkcode fs.glob}, narrowing its return type based on `withFileTypes`/`lazy`. */
 export interface GlobFn {
-  (pattern: string | readonly string[], options: GlobOptions & {withFileTypes: true; lazy: true}): AsyncGenerator<Dirent>;
+  (
+    pattern: string | readonly string[],
+    options: GlobOptions & {withFileTypes: true; lazy: true},
+  ): AsyncGenerator<Dirent>;
   (pattern: string | readonly string[], options: GlobOptions & {withFileTypes: true}): Promise<Dirent[]>;
   (pattern: string | readonly string[], options: GlobOptions & {lazy: true}): AsyncGenerator<string>;
   (pattern: string | readonly string[], options?: GlobOptions): Promise<string[]>;
--- a/packages/support/lib/fs.ts
+++ b/packages/support/lib/fs.ts
@@ -273,7 +273,7 @@ export const fs = {
     const {cwd, withFileTypes, absolute, lazy} = options;
     async function* generate(): AsyncGenerator<string | Dirent> {
       for await (const entry of fsPromises.glob(pattern, {cwd, withFileTypes})) {
-        yield absolute && !withFileTypes ? path.join(cwd ?? process.cwd(), entry as string) : entry;
+        yield absolute && !withFileTypes ? path.resolve(cwd ?? process.cwd(), entry as string) : entry;
       }
     }
     if (lazy) {
--- a/packages/support/lib/fs.ts
+++ b/packages/support/lib/fs.ts
@@ -78,15 +78,26 @@ export interface GlobOptions {
   lazy?: boolean;
 }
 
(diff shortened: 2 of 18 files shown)


packages/support/lib/fs.ts
```diff
@@ -237,12 +262,31 @@ export const fs = {
...
+   * traverse through symlinks.
    */
-  glob(pattern: string, options?: GlobOptions): Promise<string[]> {
-    return Promise.resolve((options ? glob(pattern, options) : glob(pattern)) as Promise<string[]>);
-  },
+  glob: ((pattern: string | readonly string[], options: GlobOptions = {}) => {
+    const {cwd, withFileTypes, absolute, lazy} = options;
+    async function* generate(): AsyncGenerator<string | Dirent> {
+      for await (const entry of fsPromises.glob(pattern, {cwd, withFileTypes})) {
+        yield absolute && !withFileTypes ? path.join(cwd ?? process.cwd(), entry as string) : entry;
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch, fixed in f7d1d4855 by switching to `path.resolve`.

packages/support/lib/fs.ts
```diff
@@ -64,6 +63,32 @@ export interface MvOptions {
...
+
+/** Overloaded call signature for {@linkcode fs.glob}, narrowing its return type based on `withFileTypes`/`lazy`. */
+export interface GlobFn {
+  (
+    pattern: string | readonly string[],
+    options: GlobOptions & {withFileTypes: true; lazy: true},
+  ): AsyncGenerator<Dirent>;
+  (pattern: string | readonly string[], options: GlobOptions & {withFileTypes: true}): Promise<Dirent[]>;
+  (pattern: string | readonly string[], options: GlobOptions & {lazy: true}): AsyncGenerator<string>;
+  (pattern: string | readonly string[], options?: GlobOptions): Promise<string[]>;
```
KazuCocoa: [P2] Return unions when the option flags are not literal types

The fallback overload promises `Promise<string[]>` even when the supplied options can enable `lazy` or `withFileTypes`. For example, `const opts: GlobOptions = {lazy: true}; fs.glob('*', opts).then(...)` passes TypeScript 6.0.3 but throws `TypeError: ...then is not a function`, because the implementation returns an async generator. Similarly, a `GlobOptions` variable containing `withFileTypes: true` is inferred to yield strings although it yields `Dirent`s. The preceding overload also claims a promise for `{withFileTypes: true, lazy: someBoolean}`. Please constrain the precise overloads with the corresponding `false`/omitted flags and add a union-returning fallback for boolean-valued options, so valid typed callers cannot silently select the wrong consumption API.

packages/appium/lib/extension/manifest/manifest.ts
```diff
@@ -356,6 +356,52 @@ export class Manifest {
...
+    if ((err as NodeJS.ErrnoException)?.code === 'ENOENT') {
+      return [];
+    }
+    throw err;
+  }
+
+  const pkgJsonPathsByEntry = await asyncmap(
+    entries,
+    async (entry): Promise<string[]> => {
+      const entryPath = path.join(nodeModulesDir, entry);
```
KazuCocoa, line 378: [P2] Keep hidden package directories out of extension discovery

The old `node_modules/{*,@*/*}/package.json` glob skipped dot-prefixed entries, but this walk now includes them at both levels. npm uses directories such as `node_modules/.appium-foo-driver-<hash>` (and `@scope/.foo-<hash>`) to retire packages during installation; interrupted installs can leave these behind with valid extension metadata. I reproduced the new helper returning these paths while glob 13.0.6 excludes them. `syncWithInstalledExtensions()` will then register a retired extension, or race it against the live copy under the same driver/plugin name and persist the stale install path. Please skip dot-prefixed entries in both `entries` and `scopedEntries` while preserving the new symlink traversal.

KazuCocoa (commented):
Otherwise lg entirely to me

packages/support/lib/fs.ts
mykola-mokhnach: Fixed in 0c1c7f33b: widened the fallback overload to a union return type so a non-literal `GlobOptions` (or `{withFileTypes: true, lazy: someBoolean}`) can no longer silently pick the wrong consumption API.

packages/appium/lib/extension/manifest/manifest.ts
mykola-mokhnach: Good catch, fixed in 0c1c7f33b by filtering out dot-prefixed entries at both the top-level and scoped-package level, and added a regression test.

KazuCocoa (approved):

PR #22742, opened 2026-09-07
fix(support)!: route non-error CLI console output to stdout

mykola-mokhnach:
## Summary

Part of the Appium 4 logging cleanup tracked in #22051: "Change CLI logging to output non-error logs to stdout (currently everything goes to stderr)".

`CliConsole` (`@appium/support`'s `console`, used by `appium driver`/`plugin` install/list/update/etc., `appium doctor`, and other CLI-only output) wrote **every** message level — `log`, `ok`, `info`, `warn`, and `error` — to `STDERR`. Only `json()` output went to `STDOUT`.

This PR splits the two channels the way most CLI tools do:
- `log()`, `ok()`, `debug()`, `info()`, `warn()`, and `json()` → **STDOUT**
- `error()` → **STDERR**

`jsonMode` behavior is preserved: non-JSON output is still fully squelched, and `json()` is guaranteed to land on real `STDOUT` even in that mode (now via a dedicated internal stream instead of relying on the old accidental routing).

Color-support auto-detection (`useColor`) now defaults off of `stdout`'s TTY-ness instead of `stderr`'s, since stdout is now the primary output channel.

Note: the server's Winston-based log sink (`packages/appium/lib/logsink.ts`) already split `error` to stderr and everything else to stdout via `stderrLevels: ['error']`, so this PR only needed to touch the CLI console path.

BREAKING CHANGE: Any script or tool that consumes Appium CLI output (`driver`/`plugin` list/install/update, `doctor`, etc.) and relies on the previous stream layout will need to be updated. Previously, non-JSON human-readable output (info/success/warning messages) was written to `STDERR`, with `STDOUT` reserved almost exclusively for `--json` output. Now that output goes to `STDOUT`, and only actual errors go to `STDERR`.
BREAKING CHANGE: Scripts that discarded `STDOUT` (`1>/dev/null`) to suppress "chatter" while relying on `STDERR` for status text will now see that text on `STDOUT` instead.
BREAKING CHANGE: Scripts that captured `STDERR` to detect/log informational or success messages will no longer see them there — only genuine errors will appear on `STDERR`. `--json` output and error behavior are unaffected: `--json` continues to write clean JSON to `STDOUT` only, and errors continue to go to `STDERR`.
BREAKING CHANGE: **`@appium/logger`'s `Log` class changes its default output stream** from `STDERR` to `STDOUT` for all levels except `error`, and adds an `errorStream`/`stderrLevel` pair of properties. Any code that set `.stream` on a `Log` instance expecting it to capture *all* levels (including `error`) will now need to also set `.errorStream` to the same target.

### DIFF
--- a/packages/appium/lib/bootstrap/appium-initializer.ts
+++ b/packages/appium/lib/bootstrap/appium-initializer.ts
@@ -1,4 +1,4 @@
-import {util} from '@appium/support';
+import {console as supportConsole, util} from '@appium/support';
 import type {DriverOpts} from '@appium/types';
 import type {
   Args,
@@ -77,6 +77,8 @@ export class AppiumInitializer {
     }
 
     if (isSetupCommandArgs(preConfigArgs)) {
+      driverConfig.printValidationSummary(logger);
+      pluginConfig.printValidationSummary(logger);
       await runSetupCommand(preConfigArgs, driverConfig, pluginConfig);
       return {} as InitResult<Cmd>;
     }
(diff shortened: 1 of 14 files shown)


KazuCocoa (approved):
lgtm

PR #22702, opened 2026-09-02
docs: fix docs build

KazuCocoa:
to fix https://github.com/appium/appium/actions/runs/33586751950/job/100112397133

PR #22749, opened 2026-09-09
feat: backport diff from master

mykola-mokhnach:
## Summary

Compared `appium4` against upstream `appium/appium` master (31 commits ahead) and cherry-picked the fixes that still apply and aren't already covered by appium4's refactors. Dependency bumps, lockfile regenerations, and docs-translation commits were left out as routine, branch-independent noise.

## Backported

- fix(base-driver): reject invalid timeout with `InvalidArgumentError` (#22679)
- fix(appium): do not treat a directory as a CLI argument file (#22709)
- fix(strongbox): reject item names that slugify to an empty path (#22717)
- fix(appium): match insecure feature prefixes by automationName, case-insensitively (#22710)
- fix(storage-plugin): map `StorageArgumentError` to `InvalidArgumentError` (#22734)
- test(images-plugin): give the flaky implicit-wait unit test room to retry (#22743)
- fix: add missing `.js` extension in a cherry-picked test import, required by appium4's NodeNext module resolution

## Not backported (already covered or inapplicable)

- fix(base-driver): skip unknown events in `session.unsubscribe` (#22671) — appium4's mixin-removal refactor already contains this fix
- test(base-driver): fix invalid import (#22729) — the bug it fixes (self-referencing package import) doesn't exist in appium4's version of the file
- 18 `chore(deps)`/lockfile-regeneration commits and 3 docs-translation commits — each branch manages these independently

### DIFF
--- a/packages/appium/docs/en/blog/.authors.yml
+++ b/packages/appium/docs/en/blog/.authors.yml
@@ -17,3 +17,8 @@ authors:
     avatar: 'https://avatars.githubusercontent.com/u/4045041?v=4'
     slug: 'sai'
     url: 'https://saikrishna.tech'
+  mykola:
+    name: 'Mykola Mokhnach'
+    description: 'Appium Core Maintainer'
+    avatar: https://avatars.githubusercontent.com/u/7767781?v=4
+    slug: 'mykola'
--- a/packages/appium/docs/en/blog/posts/announcing-watchos-simulator-support.md
+++ b/packages/appium/docs/en/blog/posts/announcing-watchos-simulator-support.md
@@ -1,0 +1,38 @@
+---
+authors:
+  - mykola
+date: 2026-09-01
+---
+
+# Announcing watchOS Simulator Support in the XCUITest Driver
+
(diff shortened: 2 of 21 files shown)


eglitise:
* Interesting that the self-referencing import bug doesn't manifest for Appium 4
* For the PR related to `session.unsubscribe` that was not backported PR, maybe it's worth backporting the added unit tests?
* I think it's also worth backporting the docs updates (including translations), otherwise the English language files and #22702 would need to be manually backported at some point.

mykola-mokhnach:
Re the self-referencing import point: confirmed — appium4's `test/types/basedriver/driver.test-d.ts` already used a plain relative import (`../../../lib/basedriver/driver.js`) rather than importing `@appium/base-driver` by package name, so the self-reference the original PR fixed never existed here. That's why it wasn't backported.

eglitise:
Indeed, but the original fix also began with applying only the relative import, which resulted in many more errors (https://github.com/appium/appium/pull/22729#issuecomment-5551236498).

eglitise (approved):

PR #22750, opened 2026-09-09
refactor(base-driver): split protocol.ts into focused modules

mykola-mokhnach:
## Summary

`base-driver`'s `protocol/protocol.ts` had grown to ~590 lines mixing several unrelated concerns: payload validation, session/protocol resolution, WD-proxy handling, and Express route wiring, plus a request handler (`buildHandler`) that did all of that inline in one long closure. This PR splits it into focused modules and gives the request handler a readable, step-by-step structure, with no public API or behavior changes.

- **Split `protocol.ts` into focused modules** (renamed remainder to `route-handler.ts`, since `protocol/protocol.ts` duplicated the directory name):
  - `params.ts` — payload validation/shaping: `checkParams`, `makeArgs`, `validateExecuteMethodParams`, `wrapParams`, `unwrapParams`.
  - `session.ts` — session/protocol/logger resolution: `getSessionId`, `isSessionCommand`, `extractProtocol`, `getLogger`.
  - `proxy.ts` — WD-proxy decision and execution: `driverShouldDoWdProxy`, `doWdProxy`, and a new `tryWdProxy(...)` that folds the proxy decision, plugin-override check, and proxy call into one `'handled' | 'plugin-override' | 'not-proxied'` outcome.
  - `route-handler.ts` — route wiring (`routeConfiguringFunction`) and the request handler.

- **Decomposed `buildHandler`'s ~210-line request-handling closure** into named steps: `logDeprecationWarning`, `ensureSessionExists`, `preparePayload`, `buildCommandArgs`, `runDriverCommand`, `normalizeError`/`buildErrorResponse`, `sendHandlerResponse`.

- **Removed duplication in `protocol/index.ts`**: switched from import-then-re-export to `export {...} from './x.js'` throughout, matching the convention already used in `lib/index.ts`.

- **Colocated command-name constants with their routes**: `CREATE_SESSION_COMMAND`, `DELETE_SESSION_COMMAND`, `GET_STATUS_COMMAND` now live in `routes/w3c.ts`; `LIST_DRIVER_COMMANDS_COMMAND`, `LIST_DRIVER_EXTENSIONS_COMMAND` in `routes/appium.ts` — next to the route definitions that reference them, replacing the previously duplicated string literals there. `routes/index.ts` re-exports them from their owning files.

- **Gave the `validators` map a proper type**: added a `CommandValidator` type, checked `validators` against it with `satisfies` (keeping `setUrl`/`setNetworkConnection`'s precise per-key signatures for direct/static access), and moved the one necessary cast for dynamic lookup-by-command-name into a new `getCommandValidator(command)` helper next to the map it indexes — removing the inline cast that used to live at the `buildCommandArgs` call site.

- Flagged the deprecated `setNetworkConnection` validator with `@deprecated`, matching its route's `deprecated: true`.

No public API changes — `protocol/index.ts`'s re-exports are unchanged.


packages/base-driver/lib/protocol/validators.ts
```diff
@@ -2,19 +2,26 @@ import {util} from '@appium/support';
...
   setUrl: (url: any) => {
-    if (!URL.canParse(url)) {
+    if (typeof url !== 'string' || !URL.canParse(url)) {
       throw new Error(`'${util.truncateString(String(url), {length: MAX_URL_ERROR_LENGTH})}' must be a valid URL`);
     }
   },
+  /** @deprecated Only used by the deprecated `setNetworkConnection` MJSONWP route. */
   setNetworkConnection: (type: any) => {
-    if (!isNumber(type) || [0, 1, 2, 4, 6].indexOf(type) === -1) {
+    if (![0, 1, 2, 4, 6].includes(Number(type))) {
```
KazuCocoa: **[P2] Preserve strict network-type validation**

`Number(type)` now accepts inputs such as `null`, booleans, `""`, `[]`, and `[2]`, all of which the previous validator rejected. The conversion only affects the membership check: the original value still reaches `setNetworkConnection`, whose contract expects a number. For example, `{"parameters":{"type":null}}` now invokes the driver instead of rejecting the request. Please preserve the numeric membership check without coercion and add regression coverage for nonnumeric inputs. I verified these cases against the base and PR validators.

mykola-mokhnach: added Number.isInteger check

PR #22753, opened 2026-09-10
feat(appium): add --app-url-rules server arg to restrict remote app URLs

### LINKED ISSUE #22744: 💡 feat: add server config option for URL rules used in 'appium:app' and similar capabilities
### Is your feature request related to a problem?

The `appium:app` capability used in various drivers can specify either a path to a local application package, or a URL to a remote package. For the URL option, the server only checks whether it uses a valid URL syntax, but otherwise allows providing any URL.

Users deploying Appium servers for local or public usage may want to configure which URLs are allowed, using approaches including, but not limited to:
* Whitelist/blacklist
* Allow HTTPS only
* Reject URLs containing username/password
* Limit redirects

### Describe the solution you'd like.

Add a server config option (applicable either via command-line or config file) that allows server hosts to specify rules that URLs must adhere to. If a user attempts to create a new session using a capability set with a URL that does not satisfy these rules, the server should reject the request.

Important to note: this option would not be able to cover all URLs in all possible capability values, as drivers can define their own custom capabilities. However, it would still be beneficial for common capabilities like `appium:app`.

### Describe alternatives you've considered.

Server hosts can set up third-party tools to limit outgoing URLs, which may not be as convenient.

### Additional context

_No response_

gioboa:
Closes #22744

## Proposed changes

New server arg `--app-url-rules` (or `server.app-url-rules` in the config file). It's a JSON object (or a path to a JSON file) with the rules a remote app URL (e.g. `appium:app`) has to pass before the server downloads it:

- `allow` / `deny`: hostname glob patterns (`*.example.com`), IP addresses, or CIDR subnets (`10.0.0.0/8`, `2001:db8::/32`). Hostnames are normalized before matching; address rules are applied to literal addresses and to every dynamically resolved address on each download. `deny` wins over `allow`
- `httpsOnly`: reject non-https URLs
- `allowCredentials`: `false` rejects URLs with user/password
- `maxRedirects`: cap on followed redirects (default `21`), `0` rejects any. Every redirect target is validated against the same rules

```bash
appium --app-url-rules='{"allow": ["artifacts.example.com", "10.0.0.0/8"], "httpsOnly": true, "allowCredentials": false, "maxRedirects": 0}'
```

## Types of changes

## Checklist

- [x] I have signed the CLA


eglitise:
Thanks for the PR. However, since the team is currently focusing on Appium 4, we have decided to halt new feature PRs for the default Appium 3 branch. You are still welcome to update this PR to target the `appium4` branch instead, and we'll review it accordingly.
Same applies to your other PR.

gioboa:
> You are still welcome to update this PR to target the `appium4` branch instead, and we'll review it accordingly. Same applies to your other PR.

Thanks, there we go :+1:

packages/appium/docs/en/guides/security.md
```diff
@@ -85,3 +85,49 @@ Turn on multiple insecure features for specific drivers:
...
+Several drivers accept a URL instead of a local path in capabilities such as `appium:app`, in which
+case the Appium server downloads the application package from that URL before starting the session.
+By default, any `http:` or `https:` URL is accepted. Server hosts who want to control which URLs
+their server is allowed to download from can do so with the `--app-url-rules` argument (or the
+`server.app-url-rules` property in a [configuration file](./config.md)). Its value is a JSON object
+(or a path to a JSON file) with any of the following properties:
+
+|<div style="width:10em">Rule</div>|Description|Default|
+|----------------------------------|-----------|-------|
+|`allow`|List of regular expressions matched against the full URL. If the list is not empty, a URL must match _at least one_ of them to be accepted|`[]`|
```
mykola-mokhnach, line 100: regexps are error prone and might be unreliable. Consider matching against normalized hostnames instead. Maybe use picomatch to allow simple wildcards

packages/appium/docs/en/guides/security.md
```diff
@@ -85,3 +85,49 @@ Turn on multiple insecure features for specific drivers:
...
+|`deny`|List of regular expressions matched against the full URL. A URL matching _any_ of them is rejected, even if it also matches an `allow` rule|`[]`|
```
mykola-mokhnach, line 101: there should be a way to allow/deny specific ipv6/ipv4 addresses or whole subnetworks

mykola-mokhnach: for that we must also be able to resolve given domains dynamically

packages/appium/lib/appium.ts
```diff
@@ -123,6 +124,14 @@ export class AppiumDriver extends DriverCore<AppiumDriverConstraints> {
   configureGlobalFeatures = insecureFeatures.configureGlobalFeatures;
   configureDriverFeatures = insecureFeatures.configureDriverFeatures;
 
+  /**
+   * Applies the rules from `--app-url-rules` (if any) to all remote application URLs
+   * downloaded by drivers in this server process.
+   */
+  configureAppUrlRules(): void {
```
mykola-mokhnach, line 131: this functionality should be encapsulated into a separate mixin, similarly to other commands.
It also makes sense to make it internal, so only appium/umbrella driver could see it

packages/base-driver/lib/basedriver/app-url-rules.ts
```diff
@@ -0,0 +1,144 @@
+import {util} from '@appium/support';
```
mykola-mokhnach, line 1: this should be in a mixin/command

packages/base-driver/lib/basedriver/app-url-rules.ts
```diff
@@ -0,0 +1,144 @@
...
+import type {AppUrlRulesConfig} from '@appium/types';
+
+import {log as logger} from './logger.js';
```
mykola-mokhnach, line 4: it should use the driver logger

packages/base-driver/lib/basedriver/app-url-rules.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Compiles the given rules so they can be applied to remote application URLs.
+ *
+ * @param rules - Raw rules, as provided via the `--app-url-rules` server argument.
+ * @returns Compiled rules.
+ * @throws {TypeError} If any rule has an invalid value (e.g. a malformed regular expression).
+ */
+export function compileAppUrlRules(rules: AppUrlRulesConfig): CompiledAppUrlRules {
+  if (!util.isPlainObject(rules)) {
```
mykola-mokhnach, line 27: why do we check it here, shouldn't it be already validated by the schema?

packages/base-driver/lib/basedriver/app-url-rules.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+  const {allow, deny, httpsOnly, allowCredentials, maxRedirects} = rules;
+  const compiled: CompiledAppUrlRules = {
+    allow: toRegExps('allow', allow),
+    deny: toRegExps('deny', deny),
+    httpsOnly: toBoolean('httpsOnly', httpsOnly, false),
+    allowCredentials: toBoolean('allowCredentials', allowCredentials, true),
+  };
+  if (maxRedirects !== undefined && maxRedirects !== null) {
+    if (typeof maxRedirects !== 'number' || !Number.isInteger(maxRedirects) || maxRedirects < 0) {
```
mykola-mokhnach, line 38: same as above - schema validation should cover that

packages/base-driver/lib/basedriver/helpers.ts
```diff
@@ -129,7 +130,9 @@ export async function configureApp(
   };
   const {protocol, pathname} = parseAppLink(app);
   const isUrl = isSupportedUrl(app);
-  if (!isUrl && !path.isAbsolute(newApp)) {
+  if (isUrl) {
+    assertAppUrlAllowed(new URL(app));
```
mykola-mokhnach, line 134: is it safe to perform `new URL` on `app` here? are we sure it won't throw?

gioboa (commented):
Addressed all the comments with [127e2f8](https://github.com/appium/appium/pull/22753/commits/127e2f8d6919a4936d3892adbf3126817be17a55) cc @mykola-mokhnach

packages/appium/lib/app-url-rules.ts
```diff
@@ -0,0 +1,9 @@
+// @ts-expect-error Internal runtime export is intentionally absent from base-driver declarations.
```
mykola-mokhnach: what is the purpose of this module?

packages/appium/lib/app-url-rules.ts
```diff
@@ -0,0 +1,9 @@
...
+import {configureAppUrlRules as configureBaseDriverAppUrlRules} from '@appium/base-driver';
```
mykola-mokhnach: I prefer to just use internal interface/declare instead of having `@ts-expect-error`. Check how it's done for `IpcAssignable`

packages/appium/lib/appium.ts
```diff
@@ -129,6 +130,9 @@ export class AppiumDriver extends DriverCore<AppiumDriverConstraints> {
   configureGlobalFeatures = insecureFeatures.configureGlobalFeatures;
   configureDriverFeatures = insecureFeatures.configureDriverFeatures;
 
+  /** @internal */
+  configureAppUrlRules = appUrlRules.configureAppUrlRules;
```
mykola-mokhnach: I am not convinced this should be a method on the driver, since app downloader is located in the shared helper. This means base driver package may expose some internal interface, which allows to configure the rules globally. Then its app download helpers may consume the same singleton to apply the necessary check. This would also guarantee all validation logic will be incapsulated in that singleton and won't leak

packages/base-driver/lib/basedriver/commands/app-url-rules.ts
```diff
@@ -0,0 +1,201 @@
+import type {LookupAddress, LookupAllOptions, LookupOptions} from 'node:dns';
```
mykola-mokhnach: this could probably still be improved. See my comment above about singletone rules manager/validator

packages/base-driver/lib/basedriver/helpers/app-cache.ts
```diff
@@ -311,41 +311,46 @@ function isSupportedUrl(app: string): boolean {
...
+  const requestOpts: AxiosRequestConfig = {
     url: axiosUrl,
     auth: axiosAuth,
-    responseType: 'stream' as const,
+    responseType: 'stream',
     timeout: APP_DOWNLOAD_TIMEOUT_MS,
     validateStatus: (status: number) => (status >= 200 && status < 300) || status === HTTP_STATUS_NOT_MODIFIED,
     headers: reqHeaders,
   };
+  const urlRules = getAppUrlRules();
```
mykola-mokhnach: this looks like a logic leak to me. The corresponding object should be able to just accept the requestOpts object and then return the validated one

packages/base-driver/lib/basedriver/commands/app-url-rules.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+  return compiled;
+}
+
+/** Configures process-wide app URL rules using the invoking driver's logger. */
+export function configureAppUrlRules<C extends Constraints>(
+  this: DriverCore<C>,
+  rules?: AppUrlRulesConfig | null,
+): void {
+  currentRules = !rules || Object.keys(rules).length === 0 ? undefined : compileAppUrlRules(rules);
```
KazuCocoa: [P1] Share the rules with separately installed drivers

`currentRules` is local to this module, so configuring Appium's copy of `@appium/base-driver` does not configure another installed copy. This matters for the normal extension layout: Appium loads drivers from `APPIUM_HOME`, and those drivers can resolve their own `@appium/base-driver` dependency independently of the server's installation. Their standard `configureApp` helper then sees `undefined` rules and permits downloads that the server configuration should reject. The new tests use the same module copy for configuration and downloading, so they do not exercise this boundary.

Could we pass the policy into each driver's download implementation or share it across module copies? For the intended process-wide behavior, a store on `globalThis` keyed by `Symbol.for(...)` would be one option, provided the driver also uses a base-driver version that reads that store. A regression test should load two separate copies, configure one with `{httpsOnly: true}`, and verify that the other rejects an HTTP URL. I reproduced the current gap with two copies of this PR's rule module.

packages/base-driver/lib/basedriver/helpers/app-cache.ts
```diff
@@ -311,41 +311,46 @@ function isSupportedUrl(app: string): boolean {
...
+  if (urlRules) {
+    requestOpts.lookup = createAppUrlLookup(urlRules);
```
KazuCocoa: [P1] HTTP proxies bypass destination IP restrictions

Axios honors proxy environment variables, so the custom `lookup` checks the proxy hostname—or never runs when the proxy uses an IP address. The proxy resolves the destination without applying these rules. I reproduced a successful request to `localhost` despite denying both loopback ranges, with zero lookup calls. Proxy routing needs to preserve destination enforcement or reject requests it cannot validate.

packages/appium/docs/en/guides/security.md
```diff
@@ -85,3 +85,52 @@ Turn on multiple insecure features for specific drivers:
...
+`server.app-url-rules` property in a [configuration file](./config.md)). Its value is a JSON object
+(or a path to a JSON file) with any of the following properties:
+
+|<div style="width:10em">Rule</div>|Description|Default|
+|----------------------------------|-----------|-------|
+|`allow`|List of hostname glob patterns, IP addresses, or CIDR subnets. If non-empty, a hostname or its resolved address must match _at least one_|`[]`|
+|`deny`|List of hostname glob patterns, IP addresses, or CIDR subnets. A matching hostname or resolved address is rejected, even if an `allow` rule also matches|`[]`|
+|`httpsOnly`|Only accept `https:` URLs|`false`|
+|`allowCredentials`|Accept URLs containing a username and/or password (e.g. `https://user:pass@host/app.apk`)|`true`|
+|`maxRedirects`|Maximum number of HTTP redirects to follow while downloading. Set to `0` to reject any redirect|_(unlimited)_|
```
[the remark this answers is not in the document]
gioboa: Fixed

gioboa (commented):
Comments addressed with 98f699c. Thanks 🙏

KazuCocoa (approved):

packages/base-driver/lib/index.ts
```diff
@@ -33,6 +33,9 @@ export {DEFAULT_WS_PATHNAME_PREFIX} from './express/websocket.js';
 // BiDi exports
 export {BIDI_COMMANDS} from './protocol/bidi-commands.js';
 
+/** @internal Used by the Appium server to apply the `--app-url-rules` argument process-wide */
+export {appUrlRules} from './basedriver/helpers/app-url-rules.js';
```
mykola-mokhnach: lets just add the corresponding internal method to base driver helpers, so it could be called as `this.helpers.setAppUrlRules(value)`, but only visible from this umbrella driver

mykola-mokhnach: this will also allow to avoid adding new exports to base driver

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,382 @@
...
+    return addresses;
+  }
+  const allowed = addresses.filter(({address, family}) => matchesAddress(rules.allow, address, family));
+  if (!allowed.length) {
+    reject(`no resolved IP address matches an allow rule`);
+  }
+  return allowed;
+}
+
+function hasRules(rules: CompiledRuleList): boolean {
```
mykola-mokhnach: consider adding corresponding methods/getter to the rules instance itself rather than having them as static accessors. Make sure logic is encapsulated properly for each entity

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,382 @@
...
+   */
+  applyToRequest(requestOpts: AxiosRequestConfig): AxiosRequestConfig {
+    const rules = this._getCompiledRules();
+    if (!rules) {
+      return requestOpts;
+    }
+    const url = new URL(String(requestOpts.url), requestOpts.baseURL);
+    assertRequestAllowed(url, requestOpts.proxy, rules);
+    const result: AxiosRequestConfig = {...requestOpts};
+    if (hasAddressRules(rules)) {
```
mykola-mokhnach: not sure why these are static functions - we only expect to have a single rules copy at a time, so it makes sense to keep it as a property of this class and access `rules` as a properly

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,382 @@
...
+    return this._compiled.rules;
+  }
+}
+
+/**
+ * The process-wide app URL rules singleton
+ *
+ * @internal
+ */
+export const appUrlRules = new AppUrlRules();
```
mykola-mokhnach, line 349: I expected only the raw rules setter to be exported/visible from the umbrella driver. This instance should remain internal and used by app cache for validation

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,382 @@
...
+function filterResolvedAddresses(
+  hostname: string,
+  addresses: LookupAddress[],
+  rules: CompiledAppUrlRules,
+): LookupAddress[] {
+  const reject = createReject(new URL(`http://${formatHostname(hostname)}`));
+  if (addresses.some(({address, family}) => matchesAddress(rules.deny, address, family))) {
+    reject(`a resolved IP address matches a deny rule`);
+  }
+  if (!hasRules(rules.allow) || matchesHostname(rules.allow, hostname)) {
```
mykola-mokhnach: this could have a better abstraction, where we extract the validation logic itself to a separate class instead of having static functions spreaded across the module

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,382 @@
...
+    });
+  };
+}
+
+function filterResolvedAddresses(
+  hostname: string,
+  addresses: LookupAddress[],
+  rules: CompiledAppUrlRules,
+): LookupAddress[] {
+  const reject = createReject(new URL(`http://${formatHostname(hostname)}`));
```
mykola-mokhnach: why do we assume http prefix?

packages/base-driver/test/e2e/basedriver/helpers.e2e.spec.ts
```diff
@@ -146,6 +163,145 @@ describe('app download and configuration', function () {
...
+            await assert.rejects(
+              configureApp(`${serverUrl}/FakeAndroidApp.apk`, '.apk'),
+              /is not allowed by the server configuration: the IP address does not match any allow rule/,
+            );
+          });
+          it('should reject a URL matching a deny rule', async function () {
+            appUrlRules.configure({deny: [new URL(serverUrl).hostname]});
+            await assert.rejects(
+              configureApp(`${serverUrl}/FakeAndroidApp.apk`, '.apk'),
+              /is not allowed by the server configuration: the IP address matches a deny rule/,
```
mykola-mokhnach: I think we should not expose such details, like `matches a deny rule` to the client. It's just enough to say `is not allowed by the server configuration` without stating the details of it

packages/base-driver/package.json
```diff
@@ -67,6 +67,8 @@
     "method-override": "3.0.0",
     "morgan": "1.11.0",
     "path-to-regexp": "8.4.2",
+    "picomatch": "4.0.7",
+    "proxy-from-env": "2.1.0",
```
mykola-mokhnach: can we avoid adding more dependencies?

gioboa: Fixed with [1e95c34](https://github.com/appium/appium/pull/22753/commits/1e95c341c512138e252d1e81208cb42950e00678)

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+    getEnv('all_proxy');
+  return proxy && !proxy.includes('://') ? `${proto}://${proxy}` : proxy;
+}
+
+/**
+ * @param hostname - The hostname of the URL (IPv6 addresses are wrapped in brackets)
+ * @param port - The effective port of the URL
+ */
+function shouldProxy(hostname: string, port: number): boolean {
+  const noProxy = (getEnv('npm_config_no_proxy') || getEnv('no_proxy')).toLowerCase();
```
KazuCocoa: Could we align this proxy detection with the version used by Axios? The locked `proxy-from-env@2.1.0` does not read `npm_config_no_proxy`, but this implementation gives it precedence over `NO_PROXY`. With `HTTP_PROXY` set and `npm_config_no_proxy=*`, this validator treats the request as direct and allows it even with IP rules configured, while Axios actually sends it through the proxy. The destination is then resolved by the proxy, so the configured IP rules cannot be enforced. I reproduced the mismatch with a local HTTP proxy and Axios 1.19.0. Please match Axios's proxy-selection behavior and add a regression test for this combination.

gioboa: Thanks, fixed with [1a6e469](https://github.com/appium/appium/pull/22753/commits/1a6e469f1d948cdeb92c319883cabc53096eac89)

packages/appium/docs/en/guides/security.md
```diff
@@ -85,3 +85,61 @@ Turn on multiple insecure features for specific drivers:
...
+appium --app-url-rules='{"allow": ["artifacts.example.com", "10.0.0.0/8"], "httpsOnly": true, "allowCredentials": false, "maxRedirects": 0}'
+```
+
+The same configuration as part of a config file:
+
+```yaml
+server:
+  app-url-rules:
+    allow:
+      - 'artifacts.example.com'
```
mykola-mokhnach, line 159: Please add more examples to reflect picomatch-based patterns, ipv6 subnet patterns and other features supported by this argument

packages/appium/lib/insecure-features.ts
```diff
@@ -40,6 +40,21 @@ export function configureGlobalFeatures(this: AppiumDriver) {
...
 }
 
+/**
+ * Applies the `--app-url-rules` server argument to the process-wide app URL rules, which are
+ * enforced by the shared app download helper of `@appium/base-driver` for every driver.
+ * Uses `logger` instead of `this.log` to reduce user confusion.
+ *
+ * @throws {TypeError} If any rule is invalid
+ */
+export function configureAppUrlRules(this: AppiumDriver) {
```
mykola-mokhnach: this helper method is redundant as well as the new public driver method
consider calling `appiumDriver.helpers.setAppUrlRules(rules)` directly from packages/appium/lib/bootstrap/appium-main-runner.ts

packages/appium/test/unit/appiumdriver.spec.ts
```diff
@@ -137,6 +137,39 @@ describe('AppiumDriver', function () {
 
       return [appium, mockFakeDriver];
     }
+    describe('configureAppUrlRules', function () {
```
mykola-mokhnach: these tests are not needed, the functionality is already tested in the base driver

packages/appium/test/unit/parser.spec.ts
```diff
@@ -99,6 +99,20 @@ describe('parser', function () {
...
       });
 
+      it('should parse app URL rules correctly from a string', function () {
+        const appUrlRules = {allow: ['*.example.com', '10.0.0.0/8'], httpsOnly: true, maxRedirects: 0};
+        const args = p.parseArgs(['--app-url-rules', JSON.stringify(appUrlRules)]);
+        assert.deepStrictEqual(args.appUrlRules, appUrlRules);
+      });
+
+      it('should throw an error with invalid arg to app URL rules', function () {
+        assert.throws(() => p.parseArgs(['--app-url-rules', '42']));
```
mykola-mokhnach: should we also check the actual error text or at least part of it?

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+import {log as logger} from '../../helpers/logger.js';
+
+/**
+ * Key of the process-wide store holding the active rules.
+ *
+ * A `Symbol.for` key is shared by every copy of this module loaded into the process, so rules
+ * configured by the Appium server also apply to drivers which resolve their own copy of
+ * `@appium/base-driver` (e.g. from `APPIUM_HOME`), as long as that copy supports app URL rules.
+ */
+const GLOBAL_STORE_KEY = Symbol.for('@appium/base-driver:app-url-rules');
```
mykola-mokhnach, line 24: can this bite us back if we bump the major package version?

packages/base-driver/lib/basedriver/helpers/index.ts
```diff
@@ -1,4 +1,5 @@
 export {configureApp} from './app-cache.js';
+export {setAppUrlRules} from './app-url-rules.js';
```
mykola-mokhnach, line 2: should this export be set to internal?

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+ * @throws {TypeError} If any rule is invalid
+ */
+export function setAppUrlRules(rules?: AppUrlRulesConfig | null): void {
+  appUrlRules.configure(rules);
+}
+
+/**
+ * A compiled `allow` or `deny` list of hostname patterns, IP addresses and subnets
+ */
+class HostRuleList {
```
mykola-mokhnach, line 59: 👍

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+      reject('only https: URLs are accepted');
+    }
+    if (!this._allowCredentials && (url.username || url.password)) {
+      reject('URLs containing credentials are not accepted');
+    }
+
+    const hostname = normalizeHostname(url.hostname);
+    const addressFamily = net.isIP(hostname);
+    if (this._deny.matchesHostname(hostname)) {
+      reject('the hostname matches a deny rule');
```
mykola-mokhnach: lets not expose details about deny/allow rules due to security considerations.
Also, we need to make sure the actual server command line or the config is not logged, so end users don't see these allow/deny lists - this information is specific to the server and/or internal network architecture and must not leak.

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+   * @throws {Error} If the URL violates any rule or the rules cannot be enforced for it
+   */
+  assertRequestAllowed(url: URL, proxy: AxiosRequestConfig['proxy']): void {
+    this.assertUrlAllowed(url);
+    if (!this.hasAddressRules) {
+      return;
+    }
+    const proxyUrl = proxy ? String(proxy.host) : proxy === false ? '' : getEnvProxyUrl(url);
+    if (proxyUrl) {
+      this._reject(
```
mykola-mokhnach: if appium is running as a service on a host provider VM then customers anyway cannot change env variables there since it would be not secure. And if they run appium locally such restriction would anyway not make much sense. Please design error messages and logging behavior keeping the above fact in mind.

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+    this._compiled = {source, validator: new AppUrlRulesValidator(source)};
+    this._store.rules = source;
+  }
+
+  /**
+   * Validates a URL against the static rules (scheme, credentials, hostname and literal addresses).
+   *
+   * @throws {Error} If the URL violates any rule
+   */
+  assertUrlAllowed(url: URL): void {
```
mykola-mokhnach: this method could still return the original url if the validation passes for chaining purposes

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+   * unchanged if no rules are configured.
+   *
+   * @throws {Error} If the request URL violates any rule or the rules cannot be enforced for it
+   */
+  applyToRequest(requestOpts: AxiosRequestConfig): AxiosRequestConfig {
+    const validator = this._validator;
+    if (!validator) {
+      return requestOpts;
+    }
+    validator.assertRequestAllowed(new URL(String(requestOpts.url), requestOpts.baseURL), requestOpts.proxy);
```
mykola-mokhnach: this looks like an unexpected side effect. Can the validation be separated from the request patching?

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+/**
+ * Returns the proxy URL the HTTP client would route a request to the given URL through,
+ * according to the `<scheme>_proxy`, `all_proxy` and `no_proxy` environment variables,
+ * or an empty string if none applies.
+ *
+ * This mirrors the logic of `proxy-from-env@2` used by axios. In particular, the `npm_config_*`
+ * variables are NOT consulted anymore (unlike `proxy-from-env@1`), so they must not be consulted
+ * here either: otherwise the validator would disagree with axios about whether a request is proxied.
+ */
+function getEnvProxyUrl(url: URL): string {
```
mykola-mokhnach: proxy helper can also be extracted to a separate class

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,477 @@
...
+
+function getEnv(key: string): string {
+  return process.env[key.toLowerCase()] || process.env[key.toUpperCase()] || '';
+}
+
+/**
+ * Converts a hostname pattern to a regular expression, where `*` matches any sequence of
+ * characters (including dots) and `?` matches any single character.
+ */
+function toHostnameRegExp(pattern: string): RegExp {
```
mykola-mokhnach: why do we need to use regexp instead of picomatch?

sorry, I was probably not explicit enough - we can keep picomatch as a new dependency, but not more than that. Adding regexp support adds a new class of problems, where one could, for example, add broken/invalid regexp or craft some malicious one, which means more work for us to parse and validate that

gioboa: Fixed this and the other comments with [75f3589](https://github.com/appium/appium/pull/22753/commits/75f3589115b30e863967db6d4f5d615db173953f)

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,473 @@
...
+  }
+
+  /**
+   * Replaces the process-wide rules. Empty or absent rules lift all restrictions.
+   *
+   * @throws {TypeError} If any rule is invalid
+   */
+  configure(rules?: AppUrlRulesConfig | null): void {
+    if (!rules || Object.keys(rules).length === 0) {
+      delete this._store.rules;
```
mykola-mokhnach, line 317: is it safe to call if `this._store.rules` is already undefined?

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
```diff
@@ -0,0 +1,473 @@
...
+    if (parts.length > 2) {
+      this._throwInvalid('IP address or hostname', pattern);
+    }
+    if (family) {
+      try {
+        if (prefix === undefined) {
+          this._addresses.addAddress(address, toFamilyName(family));
+        } else {
+          if (!/^\d+$/.test(prefix)) {
+            throw new Error('prefix must be an integer');
```
mykola-mokhnach: should we add more information about the value that caused this error, so it's easier to debug it?

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
mykola-mokhnach: this string is still passed to logs, which should not happen, because cloud providers usually hand over these logs to end users without redaction or with minimal redaction.

mykola-mokhnach: same below

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
gioboa: Thanks for your review. This new commit [0d46cb1](https://github.com/appium/appium/pull/22753/commits/0d46cb1def7ecdf3b5e63f0c46bd39832c3348b9) should solve this and the other comments.

packages/appium/lib/bootstrap/main-helpers.ts
```diff
@@ -206,9 +206,18 @@ export async function createAppiumServer(
...
+ */
+const SENSITIVE_ARGS: ReadonlySet<keyof Args> = new Set(['appUrlRules']);
+
 function logNonDefaultArgsWarning(args: Args): void {
+  const shownArgs = Object.fromEntries(
+    Object.entries(args).map(([name, value]) => [name, SENSITIVE_ARGS.has(name as keyof Args) ? '<redacted>' : value]),
+  );
   logger.info('Non-default server args:');
-  inspect(args);
+  inspect(shownArgs);
```
mykola-mokhnach, line 220: 👍

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
mykola-mokhnach: how about the comment https://github.com/appium/appium/pull/22753#discussion_r4020049043 ?

packages/base-driver/lib/basedriver/helpers/app-url-rules.ts
mykola-mokhnach: I've checked it separately - should be safe, - delete does not care about property existence

mykola-mokhnach (approved):

PR #22755, opened 2026-09-10
fix(base-driver): honor unquoted and RFC 5987 Content-Disposition filenames

Mochxd:
`configureApp` only matched `filename="..."` with quotes. A lot of CDNs send `attachment; filename=app.apk` or `filename*=UTF-8''app.apk`, so those fell through to the URL basename and you could get the wrong name or a defaulted `.apk` / `.ipa`.

This prefers RFC 5987 `filename*`, then a quoted `filename`, then an unquoted token. Empty `filename=""` is treated as missing so we keep the URL fallback instead of capturing the quote characters.

A name taken from the header still skips the URL-path extension defaulting. That was already true for quoted filenames; it now also applies when the name was unquoted. So `filename=app.bin` with a driver that wants `.apk` fails `verifyAppExtension` instead of being renamed.

### DIFF
--- a/packages/base-driver/lib/basedriver/helpers.ts
+++ b/packages/base-driver/lib/basedriver/helpers.ts
@@ -41,7 +41,10 @@ const APP_DOWNLOAD_TIMEOUT_MS = 120 * 1000;
 // RFC 5987: filename*=UTF-8''percent-encoded-name  (language tag is optional)
 const FILENAME_STAR_PATTERN = /(?:^|;)\s*filename\*\s*=\s*([^';\s]+)'[^']*'([^;]+)/i;
 const QUOTED_FILENAME_PATTERN = /(?:^|;)\s*filename\s*=\s*"((?:\\.|[^"\\])*)"/i;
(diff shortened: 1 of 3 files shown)


packages/base-driver/lib/basedriver/helpers.ts
```diff
@@ -485,6 +485,39 @@ async function fetchApp(srcStream: Readable, dstPath: string): Promise<string> {
...
 }
 
+/**
+ * Picks a download filename from a Content-Disposition header.
+ * Prefers RFC 5987 `filename*` over `filename`, and accepts both quoted and unquoted `filename`.
+ * @internal
+ */
+export function filenameFromContentDisposition(header: string): string | undefined {
+  // RFC 5987: filename*=UTF-8''percent-encoded-name  (language tag is optional)
+  const encoded = /(?:^|;)\s*filename\*\s*=\s*([^';\s]+)'[^']*'([^;]+)/i.exec(header);
```
mykola-mokhnach: Please move regexps to constants

mykola-mokhnach (approved):

PR #22757, opened 2026-09-10
feat(base-driver)!: remove deprecated JSONWP/MJSONWP routes

mykola-mokhnach:
## Summary

Removes the deprecated JSONWP/MJSONWP route definitions (`packages/base-driver/lib/protocol/routes/jsonwp.ts` and `mjsonwp.ts`) from `@appium/base-driver`'s global method map, along with the now-dead `receive_async_response` CORS carve-out in the Express middleware.

BREAKING CHANGE: Dropped APIs: `receive_async_response`, `element/:elementId` GET stub, IME endpoints, `orientation`, `location` (geolocation), `context`/`contexts`, `network_connection`, and `rotation`.
BREAKING CHANGE: Dropped the `allowCrossDomainAsyncExecute` middleware + its dedicated test suite, which existed solely to CORS-enable the now-removed `receive_async_response` endpoint.


KazuCocoa (approved):

KazuCocoa:
btw, maybe the context endpoint removal will impact many clients as it was probably not marked as deprecated so long. Probably drivers would need to keep them, or need to check our clients and webdriverio side to use `/appium/context` and drop lower version of appium versions

mykola-mokhnach:
> btw, maybe the context endpoint removal will impact many clients as it was probably not marked as deprecated so long. Probably drivers would need to keep them, or need to check our clients and webdriverio side to use `/appium/context` and drop lower version of appium versions

Don't worry, I've added it to corresponding driver's local maps where this API is used, so nothing will be broken

PR #22764, opened 2026-09-12
refactor(base-driver)!: split basedriver helpers into a barrel

mykola-mokhnach:
`basedriver/helpers.ts` had grown into a 580-line grab bag (app download/caching, capability parsing, log prefixing, version lookup). Split it into focused modules under `basedriver/helpers/`, re-exported through `basedriver/helpers/index.ts` so existing consumers of the `DriverHelpers`-shaped module are unaffected.

`generateDriverLogPrefix` and the base `logger` are cross-cutting (used by `protocol/` as well as `basedriver/`), so they moved to the shared `lib/helpers/` directory instead, removing `protocol/`'s backward dependency on `basedriver/` internals and the resulting risk of circular imports.


packages/base-driver/lib/helpers/log-prefix.ts
```diff
@@ -0,0 +1,17 @@
+import {node} from '@appium/support';
+
+/**
+ * Builds a short log prefix for a driver instance (e.g. `UiAutomator2@a1b2`).
+ *
+ * @param obj - Driver or other object; its constructor name and a short id are used.
+ * @param _sessionId - Deprecated and unused; kept for {@link DriverHelpers} interface compatibility.
```
eglitise: Should we keep this method if it's deprecated?

packages/base-driver/lib/basedriver/helpers/index.ts
```diff
@@ -0,0 +1,4 @@
+export {configureApp} from './app-cache.js';
+export {duplicateKeys, isPackageOrBundle, parseCapsArray} from './caps.js';
+export {generateDriverLogPrefix} from '../../helpers/log-prefix.js';
```
eglitise, line 3: Not sure about a barrel file exporting something from a different directory. If the method is still needed, perhaps it's better to adjust the caller imports instead?

packages/base-driver/lib/helpers/log-prefix.ts
mykola-mokhnach: it's not the method is deprecated, but this particular arg. Let me remove it

packages/base-driver/lib/basedriver/helpers/index.ts
mykola-mokhnach: it's fine to do it since top level helpers contain shared package exports

eglitise (approved):

PR #22767, opened 2026-09-12
fix(appium): lock the extension manifest while driver/plugin commands run

mykola-mokhnach:
Serializes `appium driver`/`plugin` CLI commands against a per-`APPIUM_HOME` lock file and re-reads `extensions.yaml` after acquiring it, so concurrent processes can no longer race to read-modify-write the manifest and silently clobber each other's changes. `ExtensionConfig.installedExtensions` becomes a live getter so it stays in sync with the manifest across that re-read.

Also fixes a TOCTOU race in `@appium/support`'s `getLockFileGuard`: it checked `checkSync()` then separately called `lockSync()`, so a lock created by another process in that gap surfaced as an uncaught `EEXIST` failure instead of being waited on.

### DIFF
--- a/packages/appium/lib/cli/extension.ts
+++ b/packages/appium/lib/cli/extension.ts
@@ -1,12 +1,12 @@
 /* eslint-disable no-console */
-import {util} from '@appium/support';
 import type {Class, DriverType, ExtensionType, PluginType} from '@appium/types';
 import type {Args, CliExtensionCommand, CliExtensionSubcommand} from 'appium/types/index.js';
 
 import {DRIVER_TYPE, PLUGIN_TYPE} from '../constants.js';
 import type {ExtensionConfig} from '../extension/extension-config.js';
+import {reloadManifest} from '../extension/index.js';
 import {isExtensionCommandArgs} from '../schema/cli-args-guards.js';
-import {resolveManifestLockfilePath} from '../utils/index.js';
+import {withManifestLock} from '../utils/index.js';
 import DriverCliCommand from './driver-command.js';
 import PluginCliCommand from './plugin-command.js';
 import {errAndQuit, JSON_SPACES} from './utils.js';
@@ -46,13 +46,11 @@ export async function runExtensionCommand<Cmd extends CliExtensionCommand, SubCm
   // Serialize this against any other `driver`/`plugin` CLI command running (in this or another
   // process) against the same `APPIUM_HOME`, so concurrent commands can't race to read, mutate,
   // and write the same extension manifest out from under each other.
-  const lockFile = await resolveManifestLockfilePath(config.appiumHome);
-  const acquireLock = util.getLockFileGuard<Record<string, unknown>>(lockFile);
-
-  return acquireLock(async () => {
+  return withManifestLock<Record<string, unknown>>(config.appiumHome, async () => {
     // Refresh from disk while holding the lock, in case another process wrote to the manifest
-    // between this process's startup read and now.
-    await config.manifest.read();
+    // between this process's startup read and now, and revalidate so derived state (installed
+    // extensions, pending validation summary, duplicate-automationName tracking) is current too.
+    await reloadManifest(config.manifest);
 
     let jsonResult: Record<string, unknown> = {};
     const CommandClass = commandClasses[type] as ExtCommand<Cmd>;
--- a/packages/appium/lib/cli/extension.ts
+++ b/packages/appium/lib/cli/extension.ts
@@ -1,10 +1,12 @@
 /* eslint-disable no-console */
+import {util} from '@appium/support';
 import type {Class, DriverType, ExtensionType, PluginType} from '@appium/types';
 import type {Args, CliExtensionCommand, CliExtensionSubcommand} from 'appium/types/index.js';
 
 import {DRIVER_TYPE, PLUGIN_TYPE} from '../constants.js';
 import type {ExtensionConfig} from '../extension/extension-config.js';
 import {isExtensionCommandArgs} from '../schema/cli-args-guards.js';
+import {resolveManifestLockfilePath} from '../utils/index.js';
 import DriverCliCommand from './driver-command.js';
 import PluginCliCommand from './plugin-command.js';
 import {errAndQuit, JSON_SPACES} from './utils.js';
@@ -30,9 +32,6 @@ export async function runExtensionCommand<Cmd extends CliExtensionCommand, SubCm
   args: Args<Cmd, SubCmd>,
   config: ExtensionConfig<Cmd>,
 ) {
-  // TODO driver config file should be locked while any of these commands are
-  // running to prevent weird situations
-  let jsonResult: Record<string, unknown> = {};
   const {extensionType: type} = config; // NOTE this is the same as `args.subcommand`
   if (!isExtensionCommandArgs(args)) {
     throw new TypeError(`Cannot call ${type} command without a subcommand like 'install'`);
@@ -43,23 +42,37 @@ export async function runExtensionCommand<Cmd extends CliExtensionCommand, SubCm
   if (suppressOutput) {
     json = true;
   }
-  const CommandClass = commandClasses[type] as ExtCommand<Cmd>;
-  const cmd = new CommandClass({config, json} as any);
-  cmd.printPendingValidationSummary();
-  try {
-    jsonResult = (await cmd.execute(args)) as Record<string, unknown>;
-  } catch (err) {
-    // in the suppress output case, we are calling this function internally and should
-    // just throw instead of printing an error and ending the process
-    if (suppressOutput) {
-      throw err;
+
+  // Serialize this against any other `driver`/`plugin` CLI command running (in this or another
+  // process) against the same `APPIUM_HOME`, so concurrent commands can't race to read, mutate,
+  // and write the same extension manifest out from under each other.
+  const lockFile = await resolveManifestLockfilePath(config.appiumHome);
+  const acquireLock = util.getLockFileGuard<Record<string, unknown>>(lockFile);
+
+  return acquireLock(async () => {
+    // Refresh from disk while holding the lock, in case another process wrote to the manifest
+    // between this process's startup read and now.
+    await config.manifest.read();
+
+    let jsonResult: Record<string, unknown> = {};
+    const CommandClass = commandClasses[type] as ExtCommand<Cmd>;
+    const cmd = new CommandClass({config, json} as any);
+    cmd.printPendingValidationSummary();
+    try {
+      jsonResult = (await cmd.execute(args)) as Record<string, unknown>;
+    } catch (err) {
+      // in the suppress output case, we are calling this function internally and should
+      // just throw instead of printing an error and ending the process
+      if (suppressOutput) {
+        throw err;
+      }
+      errAndQuit(json, err);
     }
-    errAndQuit(json, err);
-  }
 
-  if (json && !suppressOutput) {
-    console.log(JSON.stringify(jsonResult, null, JSON_SPACES));
-  }
+    if (json && !suppressOutput) {
+      console.log(JSON.stringify(jsonResult, null, JSON_SPACES));
+    }
 
-  return jsonResult;
+    return jsonResult;
+  });
 }
--- a/packages/appium/lib/extension/index.ts
+++ b/packages/appium/lib/extension/index.ts
@@ -5,7 +5,7 @@ import {asyncmap} from 'asyncbox';
 
 import {USE_ALL_PLUGINS} from '../constants.js';
 import {log} from '../logger.js';
-import {zip} from '../utils/index.js';
+import {withManifestLock, zip} from '../utils/index.js';
 import {DriverConfig} from './driver-config.js';
 import {Manifest} from './manifest/index.js';
 import {PluginConfig} from './plugin-config.js';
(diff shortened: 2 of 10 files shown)


packages/appium/lib/cli/extension.ts
```diff
@@ -43,23 +42,37 @@ export async function runExtensionCommand<Cmd extends CliExtensionCommand, SubCm
...
-  } catch (err) {
-    // in the suppress output case, we are calling this function internally and should
-    // just throw instead of printing an error and ending the process
-    if (suppressOutput) {
-      throw err;
+
+  // Serialize this against any other `driver`/`plugin` CLI command running (in this or another
+  // process) against the same `APPIUM_HOME`, so concurrent commands can't race to read, mutate,
+  // and write the same extension manifest out from under each other.
+  const lockFile = await resolveManifestLockfilePath(config.appiumHome);
```
KazuCocoa: [P1] Include startup manifest writes in the critical section

The initial `loadExtensions(appiumHome)` in `AppiumInitializer.init()` runs before this lock is acquired, and its `Manifest.read()` can also write the manifest (initialization, migration, or synchronization with installed packages). A second process can therefore perform an unlocked read-modify-write while the first process holds this lock, potentially overwriting its changes. Re-reading here cannot recover changes already overwritten by that startup path. Could we protect the initial load and all write-capable manifest paths with the same lock, and add a concurrent-process test that exercises startup synchronization?

packages/appium/lib/cli/extension.ts
```diff
@@ -43,23 +42,37 @@ export async function runExtensionCommand<Cmd extends CliExtensionCommand, SubCm
...
+  const acquireLock = util.getLockFileGuard<Record<string, unknown>>(lockFile);
+
+  return acquireLock(async () => {
+    // Refresh from disk while holding the lock, in case another process wrote to the manifest
+    // between this process's startup read and now.
+    await config.manifest.read();
```
KazuCocoa: [P2] Revalidate extension state after reloading the manifest

`loadExtensions()` already validates the entries, removing invalid ones from the in-memory manifest, but this `read()` replaces that data with the on-disk entries without repeating validation. An entry missing required fields can consequently reappear as installed in `list --installed`, even though the pending validation summary says it was rejected. The driver's `knownAutomationNames` also still reflects the pre-lock snapshot, so changes made while this process waited are not reflected in subsequent duplicate-name checks. Could we revalidate the refreshed configs and rebuild their derived state before executing the command? A regression test should cover invalid entries and driver changes made between the initial load and lock acquisition.

packages/appium/lib/cli/extension.ts
mykola-mokhnach: Fixed in e790e546c: `loadExtensions()` now runs its manifest read (and any migration/sync writes it may trigger) under the same cross-process lock as `runExtensionCommand`, via a new shared `withManifestLock` helper in `packages/appium/lib/utils/env.ts`. Added a regression test (`loadExtensions > waits for another process holding the manifest lock instead of reading/writing around it`) that fails without this fix and passes with it.

packages/appium/lib/cli/extension.ts
mykola-mokhnach: Fixed in e790e546c: added a shared `reloadManifest()` helper (in `packages/appium/lib/extension/index.ts`) that both `runExtensionCommand` and `loadExtensions` call after re-reading the manifest -- it revalidates whichever `DriverConfig`/`PluginConfig` singletons already exist for it, so invalid entries are dropped again and `knownAutomationNames` (plus the pending validation summary) is rebuilt from the freshly-read data rather than the pre-lock snapshot. Added a regression test (`runExtensionCommand > revalidates the reloaded manifest, dropping conflicting entries and refreshing derived duplicate-name tracking`) covering a duplicate-automationName entry written to disk between the initial load and the lock-protected reload.

packages/appium/lib/extension/index.ts
```diff
@@ -29,11 +45,11 @@ export type DriverNameMap = Map<DriverClass, string>;
  */
 export async function loadExtensions(appiumHome: string): Promise<ExtensionConfigs> {
   const manifest = Manifest.getInstance(appiumHome);
-  await manifest.read();
   const driverConfig = DriverConfig.getInstance(manifest) ?? DriverConfig.create(manifest);
   const pluginConfig = PluginConfig.getInstance(manifest) ?? PluginConfig.create(manifest);
 
-  await Promise.all([driverConfig.validate(), pluginConfig.validate()]);
+  await withManifestLock(appiumHome, () => reloadManifest(manifest));
```
KazuCocoa, line 51: [P2] Preserve startup with a read-only APPIUM_HOME

`loadExtensions()` is used for server startup as well as extension commands. Acquiring this lock unconditionally now requires creating `extensions.yaml.lock` alongside the manifest, even when an existing manifest is valid and needs no migration or synchronization. A preinstalled, read-only `APPIUM_HOME` therefore fails with EACCES/EROFS before either `driver list --installed` or server startup with `--no-perms-check` can proceed; previously those paths could just read the manifest. Could we preserve read-only operation, for example by revisiting the lock location or providing an appropriate read-only loading path while keeping write-capable paths serialized? A regression test with an existing manifest in a non-writable directory would cover this.

KazuCocoa (commented):
last one comment:

packages/appium/lib/extension/index.ts
mykola-mokhnach: Good catch. Fixed in 82311f31f: added a `runUnlockedIfUnwritable` option to `@appium/support`'s `getLockFileGuard` -- when lock acquisition fails specifically because the location is unwritable (EACCES/EROFS/EPERM), it runs the behavior directly instead of throwing, since nothing else could be racing a write against a location nothing can write to. `withManifestLock` opts into this and also tolerates the same failure from its own `mkdirp` of the lock's parent directory (a no-op when that directory already exists, so it doesn't by itself require write access).

Added two regression tests with a chmod'd read-only manifest directory (verified both fail with the exact EACCES you described, without this fix): `loadExtensions > succeeds against a read-only manifest directory when no write is needed` and `runExtensionCommand > still lists installed extensions when the manifest directory is read-only` (the latter covering the `driver list --installed` case you mentioned).

packages/support/lib/util.ts
```diff
@@ -571,10 +578,21 @@ export function getLockFileGuard<T>(lockFile: string, opts: LockFileOptions = {}
...
-          } else {
+          try {
             _lockfile.lockSync(lockFile);
+          } catch (e) {
+            const lockErr = e as NodeJS.ErrnoException;
+            if (lockErr.code !== 'EEXIST') {
+              if (runUnlockedIfUnwritable && isUnwritableLocationError(lockErr)) {
+                // We haven't acquired anything yet, so there's nothing to release -- just run the
+                // behavior directly.
+                return await behavior();
```
KazuCocoa: [P2] Do not infer that the protected data is read-only from a lock creation failure

A non-writable directory prevents creating a new lock file, but does not prevent overwriting an existing writable manifest. I reproduced this locally with the same directory mode used in the new tests: with the directory set to `0555` and the existing manifest still writable, exclusive lock-file creation fails with EACCES while overwriting the manifest succeeds. Since `withManifestLock()` enables this fallback for the entire driver/plugin command, synchronization or installation can consequently write the manifest without any serialization, reintroducing the lost-update race. Could we restrict fallback to a loading path that cannot write, or use a shared writable lock location? A regression test should cover a non-writable directory containing a writable manifest, rather than only checking that listing succeeds.

KazuCocoa (commented):
found one in the last commit

packages/support/lib/util.ts
mykola-mokhnach: Good catch, and you're right that the earlier fix was unsound. Redesigned in 507a5aed2: the lock now lives under the OS temp dir, keyed by a hash of `appiumHome` (see `resolveManifestLockfilePath`), instead of alongside the manifest. Its own writability no longer depends on `APPIUM_HOME`'s permissions at all, so there's no more inference from "can't create a lock file here" to "nothing can write here" -- removed the `runUnlockedIfUnwritable` option from `@appium/support` entirely along with it.

Added the regression test you asked for: `runExtensionCommand > still serializes writes when the manifest directory is non-writable but the manifest file itself is writable` -- chmods the manifest directory to 0555 (same as the other read-only tests) while a separate lock holder is active, and asserts a concurrent `runExtensionCommand` call actually blocks on it rather than resolving immediately (verified this fails without the fix, per the wait/assert-false-then-release pattern already used for the plain "waits for another lock holder" test).

Also: while running the full e2e suite locally to validate this, I found and fixed an unrelated real regression from my earlier commit (e790e546c) that CI's e2e run also caught -- `AppiumSchema.finalize()` was mutating registered extension schemas in place, which broke re-registration equality for any extension with an inline (non-file-path) schema once `validate()` started running a second time per command. That's fixed in the same commit (schema.ts now clones before mutating), with a regression test in schema.spec.ts and the pre-existing schema tests updated -- they'd been silently relying on that mutation.

packages/appium/lib/utils/env.ts
```diff
@@ -124,3 +125,34 @@ export const resolveManifestPath = util.memoize(async function _resolveManifestP
...
+ * regardless of `appiumHome`'s own permissions, including a preinstalled, read-only one.
+ *
+ * See caveat on {@link resolveManifestPath} about pre-resolving `appiumHome`.
+ */
+export const resolveManifestLockfilePath = util.memoize(async function _resolveManifestLockfilePath(
+  appiumHome?: string,
+): Promise<string> {
+  const resolvedHome = path.resolve(appiumHome ?? (await resolveAppiumHome()));
+  const homeHash = createHash('sha256').update(resolvedHome).digest('hex').slice(0, 32);
+  return path.join(tmpdir(), `appium-manifest-${homeHash}.lock`);
```
KazuCocoa: [P2] Derive one shared lock identity for the same manifest

`path.resolve()` does not resolve symlinks, so using the real APPIUM_HOME path in one process and a symlink to it in another produces different hashes for the same manifest. Independently, `tmpdir()` varies with TMPDIR (and user/platform defaults), so even identical APPIUM_HOME strings can produce different lock locations. I checked the path calculation locally: the symlink and real path produce distinct exclusive lock files that can both be held simultaneously; differing TMPDIR values also produce different lock paths. Both cases bypass serialization.

A concrete approach would be to canonicalize the protected location with `fs.realpath()` before hashing (at least APPIUM_HOME, which startup already ensures exists), then store the lock in a stable directory shared by all participating processes instead of their environment-selected temp directories. For same-user coordination, a fixed Appium cache directory under the user's home is one option; if a shared APPIUM_HOME is supported across users, the lock root must instead be explicitly shared with suitable permissions. Keep lock acquisition mandatory for write-capable operations. Could we add tests where one process holds the lock and another uses (1) a symlink alias and (2) a different TMPDIR, verifying that neither enters the critical section until release?

mykola-mokhnach: Both good catches, thanks. Fixed in 79f5082d5:

- Canonicalize `appiumHome` with `fs.realpath()` before hashing (falling back to `path.resolve()` only if `realpath` fails, e.g. the dir doesn't exist yet), so a symlink alias and its real target now share one lock identity.
- Moved the lock off `tmpdir()` entirely, to a fixed per-user location (`~/.appium/locks`) that doesn't depend on `TMPDIR`/platform/user defaults -- while still being independent of `appiumHome`'s own permissions, so the read-only-`APPIUM_HOME` fix from the previous round still holds.

I scoped this to same-user coordination per your note, and left a comment on `resolveManifestLockfilePath` saying so explicitly -- a single `APPIUM_HOME` deliberately shared across different users' processes would need its lock pointed at a location those users jointly have write access to, which felt like a separate, explicit opt-in decision rather than something to default into.

Added both regression tests you asked for, adapted to actually hold the lock and assert a concurrent call blocks (rather than just comparing paths): `resolveManifestLockfilePath > shares one lock between a symlink alias and its real target` and `... shares one lock for the same APPIUM_HOME regardless of TMPDIR`. Verified both fail against the prior (pre-canonicalization / tmpdir-based) logic and pass now.

packages/appium/lib/utils/env.ts
```diff
@@ -124,3 +125,52 @@ export const resolveManifestPath = util.memoize(async function _resolveManifestP
 ): Promise<string> {
   return path.join(appiumHome ?? (await resolveAppiumHome()), MANIFEST_RELATIVE_PATH);
 });
+
+/**
+ * Relative path (from {@link DEFAULT_APPIUM_HOME}'s parent, i.e. the user's home dir) to the
+ * directory holding manifest lockfiles.
+ */
+const MANIFEST_LOCKS_DIRNAME = path.join('.appium', 'locks');
```
KazuCocoa: [P2] Keep the lock root outside the default APPIUM_HOME

The default APPIUM_HOME is also `~/.appium`, so placing locks in `~/.appium/locks` reintroduces the read-only startup failure when the default home is preinstalled and made read-only as a whole. The manifest can remain readable while `withManifestLock()` fails to create `locks` (or a lock file inside a read-only existing `locks` directory). I reproduced the directory-permission case locally. The current tests only chmod the manifest's immediate parent, leaving the new lock root writable, so they do not exercise this case.

A minimal fix for this layout would be to change the fixed relative lock directory to `path.join('.cache', 'appium', 'locks')`, retaining `homedir()` as the root, realpath-based hashing, and mandatory lock acquisition. This preserves same-user coordination without depending on TMPDIR and moves locks outside the default APPIUM_HOME. Please add a regression test with a simulated user home containing a valid manifest in `.appium`, make that entire Appium directory read-only while leaving the separate cache writable, and verify both loading/listing and lock contention. A fully read-only user home would require a separately configured shared writable lock root; it should not trigger an unlocked write-capable fallback.

mykola-mokhnach: Right, good catch -- I missed that the default APPIUM_HOME is exactly `~/.appium`. Fixed in 62d36204f: the lock root is now `~/.cache/appium/locks` (your suggested minimal fix), keeping `homedir()` as the root, realpath-based hashing, and mandatory lock acquisition unchanged.

Added the regression test with a simulated user home: mocks `node:os`'s `homedir()` (so nothing touches the real user's home) and re-imports `env.js` fresh so it picks up the mock, then builds `appiumHome` as `<fakeHome>/.appium` -- matching the default layout -- to assert the lock path never falls under it. I verified this fails against the prior `.appium/locks`-rooted logic and passes now.

I also added an end-to-end check with that same layout: populate a valid manifest, chmod `~/.appium` (the whole thing) read-only, then verify both a read needing no write and lock contention between two concurrent commands still work. One honest caveat on that second test: chmod isn't recursive, so a `locks` directory that already existed *inside* `.appium` before the chmod keeps its own (still-writable) permissions regardless of the parent -- so it didn't actually discriminate the bug in my own before/after check, only the first (path-identity) test did. I kept it anyway since it's still a real, valid check of the fixed end-to-end behavior the way you described it (loading/listing plus lock contention against a read-only home), just flagging that it's not independently proof of the fix by itself.

KazuCocoa (approved):
lg

PR #22769, opened 2026-09-13
fix(base-driver): improve bidi command param validation and error messages

mykola-mokhnach:
Reuse the HTTP route handlers' checkParams/makeArgs logic in executeBidiCommand instead of a bespoke loop, so missing/invalid bidi params report the same detailed multi-line error as HTTP endpoints and unknown params are filtered out consistently.


KazuCocoa (approved):

eglitise:
The `checkParams` logic adds `sessionId` and `id` in optional params, is that still needed for BiDi?

PR #22770, opened 2026-09-13
feat(base-driver): add definitions for all WebDriver BiDi commands

mykola-mokhnach:
Fill in the BIDI_COMMANDS map with every command from every module in the current spec (session, browser, browsingContext, emulation, network, script, storage, input, webExtension), split one module per file under protocol/bidi-commands/ and assembled by a barrel, mirroring protocol/routes/. Unimplemented commands already throw NotYetImplementedError via ensureBidiCommandExists, so this is introspection/validation surface only, not new behavior.


packages/base-driver/lib/protocol/bidi-commands/session.ts
```diff
@@ -0,0 +1,31 @@
+import type {BidiMethodMap} from '@appium/types';
+
+const SUBSCRIPTION_REQUEST_PARAMS = {
+  required: ['events'],
+  optional: ['contexts'],
```
eglitise: The current spec is slightly different:
* `session.subscribe` also supports `userContexts` as an optional parameter
* `session.unsubscribe` requires either `subscriptions` or `events`, and has no optional parameters

packages/base-driver/lib/protocol/bidi-commands/emulation.ts
```diff
@@ -0,0 +1,94 @@
...
+    command: 'bidiEmulationSetScrollbarTypeOverride',
+    params: {
+      required: ['scrollbarType'],
+      optional: ['contexts', 'userContexts'],
+    },
+  },
+  setTimezoneOverride: {
+    command: 'bidiEmulationSetTimezoneOverride',
+    params: {
+      required: ['timezoneId'],
```
eglitise:
```suggestion
      required: ['timezone'],
```

packages/base-driver/lib/protocol/bidi-commands/session.ts
mykola-mokhnach: Confirmed against the spec (session.SubscribeParameters / UnsubscribeByIDRequest / UnsubscribeByAttributesRequest) and fixed in 7423b6f10: subscribe now also accepts `userContexts`, and unsubscribe requires either `subscriptions` or `events` (no optional params). Had to widen `BidiMethodParams.required` to support the same alternate-required-set shape `PayloadParams` already uses for REST routes, since `checkParams`/`makeArgs` already handle it generically.

packages/base-driver/lib/protocol/bidi-commands/emulation.ts
mykola-mokhnach: Confirmed against the spec (emulation.SetTimezoneOverrideParameters uses `timezone`) — fixed in 7423b6f10.

eglitise (approved):

eglitise:
Unsure if it should be listed as a breaking change, since `session.unsubscribe` no longer accepts `contexts`

mykola-mokhnach:
Good catch — restored `contexts` as a non-standard optional param on `session.unsubscribe` in 6cff5eb27, so this isn't a breaking change after all. `bidiUnsubscribe` genuinely uses it to scope which contexts' subscriptions get removed; only the spec-accurate `subscriptions`/`events` required-alternation from the earlier fix stays.

eglitise:
But since the BiDi protocol does not define this filtering by context when unsubscribing, would it not be more accurate to instead remove that functionality here?
Edit: can also be a separate PR if preferred

mykola-mokhnach:
Follow-up: after checking the actual driver ecosystem (xcuitest, uiautomator2, android, mac2, chromium-driver, geckodriver — none call `session.unsubscribe` with `contexts`, and the browsers that proxy the whole BiDi socket bypass this validation entirely), decided to drop `contexts` again in d27adc36c and mark it explicitly as a breaking change instead of quietly restoring it.

eglitise:
Should the signature and implementation of `bidiUnsubscribe` also be updated?
Edit: I guess `bidiSubscribe` too. Perhaps best left for a separate PR

mykola-mokhnach:
> Should the signature and implementation of `bidiUnsubscribe` also be updated? Edit: I guess `bidiSubscribe` too. Perhaps best left for a separate PR

Updated the docs

eglitise:
I was mainly referring to their actual implementation in `packages/base-driver/lib/basedriver/commands/bidi.ts`, plus related tests and types

mykola-mokhnach:
> I was mainly referring to their actual implementation in `packages/base-driver/lib/basedriver/commands/bidi.ts`, plus related tests and types

Would it be possible for you to add them in a separate PR? I probably don't know what's expected.

eglitise:
That's why I was suggesting to have a separate PR: this one can add all definitions while leaving the subscribe/unsubscribe APIs as is, without being breaking; and another breaking change PR can focus specifically on updating subscribe/unsubscribe

mykola-mokhnach:
> That's why I was suggesting to have a separate PR: this one can add all definitions while leaving the subscribe/unsubscribe APIs as is, without being breaking; and another breaking change PR can focus specifically on updating subscribe/unsubscribe

got it. Is it ok to merge this one as is?

eglitise:
I think it would be best if the subscribe/unsubscribe related docs and signature changes were reverted here (again, so as not to make this PR a breaking change)

packages/base-driver/lib/protocol/bidi-commands/session.ts
```diff
@@ -0,0 +1,32 @@
...
+  subscribe: {
+    command: 'bidiSubscribe',
+    params: {
+      required: ['events'],
+      optional: ['contexts', 'userContexts'],
+    },
+  },
+  unsubscribe: {
+    command: 'bidiUnsubscribe',
+    params: {
```
KazuCocoa, line 27: [P2] Update the unsubscribe handler when removing contexts

Removing `contexts` from the parameter map makes `bidiUnsubscribe` always use its default `['']`, but the handler still only filters out the specified context strings. After `session.subscribe` with `{events: ['log.entryAdded'], contexts: ['context-1']}`, unsubscribing returns success while leaving that subscription active. I reproduced this with the repository's `checkParams`, `makeArgs`, and subscription handlers: the base map removes the subscription, whereas this map leaves it intact. Could the handler also be updated to remove all contexts for the requested events when dropping this parameter?

packages/base-driver/lib/protocol/bidi-commands/session.ts
```diff
@@ -0,0 +1,32 @@
...
+      // either unsubscribe by previously-returned subscription ids, or by event names
+      required: [['subscriptions'], ['events']],
```
KazuCocoa, line 29: [P2] Preserve the distinction between subscription IDs and event names

`makeArgs` converts both `{subscriptions: ['abc']}` and `{events: ['abc']}` into the same positional argument list, `[['abc']]`. The existing `bidiUnsubscribe` handler interprets that array as event names, so the newly accepted ID-based form silently succeeds without removing the intended subscription. The handler also cannot distinguish the two forms from these arguments. Could ID-based handling and a way to preserve which parameter was supplied be added before accepting this form?

packages/base-driver/lib/protocol/bidi-commands/session.ts
mykola-mokhnach: I've restored contexts param

packages/base-driver/lib/protocol/bidi-commands/session.ts
mykola-mokhnach: this is probably a topic for another pr

PR #22776, opened 2026-09-14
feat(base-driver)!: align timeouts endpoints closer to W3C standard

eglitise:
The current implementation of the `GET/SET /timeouts` endpoints is somewhat non-standard:
* `POST /timeouts` accepts the standard `script`, `pageLoad`, and `implicit` parameters, as well as legacy `type` and `ms` parameters
* `GET /timeouts` returns an object with the W3C standard `implicit` value, plus the Appium custom `command` value. The W3C standard `script` and `pageLoad` values are omitted.

This PR aims to make the implementation more closely aligned with [the W3C WebDriver standard](https://w3c.github.io/webdriver/#timeouts), while still retaining the custom Appium parameters, since they are used across many drivers.
* `POST /timeouts` now also accepts the `command` parameter. The `type` and `ms` parameters are deprecated.
* `GET /timeouts` now also returns `script` and `pageLoad` values

I have also removed the error if none of `script`/`pageLoad`/`implicit` are provided, as per the WebDriver spec.

Important to note that if the endpoint handling is proxied to the driver's native bindings while in web context, both the GET and SET endpoints are expected to accept/return only the standard W3C parameters and values.

BREAKING CHANGE: the `POST /timeouts` endpoint no longer raises an error if none of the W3C standard parameters are specified, as per the WebDriver protocol
BREAKING CHANGE: `driver.timeouts` no longer accepts the `implicit` parameter in `string` format, to align with the `script` and `pageLoad` parameters
BREAKING CHANGE: removed the `scriptTimeoutW3C`/`pageLoadTimeoutW3C`/`implicitWaitW3C`/`newCommandTimeout` driver methods. Please use `setScriptTimeout`/`setPageLoadTimeout`/`setImplicitWait`/`setNewCommandTimeout` instead. Note that the `appium:newCommandTimeout` capability is unaffected.


mykola-mokhnach:
> BREAKING CHANGE: the GET /timeouts endpoint now also returns the script and pageLoad values, as per the WebDriver protocol

is this really a breaking change?

packages/base-driver/lib/basedriver/core.ts
```diff
@@ -56,7 +54,9 @@ export class DriverCore<const C extends Constraints, Settings extends StringReco
   allowInsecure: string[] = [];
   denyInsecure: string[] = [];
   newCommandTimeoutMs: number = NEW_COMMAND_TIMEOUT_MS;
-  implicitWaitMs: number = 0;
+  scriptTimeoutMs: number = W3C_TIMEOUTS_MS.SCRIPT;
```
mykola-mokhnach, line 57: should they be readonly?

mykola-mokhnach: since we already have corresponding setters

eglitise: the setters directly assign to these variables, so adding readonly causes TS to complain

eglitise:
> is this really a breaking change?

I guess only for users who expect only the existing two properties, but fair enough, I'll remove the line

packages/base-driver/lib/basedriver/commands/timeout.ts
```diff
@@ -100,8 +113,7 @@ export async function pageLoadTimeoutW3C<C extends Constraints>(this: BaseDriver
  * @param ms - the timeout in ms
  */
 export async function scriptTimeoutW3C<C extends Constraints>(this: BaseDriver<C>, ms: number): Promise<void> {
```
mykola-mokhnach: do we need to keep these extra wrappers with W3C suffix?

packages/base-driver/test/unit/basedriver/timeout.spec.ts
```diff
@@ -18,19 +21,28 @@ describe('timeout', function () {
...
+    });
+  });
+
   describe('timeouts', function () {
-    describe('errors', function () {
-      it('should reject an invalid timeout with InvalidArgumentError', function () {
-        assert.throws(() => driver.parseTimeoutArgument('abc'), errors.InvalidArgumentError);
-        assert.throws(() => driver.parseTimeoutArgument(-1), errors.InvalidArgumentError);
-      });
+    describe('JSONWP errors', function () {
```
mykola-mokhnach, line 45: is this suite still relevant?

eglitise: It verifies the functionality of the `type` and `ms` properties. This PR only deprecates them, since I assume we want to keep them for now.

packages/base-driver/lib/basedriver/commands/timeout.ts
eglitise: removed them

mykola-mokhnach (approved):

PR #22777, opened 2026-09-14
feat(appium): let plugins intercept BiDi commands and events when proxying

mykola-mokhnach:
## Summary

Previously, a driver that proxies its BiDi websocket connection to an upstream server (e.g. a real browser) bypassed plugins entirely -- commands and events were piped through as raw, unparsed bytes with an explicit `// too complex for now, so just ignore plugins in this case` TODO. This PR closes that gap: plugins can now intercept, modify, or veto BiDi **commands** and BiDi **events** identically, whether the driver handles BiDi locally or proxies it upstream.

## What changed

**Proxy transport**
- Add `BidiProxyClient` (`packages/appium/lib/bidi/proxy-client.ts`), a WS analog of `WebDriverProxy`: it turns the previous raw byte-pipe into an id-correlated request/response client, so proxied commands can flow through the existing plugin `next()`-chain instead of being forwarded blindly.

**Command interception while proxying**
- `onBidiMessage`'s base (innermost) handler now forwards through `BidiProxyClient.executeCommand` when the driver is proxying, instead of defaulting straight to `driver.executeBidiCommand`. Plugins wrap this base handler exactly as they do for local commands.
- Known/registered commands are still param-validated locally before being forwarded; unrecognized/vendor commands pass through permissively so upstream-only functionality isn't blocked by Appium's static command knowledge.
- `session.subscribe`/`session.unsubscribe` sent through the proxy also update the driver's local `bidiEventSubs` bookkeeping (dual bookkeeping), since the event dispatcher's subscription gate depends on it uniformly for both local and proxied events.

**Event interception (new, for all origins)**
- New `Plugin.handleBidiEvent` hook (`packages/types/lib/plugin.ts`) mirrors `Plugin.handle` for commands: `(next, driver, event, origin) => Promise<void>`. Plugins can pass an event through unchanged, replace its payload via `next(modifiedEvent)`, or veto it by not calling `next` at all.
- New `createBidiEventDispatcher` (`packages/appium/lib/bidi/events.ts`) builds a single per-connection dispatcher shared by driver-emitted, plugin-emitted, and (new) upstream-proxy-pushed events. Every event, regardless of origin, folds through the same plugin chain (last-declared plugin runs first, matching command-chain ordering) before the `bidiEventSubs` subscription gate decides delivery. Events are processed via a per-connection FIFO queue so wire order always matches emission order, and a throwing handler drops only that one event (logged) without stalling the queue.
- `origin.type` is one of `'driver' | 'plugin' | 'proxy'`, so a plugin can distinguish where an event came from (including its own emitted events).

**Refactor**
- Split the previously monolithic `bidi-commands.ts`/`bidi-proxy-client.ts` into a `packages/appium/lib/bidi/` module: `types.ts`, `constants.ts`, `proxy-client.ts`, `commands.ts` (command dispatch), `events.ts` (event dispatch), `socket.ts` (connection lifecycle), with an `index.ts` barrel. No behavior change; purely organizational.

**Test infrastructure**
- `fake-driver` gains a `bidiProxyUrl` capability (wires the previously-dead `bidiProxyUrl` getter), so e2e tests can point a real session at a local upstream `ws` server fixture.
- `fake-plugin` gains an opt-in `interceptBidiEvents` cliArg demonstrating `handleBidiEvent`: modifies driver-emitted clock-tick events, and vetoes a dedicated test event.

## Testing

- Unit: `packages/appium/test/unit/bidi/proxy-client.spec.ts` (id correlation, error unwrap, unsolicited-message routing, concurrent in-flight commands, close-rejects-pending) and `packages/appium/test/unit/bidi/events.spec.ts` (plugin chain ordering, modify/veto, self-interception, subscription-gate-after-interception, error containment, FIFO ordering).
- E2E: extended `packages/appium/test/e2e/plugin.e2e.spec.ts` with a proxying-driver `describe` block (real upstream `ws.WebSocketServer` fixture) and a BiDi-event-interception `describe` block.
- Verified: `tsc -b --force` (clean), `oxlint -c oxlint.config.mjs` (clean), `oxfmt --check` (clean), full unit suite (513 passing) and e2e `plugin.e2e.spec.ts` suite (39 passing) for the `appium` package.

## Out of scope

- A typed `BIDI_EVENTS` registry parallel to `BIDI_COMMANDS` -- deferred in favor of the freeform `handleBidiEvent` catch-all, since BiDi's module system is open-ended and proxied upstream servers may emit vendor/experimental events Appium has no static knowledge of.
- Generic interception of any unregistered BiDi command beyond what `updateBidiCommands` already provides.


mykola-mokhnach:
User Stories

## Command interception (proxy mode)

- [ ] **1.** As a plugin author, I want to intercept a BiDi command even when the driver proxies BiDi to an upstream server, so that my plugin works the same regardless of whether the driver implements BiDi locally.
  - *Check:* start a session with `appium:bidiProxyUrl` pointing at a real/fake upstream BiDi server; a plugin implementing `executeBidiCommand` for that method gets called before the command reaches upstream.

- [ ] **2.** As a plugin author, I want to short-circuit a proxied command by not calling `next()`, so that I can fully replace its behavior instead of forwarding it upstream.
  - *Check:* plugin returns its own result without calling `next`; assert the upstream server never received the message.

- [ ] **3.** As a driver author whose driver proxies BiDi, I want unrecognized/vendor BiDi commands to pass through to the upstream server without Appium rejecting them, so that upstream-specific commands aren't blocked by Appium's static command knowledge.
  - *Check:* send a command like `vendor.customThing` with no matching entry in `driver.bidiCommands`; confirm it's forwarded upstream unmodified and the response comes back to the client.

- [ ] **4.** As Appium, I want to validate params locally against the canonical/registered BiDi command shape when the command IS known, even while proxying, so that malformed commands fail fast instead of round-tripping to the upstream server.
  - *Check:* send a known command (e.g. `session.subscribe`) with missing/invalid required params while proxying; confirm it errors locally without hitting the upstream socket.

- [ ] **5.** As Appium, I want `session.subscribe`/`session.unsubscribe` sent through a proxy to also update the local driver's `bidiEventSubs`, so that the event-delivery gate (which only exists locally) still works correctly for proxied event push.
  - *Check:* subscribe to an event via a proxied session, then trigger that event upstream; confirm it's delivered to the client. Unsubscribe and confirm delivery stops.

- [ ] **6.** As a client, I want a proxied command that never gets a response (upstream closes mid-flight) to fail cleanly with an error instead of hanging forever.
  - *Check:* kill/close the upstream connection after sending a command but before it replies; confirm the client gets an error response, not a timeout/hang.

## Event interception (all origins)

- [ ] **7.** As a plugin author, I want a generic `handleBidiEvent` hook that fires for BiDi events regardless of origin (driver-emitted, another plugin-emitted, or proxied from upstream), so I have one consistent interception point.
  - *Check:* trigger a driver-emitted event, a plugin-emitted event, and (in proxy mode) an upstream-pushed event; confirm the same `handleBidiEvent` implementation sees all three, with the correct `origin.type`.

- [ ] **8.** As a plugin author, I want to modify an event's payload before it reaches the client (via `next(modifiedEvent)`), so I can enrich or transform data in flight.
  - *Check:* plugin adds a field to `params`; confirm the client-received event includes it.

- [ ] **9.** As a plugin author, I want to veto an event by not calling `next()`, so subscribed clients never see it even though they're subscribed.
  - *Check:* client subscribes to an event; plugin vetoes it; confirm the client never receives that event, despite being subscribed.

- [ ] **10.** As a plugin author, I want to see (and be able to modify/veto) events emitted by my own plugin, not just driver events, so interception is symmetric.
  - *Check:* plugin emits its own BiDi event; confirm its own `handleBidiEvent` runs on it with `origin.type === 'plugin'` and `pluginName` set correctly.

- [ ] **11.** As Appium, I want multiple plugins' `handleBidiEvent` hooks to chain in the same last-declared-runs-first order as command interception, so behavior is predictable and consistent with the command chain.
  - *Check:* register two plugins that both implement `handleBidiEvent`; confirm execution order matches plugin declaration order (reversed), same as REST/command plugin chaining.

- [ ] **12.** As Appium, I want the client's subscription (`bidiEventSubs`) checked *after* all plugin interception, uniformly for every origin, so plugins can't force delivery of an event nobody subscribed to.
  - *Check:* plugin calls `next()` (passthrough) on an event the client never subscribed to; confirm it's still not delivered.

- [ ] **13.** As Appium, I want a throwing `handleBidiEvent` handler to drop just that one event (logged as a warning) without breaking delivery of subsequent events on the same connection.
  - *Check:* plugin throws on event A; confirm event A is dropped, a warning is logged, and a later unrelated event B still arrives normally.

- [ ] **14.** As a client, I want events delivered in the same order they were emitted (FIFO), even if plugin handlers have different processing delays, so event ordering isn't scrambled by interception.
  - *Check:* emit several events back-to-back where handling them takes staggered async time; confirm wire order still matches emission order.

## Session lifecycle / cleanup

- [ ] **15.** As Appium, I want deleting a session that was proxying BiDi to also close the upstream connection (not just the client-facing socket), so upstream resources aren't leaked.
  - *Check:* delete a session with an active BiDi proxy connection; confirm both the client-facing WS and the upstream WS close.

- [ ] **16.** As Appium, I want the client connection to close (with a sane, WS-spec-valid close code) when the upstream proxy connection closes unexpectedly, so clients get a clear signal instead of a silently-dead socket.
  - *Check:* close the upstream server's socket directly (not via session deletion); confirm the client-facing WS also closes, with a valid close code (invalid upstream codes get rewritten to the fallback code).

## Regression / non-proxy baseline

- [ ] **17.** As an existing plugin/driver author not using BiDi proxying, I want all existing non-proxy BiDi command and event behavior to work exactly as before, so this change is non-breaking for the common case.
  - *Check:* run the existing (pre-PR) non-proxy BiDi e2e/unit tests; confirm no behavior changes when `bidiProxyUrl` is unset and no plugin implements `handleBidiEvent`.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (known?.params) {
+      checkParams(known.params, params, {ensureSessionArgs: false});
+    }
+    const result = await bidiProxyClient.executeCommand(method, params);
+    // Keep Appium's local bidiEventSubs bookkeeping in sync, since the event dispatcher's
+    // send-gate relies on it uniformly for both locally-emitted and proxied events.
+    if (method === SESSION_SUBSCRIBE && typeof driver.bidiSubscribe === 'function') {
+      await driver.bidiSubscribe(params.events, params.contexts);
+    } else if (method === SESSION_UNSUBSCRIBE && typeof driver.bidiUnsubscribe === 'function') {
+      await driver.bidiUnsubscribe(params.events, params.contexts);
```
KazuCocoa: [P2] Handle unsubscribe-by-ID before updating local subscriptions

For `session.unsubscribe` with `{subscriptions: [subscriptionId]}`, the registered command schema accepts the request and the upstream server can successfully unsubscribe, but `params.events` is undefined here. The inherited `bidiUnsubscribe` iterates `events`, so this throws `TypeError: events is not iterable`. The client receives `unknown error` after the upstream operation has already succeeded, and the local subscription entry remains.

Could we track the upstream subscription IDs and their local event/context entries, then handle the `subscriptions` form explicitly? A regression test should subscribe, use the returned ID to unsubscribe, and assert both a successful response and removal of the local subscription. I reproduced the error using the PR functions with a stubbed successful upstream response.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,247 @@
...
+      return;
+    }
+    if (!parsed.method) {
+      driverLog.warn(
+        `Ignoring unsolicited upstream BiDi message without a method: ${util.truncateString(data.toString(), {length: MAX_LOGGED_DATA_LENGTH})}`,
+      );
+      return;
+    }
+    void dispatchBidiEvent(
+      {method: parsed.method, params: parsed.params ?? {}, context: parsed.context},
```
KazuCocoa: [P1] Preserve the context of standard upstream BiDi events

Subscribing with `{events: ["browsingContext.load"], contexts: ["tab-1"]}` succeeds upstream, but a standard load event carries the context in `params.context`, not at the top level. Reading `parsed.context` here therefore produces undefined, which the terminal dispatcher turns into an empty string. The subscription gate then rejects the event because `["tab-1"]` does not include that empty string. This silently drops matching context-scoped events even with no plugins enabled, whereas the previous raw proxy forwarded them.

Could we normalize upstream event context information and apply the appropriate BiDi subscription matching semantics before gating delivery? Please add a test that subscribes to a specific context and pushes a standard-shaped event with `params.context`. I reproduced the drop with the PR dispatcher and proxy handler.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,247 @@
...
+): BidiDispatch {
+  let queue: Promise<void> = Promise.resolve();
+
+  const terminal = async (event: BidiEventPayload): Promise<void> => {
+    if (ws.readyState !== WebSocket.OPEN) {
+      return;
+    }
+    const context = event.context || '';
+    const {method, params} = event;
+    const eventSubs = bidiHandlerDriver.bidiEventSubs[method];
```
KazuCocoa: [P1] Match module-wide subscriptions when filtering proxied events

BiDi allows subscribing to an entire module, for example `{events: ["browsingContext"]}`. The new dual bookkeeping stores that literal key in `driver.bidiEventSubs`, but an upstream event arrives with a method such as `browsingContext.load`. This exact-key lookup returns undefined, so the following gate drops every event covered only by the module subscription. Previously the raw proxy delivered these upstream events.

Could the local gate account for module-wide subscriptions, including vendor modules, rather than requiring an exact event-name entry? Please add a proxy regression test that subscribes to `browsingContext` and verifies delivery of `browsingContext.load`. I reproduced this with the PR dispatcher and inherited subscription bookkeeping; an exact event-name subscription passes the same test.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in 14c4b098e: track each `session.subscribe`'s returned subscription id -> {events, contexts} per `BidiProxyClient`, and resolve it back on `session.unsubscribe` with `{subscriptions: [id]}`, falling back to the events/contexts form otherwise. Added an e2e regression test (subscribe, capture the id, unsubscribe by id, verify delivery stops).

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in 14c4b098e: context now falls back to `params.context` when the top-level `context` field is absent, matching the standard BiDi event envelope shape. Added an e2e regression test that subscribes to a context and pushes a standard-shaped event with `params.context`.

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in 14c4b098e: the subscription gate now checks both the exact event name and its module name, so a module-wide subscription (e.g. {events: ["browsingContext"]}) matches events like browsingContext.load. Added a unit test for the gate plus an e2e proxy regression test subscribing to a bare module name.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,284 @@
...
+        `Ignoring unsolicited upstream BiDi message without a method: ${util.truncateString(data.toString(), {length: MAX_LOGGED_DATA_LENGTH})}`,
+      );
+      return;
+    }
+    const params = parsed.params ?? {};
+    // Standard BiDi event envelopes don't carry a top-level `context` -- context-scoped events
+    // (e.g. `browsingContext.load`) nest it inside `params.context` instead. Normalize a missing
+    // context to '' here (rather than leaving it undefined), matching how driver/plugin-origin
+    // events are normalized in initBidiEventListeners, so origin never changes what plugins see.
+    const context = parsed.context ?? (params.context as string | undefined) ?? '';
```
KazuCocoa: [P2] Handle source.context when normalizing log events

The fallback to `params.context` fixes browsing-context events, but `log.entryAdded` carries its context in `params.source.context`. With `{events: ["log.entryAdded"], contexts: ["tab-1"]}`, an upstream console event whose source is `{realm: "realm-1", context: "tab-1"}` is still normalized to an empty context and rejected by the subscription gate, even though the upstream subscription matches.

Could context normalization account for the event-specific BiDi payload shapes, including `source.context`, with a regression test for a context-scoped console log? I reproduced the drop using the current proxy handler and dispatcher with stubbed dependencies.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,255 @@
...
+  driver: ExtensionCore & Partial<IBidiCommands>,
+  bidiProxyClient: BidiProxyClient,
+  method: string,
+  params: StringRecord,
+  result: BiDiResultData,
+): Promise<void> {
+  if (method === SESSION_SUBSCRIBE && typeof driver.bidiSubscribe === 'function') {
+    const events = (params.events ?? []) as string[];
+    const contexts = (params.contexts ?? ['']) as string[];
+    await driver.bidiSubscribe(events, contexts);
```
KazuCocoa: [P1] Preserve coverage from earlier subscriptions to other contexts

The inherited `bidiSubscribe` assigns `this.bidiEventSubs[event] = contexts`, so this call replaces existing coverage rather than adding to it. Subscribe to `browsingContext.load` for `["tab-1"]`, then subscribe to the same event for `["tab-2"]`: both subscriptions remain active upstream, but locally only tab-2 remains and tab-1 events are dropped. Unsubscribing the second ID then removes the local entry entirely instead of restoring the still-active tab-1 subscription.

Could we derive local coverage from the union of all active tracked subscriptions (including the global wildcard) when adding/removing subscriptions? The new overlap test uses identical contexts and therefore misses this case. I reproduced both the lost tab-1 delivery and the empty local state after removing tab-2 using the current functions with a stubbed upstream.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,255 @@
...
+      throw new errors.UnknownCommandError(
+        `Did not receive a valid BiDi module and method name of the form moduleName.methodName. ` +
+          `Instead received '${moduleName}.${methodName}'`,
+      );
+    }
+    const known = driver.bidiCommands[moduleName]?.[methodName];
+    if (known?.params) {
+      checkParams(known.params, params, {ensureSessionArgs: false});
+    }
+    const result = await bidiProxyClient.executeCommand(method, params);
```
KazuCocoa, line 388: [P1] Preserve events emitted before the subscribe response

Local subscription state is updated only after this await completes. An upstream BiDi server can emit events while executing `session.subscribe`: for example, the standard subscribe steps for `browsingContext.contextCreated` emit notifications for existing contexts before returning the subscription result. At that point the local gate still sees no subscription and drops those initial notifications. They are not replayed when `syncLocalSubscriptionState` eventually runs, so clients miss the existing contexts even though subscribing succeeds.

Could we buffer events while the subscription is pending, or otherwise coordinate subscription activation with delivery without losing those notifications? Please add a proxy test that emits contextCreated before the subscribe success response, rather than only triggering events after it. I reproduced the loss with the current functions and a stubbed upstream that emits before resolving the command; events emitted afterward are delivered.

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in 450139e5e: context normalization now also falls back to params.source.context (where log events carry it), in addition to params.context. Added an e2e regression test with a context-scoped log event.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in 450139e5e: replaced the ad-hoc subscribe/unsubscribe bookkeeping with a single syncCoverage() that always recomputes bidiEventSubs[event] as the union of every currently-tracked subscription's contexts, for both subscribe and unsubscribe-by-id. Added an e2e regression test subscribing the same event for two different contexts and verifying the first context's delivery survives unsubscribing the second.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in 450139e5e: executeProxiedSubscribe now opens local delivery optimistically (under a placeholder tracking id) before forwarding session.subscribe upstream, swaps in the real subscription id on success, and rolls the placeholder back if the upstream call fails -- so events pushed while the subscribe is still in flight are no longer dropped. Added an e2e regression test with a stubbed upstream that emits before sending the subscribe success response.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,287 @@
...
+ * or an entire module (`browsingContext`, covering all of its events). `bidiEventSubs` may
+ * therefore hold either kind of key, so a matching event must be checked against both. An
+ * empty-string context entry means "all contexts" (per `bidiSubscribe`'s contract), so it
+ * matches any actual context id, not just the literal empty string.
+ */
+function isEventSubscribed(bidiEventSubs: Record<string, string[]>, method: string, context: string): boolean {
+  const moduleName = method.split('.')[0];
+  return [method, moduleName].some((key) => {
+    const subs = bidiEventSubs[key];
+    return Array.isArray(subs) && (subs.includes('') || subs.includes(context));
```
KazuCocoa: [P1] Preserve events from descendants of a subscribed context

This exact-ID check drops valid upstream events from iframes. For example, after subscribing to `browsingContext.contextCreated` with `contexts: ['tab-1']`, an upstream event with `params.context: 'frame-1'` and `params.parent: 'tab-1'` is discarded because only `tab-1` is recorded locally. BiDi context-scoped subscriptions cover the associated top-level traversable, including its descendants, so the upstream subscription correctly includes this event. Please account for that relationship when applying the local gate (or preserve the upstream filtering semantics). I reproduced this with the current dispatcher and proxy handlers. This is an existing issue in this PR's proxy event gating, rather than a regression introduced by the latest subscription fixes.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+ * Drops tracked subscriptions whose (event, context) coverage is no longer reflected in
+ * `bidiEventSubs`. Needed after an `events`/`contexts`-form unsubscribe (which bypasses id
+ * tracking and mutates `bidiEventSubs` directly), so a stale tracked id doesn't later cause
+ * {@link syncCoverage} to think an already-removed subscription is still active.
+ */
+function pruneStaleTracking(bidiEventSubs: Record<string, string[]>, tracking: Map<string, TrackedSubscription>): void {
+  for (const [subscriptionId, sub] of tracking) {
+    const stillCovered = sub.events.some((event) => {
```
KazuCocoa: [P2] Remove partially unsubscribed coverage from tracked records

`pruneStaleTracking` retains the entire record whenever any event/context is still covered. With the newly added `syncCoverage`, those stale entries can now recreate a subscription that was explicitly removed. I reproduced this sequence: (1) subscribe globally to `['appium:clock.currentTime', 'vendor.keep']` as `sub1`; (2) unsubscribe by event name from `appium:clock.currentTime`; (3) subscribe to that clock event for `contexts: ['tab-1']` as `sub2`; (4) unsubscribe `sub2` by ID. Because `sub1` still contains the removed clock event, `syncCoverage` restores its global `['']` coverage, and a subsequent driver-emitted clock event is delivered even though no clock subscription remains upstream. Please update the retained records to remove the specific event/context coverage on partial unsubscribe, rather than only deleting wholly stale records. This becomes a regression with the latest coverage-recomputation change.

KazuCocoa (commented):
Some additions

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in 4982b196d: added per-driver browsing-context ancestry tracking, built by observing browsingContext.contextCreated/contextDestroyed events as they pass through the dispatcher (before plugin interception, so a plugin veto/modify doesn't blind the gate). isEventSubscribed now walks that ancestry when a context doesn't match a subscription directly. Added unit coverage for descendant-context matching.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in 4982b196d: pruneStaleTracking now shrinks a record's event list to just the events still actually reflected in bidiEventSubs (deleting the whole record only once none remain), instead of only ever deleting the record wholesale. Added an e2e regression test reproducing the exact sequence from the report.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,316 @@
...
+
+interface TrackedSubscription {
+  events: string[];
+  contexts: string[];
+}
+
+// Per-proxy-client bookkeeping of upstream subscription ids -> the events/contexts they cover,
+// so a later `session.unsubscribe` by id (the standard BiDi form) can be translated into the
+// event/context form that the local `bidiUnsubscribe` bookkeeping understands.
+const SUBSCRIPTION_TRACKING: WeakMap<BidiProxyClient, Map<string, TrackedSubscription>> = new WeakMap();
```
KazuCocoa: [P2] Keep subscription tracking at the same scope as the driver state

This map belongs to a `BidiProxyClient`, but `driver.bidiEventSubs` belongs to the whole session. With two connections to the same session, subscribing to `log.entryAdded` for tab-1 on A and then tab-2 on B makes B rebuild coverage from only its own records, removing tab-1 and dropping valid events for A. Unsubscribing an existing subscription ID after reconnecting has the inverse problem: upstream accepts it, but the new client has no record to remove locally. Please keep the tracking and coverage at the same session scope, including across reconnects.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,316 @@
...
+async function executeProxiedSubscribe(
+  driver: ExtensionCore & Partial<IBidiCommands>,
+  bidiProxyClient: BidiProxyClient,
+  params: StringRecord,
+): Promise<BiDiResultData> {
+  if (typeof driver.bidiSubscribe !== 'function') {
+    return await bidiProxyClient.executeCommand(SESSION_SUBSCRIBE, params);
+  }
+  const events = (params.events ?? []) as string[];
+  const contexts = (params.contexts ?? ['']) as string[];
```
KazuCocoa: [P2] Normalize subscription contexts to their top-level traversables

BiDi normalizes `session.subscribe` context IDs to their top-level traversables. If frame-1 and frame-2 belong to tab-1, subscribing with `contexts: ["frame-1"]` also covers tab-1 and frame-2 upstream. Storing frame-1 verbatim here makes the local ancestry check reject those valid events, even when the complete parent map is known. Please normalize the stored scope as well. This is separate from missing ancestry information: I reproduced it with both parent relationships already populated.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,316 @@
...
+ */
+async function executeProxiedSubscribe(
+  driver: ExtensionCore & Partial<IBidiCommands>,
+  bidiProxyClient: BidiProxyClient,
+  params: StringRecord,
+): Promise<BiDiResultData> {
+  if (typeof driver.bidiSubscribe !== 'function') {
+    return await bidiProxyClient.executeCommand(SESSION_SUBSCRIBE, params);
+  }
+  const events = (params.events ?? []) as string[];
```
KazuCocoa, line 188: [P2] Keep user-context subscriptions distinct from global subscriptions

`params.userContexts` is not tracked, so a subscription scoped to user-1 is recorded with the same empty-context wildcard as a global subscription. Subscribe to `log.entryAdded` for user-1, then globally, and unsubscribe by event name: upstream removes the global subscription while retaining the user-context subscription, but local tracking removes both and drops subsequent valid user-1 events. The wildcard also lets locally generated events from unrelated user contexts pass. Please preserve this scope in the record and apply the attribute-based unsubscribe rules to the appropriate records.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,316 @@
...
+  tracking.set(placeholderId, {events, contexts});
+  try {
+    await syncCoverage(driver, tracking, events);
+    const result = await bidiProxyClient.executeCommand(SESSION_SUBSCRIBE, params);
+    // Standard BiDi `session.subscribe` results carry the new subscription's id, which a later
+    // `session.unsubscribe` may reference instead of repeating events/contexts. The coverage
+    // itself is unchanged (same events/contexts as the placeholder), so no need to re-apply it.
+    const subscriptionId = (result as {subscription?: string} | undefined)?.subscription;
+    tracking.delete(placeholderId);
+    tracking.set(subscriptionId ?? placeholderId, {events, contexts});
```
KazuCocoa: [P2] Preserve changes to a pending subscription when its response arrives

The placeholder can already have been changed by a concurrent unsubscribe. For example, let upstream accept a subscription to `["browsingContext.load", "browsingContext.domContentLoaded"]` but delay its response; an unsubscribe for `browsingContext.load` then succeeds and prunes the placeholder. When the delayed subscribe response arrives, this assignment restores the original event list. A later coverage rebuild resurrects the unsubscribed event locally. BiDi responses can arrive out of order, so please promote the current placeholder record instead of recreating it from the original arguments, and do not restore a placeholder that was fully removed.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,316 @@
...
+        continue;
+      }
+      tracking.delete(subscriptionId);
+      for (const event of tracked.events) {
+        affectedEvents.add(event);
+      }
+    }
+    await syncCoverage(driver, tracking, affectedEvents);
+  } else if (Array.isArray(params.events)) {
+    await driver.bidiUnsubscribe(params.events as string[], (params.contexts as string[] | undefined) ?? ['']);
```
KazuCocoa: [P2] Apply unsubscribe to expanded event names

Module names and individual events are currently treated as different keys. Subscribe to `log`, then unsubscribe from `log.entryAdded`: the `log` key remains locally and driver/plugin-generated `log.entryAdded` events are still delivered after upstream has stopped them. The reverse sequence (subscribe to `log.entryAdded`, unsubscribe from `log`) also leaves coverage behind. Please expand module subscriptions consistently when updating tracking and checking coverage, so unsubscribing an individual event from a module is represented correctly.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,316 @@
...
+  for (const [subscriptionId, sub] of tracking) {
+    const isCovered = (event: string) => {
+      const activeContexts = bidiEventSubs[event];
+      return Array.isArray(activeContexts) && sub.contexts.some((context) => activeContexts.includes(context));
+    };
+    const remainingEvents = sub.events.filter(isCovered);
+    if (remainingEvents.length === 0) {
+      tracking.delete(subscriptionId);
+    } else if (remainingEvents.length !== sub.events.length) {
+      tracking.set(subscriptionId, {events: remainingEvents, contexts: sub.contexts});
```
KazuCocoa: [P2] Prune partially removed contexts as well as event names

The event-list fix addresses the earlier global case, but the backward-compatible `events` + `contexts` form still leaves stale context coverage when the upstream supports that form. Subscribe to an event for `["tab-1", "tab-2"]`, unsubscribe it only for tab-1, then subscribe to the same event for tab-3 and unsubscribe that new ID. Because the original record retains both contexts here, the final coverage rebuild resurrects tab-1 and locally generated events are delivered there again. Please retain coverage at the event/context-pair level when pruning. This concerns Appium’s compatible context-based unsubscribe form, not the current standard BiDi unsubscribe schema.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,352 @@
...
+ * context ancestry map current. Tracking happens independently of delivery so a plugin that
+ * modifies/vetoes these events doesn't blind the gate to context relationships it needs for
+ * *other* events.
+ */
+function trackContextHierarchy(driver: AnyDriver, event: BidiEventPayload): void {
+  if (event.method === 'browsingContext.contextCreated') {
+    const context = event.params?.context as string | undefined;
+    const parent = event.params?.parent as string | undefined;
+    if (context && parent) {
+      getContextParents(driver).set(context, parent);
```
KazuCocoa: [P1] Do not depend on a client subscription to learn context ancestry

The earlier `contextCreated` case is fixed, but ancestry is learned only from events that upstream happens to send. A client subscribing only to `log.entryAdded` for tab-1 has not subscribed to `browsingContext.contextCreated`, so upstream need not send the events that populate this map. A valid log event with `source.context: "frame-1"` from a child of tab-1 is then discarded by the local gate. The same issue affects network/load-only subscriptions and already-existing frames. Please initialize and maintain ancestry independently of the client’s event subscriptions, or preserve upstream filtering semantics without requiring that map. A regression test should subscribe to child-capable events without subscribing to `contextCreated`.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,352 @@
...
+    }
+  } else if (event.method === 'browsingContext.contextDestroyed') {
+    const context = event.params?.context as string | undefined;
+    if (context) {
+      getContextParents(driver).delete(context);
```
KazuCocoa: [P2] Retain ancestry until queued destruction events have been delivered

Deleting this relationship before the event passes through the queue/gate drops the child’s own `browsingContext.contextDestroyed` event for a subscription scoped to its parent, even if its creation was observed correctly. Since `trackContextHierarchy` also runs outside the FIFO, a destruction message can remove ancestry needed by an earlier child event still awaiting plugin processing in the queue. Please apply lifecycle updates in delivery order and retain the relationship while filtering/delivering the destruction event and preceding events.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,352 @@
...
+    }
+    const params = parsed.params ?? {};
+    // Standard BiDi event envelopes don't carry a top-level `context`. Most events nest it
+    // directly in `params.context` (e.g. `browsingContext.load`), but log events instead carry
+    // it in `params.source.context` (a log entry's context lives on its `source`, not the entry
+    // itself). Normalize a missing context to '' here (rather than leaving it undefined),
+    // matching how driver/plugin-origin events are normalized in initBidiEventListeners, so
+    // origin never changes what plugins see.
+    const source = params.source as {context?: string} | undefined;
+    const context = parsed.context ?? (params.context as string | undefined) ?? source?.context ?? '';
```
KazuCocoa: [P2] Preserve scoped events that identify a realm rather than a context

Some standard events have no context field to extract. With a `script` subscription scoped to tab-1, `script.realmCreated` for `{realm: "r1", context: "tab-1", type: "window", ...}` passes, but the corresponding `script.realmDestroyed` contains only `{realm: "r1"}` and is dropped as context `""`. Dedicated-worker realm creation uses `owners`, and worker log entries may similarly identify only `source.realm`. Please track realm/owner relationships or preserve upstream eligibility for these events; adding more context-field fallbacks cannot recover this missing association.

packages/appium/lib/bidi/proxy-client.ts
```diff
@@ -0,0 +1,222 @@
...
+   * Sends `{id, method, params}` upstream (with a freshly-minted internal id) and waits for the
+   * correlated response, unwrapping `result` on success or throwing a `bidiErrObject`-capable
+   * error (built via {@link errorFromW3CJsonCode}) on a `type: 'error'` response. Rejects if no
+   * response is received within `commandTimeoutMs`.
+   */
+  async executeCommand(method: string, params: StringRecord): Promise<BiDiResultData> {
+    await this.waitUntilOpen();
+    const id = this.nextId++;
+    return await new Promise<BiDiResultData>((resolve, reject) => {
+      const timeoutId = setTimeout(() => {
```
KazuCocoa: [P1] Allow legitimate commands to run longer than 30 seconds

The default timeout now makes valid long-running commands fail. For example, `input.performActions` with `{context: "tab-1", actions: [{type: "none", id: "wait", actions: [{type: "pause", duration: 45000}]}]}` legitimately needs at least 45 seconds. This timer returns `unknown error` after 30 seconds while upstream continues executing; its eventual success no longer has a pending request and is discarded by the event path. `socket.ts` constructs this client with only `log`, so users cannot adjust this policy through the current integration. Please support an appropriate configurable/command-aware timeout rather than imposing a fixed 30-second cap. I verified the default-timer behavior with fake timers and a delayed upstream response.

packages/appium/lib/bidi/proxy-client.ts
```diff
@@ -0,0 +1,222 @@
...
+    const id = parsed?.id;
+    const entry = typeof id === 'number' ? this.pending.get(id) : undefined;
+    if (!entry) {
+      this.dispatchUnsolicited(data);
+      return;
+    }
+    this.pending.delete(id);
+    clearTimeout(entry.timeoutId);
+    if (parsed.type === 'error') {
+      entry.reject(errorFromW3CJsonCode(parsed.error, parsed.message, parsed.stacktrace));
```
KazuCocoa: [P2] Preserve upstream BiDi-specific error codes

`errorFromW3CJsonCode` uses the Classic WebDriver error map, which does not include BiDi codes such as `no such handle`, `no such request`, `no such script`, or `no such user context`. Valid upstream errors with these codes therefore become `unknown error` when forwarded, preventing clients from identifying the failure or applying the appropriate recovery. I reproduced all four conversions using the actual error classes. Please preserve the upstream BiDi error code, message, and stack trace, or use a BiDi-aware conversion here.

KazuCocoa (commented):
I let codex review widely. I think there are everything..

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in ea62021bb: subscription tracking is now keyed by the driver (WeakMap<AnyDriver, ...>), the same session-wide scope as bidiEventSubs, instead of per-BidiProxyClient. Coverage now survives across a new connection's own BidiProxyClient rather than being rebuilt from that connection's tracking alone.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in ea62021bb: executeProxiedSubscribe now normalizes each subscribed context to its top-level traversable (via a new getTopLevelContext helper built on the existing context-ancestry map from events.ts) before storing/syncing it, matching BiDi's own session.subscribe normalization.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Looked into this one but didn't fix it: representing a userContexts-scoped subscription distinctly from a global one would require extending bidiEventSubs/bidiSubscribe/bidiUnsubscribe's shape (packages/base-driver's IBidiCommands, which only accepts contexts today) to carry userContexts too, since that's the actual local state the gate reads -- not something addressable within the bidi/ proxy module alone. That's a public, cross-driver API surface change, which felt out of scope for this PR to make unilaterally. Leaving this thread open -- flagging as a known gap rather than closing it.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in ea62021bb: executeProxiedSubscribe now reads the placeholder's *current* tracked value (which a concurrent unsubscribe may have already narrowed or fully removed) when promoting it to the real subscription id, instead of recreating it from the original subscribe arguments.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in ea62021bb, conservatively: without a full BiDi event registry (deliberately out of scope, since BiDi's module system is open-ended), there's no way to know a module's complete membership, so narrowing one event out of an active module-wide subscription (or vice versa) can't be represented precisely. expandRelatedEventKeys now detects the overlap (bare module <-> its specific-event keys already reflected in bidiEventSubs) and drops local coverage across every related key together, rather than leaving the other key wrongly granting/lacking coverage. Under-delivering a sibling event this way is a safer failure than over-delivering one the client never asked for. Added an e2e regression test documenting this conservative behavior.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in ea62021bb: TrackedSubscription now stores contexts per-event (Record<event, contexts>) instead of one shared contexts list, and pruneStaleTracking shrinks a specific event's context list to just what's still reflected in bidiEventSubs, so a partially-unsubscribed context can't be resurrected by a later, unrelated coverage rebuild. Added an e2e regression test reproducing the exact sequence from the report.

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in ea62021bb: when proxying, Appium now bootstraps browsing-context ancestry from the upstream browsingContext.getTree response and auto-subscribes upstream to browsingContext.contextCreated/contextDestroyed internally (bypassing the client-facing dual-bookkeeping entirely), independent of whatever the client itself subscribes to. Added unit coverage for both the getTree bootstrap and delivery of a child-context event without the client ever subscribing to contextCreated.

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in ea62021bb: context/realm destruction bookkeeping (trackContextDestroyed/trackRealmDestroyed) now runs after chain(event, origin) resolves for that event (in a finally block), rather than synchronously before the event enters the FIFO, so an earlier-queued event -- and the destroy event's own gate check -- still see the relationship. Added a unit test reproducing the ordering race.

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Partially fixed in ea62021bb: added realm -> context tracking (from script.realmCreated) so events that identify only a realm (e.g. script.realmDestroyed, or a log entry with only source.realm) resolve to a context via that map. I did not attempt dedicated-worker realm ownership chains (the 'owners' field) -- that needs a full ownership graph and felt like its own deeper piece of work; happy to follow up if that's a real gap for you. Added a unit test for the realmCreated/realmDestroyed case.

packages/appium/lib/bidi/proxy-client.ts
mykola-mokhnach: Fixed in ea62021bb: removed the fixed 30s default command timeout entirely (commandTimeoutMs now defaults to disabled/unlimited); it's still available as an opt-in via BidiProxyClientOptions for callers that want a cap. Added a regression test with a 100ms-delayed response and no configured timeout.

packages/appium/lib/bidi/proxy-client.ts
mykola-mokhnach: Fixed in ea62021bb: BidiProxyClient now preserves the original upstream error signature when errorFromW3CJsonCode falls back to UnknownError for a code the classic WebDriver map doesn't recognize (e.g. 'no such handle'), instead of relabeling it to 'unknown error'. Added a regression test.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,403 @@
...
+    return await bidiProxyClient.executeCommand(SESSION_SUBSCRIBE, params);
+  }
+  const events = (params.events ?? []) as string[];
+  const rawContexts = (params.contexts ?? ['']) as string[];
+  // BiDi normalizes session.subscribe context ids to their top-level traversable (subscribing
+  // with a frame's id also covers its whole tab and sibling frames), so local coverage has to
+  // track the same top-level scope -- otherwise a valid event from a sibling/ancestor context the
+  // upstream subscription actually covers would be rejected by the local gate.
+  const contexts = [
+    ...new Set(rawContexts.map((context) => (context ? getTopLevelContext(driver, context) : context))),
```
KazuCocoa, line 243: [P2] Wait for ancestry initialization before normalizing a subscription

`primeContextAncestry` runs without being awaited, while client commands are already accepted. If a client subscribes using an existing iframe ID before the initial `browsingContext.getTree` response arrives, this records that iframe as its own top-level context. When the tree eventually supplies its parent, the stored subscription is never normalized again, so valid events from the parent and sibling frames continue to be dropped for the lifetime of the subscription. I reproduced this by delaying only the initial tree response, subscribing to frame-1, then returning tab-1 with frame-1/frame-2: subsequent events for tab-1 and frame-2 were both lost. Please coordinate initialization with subscription normalization (or reconcile records when ancestry becomes available).

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,403 @@
...
+    // upstream round trip was in flight, so promote its *current* tracked value rather than
+    // recreating it from the original arguments -- otherwise a concurrent narrowing would be
+    // undone -- and don't resurrect a placeholder a concurrent unsubscribe fully removed.
+    const subscriptionId = (result as {subscription?: string} | undefined)?.subscription;
+    const current = tracking.get(placeholderId);
+    tracking.delete(placeholderId);
+    if (current) {
+      tracking.set(subscriptionId ?? placeholderId, current);
+    }
+    return result;
```
KazuCocoa, line 263: [P2] Do not prune a subscription that has not yet been sent upstream

Promoting the narrowed placeholder fixes the delayed-response case, but introduces a failure for the opposite ordering. Start with an existing global subscription to event A, then start `subscribe([A, B])` and `unsubscribe([A])` concurrently. The new subscribe awaits local `syncCoverage`, so the actual upstream order can be unsubscribe first, subscribe second. The unsubscribe removes A from the optimistic placeholder even though it cannot remove the future upstream subscription. This success path then preserves only B, permanently dropping valid A events from the newly successful subscription. I reproduced this with the default BaseDriver subscription methods and immediate successful upstream responses, using A=`browsingContext.load` and B=`browsingContext.domContentLoaded` semantics. Please distinguish not-yet-sent subscriptions from subscriptions affected by an upstream unsubscribe, or coordinate their forwarding and bookkeeping order.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,403 @@
...
+        affectedEvents.add(event);
+      }
+    }
+    await syncCoverage(driver, tracking, affectedEvents);
+  } else if (Array.isArray(params.events)) {
+    const events = params.events as string[];
+    const contexts = (params.contexts as string[] | undefined) ?? [''];
+    const expandedEvents = expandRelatedEventKeys(driver.bidiEventSubs, events);
+    await driver.bidiUnsubscribe([...expandedEvents], contexts);
+    pruneStaleTracking(driver.bidiEventSubs, tracking, expandedEvents);
```
KazuCocoa, line 337: [P2] Preserve sibling events when narrowing a module subscription

The new conservative expansion prevents the previously reported over-delivery, but also removes subscriptions the client still needs. For example, subscribe globally to `browsingContext`, then unsubscribe only `browsingContext.load`: upstream keeps `browsingContext.domContentLoaded`, `contextCreated`, etc. enabled, but `expandedEvents` includes the bare module and this removes its local coverage entirely. All those valid upstream events are now dropped. This affects ordinary standard events, not just unknown vendor modules, and the added e2e test currently asserts the loss of the sibling event. Please represent the remaining event set (or exclusions from a module) so a successful partial unsubscribe does not silently disable the rest of the subscription.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,482 @@
...
+  } catch (err) {
+    driver.log.warn(
+      `Could not fetch the initial upstream BiDi browsing context tree for ancestry tracking: ` +
+        `${err instanceof Error ? err.message : err}`,
+    );
+  }
+  try {
+    await bidiProxyClient.executeCommand(SESSION_SUBSCRIBE, {
+      events: ['browsingContext.contextCreated', 'browsingContext.contextDestroyed'],
+      contexts: [''],
```
KazuCocoa: [P1] Omit contexts for the internal global subscription

The empty string is Appium’s local wildcard, not a valid upstream BiDi context ID. `session.subscribe` resolves every supplied context ID, so `contexts: [""]` fails with `no such frame` on a standard browser instead of creating a global subscription. The catch only logs the failure and never retries. Consequently, frames created after the initial `getTree` snapshot are never tracked unless the client separately subscribes to lifecycle events, and their logs/network/load events are still dropped for a parent-tab subscription. Please omit `contexts` in this upstream request and test with a stub that validates context IDs rather than accepting this invalid request.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,482 @@
...
+    // Normalize a missing context to '' here (rather than leaving it undefined), matching how
+    // driver/plugin-origin events are normalized in initBidiEventListeners, so origin never
+    // changes what plugins see.
+    const source = params.source as {context?: string; realm?: string} | undefined;
+    const realm = (params.realm as string | undefined) ?? source?.realm;
+    const context =
+      parsed.context ??
+      (params.context as string | undefined) ??
+      source?.context ??
+      (realm ? getRealmContext(bidiHandlerDriver, realm) : undefined) ??
```
KazuCocoa, line 440: [P2] Populate realm mappings independently of the client’s subscriptions

The observed-realm fix works when `script.realmCreated` is actually received, but a client subscribing only to `script.realmDestroyed` for tab-1 does not enable creation events upstream. There is no `script.getRealms` bootstrap or internal realm lifecycle subscription, so a valid `{realm: "r1"}` destruction notification still resolves to `""` and is discarded. This also affects realms that predate the subscription. The new unit test sends `realmCreated` from the fake upstream despite the client subscribing only to `realmDestroyed`, hiding this prerequisite. Please obtain/maintain the mapping independently of client subscriptions. This remaining window-realm case is separate from the already acknowledged worker `owners` limitation.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,482 @@
...
+          `Error while running a plugin's BiDi event interceptor for '${event.method}': ` +
+            `${err instanceof Error ? err.message : err}. Event was dropped.`,
+        );
+      } finally {
+        // Destruction is applied only after this event has been filtered/delivered (and after
+        // any earlier-queued event has had its own turn), so an ancestry/realm relationship
+        // remains available for the destruction event's own gate check, and for any event still
+        // ahead of it in the queue.
+        if (event.method === 'browsingContext.contextDestroyed') {
+          trackContextDestroyed(bidiHandlerDriver, event);
```
KazuCocoa: [P2] Keep shared ancestry alive for every connection’s queued events

Moving deletion into the FIFO fixes one connection, but the ancestry map is shared by driver/session while each connection has its own FIFO. With two connections to the same session, let A’s plugin pause on a child-frame `browsingContext.load`. B can process that child’s subsequent `contextDestroyed` immediately and delete the shared relationship here. When A resumes, its earlier load fails the parent-tab subscription gate and is lost. I reproduced this with two dispatchers sharing one driver, a parent-scoped subscription, and a deferred plugin on A. Please coordinate lifecycle cleanup across the connection queues, or retain the relevant ancestry for events already queued on each connection.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in 29a575f05: exported waitForContextTreeReady from events.ts (tracks the getTree bootstrap promise per BidiProxyClient) and executeProxiedSubscribe now awaits it before normalizing contexts, so a subscribe can't race ahead of the bootstrap.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in 29a575f05, at the root cause: added a per-driver mutex (withSubscriptionLock) wrapping both the local bookkeeping and the upstream round trip for session.subscribe/unsubscribe, so two commands on the same connection can no longer interleave their bookkeeping mutations -- whichever command's message the server received first now fully completes before the next one starts. Added an e2e regression test.

packages/appium/lib/bidi/commands.ts
mykola-mokhnach: Fixed in 29a575f05: added a small, explicit table of known member events (KNOWN_MODULE_EVENTS) for the standard browsingContext/log/network/script modules. When one of those modules' members is individually narrowed out, the module-wide tracked record is now split into precise per-sibling entries instead of dropping the whole module -- so siblings keep their coverage. Modules outside that table (vendor/experimental) still fall back to the coarser conservative behavior, since their membership genuinely can't be known. Added an e2e regression test.

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in 29a575f05: the internal lifecycle subscribe no longer passes contexts: [''] -- it omits contexts entirely now, which is how BiDi itself expresses a global subscription.

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in 29a575f05: added a script.getRealms bootstrap (alongside the existing browsingContext.getTree one) and an internal script.realmCreated/realmDestroyed subscribe, both independent of client subscriptions, so realm->context mappings are available even for realms that predate the connection or that the client never explicitly subscribed to creation events for. Updated the unit test to no longer rely on the client's own realmCreated subscription.

packages/appium/lib/bidi/events.ts
mykola-mokhnach: Fixed in 29a575f05, by removing the premise: ancestry/realm entries are no longer deleted on contextDestroyed/realmDestroyed at all, since the map is shared across a session's connections but each has its own independent delivery queue, which made any deletion point racy against another connection's still-queued events. This trades a small, session-bounded amount of memory for avoiding that whole class of cross-connection races. Added a regression test with two dispatchers sharing one driver.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,475 @@
...
+      splitEvents.add(moduleName);
+      for (const sibling of KNOWN_MODULE_EVENTS[moduleName]) {
+        splitEvents.add(sibling);
+      }
+    }
+
+    const expandedEvents = expandRelatedEventKeys(driver.bidiEventSubs, events);
+    await driver.bidiUnsubscribe([...expandedEvents], contexts);
+    if (splitEvents.size) {
+      await syncCoverage(driver, tracking, splitEvents);
```
KazuCocoa, line 332: [P2] Prune explicit event records before rebuilding split coverage

With a global `browsingContext` subscription and a separate global `browsingContext.load` subscription, unsubscribing `{events: ["browsingContext.load"]}` removes load from both subscriptions upstream. Locally, however, the explicit load record is still in `tracking` when `syncCoverage` runs, so it restores load coverage immediately after `bidiUnsubscribe` removed it. The subsequent pruning sees that restored coverage and retains the stale record. I reproduced a driver/plugin-emitted load event still reaching the client after the unsubscribe succeeds. Please invalidate the affected explicit records before rebuilding coverage from the surviving subscriptions.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,475 @@
...
+): void {
+  for (const [id, sub] of tracking) {
+    const moduleContexts = sub.events[moduleName];
+    if (!moduleContexts) {
+      continue;
+    }
+    const nextEvents = {...sub.events};
+    delete nextEvents[moduleName];
+    for (const sibling of siblings) {
+      if (excluding.has(sibling)) {
```
KazuCocoa, line 116: [P2] Keep context-scoped module coverage when unsubscribing globally

Subscribe to `browsingContext` for `contexts: ["tab-1"]`, also subscribe to `browsingContext` globally, then unsubscribe `{events: ["browsingContext.load"]}` without contexts. That unsubscribe only changes global subscriptions upstream; the tab-scoped subscription must still cover load. This loop nevertheless excludes load from every matching module record, including the tab-scoped one, and rebuilding coverage drops valid load events for tab-1. I reproduced this with ordinary `contexts` (independent of the acknowledged `userContexts` limitation). Please restrict splitting/removal to the subscription scopes actually affected by the unsubscribe.

packages/appium/lib/bidi/commands.ts
```diff
@@ -0,0 +1,475 @@
...
+    'browsingContext.contextDestroyed',
+    'browsingContext.domContentLoaded',
+    'browsingContext.load',
+    'browsingContext.navigationStarted',
+    'browsingContext.fragmentNavigated',
+    'browsingContext.historyUpdated',
+    'browsingContext.userPromptClosed',
+    'browsingContext.userPromptOpened',
+    'browsingContext.downloadWillBegin',
+    'browsingContext.navigationAborted',
```
KazuCocoa, line 35: [P2] Include downloadEnd when expanding browsingContext subscriptions

This registry includes `browsingContext.downloadWillBegin` but omits the standard `browsingContext.downloadEnd` event. After subscribing to `browsingContext` and unsubscribing only `browsingContext.load`, the wildcard is replaced with these listed siblings, so downloadEnd is silently filtered out even though the upstream subscription still covers it. I reproduced downloadWillBegin being delivered while downloadEnd is dropped, which can leave a client waiting indefinitely for download completion. Please include the full event set, or represent exclusions without losing other module members.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,477 @@
...
+  }
+}
+
+async function subscribeInternalLifecycleEvents(driver: AnyDriver, bidiProxyClient: BidiProxyClient): Promise<void> {
+  try {
+    // no `contexts`/`userContexts` here -- omitting both is how BiDi expresses "every context";
+    // Appium's local '' wildcard is not a valid upstream context id
+    await bidiProxyClient.executeCommand(SESSION_SUBSCRIBE, {
+      events: [
+        'browsingContext.contextCreated',
```
KazuCocoa, line 370: [P1] Preserve initial creation notifications for client subscriptions

Installing this global subscription before any client subscription consumes the initial contextCreated/realmCreated notifications: they reach the local gate while it has no client coverage and are discarded. When the client later subscribes, BiDi only replays existing contexts/realms for navigables where the event was not already enabled; this internal global subscription has already enabled it everywhere. Consequently the client never learns about existing tabs/realms through its subscription. I reproduced both missing initial notifications with an upstream stub implementing those first-enable replay semantics. Please preserve the client-visible initial notifications when adding these internal subscriptions.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,477 @@
...
+        'browsingContext.contextDestroyed',
+        'script.realmCreated',
+        'script.realmDestroyed',
+      ],
+    });
```
KazuCocoa, line 375: [P2] Protect the internal lifecycle subscription from client unsubscriptions

This creates a global subscription in the same upstream session, but its returned ID is discarded and client attribute-based unsubscriptions are forwarded unchanged. A client subscribing and then unsubscribing `{events: ["browsingContext.contextCreated"]}` removes that event from all global subscriptions, including this internal one. If the client still subscribes to tab-scoped log events, newly created child frames are then absent from the ancestry map and their valid logs are dropped. I reproduced that sequence; unsubscribing script/browsingContext modules can similarly disable internal realm/context tracking. Please preserve internal lifecycle coverage when translating client unsubscriptions.

packages/appium/lib/bidi/events.ts
```diff
@@ -0,0 +1,477 @@
...
+
+/**
+ * Keeps context/realm ancestry current independently of whatever the client itself subscribes
+ * to, by bootstrapping from the upstream context tree and realm list, then subscribing
+ * internally to lifecycle events. Best-effort: failures are logged, not surfaced.
+ */
+function primeAncestryTracking(driver: AnyDriver, bidiProxyClient: BidiProxyClient): void {
+  const treeReady = bootstrapContextTree(driver, bidiProxyClient);
+  CONTEXT_TREE_READY.set(bidiProxyClient, treeReady);
+  void bootstrapRealms(driver, bidiProxyClient);
```
KazuCocoa, line 391: [P2] Wait for realm tracking before accepting realm-scoped subscriptions

Subscriptions wait for the browsing-context tree, but both realm bootstrapping and the internal lifecycle subscription run without being awaited. If getTree finishes first while getRealms and the internal subscribe/replay are still pending, a client subscription to `script.realmDestroyed` for tab-1 can already succeed. A destruction event containing only `{realm: "r1"}` then has no realm-to-context mapping and is dropped, even though r1 belongs to tab-1. I reproduced this by delaying those two initialization operations; completing initialization later cannot recover the lost notification. Please include the required realm/lifecycle setup in readiness, or buffer relevant events until the mapping is available.

PR #22780, opened 2026-09-17
fix(base-driver): keep existing contexts when session.subscribe is called again

Mochxd:
session.subscribe was replacing the context list for an event. A second subscribe for the same event dropped the first tab, so those events never got sent.

Subscribe now unions the contexts. Unsubscribe already filtered, so the two commands agree.


packages/base-driver/lib/basedriver/commands/bidi.ts
```diff
@@ -12,7 +12,9 @@ declare module '../driver' {
 const BidiCommands: IBidiCommands = {
   async bidiSubscribe<C extends Constraints>(this: BaseDriver<C>, events: string[], contexts: string[] = ['']) {
     for (const event of events) {
-      this.bidiEventSubs[event] = contexts;
+      // a later subscribe for the same event must keep existing contexts; the spec appends subscriptions
+      const existing = this.bidiEventSubs[event] ?? [];
+      this.bidiEventSubs[event] = [...new Set([...existing, ...contexts])];
```
mykola-mokhnach: consider using util.uniq helper

mykola-mokhnach: consider using util.uniq helper from support package

Mochxd: Switched it to util.uniq.

mykola-mokhnach (approved):

PR #22790, opened 2026-09-19
docs: add Appium 3 to 4 migration guide

mykola-mokhnach:
Summarizes the breaking changes, removed endpoints, and driver/plugin developer impact from the appium4 branch's PRs, mirroring the existing migrating-2-to-3 guide.


eglitise:
I would prefer to write this guide myself later. I assume it's not needed until closer to the full release.

mykola-mokhnach:
> I would prefer to write this guide myself later. I assume it's not needed until closer to the full release.

This is for the beta release. When we are ready for GA then the document could be polished. We are still not feature-complete anyway, I also beta-feedback later.

packages/appium/docs/en/guides/migrating-3-to-4.md
```diff
@@ -0,0 +1,269 @@
...
+The `orientation`, `context`/`contexts`, and `rotation` deprecated endpoints are unaffected and
+remain available.
+
+## Changes for Driver and Plugin Developers
+
+The following changes primarily affect authors of Appium drivers, plugins, or tools that import
+Appium's internal packages directly (`@appium/base-driver`, `@appium/base-plugin`,
+`@appium/support`, `@appium/logger`). They do not affect most Appium end users.
+
+### All Packages Are Now ESM-Only
```
eglitise: Since Appium and drivers can also be imported as npm packages by end users, this point is worth moving into the initial breaking change list

packages/appium/docs/en/guides/migrating-3-to-4.md
```diff
@@ -0,0 +1,269 @@
...
+(`createSession(jwpCaps, reqCaps, w3cCaps)`), a holdover from JSONWP. Only the single
+`w3cCapabilities` argument is supported now.
+
+!!! info "Actions Needed"
+
+    Update any driver or plugin overriding `createSession` to use the single-argument signature
+
+### `getSession` Command Removed
+
+`BaseDriver.getSession`, `ISessionHandler.getSession`, and the `GET /session/:sessionId` route
```
eglitise: The removed route should be moved to the removed routes list earlier

packages/appium/docs/en/guides/migrating-3-to-4.md
```diff
@@ -0,0 +1,269 @@
...
+have been removed, along with the `appium:eventTimings` capability that controlled whether its
+response included event history.
+
+!!! info "Actions Needed"
+
+    Use `mobile: getAppiumSessionCapabilities` to retrieve session capabilities, and
+    `mobile: getLogEvents` to retrieve event history (now returned unconditionally)
```
eglitise: Both of these are endpoints, not execute methods

packages/appium/docs/en/guides/migrating-3-to-4.md
```diff
@@ -0,0 +1,269 @@
...
+* `GET/POST /session/:sessionId/ime/*` (all IME endpoints)
+* `GET /session/:sessionId/location`, `POST /session/:sessionId/location` (legacy geolocation)
+* `GET /session/:sessionId/network_connection`, `POST /session/:sessionId/network_connection`
+* `POST /session/:sessionId/receive_async_response`
+* `GET /session/:sessionId/element/:elementId` (undocumented legacy stub)
+
+The `orientation`, `context`/`contexts`, and `rotation` deprecated endpoints are unaffected and
+remain available.
+
+## Changes for Driver and Plugin Developers
```
eglitise, line 163: It may also be worth adding the removed JSONWP support, such as JSONWP error codes no longer being included in responses

packages/appium/docs/en/guides/migrating-3-to-4.md
mykola-mokhnach: Moved this into the initial Breaking Changes list (right after the Node version bump), since it affects anyone importing Appium/driver packages directly. Fixed in a51260bf2.

packages/appium/docs/en/guides/migrating-3-to-4.md
mykola-mokhnach: Moved `GET /session/:sessionId` into the removed-routes list, with a cross-reference from the getSession section.

packages/appium/docs/en/guides/migrating-3-to-4.md
mykola-mokhnach: Good catch — replaced with the actual endpoints (`GET .../appium/capabilities`, `POST .../appium/events`) instead of nonexistent execute methods.

packages/appium/docs/en/guides/migrating-3-to-4.md
mykola-mokhnach: Added a new subsection documenting the removed jsonwpCode/errorFromMJSONWPStatusCode, legacy ELEMENT key duplication, downstream MJSONWP proxy conversion, and the legacy {sessionId, status, value} response shape.

packages/appium/docs/en/guides/migrating-3-to-4.md
```diff
@@ -0,0 +1,293 @@
...
+of specific removed flags/endpoints.
+
+## Installation
+
+The installation method for Appium 4 remains the same as for Appium 3. If you want to upgrade, you
+can simply install Appium 4 on top of your existing installation:
+
+```bash
+# optional: `appium setup reset`
+npm install -g appium
```
KazuCocoa: should we add `@beta`

mykola-mokhnach: Good point — updated the install command to `npm install -g appium@beta` and noted that Appium 4 is currently only published under the `beta` npm dist-tag. Fixed in 6a968ac6f.

eglitise (approved):

PR #22796, opened 2026-09-20
feat(base-driver)!: exempt getStatus from the commands queue by default

mykola-mokhnach:
## Summary
- Excludes `getStatus` from `BaseDriver`'s commands queue by default, so status can be answered immediately while another command is queued or running, instead of waiting its turn behind it.
- Adds an overridable `queueExemptCommands` getter (default: `{getStatus}`) so inherited drivers can register their own queue-exempt commands, instead of overriding `executeCommand` themselves (as `appium-xcuitest-driver` currently does).
- The rest of the per-command handling (new-command-timeout clear/restart, unexpected-shutdown race, event history) still runs as usual for exempt commands — only the `commandsQueueGuard.acquire` serialization is skipped.

BREAKING CHANGE: `getStatus` command is now exempt from the queue by default


eglitise:
Could the docs for this feature also be added to the build drivers documentation?

mykola-mokhnach:
> Could the docs for this feature also be added to the build drivers documentation?

I'll add it in a different PR

packages/base-driver/lib/basedriver/driver.ts
```diff
@@ -174,22 +174,27 @@ export class BaseDriver<
...
       );
     }
 
-    const res = this.isCommandsQueueEnabled
-      ? await this.commandsQueueGuard.acquire(synchronizationKey, runCommandPromise)
-      : await runCommandPromise();
+    const res =
+      this.isCommandsQueueEnabled && !isQueueExempt
+        ? await this.commandsQueueGuard.acquire(synchronizationKey, runCommandPromise)
+        : await runCommandPromise();
```
KazuCocoa, line 211: [P1] Avoid restarting the idle timeout while another command is running

Bypassing the queue still executes `runCommandPromise`'s `finally`, which calls `startNewCommandTimeout()` even when another command is still executing. If that command remains active longer than `newCommandTimeout` after the exempt command completes, the timer triggers unexpected shutdown, deletes the session, and rejects the running command. This affects `getStatus` when dispatched through `BaseDriver.executeCommand`, as well as custom queue-exempt commands.

I reproduced this with the PR's actual BaseDriver methods in an isolated harness: a 100 ms command succeeds with status serialized, but a concurrent exempt status call causes a 30 ms new-command timeout to reject the command and clear the session. Could the timeout restart account for commands still in flight, so it only starts once the driver is idle?

mykola-mokhnach: Good catch — fixed in 3e56ee5ea by tracking in-flight command count and only restarting the new command timeout once the driver is fully idle (no queued or queue-exempt commands still executing), rather than after every individual command finishes. Added a regression test that reproduces this exact race (short `newCommandTimeoutMs`, a slow in-flight command, and a concurrent `getStatus` call) and fails without the fix.

KazuCocoa (approved):

PR #22799, opened 2026-09-21
fix(base-driver): carry proxy req via AsyncLocalStorage instead of args

mykola-mokhnach:
Removes the args-list smuggling hack for reqForProxy (route-handler.ts TODO): the raw request is now threaded through an AsyncLocalStorage context, so it never pollutes the driver command's args list and no breaking change to executeCommand's signature is needed.


packages/appium/lib/appium.ts
```diff
@@ -574,13 +576,9 @@ export class AppiumDriver extends DriverCore<AppiumDriverConstraints> {
...
-    // if a plugin override proxying for this command and that is why we are here instead of just
-    // letting the protocol proxy the command entirely, determine that, get the request object for
-    // use later on, then clean up the args
-    const reqForProxy = args.at(-1)?.reqForProxy;
-    if (reqForProxy) {
-      args.pop();
-    }
+    // if a plugin overrode proxying for this command and that is why we are here instead of just
+    // letting the protocol proxy the command entirely, get the request object for use later on
+    const reqForProxy = getProxyReq();
```
KazuCocoa, line 582: [P2] Clear the proxy context before running plugin callbacks

The stored request is inherited by nested `AppiumDriver.executeCommand` calls. If a plugin or embedding integration holds a reference to the umbrella driver and re-enters it to run another command, that command also sees `reqForProxy` and proxies the outer request instead. For example, a click handler calling `appium.executeCommand('getPageSource', sessionId)` before `next()` executes the click twice and returns the click result as page source. This concerns calls through the umbrella driver, not calls directly to the inner driver passed to a plugin.

I reproduced this using the actual dispatch methods with stubbed dependencies: the previous argument-based implementation dispatches `getPageSource` and then proxies the click, while this implementation proxies the click twice. Could we capture the request for the current command, then clear the ambient context while running its plugin/default chain, and add a nested-command regression test?

mykola-mokhnach: Good catch, thanks — fixed in fcc9f8357.

You were right: `getProxyReq()` was ambient for the whole call, so a plugin re-entering `executeCommand` for a different command mid-chain would inherit the outer `reqForProxy` and proxy the wrong request.

Fix: capture `reqForProxy` at the top of `executeCommand` (as before), then clear the ambient context via a new `withoutProxyReq()` helper while the plugin/default chain (`executeWrappedCommand`) runs, so any reentrant `executeCommand` call sees no proxy req.

Added a unit test (`packages/appium/test/unit/appiumdriver.spec.ts`, "should not leak the ambient proxy request into a command a plugin re-enters via executeCommand") that reproduces exactly the scenario you described — a plugin calling back into `executeCommand('getWindowHandle', ...)` before `next()` — and verified it fails without the fix (the nested call gets proxied instead of running normally) and passes with it.

KazuCocoa (approved):

PR #22800, opened 2026-09-21
refactor(base-driver)!: extract the HTTP/Express server into @appium/http-server

mykola-mokhnach:
@appium/base-driver dropped its express, morgan, body-parser, method-override, and path-to-regexp dependencies. Express app/HTTP(S) server construction (server(), configureServer(), the morgan-based logging middleware, and the CORS/idempotency/websocket-upgrade middleware) moved to a new @appium/http-server package, since appium is the only component that actually operates the server -- inherited drivers only ever needed BaseDriver, the protocol/command-dispatch layer, and WebDriverProxy from base-driver.

DEFAULT_WS_PATHNAME_PREFIX moves into base-driver's own constants instead: it's a plain string with no server dependency, and real drivers reference it directly. routeConfiguringFunction, WebDriverProxy, and all protocol/ command-dispatch logic stay in base-driver unchanged, since real out-of-tree drivers depend on them directly and moving them would be a much bigger break for no bundle-size benefit (they only ever type-imported Express).

appium now depends on @appium/http-server instead of the express-family packages directly. Checked every locally available driver/plugin repo (android, uiautomator2, xcuitest, mac2, windows, espresso, safari, gecko, novawindows, chromium, flutter-integration, devtools-plugin): none import server/normalizeBasePath/configureServer/ServerOpts, so appium/driver.js and driver.d.ts no longer re-export them at all, and now use explicit named exports instead of `export *`. fake-driver (the only consumer, an internal test fixture) now depends on @appium/http-server directly instead of going through the driver.js re-export.

Server-dependent e2e/unit tests (protocol dispatch, websockets, proxying, idempotency/aborted-session) moved from base-driver's test suite into @appium/http-server's, since they now exercise the base-driver + http-server integration.

BREAKING CHANGE: @appium/base-driver no longer exports server, normalizeBasePath, configureServer, ServerOpts, ConfigureServerOpts, ConfigureHttpOpts, or StartServerOpts, and no longer depends on express/morgan/body-parser/ method-override/path-to-regexp. Use the new @appium/http-server package instead.

BREAKING CHANGE: appium/driver.js and driver.d.ts no longer re-export server/normalizeBasePath/configureServer/ServerOpts (or their types), and no longer use a blanket `export *`. Import from @appium/http-server directly if you were relying on this undocumented path.


KazuCocoa (approved):
lgtm

PR #22805, opened 2026-09-22
refactor(base-driver): vendor Levenshtein distance and drop http-status-codes dep

mykola-mokhnach:
Both deps were used for a handful of small things: distance() over short command/capability names, and 4 numeric status codes. Replacing them with a local DP implementation and a small local constants object removes two external dependencies with negligible added code.


packages/base-driver/lib/helpers/levenshtein-match.ts
```diff
@@ -1,12 +1,38 @@
...
+/**
+ * Levenshtein edit distance between two strings (classic O(n*m) DP, single-row optimized).
+ * Inputs here are short command/capability names, so no need for a bit-vector algorithm.
+ */
+function distance(a: string, b: string): number {
+  if (a === b) {
+    return 0;
+  }
+  if (!a.length) {
+    return b.length;
```
eglitise: `b.length` is not guaranteed to be truthy here

mykola-mokhnach: Good catch on the ambiguity — that branch was only reachable after `a === b` and `!a.length` both failed, so `b.length` was actually guaranteed non-zero there, but the guarantee wasn't obvious from the code. Turns out the DP loop already produces the correct result for empty/equal strings without any special-casing, so I just removed the three early-return branches in 1e0356a10 instead — simpler, and the ambiguity goes away with it.

eglitise: I was thinking more about the case if a or b are null/undefined, but perhaps it's fine if there's guards in every callsite

mykola-mokhnach: Checked the call sites — `distance()` isn't exported from the package's public API (not in `lib/index.ts`), and both internal callers (`rankLevenshteinCandidates` from `driver.ts`/`execute.ts`) pass strings derived from `Object.keys()` or typed `string` params, under `strict: true`. So null/undefined can't reach it without an unsafe cast somewhere upstream, same as the old `fastest-levenshtein` usage. I'll leave it without extra guards to match that — happy to add them if you'd rather be defensive here regardless.

eglitise (approved):

